Issue #87, opened 2016-04-26
Support .gzipped benchmark data


danielmitterdorfer:
Currently we only support .zip and .bz2. We should maybe also support gzipped data sets.

Currently, we get the following trace:

```
Traceback (most recent call last):
  File "/usr/local/bin/esrally", line 9, in <module>
    load_entry_point('esrally==0.2.1.dev0', 'console_scripts', 'esrally')()
  File "/home/dm/rally/esrally/rally.py", line 238, in main
    success = race_control.start(subcommand)
  File "/home/dm/rally/esrally/racecontrol.py", line 260, in start
    pipeline.run(t)
  File "/home/dm/rally/esrally/racecontrol.py", line 88, in run
    step.run(track)
  File "/home/dm/rally/esrally/racecontrol.py", line 27, in run
    self.command(self.ctx, track)
  File "/home/dm/rally/esrally/racecontrol.py", line 116, in prepare_track
    ctx.marshal.setup(track)
  File "/home/dm/rally/esrally/track/track.py", line 346, in setup
    unzipped_data_set_path = self._unzip(data_set_path)
  File "/home/dm/rally/esrally/track/track.py", line 406, in _unzip
    io.unzip(data_set_path, io.dirname(data_set_path))
  File "/home/dm/rally/esrally/utils/io.py", line 68, in unzip
    raise RuntimeError("Unsupported file extension '%s'. Cannot unzip '%s'" % (extension, zip_name))
RuntimeError: Unsupported file extension '.gz'. Cannot unzip '/home/dm/rally-test/data/logging/logs.json.gz'
```

danielmitterdorfer:
Closed by #88

PR #88, opened 2016-04-28
Add additional decompression options - PEP-8

monk-ee:
Addresses and closes the following issue: Support .gzipped benchmark data #87

### DIFF
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -3,7 +3,9 @@ import errno
 import glob
 import subprocess
 import bz2
+import gzip
 import zipfile
+import tarfile
 
 from esrally.utils import process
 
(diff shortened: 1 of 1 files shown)


monk-ee:
image

danielmitterdorfer:
Thanks for your PR. I'll have my colleague check what's wrong with the CLA check and will review it then.

karmi:
Hi @monk-ee, we have your signature in our records, but it seems you have used a one e-mail when signing the CLA and another one in your Git commit. Can you please add both to your Github profile (they can be hidden), so the systems properly match you against your Github username? When you comment here again, it should re-trigger the validation and all should be green.

monk-ee:
Thanks, was using my work computer, my bad.

esrally/utils/io.py
```diff
@@ -48,22 +50,56 @@ def unzip(zip_name, target_directory):
...
     elif extension == ".bz2":
-        # We rather avoid external tools as much as possible to simplify Rally's setup, hence we use the library functions
+        # We rather avoid external tools as much as possible to simplify Rally's setup, hence we use the library
+        # functions
         target_file = os.path.join(target_directory, filename)
         with open(target_file, "wb") as extracted, bz2.BZ2File(zip_name, "rb") as file:
             for data in iter(lambda: file.read(100 * 1024), b''):
                 extracted.write(data)
+    elif extension == ".gz":
+        target_file = os.path.join(target_directory, filename)
```
danielmitterdorfer: Here we have a fair deal of repetition. I wonder whether it would make sense to create a separate internal function for this. Something along the lines of:

```python
def _do_unzip(target_directory, target_filename, compressed_file):
    target_file = os.path.join(target_directory, target_filename)
    with open(target_file, "wb") as extracted:
        for data in iter(lambda: compressed_file.read(100 * 1024), b''):
            extracted.write(data)
```

and then calling it just for each supported extension, e.g. for ".gz" files: `_do_unzip(target_directory, filename, gzip.GzipFile(zip_name, "rb"))`. What do you think?

danielmitterdorfer:
@monk-ee I tested your PR locally and it looks great. Thanks for also supporting lots of other file extensions too! I left on comment on the code. Would be great if you could change that and finally squash into one commit, then I think we're good to merge.

monk-ee:
Yeah OK, I totally agree, it has too much repetition. I was not sure how you wanted to approach it. I like your suggestion will implement it and squash away.

danielmitterdorfer:
Great that you like it. Thanks!

danielmitterdorfer:
Thanks again for your PR! I've just merged it to master. I'll also create an authors list and add you there if you're fine with that.

PR #113, opened 2016-07-08
Add instructions for installing Rally into a Python VirtualEnv.

joshuar:
This PR adds instructions for running Rally in a Python VirtualEnv, which means no sudo access required and less issues with the Python dependencies getting upgraded with system updates.


jasontedor:
+1, this is how I run Rally.

danielmitterdorfer:
Very helpful, thanks for your PR @joshuar!

PR #130, opened 2016-08-12
Increase `maxage` of the `jfr` telemetry.

jpountz:
By default, the `jfr` telemetry only keeps information about the last 15
minutes. This commit increases it to one day so that we keep information about
the whole benchmark in most cases.

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -203,7 +203,7 @@ class FlightRecorder(TelemetryDevice):
         #
         # in that case change to: -XX:StartFlightRecording=defaultrecording=true,settings=es-memory-profiling
         return {"ES_JAVA_OPTS": "-XX:+UnlockDiagnosticVMOptions -XX:+UnlockCommercialFeatures -XX:+DebugNonSafepoints -XX:+FlightRecorder "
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
It took a little bit to verify the current behaviour. I ran the PMC track for multiple times (took about 3 hours each time) with JFR and here's what I've found out:

* The documented defaults according to the [Oracle JMC 5.5 docs](http://docs.oracle.com/javacomponents/jmc-5-5/jfr-runtime-guide/comline.htm#JFRRT186) should already avoid limiting the flight recording. The flight recording showed up as three hours long in Mission Control but I had only data for the last 50 minutes or so.
* By explicitly specifying `maxage=0s,maxsize=0`as additional options in `FlightRecorderOptions` (which should actually be set by default according to the docs...) I could successfully get data for the whole duration of the benchmark, so I'd prefer setting these parameters instead. Could you please update your PR with these settings?

Hint: I tested with Oracle JDK 1.8.0_92-b14 on Linux.

jpountz:
Thanks for looking, I updated the PR.

danielmitterdorfer:
LGTM

PR #135, opened 2016-08-23
Fix typo in rally message.

nicholaskuechler:


karmi:
Hi @nicholaskuechler, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/rally/pull/135.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #136, opened 2016-08-23
Fix typo in rally message.

nicholaskuechler:


danielmitterdorfer:
Thanks for the fix @nicholaskuechler. It will be part of the next release of Rally (0.4.0)

PR #140, opened 2016-08-24
Handle $TERM being set to 'dumb' in progress reporting

dakrone:
Some shells set their `TERM` environment variable to 'dumb' when they do
not support all shell escape codes. This changes the progress reporting
to check if the `$TERM` is 'dumb', and if so, report progress like:

```
  Benchmarking indexing at 0.0 docs/s                                [  0% done]
  Benchmarking indexing at 1425.2 docs/s                             [100% done]

  Benchmarking stats (warmup iteration 0/100)                        [  0% done]
  Benchmarking stats (warmup iteration 5/100)                        [  5% done]
  Benchmarking stats (warmup iteration 10/100)                       [ 10% done]
  Benchmarking stats (warmup iteration 15/100)                       [ 15% done]
  Benchmarking stats (warmup iteration 20/100)                       [ 20% done]
  Benchmarking stats (warmup iteration 25/100)                       [ 25% done]
... etc ...
```

instead of using escape codes to overwrite the previous progress line.

Additionally, this adds the `plain_output` parameter to
`CmdLineProgressReporter` (defaulting to `False`) in hope of hooking
this up as a `--plain_output=true` CLI parameter as future work.

### DIFF
--- a/esrally/utils/progress.py
+++ b/esrally/utils/progress.py
@@ -1,4 +1,5 @@
 import sys
+import os
 
 
 class CmdLineProgressReporter:
@@ -6,9 +7,12 @@ class CmdLineProgressReporter:
     CmdLineProgressReporter supports displaying an updating progress indication together with an information message.
     """
 
-    def __init__(self, width=80):
+    def __init__(self, width=80, plain_output=False):
         self._width = width
         self._first_print = True
+        self._plain_output = plain_output
+        if (os.environ.get('TERM') == 'dumb'):
+            self._plain_output = True
 
     def print(self, message, progress):
         """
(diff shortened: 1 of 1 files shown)


esrally/utils/progress.py
```diff
@@ -1,14 +1,18 @@
...
     """
     CmdLineProgressReporter supports displaying an updating progress indication together with an information message.
     """
 
-    def __init__(self, width=80):
+    def __init__(self, width=80, plain_output=False):
         self._width = width
         self._first_print = True
+        self._plain_output = plain_output
+        if (os.environ['TERM'] == 'dumb'):
```
danielmitterdorfer: This relies that the variable "TERM" is set. Evil me did `unset TERM` and it broke with an ambiguous error. I know TERM should normally be set though...

danielmitterdorfer:
Thanks for your PR @dakrone. I left one minor comment, otherwise LGTM.

esrally/utils/progress.py
dakrone: Good idea, I pushed a change to use `.get` which will return `None` so this doesn't throw a `KeyError`

danielmitterdorfer:
LGTM

dakrone:
@danielmitterdorfer thanks for taking a look, do you want to merge this or should I? I don't know what your git workflow is on this project so I want to merge it however you'd like it merged :)

danielmitterdorfer:
@dakrone I'd like to handle it as close as possible to the Elasticsearch workflow. I guess this is easiest option for all of us. :) The only special case: It does not make too much sense to create a PR for my own changes. ;) So you're good to merge at any time that's convenient for you.

PR #156, opened 2016-09-20
Add logger prefixes to log configuration

jasontedor:
In elastic/elasticsearch@fbe27664a65ca7ab5b7dbe1c328ab27423e4313c,
logger prefixes (e.g., node name, index name, shard IDs) are no longer
part of the message but are now markers. This commit adapts the layout
patterns so that these markers are included in log messages.


danielmitterdorfer:
Thanks for the PR @jasontedor! LGTM. You can just merge the PR at any time.

Issue #121, opened 2016-07-19
Allow specifying bulk_size in doc size, not number


eugenemiretsky:

danielmitterdorfer:
Beg my pardon but I don't understand what you mean by that. :)

Here is an example snippet from the geonames track:

```json
{
      "name": "index-append-default-settings",
      "type": "index",
      "index-settings": {
        "index.number_of_replicas": 0
      },
      "bulk-size": 5000,
      "force-merge": true,
      "clients": {
        "count": 8
      }
}
```

This means that the intended bulk size is 5000 documents per bulk.

nik9000:
I think it'd be useful to have something like
```
{
      "name": "index-append-default-settings",
      "type": "index",
      "index-settings": {
        "index.number_of_replicas": 0
      },
      "bulk-size": '5mb',
      "force-merge": true,
      "clients": {
        "count": 8
      }
}
```

When I talk to folks about picking the right bulk size I almost always talk in terms of megabytes, not documents. I know both numbers matter, but the upper limit in mb is very important.

danielmitterdorfer:
@nik9000 thanks. That makes sense now. Should be doable but probably not top priority tbh.

danielmitterdorfer:
We also should not start implementing this before #108 as the code base changes a bit.

PR #186, opened 2016-11-22
support bulk-size/batch_size with unit kb|mb|gb

jomenxiao:
support bulk-size/batch_size with unit kb|mb|gb   #121

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -2,6 +2,7 @@ import logging
 import random
 import time
 import types
+import re
 from enum import Enum
 
 from esrally import exceptions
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
@jomenxiao Thank you for your PR! Can I please ask you to [sign our CLA](https://www.elastic.co/contributor-agreement) with the same e-mail address as you've used in your PR? Afterwards, I'll review it.

jomenxiao:
@danielmitterdorfer  done

esrally/track/params.py
```diff
@@ -197,25 +198,39 @@ def __init__(self, indices, params):
...
+                    temp_size = int(params_size)
+                    if temp_size <=0:
+                        raise exceptions.InvalidSyntax("'%s' must be positive but was %d" % (params_key, params_size))
+                except ValueError:
+                    raise exceptions.InvalidSyntax("'%s' must be numeric"% (params_key) )
+            else:
+                temp_size = if_with_unit[0][0]
+                temp_unit = if_with_unit[0][1].upper()
+                if temp_unit == "GB":
+                    temp_size = int(float(temp_size) * 1024 * 1024 * 1024)
```
danielmitterdorfer: you can use `convert.bytes_to_gb()` here (and similarly for MB and KB).

esrally/track/params.py
```diff
@@ -197,25 +198,39 @@ def __init__(self, indices, params):
...
+                elif temp_unit == "MB":
+                    temp_size = int(float(temp_size) * 1024 * 1024)
+                elif temp_unit == "KB":
+                    temp_size = int(float(temp_size) * 1024)
+                else:
+                    raise AttributeError("'%s' unit '%s' must be KB/MB/GB" % (params_size, temp_size))
```
danielmitterdorfer: I see that you started to use Python standard exceptions that are not subclassed from `exceptions.RallyError`. While I think this is a good idea (as your exception type indicates the cause more specifically), Rally will behave differently when such an error is thrown. I suggest we have a catch-all error handler where the parser is called and wrap the thrown error in an `InvalidSyntaxError`.

esrally/track/params.py
```diff
@@ -197,25 +198,39 @@ def __init__(self, indices, params):
...
+            return temp_size
+
+        for key in ["bulk-size", "batch_size"]:
+            if key not in params:
+                raise KeyError("Mandatory parameter '%s' is missing" % (key))
+        self.bulk_size = handler_size_number(params["bulk-size"], "bulk-size")
```
danielmitterdorfer: Hmm, suppose the user has specified a bulk size of "500 KB". `handler_size_number()` will return 500 * 1.024 = 512.000 (bytes). But the code down the line (specifically `params.IndexDataReader` and `runner.BulkIndex` has no clue that it should handle bytes now. So what will happen instead is that `params.IndexDataReader` will read 512.000 lines from the source file and `params.IndexDataReader` will report a bulk size of 512.000 docs. So we need to represent the unit too + enable `params.IndexDataReader` to read lines until the limit is exhausted. 

Additionally, you need to adapt `params.Slice` which implements the skipping logic (based on lines currently). A little bit of background on this skipping logic: If a user specifies that N clients should be used for bulk-indexing (where N > 1), Rally will create N sub-processes and each sub-process will read one-Nth of the source file. The skipping logic with the file offset table is necessary because all clients should start roughly at the same time to maximize load on the cluster and my experiments have shown that without this skipping logic it takes clients up to a minute to skip to the relevant slice of a large file (even on a very fast SSD!). I hope it makes sense now with this description.

Finally, you should never reach this point in the code due to a JSON schema validation error because `bulk-size` is declared as `integer` in `resources/track-schema.json`. I guess we need to remove the specification of `bulk-size` entirely there and reimplement the necessary constraint checks (only positive integer values or number + "B/KB/MB/GB" values are allowed). Btw, I think we should not support GB bulk sizes as it makes no practical sense.

esrally/track/params.py
```diff
@@ -197,25 +198,39 @@ def __init__(self, indices, params):
...
+        self.batch_size = handler_size_number(params["batch_size"], "batch_size")
```
danielmitterdorfer: Some of my comments on `self.bulk_size` above apply here similarly.

danielmitterdorfer:
@jomenxiao Are you still interested in doing this PR?

danielmitterdorfer:
I'm closing this PR now due to lack of feedback. If you want to pick it up again, please just reopen.

PR #198, opened 2017-01-13
Fixing typos in track.json example

sstults:
A couple of changes were necessary to get the example track.json to work:

* Move the meta values up a level
* Delete a wild comma that appeared
* Rename "operation" to "operation-type" a couple of times in "operations" block.


danielmitterdorfer:
Thanks for your corrections @sstults, much appreciated! I'll merge.

PR #232, opened 2017-02-19
Add python dev as requirement too

ebuildy:
Following install tutorial from a fresh Debian, had to install python3-dev as well to fix ``psutil/_psutil_linux.c:12:20: fatal error: Python.h: No such file or directory ``error when running ``pip3 install esrally``


danielmitterdorfer:
Thanks for your PR @ebuildy! I'll merge soon.

PR #250, opened 2017-03-08
Saving tournament report [Closes #249]

### LINKED ISSUE #249: Saving tournament report
Hello,

It would be great if we could save the tournament report in markdown file. This command is not working `esrally compare --baseline=XX --contender=XX --report-file=tournament.md`

keyur9:
Added parameters in compare parsers to enable saving tournament report in csv or md format.

```
compare_parser.add_argument(
        "--report-format",
        help="define the output format for the command line report (default: markdown).",
        choices=["markdown", "csv"],
        default="markdown")
compare_parser.add_argument(
        "--report-file",
        help="write the command line report also to the provided file",
        default="")

```

### DIFF
--- a/esrally/reporter.py
+++ b/esrally/reporter.py
@@ -34,7 +34,7 @@ def print_internal(message):
 def print_header(message):
     print_internal(console.format.bold(message))
 
-def write_single_report(report_file, report_format, cwd, headers, data, write_header=True, show_also_in_console=True):
+def write_single_report(report_file, report_format, cwd, headers, data_plain, data_rich, write_header=True, show_also_in_console=True):
     
     if report_format == "markdown":
         formatter = format_as_markdown
@@ -44,7 +44,7 @@ def write_single_report(report_file, report_format, cwd, headers, data, write_he
         raise exceptions.SystemSetupError("Unknown report format '%s'" % report_format)
 
     if show_also_in_console:
-        print_internal(formatter(headers, data))
+        print_internal(formatter(headers, data_rich))
     if len(report_file) > 0:
         normalized_report_file = rio.normalize_path(report_file, cwd)
         logger.info("Writing report to [%s] (user specified: [%s]) in format [%s]" %
@@ -52,7 +52,7 @@ def write_single_report(report_file, report_format, cwd, headers, data, write_he
         # ensure that the parent folder already exists when we try to write the file...
         rio.ensure_dir(rio.dirname(normalized_report_file))
         with open(normalized_report_file, mode="a+", encoding="UTF-8") as f:
-            f.writelines(formatter(headers, data, write_header))
+            f.writelines(formatter(headers, data_plain, write_header))
 
 def format_as_markdown(headers, data, write_header=True):
     rendered = tabulate.tabulate(data, headers=headers, tablefmt="pipe", numalign="right", stralign="right")
@@ -270,10 +270,10 @@ class SummaryReporter:
         report_file = self._config.opts("reporting", "output.path")
         report_format = self._config.opts("reporting", "format")
         cwd = self._config.opts("node", "rally.cwd")
-        write_single_report(report_file, report_format, cwd, headers=["Lap", "Metric", "Operation", "Value", "Unit"], data=metrics_table,
-                                 write_header=self.needs_header())
+        write_single_report(report_file, report_format, cwd, headers=["Lap", "Metric", "Operation", "Value", "Unit"], data_plain=metrics_table,
+                                 data_rich = metrics_table, write_header=self.needs_header())
         if self.is_final_report() and len(report_file) > 0:
-            write_single_report("%s.meta" % report_file, headers=["Name", "Value"], data=meta_info_table, show_also_in_console=False)
+            write_single_report("%s.meta" % report_file, headers=["Name", "Value"], data=meta_info_table, data_rich = meta_info_table, show_also_in_console=False)
 
     def report_throughput(self, stats, operation):
         min, median, max, unit = stats.op_metrics[operation.name]["throughput"]
@@ -418,6 +418,7 @@ class ComparisonReporter:
 
         metric_table_plain = self.metrics_table(baseline_stats, contender_stats, plain = True)
         metric_table_rich = self.metrics_table(baseline_stats, contender_stats, plain = False)
+        # Writes metric_table_rich to console, writes metric_table_plain to file
         self.write_report(metric_table_plain,metric_table_rich)
 
     def metrics_table(self, baseline_stats, contender_stats, plain):
@@ -449,9 +450,7 @@ class ComparisonReporter:
         report_format = self._config.opts("reporting", "format")
         cwd = self._config.opts("node", "rally.cwd")
         write_single_report(report_file, report_format, cwd, headers=["Metric", "Operation", "Baseline", "Contender", "Diff", "Unit"], 
-            data= metrics_table, write_header=True)
-        write_single_report(report_file, report_format, cwd, headers=["Metric", "Operation", "Baseline", "Contender", "Diff", "Unit"], 
-            data= metrics_table_console, write_header=True)
+            data_plain= metrics_table, data_rich= metrics_table_console, write_header=True)
 
     def report_throughput(self, baseline_stats, contender_stats, operation):
         b_min, b_median, b_max, b_unit = baseline_stats.op_metrics[operation]["throughput"]
--- a/esrally/reporter.py
+++ b/esrally/reporter.py
@@ -419,6 +419,7 @@ class ComparisonReporter:
         print_internal("")
 
         print_internal(self.format_as_table(self.metrics_table(baseline_stats, contender_stats)))
+        self.write_report(metrics_table())
 
     def format_as_table(self, table):
         return tabulate.tabulate(table,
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Thanks for the PR @keyur9. I had a quick look. It's a good start but you also need to use these parameters, i.e. `ComparisonReporter` (in `reporter.py`) needs to write the report file (see https://github.com/elastic/rally/blob/master/esrally/reporter.py#L242 how it is done in the `SummaryReporter` that is used for normal reports).

keyur9:
Thank you for the feedback @danielmitterdorfer. As suggested, have updated the `ComparisonReporter `(wrote the report file similar to how its done in the `SummaryReporter`) in `reporter.py`.

danielmitterdorfer:
Thanks for the follow-up. Could you please refactor it in a way that both reporters call just one implementation? I'd just implement `write_single_report` as a function and have then both reporter implementations call it.

keyur9:
Thank you for the quick response. Sure, I'll work on refactoring and complete it before end of the week.

keyur9:
Hello @danielmitterdorfer, as suggested I've implemented `write_single_report` as a function and call the same function from both reporter. I'd appreciate your feedback / suggestions if there is any way I could refactor the code.

esrally/reporter.py
```diff
@@ -34,6 +34,26 @@ def print_internal(message):
 def print_header(message):
     print_internal(console.format.bold(message))
 
+def write_single_report(self, report_file, headers, data, write_header=True, show_also_in_console=True):
```
danielmitterdorfer: `self` is an (implicit) reference to the current object. As you refactored it to be a function you need to remove `self` here and also move all related helper functions (like `format_as_markdown`, `format_as_csv`).

danielmitterdorfer: Nit: There should be 2 blank lines.

esrally/reporter.py
```diff
@@ -34,6 +34,26 @@ def print_internal(message):
...
+    report_format = self._config.opts("reporting", "format")
```
danielmitterdorfer: You need to pass the report format as a parameter instead of using the config here.

esrally/reporter.py
```diff
@@ -34,6 +34,26 @@ def print_internal(message):
...
+        cwd = self._config.opts("node", "rally.cwd")
+        normalized_report_file = rio.normalize_path(report_file, cwd)
+        logger.info("Writing report to [%s] (user specified: [%s]) in format [%s]" %
+                    (normalized_report_file, report_file, report_format))
+        # ensure that the parent folder already exists when we try to write the file...
+        rio.ensure_dir(rio.dirname(normalized_report_file))
+        with open(normalized_report_file, mode="a+", encoding="UTF-8") as f:
+            f.writelines(formatter(headers, data, write_header))
 
 class Stats:
```
danielmitterdorfer: Nit: There should be 2 blank lines.

keyur9:
Hello @danielmitterdorfer , Thank you for all the feedback. I've fixed all those issue and both implementations are working as expected. However, I faced one issue while generating Tournament report file. The output of difference between baseline and contender is shown in color in console and the diff method returns the difference in color. But when writing the color output in file it is getting converted to something like this `\[31;1m+1.49723\[0m`.  So, I've changed the code that will not print the colored output in console as it will give appropriate value in report file. Would love to hear your thoughts on it. 
Attached output [Console colored: ON](https://github.com/keyur9/Benchmarking-ES/blob/master/output_when_colored.md) and [Console colored: OFF](https://github.com/keyur9/Benchmarking-ES/blob/master/output_when_not_colored.md)

danielmitterdorfer:
@I'd really love to keep the colors in the command line output. How about you generate the report twice, once for console consumption (with the ANSI escape sequence) and once for file output (without the ANSI escape sequences). Generating the report is quite fast and I wouldn't bother doing it twice.

Something like this:

```python
class ComparisonReporter:
    def report(self, r1, r2):
        # ...
        metric_table_plain = self.metrics_table(baseline_stats, contender_stats, plain=True)
        metric_table_rich = self.metrics_table(baseline_stats, contender_stats, plain=False)
        # writes metric_table_rich to console, writes metric_table_plain to file
        self.write_report(metric_table_plain, metric_table_rich)

    def metrics_table(self, baseline_stats, contender_stats, plain):
        self.plain=plain
        metrics_table = []
        # ...
        return metrics_table

    def diff(self, baseline, contender, treat_increase_as_improvement, formatter=lambda x: x):
        diff = formatter(contender - baseline)
        if self.plain:
            color_greater = lambda x: x
            color_smaller = lambda x: x
            color_neutral = lambda x: x
        elif treat_increase_as_improvement:
            color_greater = console.format.green
            color_smaller = console.format.red
            color_neutral = console.format.neutral
        else:
            color_greater = console.format.red
            color_smaller = console.format.green
            color_neutral = console.format.neutral

        if diff > 0:
            return color_greater("+%.5f" % diff)
        elif diff < 0:
            return color_smaller("%.5f" % diff)
        else:
            # tabulate needs this to align all values correctly
            return color_neutral("%.5f" % diff)
```

You just need to add a new parameter to the function that actually writes the report data and write  `metric_table_plain` to the output file and `metric_table_rich` to console. For the summary reporter, you can just provide the same metric table twice (as there are no ANSI escape sequences). Apart from that your change looks good. What do you think?

keyur9:
Hello @danielmitterdorfer, Thanks for the feedback. Generating the report twice didn't took long and have made the changes. Now we would be able to see the console output in color and save the tournament report in md and csv file.

danielmitterdorfer:
Thanks for following-up @keyur9! I'll try to review it later today.

danielmitterdorfer:
I did a few tests and unfortunately, a few of them failed. Can you please have a look @keyur9?

Command:

```
esrally --distribution-version=5.3.0 --test-mode --report-format=csv --laps=2 --report-file=/Users/daniel/Desktop/report.csv
```

Trace:

```
  File "/Users/daniel/Projects/rally/esrally/racecontrol.py", line 303, in run
    pipeline(cfg)
  File "/Users/daniel/Projects/rally/esrally/racecontrol.py", line 42, in __call__
    self.target(cfg)
  File "/Users/daniel/Projects/rally/esrally/racecontrol.py", line 244, in from_distribution
    return race(Benchmark(cfg, distribution=True))
  File "/Users/daniel/Projects/rally/esrally/racecontrol.py", line 216, in race
    benchmark.teardown(cancelled)
  File "/Users/daniel/Projects/rally/esrally/racecontrol.py", line 160, in teardown
    reporter.summarize(self.race_store, self.metrics_store, self.cfg, self.track)
  File "/Users/daniel/Projects/rally/esrally/reporter.py", line 15, in summarize
    SummaryReporter(race_store, metrics_store, cfg, lap).report(track)
  File "/Users/daniel/Projects/rally/esrally/reporter.py", line 267, in report
    self.write_report(metrics_table, meta_info_table)
  File "/Users/daniel/Projects/rally/esrally/reporter.py", line 276, in write_report
    write_single_report("%s.meta" % report_file, headers=["Name", "Value"], data=meta_info_table, show_also_in_console=False)
TypeError: write_single_report() missing 2 required positional arguments: 'report_format' and 'cwd'
```

Command (obviously you need to choose two different timestamp depending on the output of `esrally list races`):

```
esrally compare --baseline=20170331T081942Z --contender=20170331T082445Z
```

* Expected result: The report is printed once on the console using colors. No file is written.
* Actual result: The report is printed twice: Once without colors, once with colors


Command:

```
esrally compare --baseline=20170331T081942Z --contender=20170331T082445Z --report-format=csv --report-file=/Users/daniel/Desktop/comparison.csv
```

* Expected result: The report is printed once on the console using colors. The file `/Users/daniel/Desktop/comparison.csv` is written and contains the same data but without any ANSI escape sequences.
* Actual result: The report is printed twice: Once without colors, once with colors. Also the file contents are written twice, once with ANSI escape sequences and once without.

keyur9:
Hello @danielmitterdorfer. Thank you for the feedback. I have fixed the issues and now report is printed once in console and written once to file without ANSI escape sequence.

Now the output for the command - 
```
esrally compare --baseline=20170331T081942Z --contender=20170331T082445Z
```

- Expected result: The report is printed once on the console using colors. No file is written.
- Actual result: The report is printed once on the console using colors. No file is written.

```
esrally compare --baseline=20170331T081942Z --contender=20170331T082445Z --report-format=csv --report-file=/Users/daniel/Desktop/comparison.csv
```

- Expected result: The report is printed once on the console using colors. The file /Users/daniel/Desktop/comparison.csv is written and contains the same data but without any ANSI escape sequences.
- Actual result: The report is printed once on the console using colors. The file /Users/daniel/Desktop/comparison.csv is written and contains the same data but without any ANSI escape sequences.

Please let me know if you're fine with changes.

danielmitterdorfer:
@keyur9 We're almost there. The only test case that failed for me was:

```
esrally --distribution-version=5.3.0 --test-mode --report-format=csv --laps=2 --report-file=/Users/daniel/Desktop/report.csv
```

Rally writes not only the actual report but also a second file containing the git commit hash of the Elasticsearch version that it has benchmarked. When Rally attempts to write it with your modified version, we get (trace is shortened):

```
  File "/Users/daniel/Projects/rally/esrally/reporter.py", line 276, in write_report
    write_single_report("%s.meta" % report_file, headers=["Name", "Value"], data=meta_info_table, data_rich = meta_info_table, show_also_in_console=False)
TypeError: write_single_report() got an unexpected keyword argument 'data'
```

The keyword argument is now `data_plain` instead of `data` and you also forgot to add the new arguments `report_format` and `cwd`. If you change line 276 instead to 

```python
write_single_report("%s.meta" % report_file, report_format, cwd, headers=["Name", "Value"], data_plain=meta_info_table, data_rich=meta_info_table, show_also_in_console=False)
```
everything works fine.

I think this is the last change to get this PR in.

So you need

keyur9:
Hello @danielmitterdorfer, Thank you for the feedback. I added both the missing parameters and changed the keyword to `data_plain`.

Please let me know if you're fine with those changes.

PR #272, opened 2017-04-19
Mappings should be created on a single call when the index is created

jimczi:
Mappings for a race are always known beforehand so it is possible to create the index with the final mapping rather than calling `put_mapping` right after the creation of the index.
This change creates the full mapping and adds it to the call to create_index when the index for a race is created for the first time.
This is required if we want to benchmark index sorting which requires to create the mapping when the index is created in order to apply the correct sort specification extracted from the index settings.


danielmitterdorfer:
Thanks for the PR @jimczi. I'll review it soon.

jimczi:
Thanks @danielmitterdorfer !

PR #273, opened 2017-04-21
Add support for max_num_segments in forcemerge action

jimczi:
This change adds support to provide the maximum number of segments an index should be merged to when using the forcemerge action.
It can be useful to force merge an index to 1 segment in a challenge in order to be able to compare index size between
previous runs. Merging down to 1 segment would ensure that different runs would always have the same number of segments.


jimczi:
Thanks @danielmitterdorfer !

PR #286, opened 2017-06-05
Additional metrics for bulk requests

cdahlqvist:
When comparing bulk indexing throughput for different types of data, pure EPS can be very misleading if events have different sizes. I have found the total volume of JSON records indexed per time period to sometimes be a more useful measure, as this takes event size into consideration.

This pull request adds the total size of the bulk request as well as the total size of the documents indexed to the metrics reported, allowing the amount of JSON indexed per timeframe and total volume sent over the wire to be calculated.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -255,41 +255,41 @@ class BulkIndex(Runner):
             raise exceptions.DataError(
                 "Bulk parameter source did not provide a 'bulk-size' parameter. Please add it to your parameter source.")
 
-        bulk_request_size_bytes = 0
-        total_document_size_bytes = 0
-
-        for i in range(len(params["body"])):
-            if params["action_metadata_present"]:
-                if i % 2 == 1:
-                    total_document_size_bytes += len(params["body"][i])
-            else:
-                total_document_size_bytes += len(params["body"][i])
-
-            bulk_request_size_bytes += len(params["body"][i])
-
         if with_action_metadata:
             # only half of the lines are documents
             response = es.bulk(body=params["body"], params=bulk_params)
         else:
             response = es.bulk(body=params["body"], index=index, doc_type=params["type"], params=bulk_params)
 
-        stats = self.detailed_stats(bulk_size, response) if detailed_results else self.simple_stats(bulk_size, response)
+        stats = self.detailed_stats(params, bulk_size, response) if detailed_results else self.simple_stats(bulk_size, response)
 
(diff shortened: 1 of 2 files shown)


cdahlqvist:
Have updated my git repository to correct the CLA issue.

esrally/driver/runner.py
```diff
@@ -245,6 +255,18 @@ def __call__(self, es, params):
             raise exceptions.DataError(
                 "Bulk parameter source did not provide a 'bulk-size' parameter. Please add it to your parameter source.")
 
+        bulk_request_size_bytes = 0
+        total_document_size_bytes = 0
+
+        for i in range(len(params["body"])):
```
danielmitterdorfer: This adds some overhead which is - depending on the track it's - at worst in the single-digit percentage range (I've measured up to 95ms). I think it would make sense to move this loop to the method `detailed_stats`. Users can then enable it for their tracks by setting `detailed-results` to `true. Complete example:

```json
{
  "name": "bulk-index",
  "operation-type": "index",
  "bulk-size": 5000,
  "detailed-results": true
}
```

Btw, you could also use [`enumerate`](https://docs.python.org/3.6/library/functions.html#enumerate) for the loop instead of range to simplify it a bit:

```python
for line_number, data in enumerate(params["body"]):
    line_size = len(data)
    if params["action_metadata_present"]:
        total_document_size_bytes += line_size
    #....
```

Another fine detail: the `len` function in Python returns the number of characters (although it's not really clearly documented) but combining the following two facts show that this is the case:

1. "[len] [r]eturn[s] the length (the number of items) of an object. The argument may be a sequence (such as a string [...])" ([source](https://docs.python.org/3.6/library/functions.html#len))
2. "Strings are immutable sequences of Unicode code points." ([source](https://docs.python.org/3.6/library/stdtypes.html#textseq))

Therefore, we can reason that `len` returns the number of Unicode code points (i.e. characters) of a string. For characters that can be represented with 1 byte, the number of characters is identical to the number of bytes but strictly speaking it's the number of characters. To determine the number of bytes you need to encode the string as UTF-8, e.g. `len(s.encode('utf-8'))`.

cdahlqvist:
@danielmitterdorfer Have updated the PR according to given suggestions.

cdahlqvist:
Added the assertions to the test.

danielmitterdorfer:
LGTM. You can merge at any time.

Issue #71, opened 2016-03-31
Allow to benchmark multi-machine clusters


danielmitterdorfer:
Currently, it is only possible to benchmark on a single machine (except for the special case where we run with `--pipeline=benchmark-only` which (a) puts the burden of provisioning on the user and (b) does not gather system metrics (like CPU usage or index size). Rally should be able to run benchmarks also on clusters across multiple machines.

In this ticket we want to collect the high-level ideas around making this possible. Work should be done in smaller, more focused tickets. At least these areas need to change:
- Provisioning (consider also #42 and #117)
- Gathering metrics (we need to gather them on the machines that are being benchmarked)
- Reporting: We currently report system level metrics only once (e.g. CPU usage) and need to reconsider that (although we will prepare for that case already in #54)

_Note: Distribution of the load generator is handled separately in #257._

mynameismevin:
What's the status of this issue?

danielmitterdorfer:
It's in the early concept stage. We use Github's milestone feature and it's planned for 0.5.0 (no release date). 

The current "workaround" is to use the pipeline "benchmark-only" as I've mentioned in the description of this ticket provided you are able to apply enough load with a single Rally instance. I'd expect that you can apply a bit more load when #108 is ready on which I am currently working on.

mynameismevin:
Gotcha, thanks!

danielmitterdorfer:
We'll implement this gradually and implement support for single-machine clusters in #184 first. Then we can extend Rally to multi-machine clusters. This will reduce the risk of adding too many changes at once (as we then also need an ability to define cluster topology).

danielmitterdorfer:
For the first implementation we'll assume that every node has the same configuration (i.e. there is no possibility to define dedicated node roles (master, data, coordinator, ...) and different node configurations). We will allow multiple nodes per host though.

danielmitterdorfer:
Having Rally set-up multi-node clusters can be done now by specifying the respective nodes with `--target-hosts`. It is possible to use multiple nodes per host by specifying the same IP/port pair twice.

Examples:

* `--target-hosts=127.0.0.1:39200` will set up a single ES node on the local machine which runs on port 39200.
* `--target-hosts=127.0.0.1:39200,127.0.0.1:39200` will set up two ES nodes on the local machine.
[...]

PR #299, opened 2017-07-10
Leave the host parsing on elasticsearch-py

HonzaKral:
That way you can specify 'https://user:secret@localhost:9200' if needed

There is already a comprehensive parsing logic in `elasticsearch-py` (0), no need to reimplement.

0 - https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/__init__.py#L19-L59

### DIFF
--- a/esrally/mechanic/mechanic.py
+++ b/esrally/mechanic/mechanic.py
@@ -143,8 +143,12 @@ class MechanicActor(actor.RallyActor):
                     if len(hosts) == 0:
                         raise exceptions.LaunchError("No target hosts are configured.")
                     for host in hosts:
-                        host_or_ip = host["host"]
-                        port = int(host["port"])
+                        host = host.copy()
+                        host_or_ip = host.pop("host")
+                        port = host.pop("port", 9200)
+                        if host:
+                            raise exceptions.SystemSetupError("When specifying nodes to be managed by rally you can only supply hostname:port "
+                                                              "pairs (e.g. 'localhost:9200'), any additional options cannot be supported.")
                         # user may specify "localhost" on the command line but the problem is that we auto-register the actor system
                         # with "ip": "127.0.0.1" so we convert this special case automatically. In all other cases the user needs to
                         # start the actor system on the other host and is aware that the parameter for the actor system and the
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:
Thanks for PR! There is only one problem here: You can run Rally in different modes, e.g. against an existing cluster or you can have Rally setup the cluster for you (see [docs](http://esrally.readthedocs.io/en/latest/recipes.html#benchmarking-a-remote-cluster)). In the latter case, Rally will use the same command line parameter, iterate over the target hosts and contact a remote process on that host in order to tell it to spin up a node (see https://github.com/elastic/rally/blob/master/esrally/mechanic/mechanic.py#L141-L181). And for that we need the host/IP already in a separate field. While we could (theoretically) reuse the logic of elasticsearch-py in this component, it seems to me that this function is not meant to be used outside of elasticsearch-py (and rightly so IMHO).

So while I'd really love to get rid of this parsing logic, I fear it's not doable at the moment. Once we allow more complex multi-machine setups (with #71), we need to specify the cluster nodes differently anyway because maybe not every node that we need to setup can act as a coordinating node / client node. IMHO then would be the right time to merge this PR. 

I'd leave it open for now but mark it as "blocked" until #71 is resolved. Wdyt?

HonzaKral:
I see no reason to avoid using the `_normalize_hosts` function from `elasticsearch-py`. It produces compatible output and is more robust in its parsing by using actual `urlparse` functionality. The linked code in `mechanic` can still work. The only change, if at all, might be to make sure no extra options are passed in to the `localhost` config. Like username/password etc which mechanic wouldn't know what to do with. That sounds like an easier fix while still allowing more functionality (would be happy to provide the additional code).

I do see your point though and it is a judgment call which means it's all up to you to decide what is more consistent with the rest of the project.

Thank you for your quick and thoughtful reply!

danielmitterdorfer:
I was reluctant to use it as it's not part of the public API. But then let's do it this way.

> The only change, if at all, might be to make sure no extra options are passed in to the localhost config.

I think it's fine to just use the `hosts` dict as it's returned by `_normalize_hosts`. Could you just add the necessary change in `mechanic`? Then I think we're good to merge. :)

HonzaKral:
I added a check in `mechanic` to verify that no advanced options were being passed in via the `hosts` settings. That should make this safe to merge, let me know if there is anything I have missed. Thanks!

Issue #413, opened 2018-02-09
Cannot report to ES cluster with self-signed certificate or in-house certificate authority


kesslerm:
The metrics module fails to connect to the Elasticsearch metrics store, if secure transport is used and the server has a self signed certificate or a certificate signed by a certificate authority that's not in the certificate store provided by `certifi`.

danielmitterdorfer:
We probably need to enhance the metrics store configuration to allow passing these options. I would not go so far as to ask the user for a custom certificate in Rally's setup routine but we should probably allow to configure a custom certificate directly in the config file.

PR #414, opened 2018-02-09
Fix connection to servers with self-signed certificates or unknown certificate authority

kesslerm:
The metrics module fails to connect to the Elasticsearch metrics store,
if secure transport is used and the server has a self signed certificate
or a certificate signed by a certificate authority that's not in the
certificate store provided by `certifi`.

+ Introduce a non-mandatory setting that allows turning off certificate verification.
+ Introduce a non-mandatroy setting to override the ca_certs certificate store.

Fixes https://github.com/elastic/rally/issues/413


esrally/metrics.py
```diff
@@ -130,6 +130,11 @@ def __init__(self, cfg):
         secure = self._config.opts("reporting", "datastore.secure") == "True"
         user = self._config.opts("reporting", "datastore.user")
         password = self._config.opts("reporting", "datastore.password")
+        # poor man's boolean conversion
+        verify_certs = self._config.opts("reporting", "datastore.verify_certs", default_value="True", mandatory=False) == "True"
```
danielmitterdorfer: I'd rather use `datastore.certs.verify` as name (in the config file we usually use dots as a separator).

esrally/metrics.py
```diff
@@ -130,6 +130,11 @@ def __init__(self, cfg):
...
+        ca_certs = self._config.opts("reporting", "datastore.ca_certs", default_value=None, mandatory=False)
```
danielmitterdorfer: I'd rather use `datastore.certs.path` as a name.

docs/configuration.rst
```diff
@@ -124,7 +124,7 @@ Rally will ask you a few more things in the advanced setup:
...
 * **Metrics store type**: You can choose between ``in-memory`` which requires no additional setup or ``elasticsearch`` which requires that you start a dedicated Elasticsearch instance to store metrics but gives you much more flexibility to analyse results.
-* **Metrics store settings** (only for metrics store type ``elasticsearch``): Provide the connection details to the Elasticsearch metrics store. This should be an instance that you use just for Rally but it can be a rather small one. A single node cluster with default setting should do it.
+* **Metrics store settings** (only for metrics store type ``elasticsearch``): Provide the connection details to the Elasticsearch metrics store. This should be an instance that you use just for Rally but it can be a rather small one. A single node cluster with default setting should do it. When using self-signed certificates on the Elasticsearch metrics store, certificate verification can be turned off by setting the `datastore.ssl.verification_mode` setting to `none`. Alternatively you can enter the path to the certificate authority's signing certificate in `datastore.ssl.certificate_authorities`. Both settings are optional.
```
danielmitterdorfer: One minor thing: Contrary to Markdown, reStructuredText requires two backticks for literal text. So e.g. instead of:

```
`datastore.ssl.verification_mode`
```

you need to write

```
``datastore.ssl.verification_mode``
```

PR #430, opened 2018-03-08
Allow to use track parameters in index (templates)

### LINKED ISSUE #429: Allow to use track parameters also in index / template definitions
Currently index definitions are read as is. However, it can be useful to allow track parameters also for index definition files. For consistency, we should allow the same for index templates.

Technically, this means that we will treat index definition files also as Jinja templates.

danielmitterdorfer:
Closes #429

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -757,9 +757,10 @@ class TrackSpecificationReader:
     Creates a track instances based on its parsed JSON description.
     """
 
-    def __init__(self, override_auto_manage_indices=None, source=io.FileSource):
+    def __init__(self, override_auto_manage_indices=None, track_params=None, source=io.FileSource):
         self.name = None
         self.override_auto_manage_indices = override_auto_manage_indices
+        self.track_params = track_params if track_params else {}
         self.source = source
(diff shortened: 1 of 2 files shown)


esrally/track/loader.py
```diff
@@ -757,9 +757,10 @@ class TrackSpecificationReader:
     Creates a track instances based on its parsed JSON description.
     """
 
-    def __init__(self, override_auto_manage_indices=None, source=io.FileSource):
+    def __init__(self, override_auto_manage_indices=None, track_params=None, source=io.FileSource):
         self.name = None
         self.override_auto_manage_indices = override_auto_manage_indices
+        self.track_params = track_params if track_params else {}
```
dliappis: Completely minor question, but could this be replaced by `self.track_params = track_params` if the above named parameter had the default value set to `track_params = {}`?

danielmitterdorfer: No, this is not possible because the default argument is mutable. This is a [common gotcha](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments).

PR #437, opened 2018-03-10
Bump pytest* versions and beautify make test output

dliappis:
Take advantage of the colored and verbose output options of newer
versions of pytest to make the output of `make test`/`make
coverage`/`make benchmark` easier to follow.

Also bump pytest to the latest version to support progress
percentage[1]. While at it, also bump pytest-benchmark to the latest
version.

With these changes `make test` looks like:

image

as opposed to the current output looking like:

image

[1]
https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,6 @@
+[aliases]
+test=pytest
+
+[tool:pytest]
+addopts = --verbose --color=yes
+testpaths = tests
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -0,0 +1,6 @@
+[aliases]
+test=pytest
+
+[tool:pytest]
+addopts = --verbose --color=yes
+testpaths = tests
```
dliappis: This ensures `benchmark` tests don't get execute with the regular `make test`.

PR #438, opened 2018-03-13
Create legacy corpora only if needed

danielmitterdorfer:
With this commit we create legacy corpora definitions only if they are
really needed. If a track does not define any data set these are not
needed and hence we will also not issue a warning which might confuse
users.


danielmitterdorfer:
Note that I did not set a milestone because the corresponding code will be removed anyway with 0.10.0. However, meanwhile we should not show unnecessary warnings for users on Rally master.

PR #440, opened 2018-03-13
Use curl retry features to workaround transient network problems

dliappis:
curl after a certain version allows a number of --retry options that
can improve resiliency against transient network issues.

Use curl retry features and bail out if the system doesn't honor the
minimum curl version.

Ideally we'd also use `--retry-connrefused` but this parameter got
introduced in curl 7.52 which isn't available yet on many
distributions such as Ubuntu 16.04.

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -13,6 +13,8 @@ readonly DISTRIBUTIONS=(2.4.6 5.6.7)
 readonly TRACKS=(geonames nyc_taxis http_logs nested)
 
 readonly ES_METRICS_STORE_VERSION="6.2.1"
+readonly ES_ARTIFACT_PATH="elasticsearch-${ES_METRICS_STORE_VERSION}"
+readonly ES_ARTIFACT="${ES_ARTIFACT_PATH}.tar.gz"
 readonly MIN_CURL_VERSION=(7 12 3)
 
 ES_PID=-1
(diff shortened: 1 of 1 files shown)


integration-test.sh
```diff
@@ -13,9 +13,23 @@ readonly DISTRIBUTIONS=(2.4.6 5.6.7)
...
 
 ES_PID=-1
 
+function check_prerequisites() {
+    local curl_major_version=$(curl --version | head -1 | cut -d ' ' -f 2,2 | cut -d '.' -f 1,1)
+    local curl_minor_version=$(curl --version | head -1 | cut -d ' ' -f 2,2 | cut -d '.' -f 2,2)
+    local curl_patch_release=$(curl --version | head -1 | cut -d ' ' -f 2,2 | cut -d '.' -f 3,3)
+
+    if [[ $curl_major_version < ${MIN_CURL_VERSION[0]} ]] || \
+       [[ $curl_major_version == ${MIN_CURL_VERSION[0]} && $curl_minor_version < ${MIN_CURL_VERSION[1]} ]] || \
```
danielmitterdorfer: Hmm, I don't think this will work anymore when curl 8.x will be released in the future? While the check is good, I am also not sure whether we can just avoid this check completely because we have tight control over the environments in which this runs (our CI)?

integration-test.sh
```diff
@@ -85,9 +99,11 @@ function set_up() {
     mkdir -p .rally_it/cache
     cd .rally_it/cache
     if [ ! -f elasticsearch-"${ES_METRICS_STORE_VERSION}".tar.gz ]; then
-        curl -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-"${ES_METRICS_STORE_VERSION}".tar.gz
+        # If curl fails immediately, executing all retries will take up to (2**retries)-1 seconds.
+        curl --retry 8 -O https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-"${ES_METRICS_STORE_VERSION}".tar.gz || { rm elasticsearch-"${ES_METRICS_STORE_VERSION}".tar.gz; exit 1; }
```
danielmitterdorfer: Maybe it makes sense to extract the artefact name (`elasticsearch-"${ES_METRICS_STORE_VERSION}".tar.gz`) as a constant as it is repeated multiple times now?

integration-test.sh
dliappis: It should work (I've tested it too) because we are **failing** only if:

- curl < 7 or
- curl_major == 7 and curl_minor < 12 or
- curl_major == 7 and curl_minor == 12 and curl_patch < 3

everything else passes, including curl >= 8

integration-test.sh
dliappis: Makes sense, will address.

dliappis: I pushed 20972a29aa086e4d5b1b544c8b74bb5c5ecb5011 for ^^

danielmitterdorfer:
Thanks @dliappis! LGTM.

integration-test.sh
danielmitterdorfer: Ah, I missed that. Good point.

PR #441, opened 2018-03-15
Allow to pass parameters via a file

danielmitterdorfer:
Previously we allowed users to pass track, car and plugin parameters via
a list of key-value pairs. While this covers a lot of use cases, we can
give users even more flexibility by letting them provide a file to
Rally. This allows users to pass also nested structures which is e.g.
handy to specify index settings.

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -6,6 +6,7 @@ import os
 import sys
 import time
 import uuid
+import json
 
 from esrally import version, actor, config, paths, racecontrol, reporter, metrics, track, chart_generator, exceptions, time as rtime
 from esrally import PROGRAM_NAME, DOC_LINK, BANNER, SKULL, check_python_version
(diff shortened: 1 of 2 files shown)


esrally/rally.py
```diff
@@ -587,6 +587,15 @@ def dispatch_sub_command(cfg, sub_command):
         return False
 
 
+def to_dict(arg):
+    if io.has_extension(arg, ".json"):
+        import json
```
dliappis: Minor comment: you probably have your reasons, but why not have this at the top? AFAIK [pep8](https://www.python.org/dev/peps/pep-0008/#imports) is rather strict about this. I read some literature preferring `from json import load` for inside-function uses such as this one, if it must be done.

danielmitterdorfer: I wanted to limit the scope of the import so we don't accidentally use this functionality somewhere else. I'm fine with putting it at the top. Mid-term it might make sense to split `rally.py` though.

PR #446, opened 2018-03-26
Use more specific name for mandatory plugin check selectively

dliappis:
Commit 7a051f337d419f14a1223c2a1514b57a53a34037 reverted the use of a
more specific plugin name for the mandatory plugin
check (e.g. specifying `mandatory.plugin=x-pack` instead of
``mandatory.plugin=x-pack-security`), however, more specific names are
still required for benchmarking release versions <6.3.

Selectively use a more specific name for the mandatory plugin check
depending on the Elasticsearch version.

Also add test cases for both scenarios.

Relates: 7a051f337d419f14a1223c2a1514b57a53a34037
Relates: https://github.com/elastic/elasticsearch/pull/28710

### DIFF
--- a/esrally/mechanic/provisioner.py
+++ b/esrally/mechanic/provisioner.py
@@ -188,14 +188,14 @@ class BareProvisioner:
         for installer in self.plugin_installers:
             # For Elasticsearch < 6.3 more specific plugin names are required for mandatory plugin check
             # Details in: https://github.com/elastic/elasticsearch/pull/28710
+            # TODO: Remove this section with Elasticsearch <6.3 becomes EOL.
             try:
-                if ((versions.major_version(self.distribution_version) == 6 and
-                     versions.minor_version(self.distribution_version) < 3) or
-                    (versions.major_version(self.distribution_version) < 6)):
+                major, minor, _, _ = versions.components(self.distribution_version)
+                if (major == 6 and minor < 3) or major < 6:
                     mandatory_plugins.append(installer.sub_plugin_name)
                 else:
-                    raise NameError
-            except (NameError, TypeError, exceptions.InvalidSyntax):
+                    mandatory_plugins.append(installer.plugin_name)
+            except (TypeError, exceptions.InvalidSyntax):
                 mandatory_plugins.append(installer.plugin_name)
             plugin_variables.update(installer.variables)
 
--- a/esrally/mechanic/provisioner.py
+++ b/esrally/mechanic/provisioner.py
@@ -7,12 +7,13 @@ from enum import Enum
 import jinja2
 
 from esrally import exceptions
-from esrally.utils import io, console, process, modules
+from esrally.utils import io, console, process, modules, versions
 
 logger = logging.getLogger("rally.provisioner")
 
 
 def local_provisioner(cfg, car, plugins, cluster_settings, all_node_ips, target_root, node_id):
+    distribution_version = cfg.opts("mechanic", "distribution.version", mandatory=False)
     ip = cfg.opts("provisioning", "node.ip")
     http_port = cfg.opts("provisioning", "node.http.port")
     node_name_prefix = cfg.opts("provisioning", "node.name.prefix")
@@ -24,7 +25,7 @@ def local_provisioner(cfg, car, plugins, cluster_settings, all_node_ips, target_
     es_installer = ElasticsearchInstaller(car, node_name, node_root_dir, all_node_ips, ip, http_port)
     plugin_installers = [PluginInstaller(plugin) for plugin in plugins]
 
-    return BareProvisioner(cluster_settings, es_installer, plugin_installers, preserve)
+    return BareProvisioner(cluster_settings, es_installer, plugin_installers, preserve, distribution_version=distribution_version)
 
 
 def no_op_provisioner():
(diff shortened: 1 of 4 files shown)


esrally/mechanic/provisioner.py
```diff
@@ -184,7 +186,17 @@ def _provisioner_variables(self):
...
-            mandatory_plugins.append(installer.plugin_name)
+            # For Elasticsearch < 6.3 more specific plugin names are required for mandatory plugin check
+            # Details in: https://github.com/elastic/elasticsearch/pull/28710
+            try:
+                if ((versions.major_version(self.distribution_version) == 6 and
+                     versions.minor_version(self.distribution_version) < 3) or
+                    (versions.major_version(self.distribution_version) < 6)):
+                    mandatory_plugins.append(installer.sub_plugin_name)
+                else:
+                    raise NameError
```
danielmitterdorfer: I don't understand the need to raise `NameError` here.

esrally/mechanic/provisioner.py
```diff
@@ -184,7 +186,17 @@ def _provisioner_variables(self):
         plugin_variables = {}
         mandatory_plugins = []
         for installer in self.plugin_installers:
-            mandatory_plugins.append(installer.plugin_name)
+            # For Elasticsearch < 6.3 more specific plugin names are required for mandatory plugin check
+            # Details in: https://github.com/elastic/elasticsearch/pull/28710
+            try:
```
danielmitterdorfer: I understand that it is a necessity at the moment but I think we should try to be version-agnostic in provisioner code. When earlier 6.x versions are not in use anymore I think we should remove this again to stay version-agnostic.

esrally/mechanic/provisioner.py
```diff
@@ -184,7 +186,17 @@ def _provisioner_variables(self):
...
+                if ((versions.major_version(self.distribution_version) == 6 and
```
danielmitterdorfer: Alternatively you could use `major, minor, _, _ = versions.components(self.distribution_version)`. Then there would also be no need for a new helper function `minor_version` in the `versions` module.

esrally/mechanic/provisioner.py
```diff
@@ -184,7 +186,17 @@ def _provisioner_variables(self):
...
+                     versions.minor_version(self.distribution_version) < 3) or
+                    (versions.major_version(self.distribution_version) < 6)):
+                    mandatory_plugins.append(installer.sub_plugin_name)
+                else:
+                    raise NameError
+            except (NameError, TypeError, exceptions.InvalidSyntax):
```
danielmitterdorfer: I *think* a `TypeError` should not occur at this point? If the individual components of the version string are not `int` they would not match the regex.

esrally/mechanic/provisioner.py
dliappis: I agree with you. I was hesitant to add this conditional here and wasn't 100% sure in which place to define the constraints themselves. Should I add a `TODO:` to remove this when <6.3 is EOL?

danielmitterdorfer: yes, please add a TODO.

esrally/mechanic/provisioner.py
dliappis: This is so that that if the version of Elasticsearch is >=6.3, the non-specific plugin name is applied (inside the exception code) i.e. force the execution inside the exception block.

esrally/mechanic/provisioner.py
dliappis: Yes good suggestion! I felt like simplifying the code inside the `_provisioner_variables()` method with no need to assign temporary variables but there are ways around it and I agree let's not add another helper function.

esrally/mechanic/provisioner.py
danielmitterdorfer: Hmm, but I mean it's not really an error condition if the version is >= 6.3? We could just set the plugin name with `mandatory_plugins.append(installer.plugin_name)` instead of raising an error which seems more straightforward to read to me.

esrally/mechanic/provisioner.py
dliappis: Yup it was added because the integration tests raise this, if not present:

```
[ERROR] Cannot race. expected string or bytes-like object
  File "/home/dl/source/elastic/dliappis_forks/rally/.tox/py36/lib/python3.6/site-packages/esrally/mechanic/mechanic.py", line 516, in receiveMsg_StartNodes
    nodes = self.mechanic.start_engine()
  File "/home/dl/source/elastic/dliappis_forks/rally/.tox/py36/lib/python3.6/site-packages/esrally/mechanic/mechanic.py", line 645, in start_engine
    node_configs.append(p.prepare(binaries))
  File "/home/dl/source/elastic/dliappis_forks/rally/.tox/py36/lib/python3.6/site-packages/esrally/mechanic/provisioner.py", line 164, in prepare
    provisioner_vars = self._provisioner_variables()
  File "/home/dl/source/elastic/dliappis_forks/rally/.tox/py36/lib/python3.6/site-packages/esrally/mechanic/provisioner.py", line 192, in _provisioner_variables
    if ((versions.major_version(self.distribution_version) == 6 and
  File "/home/dl/source/elastic/dliappis_forks/rally/.tox/py36/lib/python3.6/site-packages/esrally/utils/versions.py", line 25, in major_version
    major, _, _, _ = components(version)
  File "/home/dl/source/elastic/dliappis_forks/rally/.tox/py36/lib/python3.6/site-packages/esrally/utils/versions.py", line 49, in components
    matches = versions_pattern.match(version)
TypeError: expected string or bytes-like object

```

dliappis:
@danielmitterdorfer I addressed the comments, could you please have a look again?

PR #449, opened 2018-03-27
Fix venv detection with venv and Rally execution on non master

### LINKED ISSUE #447: Allow first Rally run from non master branch
Current Rally won't run from a source checkout with [disabled automatic updates](http://esrally.readthedocs.io/en/stable/developing.html?highlight=developing#automatic-updates) if the user is running it for the first time.

This is because the execution of [setup.py](https://github.com/elastic/rally/blob/master/run.sh#L65-L68) only happens when updates are enabled.

### LINKED ISSUE #448: Fix pyvenv detection when executing Rally from source
Currently `./rally` is unable to detect virtualenvs created with pyvenv (which is the case when `pyenv` is installed).

The reason is that while [sys.real_prefix](https://github.com/elastic/rally/blob/master/run.sh#L20) is defined with regular Virtualenv's (such as the one created with `python3 -m venv venv` or `virtualenv -p python3 venv`) it is not defined when `pyvenv` is in use. In other words, if a user has installed `pyvenv` locally or globally, which is a requirement for integration tests anyway, in some platforms this takes over when creating virtualenv's even with `python3 -m venv venv`.

<details>
<summary>No sys.real_path in Python3 when pyvenv is in use</summary>

```
$ pyenv shell 3.6.4
$ which python3
~/.pyenv/shims/python3
$ python3 --version
Python 3.6.4
$ python3 -m venv venv
$ source venv/bin/activate
(venv) dl@MS-7885 ~ $ python3
Python 3.6.4 (default, Mar 12 2018, 11:20:13) 
[GCC 7.3.1 20180130 (Red Hat 7.3.1-2)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> sys.real_prefix
  File "<stdin>", line 1, in <module>
AttributeError: module 'sys' has no attribute 'real_prefix'
>>> sys.base_prefix
'/home/dl/.pyenv/versions/3.6.4'
>>> sys.prefix
'/home/dl/venv'
```
</details>

dliappis:
This commit fixes two issues. Firstly, let `./rally` start when
executed for the first time with self update disabled (e.g. from a non
master branch in a fork) by ensuring the `esrally` binary has been
installed with setuptools.

Secondly, fix virtualenv detection when pyenv is in use. On some
configurations when pyenv is present, creating a Python3 virtualenv
uses pyvenv instead of Python3's own virtualenv[1] which fails the
current check for a virtualenv in the `run.sh` script.

[1] https://legacy.python.org/dev/peps/pep-0405/

Closes #447
Closes #448

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -11,13 +11,33 @@
 readonly BINARY_NAME="${__RALLY_INTERNAL_BINARY_NAME}"
 readonly HUMAN_NAME="${__RALLY_INTERNAL_HUMAN_NAME}"
 
+function install_esrally_with_setuptools() {
+    # Check if optional parameter with Rally binary path, points to an existing executable file.
+    if [[ -n $1 ]]; then
+     if [[ -f $1 && -x $1 ]]; then return; fi
+    fi
+
+    if [[ ${IN_VIRTUALENV} == 0 ]]; then
+        python3 setup.py -q develop --user
+    else
+        python3 setup.py -q develop
+    fi
+}
+
 # Attempt to update Rally itself by default but allow user to skip it.
 SELF_UPDATE=YES
 # Assume that the "main remote" is called "origin"
 REMOTE="origin"
 
 # While we could also check via the presence of `VIRTUAL_ENV` this is a bit more reliable.
-python3 -c 'import sys; print(sys.real_prefix)' >/dev/null 2>&1 && IN_VIRTUALENV=1 || IN_VIRTUALENV=0
+# Check for both pyvenv and normal venv environments
+# https://legacy.python.org/dev/peps/pep-0405/
+if python3 -c 'import sys; sys.exit(0) if ".pyenv" in sys.base_prefix else exit(0) if hasattr(sys,"real_path") else exit(1)' >/dev/null 2>&1
+then
+    IN_VIRTUALENV=1
+else
+    IN_VIRTUALENV=0
+fi
 
 # Check for parameters that are intended for this script. Note that they only work if they're specified at the beginning (due to how
 # the shell builtin `shift` works. We could make it work for arbitrary positions but that's not worth the complexity for such an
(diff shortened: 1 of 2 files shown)


run.sh
```diff
@@ -11,13 +11,33 @@
 readonly BINARY_NAME="${__RALLY_INTERNAL_BINARY_NAME}"
 readonly HUMAN_NAME="${__RALLY_INTERNAL_HUMAN_NAME}"
 
+function install_esrally_with_setuptools() {
+    # Check if optional parameter with Rally binary path, points to an existing executable file.
+    if [[ -n $1 ]]; then
+     if [[ -f $1 && -x $1 ]]; then return; fi
```
danielmitterdorfer: Do we need to enclose the parameter in double quotes to avoid issues with spaces in path names? (i.e. `"${1}"`?)

run.sh
```diff
@@ -11,13 +11,33 @@
...
+
 # Attempt to update Rally itself by default but allow user to skip it.
 SELF_UPDATE=YES
 # Assume that the "main remote" is called "origin"
 REMOTE="origin"
 
 # While we could also check via the presence of `VIRTUAL_ENV` this is a bit more reliable.
-python3 -c 'import sys; print(sys.real_prefix)' >/dev/null 2>&1 && IN_VIRTUALENV=1 || IN_VIRTUALENV=0
+# Check for both pyvenv and normal venv environments
+# https://legacy.python.org/dev/peps/pep-0405/
```
danielmitterdorfer: Is there a reason why you linked to the legacy site instead of https://www.python.org/dev/peps/pep-0405/?

run.sh
```diff
@@ -11,13 +11,33 @@
...
+if python3 -c 'import sys; sys.exit(0) if ".pyenv" in sys.base_prefix else exit(0) if hasattr(sys,"real_path") else exit(1)' >/dev/null 2>&1
```
danielmitterdorfer: I think the following would be a bit simpler? 

```python
import sys; exit(0) if ".pyenv" in sys.base_prefix or hasattr(sys,"real_path") else exit(1)
```

run.sh
dliappis: Double quotes are needed only for the backwards compatible `[`. Double square braces do not require [double quoting for the variables](http://mywiki.wooledge.org/BashGuide/TestsAndConditionals#Conditional_Blocks_.28if.2C_test_and_.5B.5B.29).

run.sh
dliappis: Will address

run.sh
dliappis: Changing the logic entirely here as per our discussion.

PR #456, opened 2018-04-05
Remove automatic index management

### LINKED ISSUE #380: Remove automatic index management
Historically, Rally has shipped only with a few core operations. Also, previous Rally versions have hard-coded the index management operations. At the beginning of a benchmark, Rally always did the following operations:

1. Delete and recreate all index template in the `templates` section of the track
2. Delete and recreate all indices that are defined in the `indices` section of the track
3. Check cluster health == green

Although doable, it was hard for users to manage indices themselves (with custom operations).

With the recent work in #293 and #361, users can just specify flexibly what they need and let Rally just execute these operations.

In the course of this ticket we will also remove the deprecated command line option `--cluster-health` which will be replaced by the track parameter mechanism (see #310).

Marked as breaking because users need to adjust their existing tracks so this will work.

danielmitterdorfer:
With this commit we remove the ability that Rally automatically deletes
and creates indices and users need to add explicit tasks now to their
tracks in order to make this work.

The rationale behind all this work is that we want the core of Rally to
be agnostic to the tasks that it executes.

Closes #380

### DIFF
--- a/tests/track/loader_test.py
+++ b/tests/track/loader_test.py
@@ -679,7 +679,7 @@ class TrackPostProcessingTests(TestCase):
             "operations": [
                 {
                     "name": "index-append",
-                    "operation-type": "index",
+                    "operation-type": "bulk",
                     "bulk-size": 5000
                 },
                 {
@@ -752,7 +752,7 @@ class TrackPostProcessingTests(TestCase):
             "operations": [
                 {
                     "name": "index-append",
-                    "operation-type": "index",
+                    "operation-type": "bulk",
                     "bulk-size": 5000
                 },
                 {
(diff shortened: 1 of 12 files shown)


danielmitterdorfer:
This should be reviewed together with https://github.com/elastic/rally-tracks/pull/41 (the change in rally-tracks is compatible with older versions of Rally so the track changes will be merged first).

tests/track/loader_test.py
```diff
@@ -679,7 +679,7 @@ def test_post_processes_track_spec(self):
             "operations": [
                 {
                     "name": "index-append",
-                    "operation-type": "index",
+                    "operation-type": "bulk",
```
dliappis: Yay!

PR #458, opened 2018-04-05
Add parameter support to telemetry devices

danielmitterdorfer:
With this commit we add a new command line flag `--telemetry-params`
which allows users to inject parameters to telemetry devices.

We also enhance the `jfr` telemetry device to choose the recording
template based on such a parameter.

### DIFF
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -122,8 +122,9 @@ class FlightRecorder(TelemetryDevice):
     human_name = "Flight Recorder"
     help = "Enables Java Flight Recorder (requires an Oracle JDK)"
 
-    def __init__(self, log_root, java_major_version):
+    def __init__(self, telemetry_params, log_root, java_major_version):
         super().__init__()
+        self.telemetry_params = telemetry_params
         self.log_root = log_root
         self.java_major_version = java_major_version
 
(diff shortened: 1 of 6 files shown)


esrally/mechanic/telemetry.py
```diff
@@ -142,22 +143,33 @@ def instrument_env(self, car, candidate_id):
...
-        # see http://stackoverflow.com/questions/34882035/how-to-record-allocations-with-jfr-on-command-line
-        #
-        # in that case change to:
-        #
-        # Java 8: -XX:StartFlightRecording=defaultrecording=true,settings=es-memory-profiling
-        # Java 9+: "-XX:StartFlightRecording=maxsize=0,maxage=0s,disk=true,dumponexit=true,settings=es-memory-profiling,filename=%s"
+
+        java_opts = self.java_opts(log_file)
+
+        logger.info("jfr: Adding JVM arguments: [%s].", java_opts)
```
dliappis: Could use the new .format() here.

esrally/mechanic/telemetry.py
```diff
@@ -142,22 +143,33 @@ def instrument_env(self, car, candidate_id):
...
         if self.java_major_version < 9:
-            return {"ES_JAVA_OPTS": "-XX:+UnlockDiagnosticVMOptions -XX:+UnlockCommercialFeatures -XX:+DebugNonSafepoints "
-                                    "-XX:+FlightRecorder "
-                                    "-XX:FlightRecorderOptions=disk=true,maxage=0s,maxsize=0,dumponexit=true,dumponexitpath=%s "
-                                    "-XX:StartFlightRecording=defaultrecording=true" % log_file}
+            java_opts += "-XX:+FlightRecorder "
+            java_opts += "-XX:FlightRecorderOptions=disk=true,maxage=0s,maxsize=0,dumponexit=true,dumponexitpath={} ".format(log_file)
+            java_opts += "-XX:StartFlightRecording=defaultrecording=true"
+            if recording_template:
+                logger.info("jfr: Using recording template [%s].", recording_template)
```
dliappis: Same comment here, could be consistent with format(), as anyway it's user later on.

esrally/mechanic/telemetry.py
```diff
@@ -142,22 +143,33 @@ def instrument_env(self, car, candidate_id):
...
+                java_opts += ",settings={}".format(recording_template)
+            else:
+                logger.info("jfr: Using default recording template.")
         else:
-            return {"ES_JAVA_OPTS": "-XX:+UnlockDiagnosticVMOptions -XX:+UnlockCommercialFeatures -XX:+DebugNonSafepoints "
-                                    "-XX:StartFlightRecording=maxsize=0,maxage=0s,disk=true,dumponexit=true,filename=%s" % log_file}
+            java_opts += "-XX:StartFlightRecording=maxsize=0,maxage=0s,disk=true,dumponexit=true,filename={}".format(log_file)
+            if recording_template:
+                logger.info("jfr: Using recording template [%s].", recording_template)
```
dliappis: .format()? AFAIU we are not in the hot path like `track/params.py`.

esrally/mechanic/telemetry.py
dliappis: Ah silly me. It's the logger. Ignore.

danielmitterdorfer: I think this is a different case here and I wanted to discuss this separately anyway. It looks as if we are using old style string formatting but we are actually not. This is the standard way the [Python logging](https://docs.python.org/3/library/logging.html) facility is used and I think it would make sense that we go with the default here.

danielmitterdorfer: Ok, I'll merge then. :)

PR #459, opened 2018-04-05
Improve documentation of track repo URL

### LINKED ISSUE #443: Improve docs about track repositories
_(based on user feedback we got at ElasticON)_

* We should clarify that the track repo URL in `rally.ini` is only used for cloning the repo. Once it is cloned, changes have no effect. Maybe we can also provide a validation and a warning?
* We should clarify the relationship between the configuration key (e.g. `mytracks.url=https://...`) and the folder name that is used in `~/.rally/benchmarks/tracks` (which would be `mytracks` in this case).

danielmitterdorfer:
Closes #443

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -70,7 +70,7 @@ If you have added a remote you should also add it in ``~/.rally/rally.ini``, oth
 
     private.url = <<URL_TO_YOUR_ORIGIN>>
 
-If you specify ``--track-repository=private``, Rally will check whether there is a directory ``~/.rally/benchmarks/tracks/private``. If there is none, it will use the provided URL to clone the repo. However, if the directory already exists the property is irrelevant and Rally will just update the local tracking branches before the benchmark starts.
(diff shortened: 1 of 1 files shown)


docs/track.rst
```diff
@@ -70,7 +70,7 @@ If you have added a remote you should also add it in ``~/.rally/rally.ini``, oth
 
     private.url = <<URL_TO_YOUR_ORIGIN>>
 
-Rally will then automatically update the local tracking branches before the benchmark starts.
+If you specify ``--track-repository=private``, Rally will check whether there is a directory ``~/.rally/benchmarks/tracks/private``. If there is none, it will use the provided URL to clone the repo. However, if the directory already exists the property is irrelevant and Rally will just update the local tracking branches before the benchmark starts.
```
dliappis: Maybe replace `if the directory already exists the property is irrelevant` with `if the directory already exists, the property gets ignored`

danielmitterdorfer: Thanks for the suggestion; makes sense to me. I'll change the wording before merging.

PR #464, opened 2018-04-10
esrally must not log clear text user passwords

suckowbiz:
esrally 0.9.4 logs the basic auth password in clear text:

```
2018-04-08 07:00:00,100 PID:42 rally.client INFO Creating ES client connected to [{'host': '1.2.3.4', 'port': 9200}, {'host': '1.2.3.5', 'port': 9200}] with options [{'use_ssl': True, 'verify_certs': False, 'basic_auth_user': 'me', 'basic_auth_password': 'mypassword', 'http_auth': ('me', 'password')}]
```

esrally must not do this to enable usage in enterprise environments.

Please accept PR (following).

### DIFF
--- a/esrally/client.py
+++ b/esrally/client.py
@@ -12,7 +12,10 @@ class EsClientFactory:
     Abstracts how the Elasticsearch client is created. Intended for testing.
     """
     def __init__(self, hosts, client_options):
-        logger.info("Creating ES client connected to %s with options [%s]" % (hosts, client_options))
+        masked_client_options = dict(client_options)
+        if masked_client_options.has_key("basic_auth_password"):
+            masked_client_options["basic_auth_password"] = "*****"
+        logger.info("Creating ES client connected to %s with options [%s]" % (hosts, masked_client_options))
         self.hosts = hosts
         self.client_options = client_options
 
--- a/esrally/client.py
+++ b/esrally/client.py
@@ -15,6 +15,8 @@ class EsClientFactory:
         masked_client_options = dict(client_options)
         if masked_client_options.has_key("basic_auth_password"):
             masked_client_options["basic_auth_password"] = "*****"
+        if masked_client_options.has_key("http_auth"):
+            masked_client_options["http_auth"] = (client_options["basic_auth_user"], "*****")
         logger.info("Creating ES client connected to %s with options [%s]" % (hosts, masked_client_options))
         self.hosts = hosts
         self.client_options = client_options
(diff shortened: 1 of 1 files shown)


esrally/client.py
```diff
@@ -12,7 +12,12 @@ class EsClientFactory:
     Abstracts how the Elasticsearch client is created. Intended for testing.
     """
     def __init__(self, hosts, client_options):
-        logger.info("Creating ES client connected to %s with options [%s]" % (hosts, client_options))
+        masked_client_options = dict(client_options)
+        if masked_client_options.has_key("basic_auth_password"):
```
danielmitterdorfer: Can you please change this to `"basic_auth_password" in masked_client_options`?

esrally/client.py
```diff
@@ -12,7 +12,12 @@ class EsClientFactory:
...
+            masked_client_options["basic_auth_password"] = "*****"
+        if masked_client_options.has_key("http_auth"):
```
danielmitterdorfer: Can you please change this to `"http_auth" in masked_client_options`?

esrally/client.py
```diff
@@ -12,7 +12,12 @@ class EsClientFactory:
...
+            masked_client_options["http_auth"] = (client_options["basic_auth_user"], "*****")
```
danielmitterdorfer: I think this should be `masked_client_options["http_auth"] = (client_options["http_auth"][0], "*****")` as you are modifying the `http_auth` entry and a `basic_auth_user` does not necessarily exist then.

esrally/client.py
```diff
@@ -12,7 +12,12 @@ class EsClientFactory:
...
+        logger.info("Creating ES client connected to %s with options [%s]" % (hosts, masked_client_options))
```
dliappis: Very minor comment/ask, the current code is not consistent at the moment, but while you are at it, would it be possible to use the comma format and have the options as arguments, as documented [here](https://docs.python.org/3/library/logging.html)? The idea is that moving forwards we move away from [% for strings](https://pyformat.info/), so in your case it'd be:

```
logger.info("Creating ES client connected to %s with options [%s]", hosts, masked_client_options)
```

Thank you!

esrally/client.py
suckowbiz: Yes. I learned it's more pyhtonic and Pyhton 3 compatible: https://stackoverflow.com/questions/1323410/has-key-or-in

esrally/client.py
suckowbiz: Will do.

esrally/client.py
suckowbiz: Of course.

suckowbiz:
Thanks for the review. I applied the requested changes.

PR #466, opened 2018-04-11
Don't require git

### LINKED ISSUE #339: Don't require git
As Rally allows recently to define tracks as simple files there is no hard requirement that git is installed (but obviously usage is restricted to Rally as load generator with custom tracks).

danielmitterdorfer:
With this commit we silence a spurious log message when git is not
present and document under which circumstances it is ok *not* to have
git installed.

Closes #339

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -69,7 +69,12 @@ We recommend that you use `Homebrew <https://brew.sh/>`_::
 git
 ~~~
 
-``git 1.9`` or better is required. Verify with ``git --version``.
+Git is not required if **all** of the following conditions are met:
+
+* You are using Rally only as a load generator (``--pipeline=benchmark-only``) or you are referring to Elasticsearch configurations with ``--team-path``
+* You create your own tracks and refer to them with ``--track-path``
+
+In all other cases, Rally requires ``git 1.9`` or better. Verify with ``git --version``.
 
 **Debian / Ubuntu**
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -71,8 +71,8 @@ git
 
 Git is not required if **all** of the following conditions are met:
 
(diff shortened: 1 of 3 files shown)


docs/install.rst
```diff
@@ -69,7 +69,12 @@ We recommend that you use `Homebrew <https://brew.sh/>`_::
 git
 ~~~
 
-``git 1.9`` or better is required. Verify with ``git --version``.
+Git is not required if **all** of the following conditions are met:
+
+* You are using Rally only as a load generator (``--pipeline=benchmark-only``) or you are referring to Elasticsearch configurations with ``--team-path``
```
dliappis: Apologies for the pedantry; full stop will be appreciated for this sentence (we do it elsewhere).

docs/install.rst
```diff
@@ -69,7 +69,12 @@ We recommend that you use `Homebrew <https://brew.sh/>`_::
...
+* You create your own tracks and refer to them with ``--track-path``
```
dliappis: Same comment here, full stop would be appreciated.

esrally/utils/git.py
```diff
@@ -8,7 +8,8 @@
 def probed(f):
     def probe(src, *args, **kwargs):
         # Probe for -C
-        if not process.exit_status_as_bool(lambda: process.run_subprocess_with_logging("git -C %s --version" % src, level=logging.DEBUG)):
+        if not process.exit_status_as_bool(lambda: process.run_subprocess_with_logging("git -C %s --version" % src, level=logging.DEBUG),
```
dliappis: Switch to `format()` for the string `"git -C {} --version".format(src)`

dliappis:
LGTM, thanks for the update.

PR #468, opened 2018-04-11
Require at least Rally 0.7.3 for config migration

danielmitterdorfer:
With this commit we remove outdated config migration code. As our tracks
require at least Rally 0.9.0 now as of the recent cleanup (see #380), we
remove all migration code for config version 11 and below (that is
versions older than Rally 0.7.3).


danielmitterdorfer:
Marked as "breaking" as it will require user action when migrating from a version older than 0.7.3.

PR #470, opened 2018-04-17
Allow to use track parameters in index definition

danielmitterdorfer:
Previously we forgot to pass track parameters to the
`TrackSpecificationReader`. As a consequence, we could not use track
parameters in index definitions as they were always empty.

With this commit we finally pass them there as well fixing the issue
above.


danielmitterdorfer:
Thanks for the review. It does not relate to #456 but rather https://github.com/elastic/rally/issues/429. I can mention it in the commit comment.

PR #473, opened 2018-04-19
Don't pass type implicitly for queries

danielmitterdorfer:
With this commit users need to explicitly define that a type should be
passed when issuing a query against Elasticsearch. The reason is that
types are deprecated with Elasticsearch 7 and will be removed in
Elasticsearch 8. Hence we already change the behavior in Rally to avoid
deprecation warnings when benchmarking against master.


danielmitterdorfer:
Thanks, merged! :)

PR #476, opened 2018-04-20
Create unique ids even for random conflicts

danielmitterdorfer:
With this commit we ensure that generated ids are always unique, even
when generating random conflicts for bulk requests. The general strategy
to simulate conflicts is as follows:

1. Generate a list of ids upfront.
2. Iterate through this list and pick a duplicate in 25% of all cases.

In order to avoid accidentally creating more conflicts than expected,
the list of ids generated in step one needs to contain *unique* ids.
However, we create ids at random without considering which ids have
already been generated. This is now changed by first generating all ids
and then shuffling them.

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -565,17 +565,16 @@ def number_of_bulks(corpora, partition_index, total_partitions, bulk_size):
     return bulks
 
 
-def build_conflicting_ids(conflicts, docs_to_index, offset, rand=random.randint):
+def build_conflicting_ids(conflicts, docs_to_index, offset, shuffle=random.shuffle):
     if conflicts is None or conflicts == IndexIdConflict.NoConflicts:
         return None
     logger.info("building ids with id conflicts of type [%s]" % conflicts)
     all_ids = [0] * docs_to_index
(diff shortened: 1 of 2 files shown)


esrally/track/params.py
```diff
@@ -565,17 +565,16 @@ def number_of_bulks(corpora, partition_index, total_partitions, bulk_size):
...
         return None
     logger.info("building ids with id conflicts of type [%s]" % conflicts)
     all_ids = [0] * docs_to_index
     for i in range(docs_to_index):
         # always consider the offset as each client will index its own range and we don't want uncontrolled conflicts across clients
-        if conflicts == IndexIdConflict.SequentialConflicts:
-            all_ids[i] = "%10d" % (offset + i)
-        else:  # RandomConflicts
-            all_ids[i] = "%10d" % rand(offset, offset + docs_to_index)
+        all_ids[i] = "%10d" % (offset + i)
```
dliappis: Sticking to `%` as we are in the hot code path right? (instead of `"{:>10d}".format(offset+i)`?)

danielmitterdorfer: yes

PR #477, opened 2018-04-20
Improve simulation of bulk-indexing conflicts

### LINKED ISSUE #422: Mechanism needed for benchmarking update-heavy workflows against create-only or index-only
Currently, rally can be configured to generate metadata that contains conflicting ids for 25% of the documents in a corpus. However, the action specified in the metadata is always `index`. Since the performance characteristics of `update` changed significantly in 5.0+, this seems like a blind spot in the current suite. 

While I understand that metadata can be interleaved with the documents in a corpus instead of being generated by rally, side-by-side comparison of performance of the same corpus with conflicting ids would be considerably simpler if the create/update workflow could use generated metadata.

danielmitterdorfer:
With this commit we introduce a new property `conflict-probability` for
the bulk-indexing parameter source. Previously we had a hard-codded
probability of 25% but now the user can control it.

We also use `update` now as bulk-indexing action when simulating a
conflict (previously the action was always `index`).

Closes #422


tests/track/params_test.py
```diff
@@ -263,6 +276,68 @@ def test_read_bulks_and_assume_metadata_line_in_source_file(self):
...
+            '{"key": "value4"}',
+            '{"key": "value5"}'
+        ]
+        bulk_size = 2
+
+        source = params.Slice(io.StringAsFileSource, 0, len(data))
+        am_handler = params.GenerateActionMetaData("test_index", "test_type",
+                                                   conflicting_ids=[100, 200, 300, 400],
+                                                   conflict_probability=25,
+                                                   rand=lambda: next(pseudo_random_conflicts),
```
dliappis: Clever use of `iter()` and `lambda` 👍

danielmitterdorfer: Hehe, thanks :)

danielmitterdorfer:
@dliappis I pushed another commit since your LGTM which adds a new parameter "on-conflict" that lets the user decide whether to use `update` or `index` in the action and meta-data line. The reason for this parameter is that we'd otherwise change the entire approach of handling id conflicts which leads to completely different results.

Can you please review that commit as well?

tests/track/params_test.py
```diff
@@ -104,6 +105,12 @@ def test_generate_action_meta_data_without_id_conflicts(self):
                          next(params.GenerateActionMetaData("test_index", "test_type")))
 
     def test_generate_action_meta_data_with_id_conflicts(self):
+        def idx(id):
+            return "index", '{"index": {"_index": "test_index", "_type": "test_type", "_id": "%s"}}' % id
```
dliappis: Very minor comment, could use `.format()` for consistency going forwards.

tests/track/params_test.py
```diff
@@ -104,6 +105,12 @@ def test_generate_action_meta_data_without_id_conflicts(self):
...
+
+        def conflict(action, id):
+            return action, '{"%s": {"_index": "test_index", "_type": "test_type", "_id": "%s"}}' % (action, id)
```
dliappis: Likewise, could use `.format()` here too.

tests/track/params_test.py
danielmitterdorfer: This is on the hot-code path where we decided against `#format()`.

tests/track/params_test.py
danielmitterdorfer: This is on the hot-code path where we decided against `#format()`.

danielmitterdorfer:
Thanks for the review @dliappis!

tests/track/params_test.py
dliappis: D'oh yes!

tests/track/params_test.py
dliappis: D'oh yes!

Issue #483, opened 2018-04-24
Improve release process


dliappis:
This is a small but annoying issue.

The current release workflow and script (`./release.sh`) rely on certain pip3 packages and is flaky while using a virtualenv.

In ticket we'll explore if there's a simpler and more reliable way to perform this.

We will start by experimenting with adding a `Makefile` target that allows us to `./release.sh` inside a Docker environment with all prerequisites satisfied.

danielmitterdorfer:
@dliappis In `make install` we setup all required dependencies that are needed to do a release so I think we've solved this by now?

dliappis:
Makes sense yes.

PR #484, opened 2018-04-24
Improve release process

dliappis:
Commit `Dockerfile`, `docker-compose` and new `release-in-docker` make target
to allow running the release process from within a Docker container.

The container tries to use the same username/uid as the local user and
bind-mounts the needed files for gpg, ssh-agent and github release
token.

The execution of `make it` is still a task left outside of the container
primarily due to execution time on macOS.

Relates #483

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM python:3.6
+
+ARG NEW_USER_UID
+ARG NEW_USER
+
+RUN apt-get -y update && \
+    apt-get install -y vim openssh-client python-tox && \
+    pip3 install twine sphinx sphinx_rtd_theme wheel && \
+    pip3 install --pre github3.py
+
+RUN useradd -u ${NEW_USER_UID} -U -d /home/${NEW_USER} -s /bin/bash -m ${NEW_USER}
+
+USER ${NEW_USER}
+WORKDIR /home/${NEW_USER}
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM python:3.6
+
+ARG NEW_USER_UID
+ARG NEW_USER
+
+RUN apt-get -y update && \
+    apt-get install -y vim openssh-client python-tox && \
```
danielmitterdorfer: out of curiosity: Why do we need `vim`?

dliappis: This was just in case something goes weird inside the container and we need to edit a file, which `exec`'ed inside the container. Since we are binding everything locally this is generally not needed, but I though maybe there's this edge case where we need to quickly editing something (had to do it myself :) ).

dliappis:
Thanks for the review; yes I am afraid we can only properly test this on a real release; there are a few things though to try esp. on a macOS system before hand esp. if the the ssh keys get passed correctly through the agent and gpg.

+1: for rebase here.

Dockerfile
danielmitterdorfer: I see. Makes sense.

dliappis:
Closing this, as passing ssh keys inside a container when keys are cached on macOS using keychain is not trivial. May revisit this in the future.

PR #487, opened 2018-04-30
Add new Makefile target to run it tests inside Docker

dliappis:
`it` tests and `tox` relying on a number of Python 3 versions via
pyenv are a good candidate for running through Docker, as setting up
pyenv and the needed Python versions (as well as required
JAVA_HOME/RUNTIME_JAVA_HOME) can be fiddly.

This is an easier task than https://github.com/elastic/rally/pull/484
as it doesn't rely on passing ssh credentials via the ssh-agent. Only
issue is that on operating systems where Docker runs as a VM (macOS,
Windows) the default configured memory will not to be increased in
order to run the entirety of the integration tests, as Rally uses the
4GB car for Elasticsearch (on top of additional tasks). However, this
target is especially suitable for running integration tests inside a
powerful server/workstation Linux environment without having to worry
about configuring all the dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM ubuntu:16.04
+
+ARG NEW_USER_UID
+ARG NEW_USER
+
+RUN apt-get -y update && \
+    apt-get install -y curl python3 python3-pip vim openssh-client \
+    git make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev \
+    openjdk-8-jdk
+
+RUN pip3 install tox virtualenvwrapper twine sphinx sphinx_rtd_theme wheel
+RUN pip3 install --pre github3.py
+
+RUN curl --progress-bar -o openjdk-10.0.1_linux-x64_bin.tar.gz \
+    https://download.java.net/java/GA/jdk10/10.0.1/fb4372174a714e6b8c52526dc134031e/10/openjdk-10.0.1_linux-x64_bin.tar.gz && \
+    mkdir -p /opt/jdk-10 ; tar xzf openjdk-10.0.1_linux-x64_bin.tar.gz -C /opt/jdk-10 --strip-components 1 && rm openjdk-10.0.1_linux-x64_bin.tar.gz
+
+RUN useradd -u ${NEW_USER_UID} -U -d /home/${NEW_USER} -s /bin/bash -m ${NEW_USER}
+
+USER ${NEW_USER}
+WORKDIR /home/${NEW_USER}
+
+RUN git clone https://github.com/yyuu/pyenv.git ~/.pyenv
+
+ENV HOME /home/${NEW_USER}
+ENV PYENV_ROOT=/home/${NEW_USER}/.pyenv
+ENV PATH=$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
+ENV JAVA_HOME=/opt/jdk-10
+ENV RUNTIME_JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
+
+RUN pyenv install 3.4.8
+RUN pyenv install 3.5.5
+RUN pyenv install 3.6.5
+RUN pyenv global 3.6.5 3.5.5 3.4.8
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,6 @@ RUN apt-get -y update && \
     openjdk-8-jdk
 
 RUN pip3 install tox virtualenvwrapper twine sphinx sphinx_rtd_theme wheel
-RUN pip3 install --pre github3.py
 
 RUN curl --progress-bar -o openjdk-10.0.1_linux-x64_bin.tar.gz \
     https://download.java.net/java/GA/jdk10/10.0.1/fb4372174a714e6b8c52526dc134031e/10/openjdk-10.0.1_linux-x64_bin.tar.gz && \
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ release: release-checks clean docs it
 	./release.sh $(release_version) $(next_version)
 
 docker-it: nondocs-clean python-caches-clean
-	@if ! export | grep UID; then export UID=$(shell id -u); fi ; \
+	@if ! export | grep UID; then -export UID=$(shell id -u) || export UID; fi ; \
 	if ! export | grep USER; then export USER=$(shell echo $$USER); fi ; \
 	if ! export | grep PWD; then export PWD=$(shell pwd); fi ; \
 	docker-compose build --pull; `# add --pull here to rebuild a fresh image` \
--- a/Makefile
+++ b/Makefile
@@ -44,7 +44,7 @@ release: release-checks clean docs it
 	./release.sh $(release_version) $(next_version)
 
 docker-it: nondocs-clean python-caches-clean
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+ARG NEW_USER_UID
+ARG NEW_USER
+
+RUN apt-get -y update && \
+    apt-get install -y curl python3 python3-pip vim openssh-client \
+    git make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev \
+    openjdk-8-jdk
+
+RUN pip3 install tox virtualenvwrapper twine sphinx sphinx_rtd_theme wheel
+RUN pip3 install --pre github3.py
```
danielmitterdorfer: I think this is not needed to run integration tests?

docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    build:
+      context: .
+      args:
+        - NEW_USER_UID=${UID}
+        - NEW_USER=${USER}
+    container_name: rally-tests
+    volumes:
+      - $HOME/.gitconfig:/home/${USER}/.gitconfig
+      - $HOME/.github:/home/${USER}/.github
+      - $HOME/.gnupg:/home/${USER}/.gnupg
```
danielmitterdorfer: Is it required to mount these directories?

Makefile
```diff
@@ -34,4 +43,11 @@ release-checks:
 release: release-checks clean docs it
 	./release.sh $(release_version) $(next_version)
 
-.PHONY: clean docs test it it34 it35 it36 benchmark coverage release release-checks
+docker-it: nondocs-clean python-caches-clean
+	@if ! export | grep UID; then export UID=$(shell id -u); fi ; \
```
danielmitterdorfer: I am not familiar with this syntax. Why is there an `@` in the first line?

Makefile
```diff
@@ -1,7 +1,16 @@
-clean:
+clean: nondocs-clean docs-clean
```
danielmitterdorfer: I wonder whether `clean` should clean everything (including Python caches).

Makefile
```diff
@@ -1,7 +1,16 @@
...
+
+nondocs-clean:
 	rm -rf .benchmarks .eggs .tox .rally_it .cache build dist esrally.egg-info logs junit-py*.xml
+
+docs-clean:
 	cd docs && $(MAKE) clean
 
+# Avoid conflicts between .pyc/pycache related files created by local Python interpreters and other interpreters in Docker
+python-caches-clean:
```
danielmitterdorfer: If I type `make python-caches-clean` locally (on MacOS), I get:

```
(.venv) daniel@io:Projects/rally ‹pr/487›$ make python-caches-clean
find ./ -name "__pycache__" -exec rm -rf -- \{\} \;
find: .//.venv/lib/python3.6/__pycache__: No such file or directory
[...]
find: .//tests/utils/__pycache__: No such file or directory
make: [python-caches-clean] Error 1 (ignored)
find ./ -name ".pyc" -exec rm -rf -- \{\} \;
```

Makefile
dliappis: `@` causes `make` to suppress echoing: https://www.gnu.org/software/make/manual/make.html#Echoing

less junk on the output esp. with `if`'s that can confuse the user.

Dockerfile
dliappis: No I thought we might keep it in case we give another try at #484 in the future with the same image. But I guess it makes sense to keep it clean for now.

docker-compose.yml
dliappis: No, this was left over with the same idea as before, i.e. maybe we could use this for another attempt at #484 in the future. I'll remove it for now.

Makefile
dliappis: As discussed, this is harmless and should always return `exit 0` due to the `-` prefix in the command.

As discussed consecutive slashes will get collapsed as per the [Single Unix Specification](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_266).

As discussed seems that using `.` instead of `./` makes the output look saner on macOS so I'll push this plus an `@` to suppress output altogether.

Makefile
dliappis: I am ambivalent about this one because it you don't use docker it will introduce some slowness every time you run `make docs` or `make it` as I've noticed some downloads happening. At the same time it ensures consistency.

Makefile
danielmitterdorfer: Thanks. Makes sense.

Makefile
danielmitterdorfer: Ok, it does not make sense to clean Python caches for `make docs`. However, I think for `make it` would make sense?

Makefile
danielmitterdorfer: Didn't know about that. I also saw now that this is actually just one long line separated by `\`, i.e. all output from this target will get suppressed.

dliappis: Just a small correction not all output will get suppressed (e.g. the output generated from the commands themselves e.g. `make it` output will get displayed as expected).

e.g. if you have a Makefile target called:

```
testme:
    @ls
```

this will display the `ls` output normally but not the `ls` command itself, which would be the case without `@`.

Makefile
dliappis: Yes it would make sense for `make it`. I think we should add it to the individual `it34/35/36` targets too?

danielmitterdorfer: Yes, please.

Makefile
danielmitterdorfer: Yes, I should have said "echoing" instead of "output". Thanks for the explanation.

dliappis:
@danielmitterdorfer I've addressed your comments, could have another pass at the PR?

Makefile
```diff
@@ -8,22 +8,22 @@ docs-clean:
...
 
-it34:
+it34: python-caches-clean
 	tox -e py34
 
-it35:
+it35: python-caches-clean
 	tox -e py35
 
 it36:
```
danielmitterdorfer: Should also need `python-caches-clean`?

docker-compose.yml
```diff
@@ -8,9 +8,6 @@ services:
         - NEW_USER=${USER}
     container_name: rally-tests
     volumes:
-      - $HOME/.gitconfig:/home/${USER}/.gitconfig
-      - $HOME/.github:/home/${USER}/.github
-      - $HOME/.gnupg:/home/${USER}/.gnupg
       - ${PWD}/.:/home/${USER}/rally
```
danielmitterdorfer: I missed this in my first pass. Doesn't this assume that the user's Rally project directory is available in `~/rally`?

dliappis: Correct. The CWD is changed here: https://github.com/dliappis/rally/blob/79c6ab8fad02f6509bb510c116c8d3de8b6b0cc0/docker-compose.yml#L14 .

The reason it's done this way is that any dot files created under `/home/$USER` inside the docker container shouldn't interfere with the dot files (e.g. `~/.rally`) of the host system.

Makefile
dliappis: Thanks forgot this one! Addressed it now.

dliappis:
@dm can you have another look now?

docker-compose.yml
danielmitterdorfer: Ah, you're right. This is the path in the container, not on the host.

danielmitterdorfer:
Thanks. LGTM! :)

PR #488, opened 2018-05-01
Add support for multiple clusters in custom runners

dliappis:
With this PR we add support for specifying multiple clusters using `target-hosts` and `client-options`.

The existing way of specifying `target-hosts`, as a csv host:port string and `client-options` as csv key:value pairs in a string is preserved. Additionally, we now allow passing parameters for `target-hosts` and `client-options` using a json filename as well as an in-line json string, which is already possible for other cli arguments. The documentation now reflects how to use this and we back this with a number of rigorous test cases.

All core runners and custom runners, by default, still receive an es client object pointing to the "default" cluster (for compatibility with the regular use case of Rally benchmarking one ES cluster) but it is now possible to specify `multi_cluster = True` as a class attribute in a custom runner. This results in an `es` dictionary instead where the user can freely choose which cluster to target inside the custom runner code.

Additionally we now wrap context-manager enabled runners and add a test case to ensure that wrapping  it in this way really works as expected. The documentation for custom runners has also been fixed to not show a half complete example of a context-manager enabled Class.


docs/command_line_reference.rst
```diff
@@ -509,3 +510,67 @@ When you run ``esrally list races``, this will show up again::
...
+      "remote":
+        {"use_ssl":true,"verify_certs": false,"ca_certs":"/path/to/cacert.pem"}
+    }
+
+* json inline string defining two clusters::
+
+    --client-options="{\"default\":{\"timeout\": 60}, \"remote\": {\"use_ssl\":true,\"verify_certs\":false,\"ca_certs\":\"/path/to/cacert.pem\"}}"
+
+.. WARNING::
+   If you use ``client-options`` you must specify options for **every** cluster name defined with ``target-hosts``. An error message will be displayed if there is a mismatch.
```
danielmitterdorfer: Nit: Use active voice? "An error message will be displayed if there is a mismatch." -> (something along the lines of): "Rally raises an error on mismatch."?

docs/command_line_reference.rst
```diff
@@ -509,3 +510,67 @@ When you run ``esrally list races``, this will show up again::
...
+        {"host": "10.127.0.8", "port": 9201}
+      ]
+    }
+
+* json inline string defining two clusters::
+
+    --target-hosts="{\"default\":[\"127.0.0.1:9200\"],\"remote\":[\"127.0.0.1:19200\",\"127.0.0.1:19201\"]}"
+
+.. NOTE::
+   **All** internal :ref:`track operations <track_operations>` will use the connection to the ``default`` cluster. However, you can utilize the client connections to the additional clusters in your :ref:`custom runners <adding_tracks_custom_runners>`.
```
danielmitterdorfer: Nit: It took me a while to understand the term "internal track operations". Maybe use "built-in operations" instead?

esrally/driver/driver.py
```diff
@@ -587,7 +587,7 @@ def receiveMsg_StartLoadGenerator(self, msg, sender):
         self.client_id = msg.client_id
         self.config = load_local_config(msg.config)
         self.abort_on_error = self.config.opts("driver", "on.error") == "abort"
-        self.es = client.EsClientFactory(self.config.opts("client", "hosts"), self.config.opts("client", "options")).create()
+        self.es = client.EsClientsFactory()(self.config.opts("client", "hosts").all_hosts, self.config.opts("client", "options").all_client_options)
```
danielmitterdorfer: That syntax looks odd to me (constructor call followed by a `__call__` method invocation). Maybe we don't need a class for this and just use a function instead?

esrally/driver/runner.py
```diff
@@ -61,16 +64,48 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+            return "user-defined context-manager enabled runner for [%s]" % self.name
+        else:
+            return "user-defined runner for [%s]" % self.name
+
+    def __enter__(self):
+        if self.context_manager_enabled:
+            self.runnable.__enter__()
+            return self
+        else:
+            return False
```
danielmitterdorfer: I think we *always* need to return `self`.

esrally/driver/runner.py
```diff
@@ -61,16 +64,48 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.context_manager_enabled:
+            return self.runnable.__exit__(exc_type, exc_val, exc_tb)
+        else:
+            return False
+
+
+class MultiClusterDelegatingRunner(Runner):
```
danielmitterdorfer: It seems `MultiClusterDelegatingRunner` does not support the context manager protocol.

esrally/mechanic/launcher.py
```diff
@@ -62,23 +62,31 @@ def start(self):
...
+        all_hosts = self.cfg.opts("client", "hosts").all_hosts
+        default_hosts = self.cfg.opts("client", "hosts").default
+
+        es = {}
+        for cluster_name, cluster_hosts in all_hosts.items():
+            all_client_options = self.cfg.opts("client", "options").all_client_options
+            cluster_client_options = all_client_options[cluster_name]
+            es[cluster_name] = self.client_factory(cluster_hosts, cluster_client_options).create()
+
+        es_default = es["default"]
```
danielmitterdorfer: I am unsure about using the "magic name" `default` here because it actually breaks encapsulation.

esrally/mechanic/launcher.py
```diff
@@ -88,7 +96,7 @@ def start(self):
             self.stop(c)
             raise exceptions.LaunchError("Elasticsearch REST API layer is not available. Forcefully terminated cluster.")
         if self.on_post_launch:
-            self.on_post_launch(es)
+            self.on_post_launch(es["default"])
```
danielmitterdorfer: Can't you use `es_default` here?

esrally/mechanic/launcher.py
```diff
@@ -184,8 +192,8 @@ def __init__(self, cfg, metrics_store, client_factory_class=client.EsClientFacto
         self.client_factory = client_factory_class
 
     def start(self, node_configurations=None):
-        hosts = self.cfg.opts("client", "hosts")
-        client_options = self.cfg.opts("client", "options")
+        hosts = self.cfg.opts("client", "hosts")()
+        client_options = self.cfg.opts("client", "options")()
```
danielmitterdorfer: I think it would be better to use an explicit property name here instead of `()`?

esrally/mechanic/mechanic.py
```diff
@@ -176,8 +176,8 @@ def cluster_distribution_version(cfg, client_factory=client.EsClientFactory):
     :param client_factory: Factory class that creates the Elasticsearch client.
     :return: The distribution version.
     """
-    hosts = cfg.opts("client", "hosts")
-    client_options = cfg.opts("client", "options")
+    hosts = cfg.opts("client", "hosts")()
+    client_options = cfg.opts("client", "options")()
```
danielmitterdorfer: I think it would be better to use an explicit property name here instead of `()`?

esrally/mechanic/mechanic.py
```diff
@@ -266,7 +266,7 @@ def receiveMsg_StartEngine(self, msg, sender):
         self.car, _ = load_team(self.cfg, msg.external)
 
         # In our startup procedure we first create all mechanics. Only if this succeeds we'll continue.
-        hosts = self.cfg.opts("client", "hosts")
+        hosts = self.cfg.opts("client", "hosts")()
```
danielmitterdorfer: I think it would be better to use an explicit property name here instead of `()`?

esrally/racecontrol.py
```diff
@@ -287,13 +287,13 @@ def race(cfg, sources=False, build=False, distribution=False, external=False, do
 
 
 def set_default_hosts(cfg, host="127.0.0.1", port=9200):
-    configured_hosts = cfg.opts("client", "hosts", mandatory=False)
-    if configured_hosts:
-        logger.info("Using configured hosts %s" % configured_hosts)
+    configured_hosts = cfg.opts("client", "hosts")
+    if len(configured_hosts()) !=0 :
```
danielmitterdorfer: I think it would be better to use an explicit property name here instead of `()`?

esrally/rally.py
```diff
@@ -436,7 +434,7 @@ def ensure_configuration_present(cfg, args, sub_command):
             exit(64)
 
 
-def list(cfg):
+def cfglist(cfg):
```
danielmitterdorfer: Use a different name like `dispatch_list`?

esrally/rally.py
```diff
@@ -11,14 +11,12 @@
 from esrally import version, actor, config, paths, racecontrol, reporter, metrics, track, chart_generator, exceptions, time as rtime
 from esrally import PROGRAM_NAME, DOC_LINK, BANNER, SKULL, check_python_version
 from esrally.mechanic import team, telemetry
-from esrally.utils import io, convert, process, console, net
+from esrally.utils import io, convert, process, console, net, config as config_helper
```
danielmitterdorfer: Maybe use a different name for this module? Instead of `config` maybe `opts` (because it is actually about handling command line options)?

esrally/utils/config.py
```diff
@@ -0,0 +1,162 @@
+import argparse
+import json
+import logging
+from esrally.utils import console, io
```
danielmitterdorfer: `console` is unused here.

esrally/utils/config.py
```diff
@@ -0,0 +1,162 @@
...
+    if io.has_extension(arg, ".json"):
+        with open(io.normalize_path(arg), mode="rt", encoding="utf-8") as f:
+            return json.load(f)
+    elif arg.startswith("{"):
+        return json.loads(arg)
+    else:
+        return default_parser(csv_to_list(arg))
+
+
+class CfgESConnectionOptions:
```
danielmitterdorfer: Shorten name e.g. to `ConnectOptions` (it is clear from the context that this is (a) about configuration and (b) about Elasticsearch)?

esrally/utils/config.py
```diff
@@ -0,0 +1,162 @@
...
+            This is needed to support single cluster use of --client-options when not
+            defined as a json string or file.
+            """
+
+            return {"default": kv_to_map(arg)}
+
+        if self.argvalue == ClientOptions.DEFAULT_CLIENT_OPTIONS and self.target_hosts != None:
+            # --client-options unset but multi-clusters used in --target-hosts? apply options defaults for all cluster names.
+            self.parsed_options = {cluster_name: kv_to_map([ClientOptions.DEFAULT_CLIENT_OPTIONS])
+                                   for cluster_name, _ in self.target_hosts.all_hosts.items()}
```
danielmitterdorfer: Simplify to: `for cluster_name in self.target_hosts.all_hosts.keys()`?

tests/driver/runner_test.py
```diff
@@ -15,24 +16,36 @@ def runner_function(es, params):
...
+    def test_runner_class_with_context_manager_should_be_wrapped_with_context_manager_enabled(self):
+        class UnitTestContextManagerRunner:
+            def __call__(self, *args):
                 return self
 
-            def __call__(self, *args):
-                pass
+            def __enter__(self):
+                self.fp = io.StringIO("many\nlines\nin\na\nfile")
+                return self.fp
```
danielmitterdorfer: I think you should still return `self` here. However, you need to forward method calls to the delegate (i.e. `self.fp`).

tests/driver/runner_test.py
```diff
@@ -15,24 +16,36 @@ def runner_function(es, params):
...
 
             def __exit__(self, exc_type, exc_val, exc_tb):
-                return False
+                return self.fp.close()
```
danielmitterdorfer: I think the implementation should be:

```python
self.fp.close()
return False
```

The return value indicates whether a potential exception has been handled (indicated by `return True`). As we don't  do anything here, `return False` is warranted IMHO.

tests/driver/runner_test.py
```diff
@@ -15,24 +16,36 @@ def runner_function(es, params):
...
         returned_runner = runner.runner_for("unit_test")
-        self.assertTrue(test_runner == returned_runner)
+        self.assertIsInstance(returned_runner, runner.SingleClusterDelegatingRunner)
+        self.assertEqual("user-defined context-manager enabled runner for [UnitTestContextManagerRunner]", repr(returned_runner))
+
+        # test that context_manager functionality gets preserved after wrapping
+        with returned_runner({"default": {}},{}) as fp:
+            file_contents = fp.read()
+        self.assertTrue(file_contents, "many\nlines\nin\na\nfile")
+        self.assertTrue(fp.closed)
```
danielmitterdorfer: I *think* as `fp` gets bound to the delegate (`self.fp` above) instead of our `UnitTestContextManagerRunner`, the file is not closed due to our implementation in `__exit()__`.

tests/mechanic/launcher_test.py
```diff
@@ -76,11 +77,15 @@ def info(self, *args, **kwargs):
 
 
 class ExternalLauncherTests(TestCase):
+    test_host = config_helper.TargetHosts('127.0.0.1:9200,10.17.0.5:19200')
```
danielmitterdorfer: Nit: So far I used double-quotes for strings across the code-base and I think we should stick just to one representation.

tests/mechanic/launcher_test.py
```diff
@@ -76,11 +77,15 @@ def info(self, *args, **kwargs):
...
+    client_options = config_helper.ClientOptions('timeout:60')
```
danielmitterdorfer: see above.

tests/utils/config_test.py
```diff
@@ -0,0 +1,165 @@
...
+        self.assertEqual(
+            {'default': ['127.0.0.1:9200','10.17.0.5:19200'],
+             'remote_1': ['88.33.22.15:19200'],
+             'remote_2': ['10.18.0.6:19200','10.18.0.7:19201']},
+            config.TargetHosts(target_hosts).all_hosts)
+
+    def test_json_file_parameter_parses(self):
+        self.assertEqual(
+            {"default": ["127.0.0.1:9200","10.127.0.3:19200"] },
+            config.TargetHosts(os.path.join(os.path.dirname(__file__), "resources/target_hosts_1.json")).all_hosts)
```
danielmitterdorfer: Nit: I think this should rather be `os.path.join(os.path.dirname(__file__), "resources", "target_hosts_1.json")`) (all path components as separate elements)

tests/utils/config_test.py
```diff
@@ -0,0 +1,165 @@
...
+              ],
+              "remote_1":[
+                {"host": "10.127.0.3", "port": 9200},
+                {"host": "10.127.0.8", "port": 9201}
+              ],
+              "remote_2":[
+                {"host": "88.33.27.15", "port": 39200}
+              ]
+            },
+            config.TargetHosts(os.path.join(os.path.dirname(__file__), "resources/target_hosts_2.json")).all_hosts)
```
danielmitterdorfer: see above w.r.t. usage of `os.path.join`.

docs/command_line_reference.rst
danielmitterdorfer: After checking the code I am also not sure where it would actually raise an error on mismatch (I spotted this because I think such a test is missing).

tests/utils/config_test.py
```diff
@@ -0,0 +1,165 @@
...
+             'remote_2': {'use_ssl': True,'verify_certs': True, 'ca_certs':'/path/to/cacert.pem'}},
+            config.ClientOptions(client_options_string).all_client_options)
+
+
+    def test_json_file_parameter_parses(self):
+        self.assertEqual(
+            {'default': {'timeout':60},
+             'remote_1': {'use_ssl': True,'verify_certs': True,'basic_auth_user':'elastic','basic_auth_password':'changeme'},
+             'remote_2': {'use_ssl': True,'verify_certs': True, 'ca_certs':'/path/to/cacert.pem'}},
+            config.ClientOptions(os.path.join(os.path.dirname(__file__), "resources/client_options_1.json")).all_client_options)
```
danielmitterdorfer: os.path.join(os.path.dirname(__file__), "resources/target_hosts_1.json")

tests/utils/config_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+        self.assertEqual(
+            {'default': {'timeout':60}},
+            config.ClientOptions(os.path.join(os.path.dirname(__file__), "resources/client_options_2.json")).all_client_options)
```
danielmitterdorfer: os.path.join(os.path.dirname(__file__), "resources/target_hosts_1.json")

tests/driver/runner_test.py
dliappis: Any particular reason for this? My understanding, and based on what I've read, this is ok and pretty standard and lets context manager definitions use the simpler `with ... as myobject` and then just reference `myobject.method()` which is for example what we do with Python's native `open()` rather than something like `as myobject` and later on `myobject.fp.read()`.

For example this:

```
import io        
                       
class Foo:
    def __enter__(self):
        self.fp = io.StringIO("many\nlines\nin\na\nfile")
        return self
    def __exit__(self, a, b, c):
        self.fp.close()
        return False
```

needs to be used as:

```
with Foo() as foo:
    print(foo.fp.read())
```

rather than:

```
with Foo() as foo:
    print(foo.read())
```

tests/driver/runner_test.py
dliappis: This is not my understanding and if `self.fp.close()` gets commented out but leaving just `return False`, for example, the file pointer remains open and the test case fails.

tests/driver/runner_test.py
dliappis: Agreed. Will address.

esrally/driver/runner.py
dliappis: Right; I _thought_ we don't really need it as it is a special case after all. Then again, if it doesn't support the context manager protocol, I should clarify this in the docs, which sounds to me like more effort than adding the support :)

esrally/rally.py
dliappis: Good suggestion. Will address.

esrally/rally.py
dliappis: Will do.

esrally/utils/config.py
dliappis: +1

esrally/utils/config.py
dliappis: Of course :) Addressing.

tests/mechanic/launcher_test.py
dliappis: All :+1:  for standardization. I like the "safety" of no-interpolation with the single quote, but the code should be standardized. Will address.

tests/mechanic/launcher_test.py
dliappis: :+1:

tests/utils/config_test.py
dliappis: Yes makes perfect sense. Addressing.

tests/utils/config_test.py
dliappis: Yes makes perfect sense. Addressing.

esrally/driver/runner.py
dliappis: I think you are right as this code will need to deal with Retry() and other wrappers. Will address.

tests/driver/runner_test.py
danielmitterdorfer: I thought that the way it is implemented now the delegate actually escapes from the wrapper.

tests/driver/runner_test.py
danielmitterdorfer: In that case it seems I was wrong. Please disregard.

tests/utils/config_test.py
danielmitterdorfer: This was a copy & paste error. I suggest the same as for the other test case above, i.e.

```python
os.path.join(os.path.dirname(__file__), "resources", "target_hosts_1.json")
```

tests/utils/config_test.py
danielmitterdorfer: This was a copy & paste error. I suggest the same as for the other test case above, i.e.

```python
os.path.join(os.path.dirname(__file__), "resources", "client_options_2.json")
```

docs/command_line_reference.rst
dliappis: The test is here: https://github.com/dliappis/rally/blob/bdc3f0e17632f5a82ae900acf9b8e142fe573934/esrally/rally.py#L690-L691

docs/command_line_reference.rst
```diff
@@ -435,7 +434,9 @@ If you run the ``benchmark-only`` :doc:`pipeline </pipelines>` or you want Rally
 
    esrally --pipeline=benchmark-only --target-hosts=10.17.0.5:9200,10.17.0.6:9200
 
-This will run the benchmark against the hosts 10.17.0.5 and 10.17.0.6 on port 9200. See ``client-options`` if you use X-Pack Security and need to authenticate or Rally should use https.
+** This will run the benchmark against the hosts 10.17.0.5 and 10.17.0.6 on port 9200. See ``client-options`` if you use X-Pack Security and need to authenticate or Rally should use https.
```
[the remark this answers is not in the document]
dliappis: Typo. Thanks for catching.

docs/command_line_reference.rst
```diff
@@ -509,3 +510,67 @@ When you run ``esrally list races``, this will show up again::
...
+Advanced topics
+---------------
+
+``target-hosts``
+~~~~~~~~~~~~~~~~
+
+Rally can also create client connections for multiple Elasticsearch clusters.
+This is only useful if you want to create :ref:`custom runners <adding_tracks_custom_runners>` that execute operations against remote clusters, for example for Cross Cluster Search or Cross Cluster Replication.
+
+To define the host:port pairs for additional clusters with ``target-hosts`` you can specify either a JSON filename (ending in ``.json``) or an inline JSON string. The JSON object should be a collection of name:value pairs. ``name`` is string for the cluster name and there **must be** one ``default``.
```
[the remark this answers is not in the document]
dliappis: :+1:

docs/command_line_reference.rst
```diff
@@ -509,3 +510,67 @@ When you run ``esrally list races``, this will show up again::
...
+.. _command_line_reference_advanced_topics:
+
+Advanced topics
+---------------
+
+``target-hosts``
+~~~~~~~~~~~~~~~~
+
+Rally can also create client connections for multiple Elasticsearch clusters.
+This is only useful if you want to create :ref:`custom runners <adding_tracks_custom_runners>` that execute operations against remote clusters, for example for Cross Cluster Search or Cross Cluster Replication.
```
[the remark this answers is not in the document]
dliappis: :+1:

docs/command_line_reference.rst
dliappis: :+1:

docs/command_line_reference.rst
```diff
@@ -509,3 +510,67 @@ When you run ``esrally list races``, this will show up again::
...
+Examples, assuming that two clusters have been specified with ``--target-hosts``:
+
+* json file: ``--client-options="client_options1.json"``::
+
+    {
+      "default":
+        {"timeout": 60},
+      "remote":
+        {"use_ssl":true,"verify_certs": false,"ca_certs":"/path/to/cacert.pem"}
+    }
```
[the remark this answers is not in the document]
dliappis: :+1:

esrally/driver/driver.py
dliappis: Will replace with a function.

esrally/driver/driver.py
```diff
@@ -582,12 +582,18 @@ def __init__(self):
 
     @actor.no_retry("load generator")
     def receiveMsg_StartLoadGenerator(self, msg, sender):
+        def EsClients(all_hosts, all_client_options):
```
danielmitterdorfer: rename to `es_clients`? (I immediately thought that this is a class when I read the code below that calls this function)

esrally/racecontrol.py
```diff
@@ -288,10 +288,10 @@ def race(cfg, sources=False, build=False, distribution=False, external=False, do
 
 def set_default_hosts(cfg, host="127.0.0.1", port=9200):
     configured_hosts = cfg.opts("client", "hosts")
-    if len(configured_hosts()) !=0 :
-        logger.info("Using configured hosts %s" % configured_hosts())
+    if len(configured_hosts.default) !=0 :
+        logger.info("Using configured hosts {}".format(configured_hosts.default))
```
danielmitterdorfer: I think we settled on using the idiom `logger.info("...%s...", parameter0, parameter1)`?

esrally/racecontrol.py
```diff
@@ -288,10 +288,10 @@ def race(cfg, sources=False, build=False, distribution=False, external=False, do
...
     else:
-        logger.info("Setting default host to [%s:%d]" % (host, port))
+        logger.info("Setting default host to [{}:{}]".format(host, port))
```
danielmitterdorfer: I think we settled on using the idiom `logger.info("...%s...", parameter0, parameter1)`?

tests/mechanic/launcher_test.py
```diff
@@ -147,7 +147,7 @@ def test_error_on_cluster_launch(self, sleep):
         cfg = config.Config()
         cfg.add(config.Scope.application, "client", "hosts", self.test_host)
         # Simulate that the client will raise an error upon startup
-        cfg.add(config.Scope.application, "client", "options", config_helper.ClientOptions('raise-error-on-info:true'))
+        cfg.add(config.Scope.application, "client", "options", opts.ClientOptions('raise-error-on-info:true'))
```
danielmitterdorfer: Nit: Use double-quotes for string?

esrally/racecontrol.py
dliappis: Ugh too fast replacing things here. Will revert.

esrally/racecontrol.py
dliappis: Right. Addressed in 0a17876dd048e0720d5104f6d5897d31de5dc94f

esrally/racecontrol.py
dliappis: Addressed in 0a17876dd048e0720d5104f6d5897d31de5dc94f

tests/mechanic/launcher_test.py
dliappis: Addressed in e04df54c7c682095b42e71ae4971cad82586eb73

esrally/driver/driver.py
dliappis: Addressed in 0a17876dd048e0720d5104f6d5897d31de5dc94f

dliappis:
Thank you for the thorough review and insightful comments. With this foundation in, adding telemetry devices for additional clusters should be a simpler task.

I'll squash and merge this in.

PR #489, opened 2018-05-02
Get distribution download URL from car config

danielmitterdorfer:
With this commit we read the download URL for an Elasticsearch
distribution from the car (and similarly for plugins). This is necessary
because with Elasticsearch 6.3 there will be two Elasticsearch
distributions: A pure OSS one and one that includes x-pack. Handling
this in rally.ini would have been possible but it is cleaner leveraging
our team repository infrastructure for this task.

### DIFF
--- a/docs/car.rst
+++ b/docs/car.rst
@@ -57,6 +57,7 @@ The default car definitions of Rally are stored in ``~/.rally/benchmarks/teams/d
         │           └── jvm.options
         ├── ea.ini
         └── vanilla
+            ├── config.ini
             └── templates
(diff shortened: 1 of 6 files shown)


danielmitterdorfer:
I have now documented `config.ini` in https://github.com/elastic/rally/pull/489/commits/8928de3339f6547be56ccad60acfb9b9f65e0790. Can you please have another look @dliappis?

dliappis:
@danielmitterdorfer Thanks for adding the documentation in. LGTM.

PR #497, opened 2018-05-11
Add record_process_stats() for process.* stats

paulcoghlan:
This PR adds the Elasticsearch node process stats (https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html#process-statsnode-stats-include-process) so that CPU metrics can be recorded when running in `pipeline=benchmark-only` mode.

There is a new `node-stats-include-process` telemetry parameter flag to control whether these stats are collected (defaults to `true`).


dliappis:
Thanks @paulcoghlan !

Issue #472, opened 2018-04-18
client-options "verify_certs:false" seems to be ignored


suckowbiz:
**Rally version** (get with `esrally --version`):
0.10.0

**Invoked command**:
```bash
rally --on-error="abort" \
    --report-file=/root/.rally/report.md \
    --track="http_logs" \
    --target-hosts="https://1.2.3.4:9200,https://1.2.3.5:9200" \
    --pipeline=benchmark-only \
    --client-options="use_ssl:true,verify_certs:false,basic_auth_user:'${ESRALLY_USERNAME}',basic_auth_password:'${ESRALLY_PASSWORD}'"
```

**Configuration file (located in `~/.rally/rally.ini`))**:
```
>_ cat rally.ini
[meta]
config.version = 15

[system]
env.name = local

[node]
root.dir = /root/.rally/benchmarks

[runtime]
java.home = /usr/lib/jvm/java-8-openjdk-amd64

[benchmarks]
local.dataset.cache = ${node:root.dir}/data

[reporting]
datastore.type = in-memory
datastore.host =
datastore.port =
datastore.secure =
datastore.user =
datastore.password =

[tracks]
default.url = https://github.com/elastic/rally-tracks

[teams]
default.url = https://github.com/elastic/rally-teams

[defaults]
preserve_benchmark_candidate = False

[distributions]
release.1.url = https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-{{VERSION}}.tar.gz
release.2.url = https://download.elasticsearch.org/elasticsearch/release/org/elasticsearch/distribution/tar/elasticsearch/{{VERSION}}/elasticsearch-{{VERSION}}.tar.gz
release.url = https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-{{VERSION}}.tar.gz
release.cache = true
```

**JVM version**:
```bash
>_ java -version
openjdk version "1.8.0_162"
OpenJDK Runtime Environment (build 1.8.0_162-8u162-b12-1~bpo8+1-b12)
OpenJDK 64-Bit Server VM (build 25.162-b12, mixed mode)
```

**OS version**:
Debian GNU/Linux 8 (jessie)

**Description of the problem including expected versus actual behavior**:
Expected: Rally runs as 0.9.4 did.
Actual: Rally fails with CERTIFICATE_VERIFY_FAILED exception.

**Steps to reproduce**:
 1. Run rally as stated
 
**Provide logs (if relevant)**:
```
[...]

PR #507, opened 2018-05-18
Set retry-on-timeout=true for ES clients used by Telemetry devices

### LINKED ISSUE #502: Set retry-on-timeout=true for telemetry device connections
As noticed in https://discuss.elastic.co/t/rally-track-report-analysis-follow-up/127266/17 , there are occasions where firewalls drop connections when there is no traffic and for benchmarks that take a long time to finish, [GcTimesSummary](https://github.com/elastic/rally/blob/3bd432054f6ab592a4b83ce90fe009eeb28f9754/esrally/mechanic/telemetry.py#L807) can timeout since it creates a connection at the beginning of the benchmark and collects again at the end of the benchmark.

After an internal discussion, we agreed that it makes sense to enable `retry-on-timeout=true` for the client connections used by telemetry devices. This will avoid an unnecessary duplication of requests performed through other connections, such as the ones created in [driver.py](https://github.com/elastic/rally/blob/3bd432054f6ab592a4b83ce90fe009eeb28f9754/esrally/driver/driver.py#L588), for example, used for bulking.

dliappis:
Closes #502

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -69,6 +69,8 @@ class ClusterLauncher:
         for cluster_name, cluster_hosts in all_hosts.items():
             all_client_options = self.cfg.opts("client", "options").all_client_options
             cluster_client_options = all_client_options[cluster_name]
+            # Use retries for telemetry connections: http://elasticsearch-py.readthedocs.io/en/latest/index.html#automatic-retries
+            cluster_client_options.update({"retry-on-timeout": True})
             es[cluster_name] = self.client_factory(cluster_hosts, cluster_client_options).create()
 
         es_default = es["default"]
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -68,9 +68,9 @@ class ClusterLauncher:
         es = {}
         for cluster_name, cluster_hosts in all_hosts.items():
             all_client_options = self.cfg.opts("client", "options").all_client_options
-            cluster_client_options = all_client_options[cluster_name]
-            # Use retries for telemetry connections: http://elasticsearch-py.readthedocs.io/en/latest/index.html#automatic-retries
-            cluster_client_options.update({"retry-on-timeout": True})
+            cluster_client_options = dict(all_client_options[cluster_name])
+            # Use retries to avoid aborts on long living connections for telemetry devices
+            cluster_client_options["retry-on-timeout"] = True
             es[cluster_name] = self.client_factory(cluster_hosts, cluster_client_options).create()
 
         es_default = es["default"]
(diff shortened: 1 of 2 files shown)


esrally/mechanic/launcher.py
```diff
@@ -69,6 +69,8 @@ def start(self):
         for cluster_name, cluster_hosts in all_hosts.items():
             all_client_options = self.cfg.opts("client", "options").all_client_options
             cluster_client_options = all_client_options[cluster_name]
```
danielmitterdorfer: You are modifying the object in the config directly which leads to side-effects, see e.g. https://github.com/elastic/rally/issues/472. Can you please make a copy with:

```python
cluster_client_options = dict(all_client_options[cluster_name])
```

esrally/mechanic/launcher.py
```diff
@@ -69,6 +69,8 @@ def start(self):
...
+            # Use retries for telemetry connections: http://elasticsearch-py.readthedocs.io/en/latest/index.html#automatic-retries
```
danielmitterdorfer: Can you please explain the reason why we're doing this in the comment (e.g. to avoid aborts on long-living connections)?

esrally/mechanic/launcher.py
```diff
@@ -69,6 +69,8 @@ def start(self):
...
+            cluster_client_options.update({"retry-on-timeout": True})
```
danielmitterdorfer: Why not:

```python
cluster_client_options["retry-on-timeout"] = True
```

tests/mechanic/launcher_test.py
```diff
@@ -141,6 +141,21 @@ def test_launches_cluster_without_post_launch_handler(self):
...
+        cfg.add(config.Scope.application, "mechanic", "telemetry.devices", [])
+        cfg.add(config.Scope.application, "mechanic", "telemetry.params", {})
+
+        cluster_launcher = launcher.ClusterLauncher(cfg, MockMetricsStore(), client_factory_class=MockClientFactory)
+        cluster = cluster_launcher.start()
+
+        retry_timeout_setting = {"retry-on-timeout": True}
+
+        for telemetry_device in cluster.telemetry.devices:
+            self.assertLessEqual(retry_timeout_setting.items(),telemetry_device.client.client_options.items())
```
danielmitterdorfer: We could use `assertDictEqual` here to check whether the client dictionary contains only expected options?

dliappis: My thought here is that this test cares about ensuring that the specific option is set and we won't need to update it, if, for whatever reason, other client option settings, specified in the Class Attributes, change in the future. I don't mind changing it to what you are suggesting though.

esrally/mechanic/launcher.py
dliappis: Great catch thanks!

dliappis:
Thanks for the review! I addressed the comments can you take another pass please?

PR #510, opened 2018-05-24
Allow conflict-probability value of 0

dliappis:
For the use case of externally generated ids (no updates), allow
setting conflict-probability to the value 0.

PR #511, opened 2018-05-25
Don't fail if conflict-probability is zero

dliappis:
PR #510 doesn't allow setting conflict probability to 0 as the
parameter check is failing, assuming only open intervals are allowed.

Fix bug and also add a number of unit tests.

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -4,6 +4,7 @@ import time
 import math
 import types
 import inspect
+import operator
 from enum import Enum
 
 from esrally import exceptions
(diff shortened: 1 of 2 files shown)


esrally/track/params.py
```diff
@@ -440,7 +441,7 @@ def __init__(self, track, params, **kwargs):
             raise exceptions.InvalidSyntax("Unknown 'conflicts' setting [%s]" % id_conflicts)
 
         if self.id_conflicts != IndexIdConflict.NoConflicts:
-            self.conflict_probability = self.float_param(params, name="conflict-probability", default_value=25, min_value=0, max_value=100)
+            self.conflict_probability = self.float_param(params, name="conflict-probability", default_value=25, min_value=0, max_value=100, min_operator=operator.lt)
```
danielmitterdorfer: Interesting idea using `operator`.

PR #512, opened 2018-05-25
Make challenges optional

### LINKED ISSUE #474: Allow to use Rally without knowing about challenges
_Note: This is a follow-up from #455._

### Purpose

In order to make it simpler for new users, "challenges" will become optional in Rally. We will however not get rid of them because it makes sense to keep them to provide structure for more complex tracks.

### Tasks

* [x] Simplify the track format so if there is only one challenge, users can specify the `schedule` element on top-level.
* [x] Hide the notion of a challenge from UI output in these simple cases.
* [x] State more clearly in the docs that challenges are a more advanced concept and that it is not necessary to use them when getting started.

danielmitterdorfer:
With this commit we make challenges an optional concept in Rally. We
modify the track syntax so it also allows to specify the `schedule`
property on top-level. We also hide the notion of a challenge from
console messages if possible.

Closes #474

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -816,7 +816,7 @@ Each challenge consists of the following properties:
 
 * ``name`` (mandatory): A descriptive name of the challenge. Should not contain spaces in order to simplify handling on the command line for users.
 * ``description`` (optional): A human readable description of the challenge.
-* ``default`` (optional): If true, Rally selects this challenge by default if the user did not specify a challenge on the command line. If your track only defines one challenge, it is implicitly selected as default, otherwise you need define ``"default": true`` on exactly one challenge.
+* ``default`` (optional): If true, Rally selects this challenge by default if the user did not specify a challenge on the command line. If your track only defines one challenge, it is implicitly selected as default, otherwise you need to define ``"default": true`` on exactly one challenge.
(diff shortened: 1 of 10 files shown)


docs/track.rst
```diff
@@ -831,6 +805,23 @@ All tasks in the ``schedule`` list are executed sequentially in the order in whi
...
+
+If your track defines only one benchmarking scenario specify the ``schedule`` on top-level. Use the ``challenge`` element if you want to specify additional properties like a name or a description. You can think of a challenge as a benchmarking scenario. If you have multiple challenges, you can define an array of ``challenges``.
+
+This section contains one or more challenges which describe the benchmark scenarios for this data set. A challenge can reference all operations that are defined in the ``operations`` section.
+
+Each challenge consists of the following properties:
+
+* ``name`` (mandatory): A descriptive name of the challenge. Should not contain spaces in order to simplify handling on the command line for users.
+* ``description`` (optional): A human readable description of the challenge.
+* ``default`` (optional): If true, Rally selects this challenge by default if the user did not specify a challenge on the command line. If your track only defines one challenge, it is implicitly selected as default, otherwise you need define ``"default": true`` on exactly one challenge.
```
dliappis: > otherwise you need define ``"default": true`` on exactly one challenge.

you need **to** define

PR #514, opened 2018-05-30
Report indexing times per shard

### LINKED ISSUE #374: Overhaul confusing index time related metrics
Currently we report various indexing time related metrics like which are based on the indices stats API. However, these numbers are the sum on index level (i.e. sum of all shards). This means that the number is dependent on shard count and will be usually larger than the benchmark's Wall clock time both of which can be confusing to users.

Hence, we should report these numbers on shard level instead and provide min/max values.

danielmitterdorfer:
With this commit we do not only show the total indexing times (across
all shards) but also report the indexing times per (primary) shard.

Closes #374

### DIFF
--- a/esrally/reporter.py
+++ b/esrally/reporter.py
@@ -3,6 +3,7 @@ import csv
 import io
 import sys
 import logging
+import statistics
 
 import tabulate
 from esrally import metrics, exceptions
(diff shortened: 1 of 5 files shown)


esrally/reporter.py
```diff
@@ -665,19 +705,38 @@ def report_merge_part_times(self, baseline_stats, contender_stats):
         )
 
     def report_total_times(self, baseline_stats, contender_stats):
+        totals = []
+        totals += self.report_total_time("indexing time",
```
dliappis: As discussed, `append()` could be a better choice as `+=` with lists may produce unexpected results if the input is a string vs a list.

danielmitterdorfer: Yes, good suggestion! I'll merge this PR and do this separately.

danielmitterdorfer: Addressed in https://github.com/elastic/rally/commit/8d7a799455c63b2ac30c728b23f946facf08e19a.

PR #520, opened 2018-06-13
Specify 1 for number of shards for metrics store indices

dliappis:
Align with the new Elasticsearch defaults[1] for number of shards for
the indices used in the metrics store.

[1] https://github.com/elastic/elasticsearch/pull/30539

### DIFF
--- a/esrally/resources/metrics-template.json
+++ b/esrally/resources/metrics-template.json
@@ -2,6 +2,7 @@
   "template": "rally-metrics-*",
   "settings": {
     "index": {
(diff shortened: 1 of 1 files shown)


dliappis:
We could probably do the same for `rally-results-*` but the amount of data is much less so didn't consider this necessary.

danielmitterdorfer:
I'd rather use one shard for all indices (metrics, results and races). Given that you are not available within the next weeks (and this is a simple change) I'd just close this one and create one that specifies it for all indices.

PR #529, opened 2018-06-28
Allow to control recency of ids in conflicts

danielmitterdorfer:
With this commit we add a new parameter to the `bulk` parameter source
called `recency` that allows to control the bias towards more recent
ids when generating id conflicts.


esrally/track/params.py
```diff
@@ -692,7 +707,21 @@ def __iter__(self):
...
-                doc_id = self.conflicting_ids[self.randint(0, self.id_up_to - 1)]
+                # a recency of zero means that we don't care about recency and just take a random number
+                # within the whole interval.
+                if self.recency == 0:
+                    idx = self.randint(0, self.id_up_to - 1)
+                else:
+                    # A recency > 0 biases id selection towards more recent ids. The recency parameter decides
+                    # by how much we bias. See docs for the resulting curve.
+                    #
+                    # idx_range is in the interval [0;1].
```
dliappis: I guess there's a typo here, `[0;1]` -> `[0,1]`

danielmitterdorfer: It was actually intentional. The German decimal separator is a comma so it could be misinterpreted by a German as 1/10 instead of the interval from zero to 1 (inclusive). However, we can change it as well.

esrally/track/params.py
```diff
@@ -692,7 +707,21 @@ def __iter__(self):
...
+                    idx_range = min(self.randexp(GenerateActionMetaData.RECENCY_SLOPE * self.recency), 1)
+                    # the resulting index is in the range [0, self.id_up_to). Note that a smaller idx_range
+                    # biases towards more recently used ids (higher indexes).
+                    idx = int(round((self.id_up_to - 1) * (1 - idx_range)))
```
dliappis: Does `round()` anyway return an `int`, if the optional [ndigits parameter](https://docs.python.org/3/library/functions.html#round) is not passed?

No harm being explicit here, but wondering about it.

danielmitterdorfer: Oh, you are right. I was probably assuming that it would behave similarly to `Math#round`in Java, hence the explicit conversion.

esrally/track/params.py
danielmitterdorfer: I changed it anyway now as I've seen that we use the comma notation in all other places.

PR #538, opened 2018-07-23
Retry metric store for more transport errors

dliappis:
Improve the resiliency of Rally when using an Elasticsearch remote metrics store by retrying certain transport errors.

Also enhance existing tests to check that retries/debug output and sleep are executed properly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,7 @@ ENV HOME /home/${NEW_USER}
 ENV PYENV_ROOT=/home/${NEW_USER}/.pyenv
 ENV PATH=$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
 ENV JAVA_HOME=/opt/jdk-10
+ENV JAVA10_HOME=/opt/jdk-10
 ENV RUNTIME_JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
 
 RUN pyenv install 3.4.8
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,6 @@ ENV HOME /home/${NEW_USER}
 ENV PYENV_ROOT=/home/${NEW_USER}/.pyenv
 ENV PATH=$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
 ENV JAVA_HOME=/opt/jdk-10
-ENV JAVA10_HOME=/opt/jdk-10
 ENV RUNTIME_JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
 
 RUN pyenv install 3.4.8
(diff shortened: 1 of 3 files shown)


tests/metrics_test.py
```diff
@@ -137,33 +138,163 @@ def raise_unknown_error():
...
 
-    def test_retries_on_timeouts(self):
-        gateway_timeout = elasticsearch.exceptions.TransportError(504, "Gateway timeout")
+    @mock.patch("esrally.time.sleep")
+    def test_retries_on_bad_gateway(self, mocked_sleep):
+        bad_gateway = elasticsearch.exceptions.TransportError(502, "Bad Gateway")
 
         # should return on first success
-        operation = mock.Mock(side_effect=[gateway_timeout, gateway_timeout, "success", gateway_timeout])
+        operation = mock.Mock(side_effect=[bad_gateway, bad_gateway, "success", bad_gateway])
```
danielmitterdorfer: Maybe randomize and vary the error type here? I think then we could also reduce the number of individual test cases?

esrally/metrics.py
```diff
@@ -93,13 +94,11 @@ def guarded(self, target, *args, **kwargs):
...
                 raise exceptions.SystemSetupError(msg)
             except elasticsearch.TransportError as e:
-                # gateway timeout - let's wait a bit and retry
-                if e.status_code == 504 and execution_count < max_execution_count:
-                    self.logger.debug("Gateway timeout in attempt [%d/%d].", execution_count, max_execution_count)
-                    time.sleep(1)
-                elif e.status_code == 429 and execution_count < max_execution_count:
-                    self.logger.debug("Execution rejected in attempt [%d/%d].", execution_count, max_execution_count)
-                    time.sleep(3)
+                retriable_responses_with_sleep = {502: 1, 503: 1, 504: 1, 429: 3}
```
danielmitterdorfer: I think it would also be ok if we settle on a single waiting period. Wdyt?

Dockerfile
```diff
@@ -25,6 +25,7 @@ ENV HOME /home/${NEW_USER}
 ENV PYENV_ROOT=/home/${NEW_USER}/.pyenv
 ENV PATH=$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH
 ENV JAVA_HOME=/opt/jdk-10
+ENV JAVA10_HOME=/opt/jdk-10
```
danielmitterdorfer: This seems unrelated?

dliappis: Yes this came from this commit https://github.com/elastic/rally/pull/538/commits/9ddb3933dbb49853ea77540fa3c198d02bdf0cbe to fix the it tests in Docker. I'll pull it out and submit separately.

dliappis:
@danielmitterdorfer Thanks for the review! I changed the logic, as discussed, to use randomized tests and fixed retry periods. Would appreciate another look when you have time.

dliappis:
Thank you for the review @danielmitterdorfer !

PR #544, opened 2018-07-31
Fix parsing of boolean values datastore.secure

### LINKED ISSUE #543: datastore.secure boolean values seem to be case sensitive in rally.ini
Specifying `datastore.secure = true` in `rally.ini` gives the misleading error message:

```
esrally.exceptions.SystemSetupError: ('Could not connect to your Elasticsearch metrics store. Please check that it is running on host [https://<url>] at port [9243] or fix the configuration in [<home_dir>/.rally/rally.ini].', None)
].
```
It works when setting it to `True`, but we should be case insensitive when parsing boolean values in this file.

dliappis:
Allow any capitalization for `datastore.secure` configuration setting in reporting section of `rally.ini`.
Also add randomized tests cases for the parsing of reporting configuration options.

Resolves #543

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -9,7 +9,7 @@ from enum import Enum, IntEnum
 
 import tabulate
 from esrally import time, exceptions, config, version, paths
-from esrally.utils import console, io, versions
+from esrally.utils import convert, console, io, versions
 from http.client import responses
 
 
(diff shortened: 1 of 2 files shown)


esrally/metrics.py
```diff
@@ -123,8 +123,7 @@ def __init__(self, cfg):
         self._config = cfg
         host = self._config.opts("reporting", "datastore.host")
         port = self._config.opts("reporting", "datastore.port")
-        # poor man's boolean conversion
-        secure = self._config.opts("reporting", "datastore.secure") == "True"
+        secure = str(self._config.opts("reporting", "datastore.secure")).capitalize() == "True"
```
danielmitterdorfer: I think we could even use `convert.to_bool()` here (similarly to treatment of other booleans)?

dliappis: Great suggestion! I've switch to that now.

PR #546, opened 2018-08-07
Update ccr stats endpoint

dliappis:
Update the endpoint and the unit tests to reflect the changes brought
in [1]

Also simplify filtering ccr-stats related documents by adding 
`meta.name=ccr-stats` to the metric documents, similarly to the work 
done for node-stats in [2].

[1] elastic/elasticsearch#32350
[2] 9b8974f

### DIFF
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -416,7 +416,8 @@ class CcrStatsRecorder:
                 shard_metadata = {
                     "cluster": self.cluster_name,
                     "index": name,
-                    "shard": shard_id
+                    "shard": shard_id,
+                    "name": "ccr-stats"
                 }
 
                 self.metrics_store.put_doc(shard_stats, level=MetaInfoScope.cluster, meta_data=shard_metadata)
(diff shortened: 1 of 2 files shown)


esrally/mechanic/telemetry.py
```diff
@@ -410,14 +410,17 @@ def record_stats_per_index(self, name, stats):
...
 
-        for shard_num, shard_stats in stats.items():
-            shard_metadata = {
-                "cluster": self.cluster_name,
-                "index": name,
-                "shard": shard_num
-            }
+        for shard_stats in stats:
+            shard_id = shard_stats.get("shard_id", None)
+            if shard_id is not None:
```
danielmitterdorfer: Probably a matter of taste but you could also write this as `if "shard_id" in shard_stats` and then use `shard_stats["shard_id"]` in the body.

dliappis: Thanks for the suggestion I prefer this as well; addressed in 581a69f85ed67f9f0d363f3166d3683c7de36ca0

Issue #548, opened 2018-08-10
Add exponential back off mechanism while retrying transport errors to metrics store


dliappis:
The solution introduced in 77245264e7b2f08d62729ef45f6accd1c98ba50b to retry certain TransportErrors while writing to the metrics store is not sufficient as there are only 3 retries, 3 seconds each.

We need an exponential back off mechanism instead with a larger amount of retries to increase resiliency for errors.

Note that we still need to investigate how this blocks which thread, but will track it in a separate issue.

dliappis:
cc @dakrone

PR #549, opened 2018-08-10
Exponential back-off for retrying transport errors to metrics store

dliappis:
Introduce exponential back-off mechanism for waiting between retries
when TransportErrors occurs towards the metrics store.  Increase
amount of execution attempts to 11 (10 retries) so that total time
spent in retries is 17.05min.

Closes https://github.com/elastic/rally/issues/548

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -2,8 +2,10 @@ import collections
 import logging
 import math
 import pickle
+import random
 import statistics
 import sys
+import time
 import zlib
 from enum import Enum, IntEnum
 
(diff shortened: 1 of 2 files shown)


esrally/metrics.py
```diff
@@ -53,11 +55,13 @@ def search(self, index, doc_type, body):
 
     def guarded(self, target, *args, **kwargs):
         import elasticsearch
-        max_execution_count = 3
+        max_execution_count = 11
         execution_count = 0
 
         while execution_count < max_execution_count:
+            time_to_sleep = 2 ** execution_count + (random.randint(0, 1000) / 1000)
```
danielmitterdorfer: Any reason you use `random.randint(0, 1000) / 1000` instead of just `random.random()`?

danielmitterdorfer: Might also be worth adding a debug log statement?

dliappis: D'oh was thinking too much in milliseconds.

PR #555, opened 2018-08-24
Load only required corpora

### LINKED ISSUE #553: Only load required corpora
Currently Rally is loading all referenced corpora of track, even if they are not needed. As it is possible that we download and extract a significant amount of data (usually several GB if not more), we should investigate whether we can load only the required ones.

danielmitterdorfer:
With this commit Rally will only load the required corpora / document
sets that are actually referenced by the tasks in the currently selected
challenge. This saves race setup time by avoiding unneeded downloads and
decompression of document sets.

Closes #553

### DIFF
--- a/esrally/track/track.py
+++ b/esrally/track/track.py
@@ -215,7 +215,8 @@ class DocumentCorpus:
         return DocumentCorpus(self.name, filtered)
 
     def union(self, other):
-        assert self.name == other.name, "Both document corpora must have the same name"
+        if self.name != other.name:
+            raise exceptions.RallyAssertionError("Both document corpora must have the same name")
         if self is other:
             return self
         else:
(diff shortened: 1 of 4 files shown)


esrally/track/track.py
```diff
@@ -214,6 +214,13 @@ def filter(self, source_format=None, target_indices=None):
             filtered.append(d)
         return DocumentCorpus(self.name, filtered)
 
+    def union(self, other):
+        assert self.name == other.name, "Both document corpora must have the same name"
```
dliappis: We've discussed this for other cases before, but would it be more consistent with existing behavior to throw an exception here?

danielmitterdorfer: I can address this, sure.

PR #560, opened 2018-08-30
Disable automatic log rotation

### LINKED ISSUE #539: Externalize log rotation
In #503 we implemented that Rally's logging can be configured via a file. Our default configuration contains the following handler definition:

```json
    "rally_log_handler": {
      "class": "logging.handlers.TimedRotatingFileHandler",
      "filename": "${LOG_PATH}/rally.log",
      "utc": true,
      "when": "midnight",
      "backupCount": 14,
      "encoding": "UTF-8",
      "formatter": "normal",
      "filters": ["isActorLog"]
    }
```

This is problematic because Rally uses multiple processes and there is a lurking race condition when rolling log files (which we had already found in #263). Hence, we will stop rotating logs from within Rally and leverage established tools like `logrotate` for that.

### Tasks

* [x] Change the current log configuration template to use [`WatchedFileHandler`](https://docs.python.org/3/library/logging.handlers.html#watchedfilehandler) which is meant to be used with tools like `logrotate`
* [x] Determine a suitable `logrotate` configuration for Rally and provide a snippet in our docs.
* [x] Implement a check on Rally startup that ensures that the new configuration file template is used  (backup the old one if present and replace it with the new one)

Note: `WatchedFileHandler` uses the `stat` system call to check whether the inode of the log file has changed to determine whether to reopen the file. The file system metadata are cached. Also, this check is done every time a log line is written. In steady state this happens only a couple of times per minute in the default logging configuration. Therefore, we consider the performance impact of using `WatchedFileHandler` by default negligible.

danielmitterdorfer:
With this commit we disable automatic log rotation in Rally's
out-of-the-box logging configuration because it can lead to race
conditions (multiple processes attempting to rollover the log file).

We also migrate existing configurations if possible (i.e. the user did
not modify it) and provide an example configuration for `logrotate`.

Closes #539

### DIFF
--- a/esrally/resources/logging_1_0_0.json
+++ b/esrally/resources/logging_1_0_0.json
@@ -19,8 +19,11 @@
   },
   "handlers": {
     "rally_log_handler": {
-      "class": "logging.handlers.WatchedFileHandler",
+      "class": "logging.handlers.TimedRotatingFileHandler",
       "filename": "${LOG_PATH}/rally.log",
+      "utc": true,
+      "when": "midnight",
+      "backupCount": 14,
       "encoding": "UTF-8",
       "formatter": "normal",
       "filters": ["isActorLog"]
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -119,7 +119,19 @@ Logging in Rally is configured in ``~/.rally/logging.json``. For more informatio
 * The Python reference documentation on the `logging configuration schema <https://docs.python.org/3/library/logging.config.html#logging-config-dictschema>`_ explains the file format.
 * The `logging handler documentation <https://docs.python.org/3/library/logging.handlers.html>`_ describes how to customize where log output is written to.
 
-By default, Rally will log all output to ``~/.rally/logs/rally.log`` and rotate this file daily.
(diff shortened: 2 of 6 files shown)


esrally/resources/logging_1_0_0.json
```diff
@@ -0,0 +1,55 @@
...
+      "level": "WARNING",
+      "propagate": false
+    },
+    "rally.profile": {
+      "handlers": ["rally_profile_handler"],
+      "level": "INFO",
+      "propagate": false
+    }
+  }
+}
```
dliappis: I think we need a newline here, not just because of POSIX, but since earlier we read the file with `read()` the result will be something like: `contents='line1\nline2'` while the `logging.json` that Rally created itself in versions <1.0.1 has `\n` in the last line, so contents wouldn't match.

dliappis: Actually checked things again on a fresh installation on both MacOS and Linux and seems that using the currently released version of Rally, the `logging.json` file does **not** have a `\n` at the last line, so ignore my comment.

danielmitterdorfer:
Thanks for the review!

PR #563, opened 2018-09-03
Make Rally compatible with Python 3.7

danielmitterdorfer:
With this commit we make Rally compatible with Python 3.7. The main
change is due to [PEP 479](https://www.python.org/dev/peps/pep-0479/).
Apart from that we do a couple of smaller cleanups and document it in
our setup file.


danielmitterdorfer:
I suggest we target this change for release 1.0.2 instead of 1.0.1.

PR #569, opened 2018-09-06
Add shrink index runner

danielmitterdorfer:
With this commit we add a new runner to Rally that executes the shrink
index API and all other relevant API calls in order to shrink an index.


danielmitterdorfer:
If you want to try this at home: I've used a [modified version of the geonames track](https://github.com/elastic/rally/files/2357046/shrink-geonames.zip) for my tests.

danielmitterdorfer:
Thanks for the review!

PR #571, opened 2018-09-14
Reduce logging when loading tracks

danielmitterdorfer:
With this commit we reduce the log level to DEBUG for several log
entries when loading tracks so we create less noise for standard cases.


danielmitterdorfer:
Thanks for the review!

PR #575, opened 2018-09-19
Always use remote branch for updates

### LINKED ISSUE #573: es download url below 6.3.0 version is not correct
- rally version: 1.0.1

when rally download es, it's url is "http://xxxx/elasticsearch-oss-6.2.2.tar.gz", but correct url should be "http://xxxx/elasticsearch-6.2.2.tar.gz".  The former url is not applied to versions below 6.3.0.

danielmitterdorfer:
With this commit we always attempt to checkout the remote branch first.
In a scenario where a user has already cloned the repository but uses
the `--offline` flag the correct branches might not be proper local
tracking branches but are already available on the target machine as
remote branches.

Closes #573

### DIFF
--- a/esrally/utils/repo.py
+++ b/esrally/utils/repo.py
@@ -35,7 +35,7 @@ class RallyRepository:
 
     def update(self, distribution_version):
         try:
-            if self.remote and not self.offline:
+            if self.remote:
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Thanks for the review and no worries. :)

Issue #327, opened 2017-09-26
Improve formatting of document count in list tracks output


danielmitterdorfer:
When running `esrally list tracks`, Rally only show the number of documents as raw number without any formatting which is hard to read. We should add thousand separators and also show a human readable summary, e.g. instead of showing `11396505`, we could print `11,396,505`.

muralimani:
Daniel, I am interested in providing fix for this improvement. Please let me know the procedure to start on the fix.

danielmitterdorfer:
Hi @muralimani! I am sorry but I missed your comment on this ticket and I just saw it because I went to this ticket from a PR that has already been opened. If you are interested in contributing in general, another issue that might be worthwhile is #382. We also have a [contribution guide](https://github.com/elastic/rally/blob/master/CONTRIBUTING.md) available in this repository.

PR #576, opened 2018-09-20
Improved formatting of document count in list tracks output

akhil-rane:
Added thousand seperators to document count in list tracks output. E.g. instead of showing 11396505, new change will show 11,396,505

Relates #327

### DIFF
--- a/esrally/utils/convert.py
+++ b/esrally/utils/convert.py
@@ -25,8 +25,8 @@ def bytes_to_human_string(b):
     return "%d bytes" % b
 
 
-def number_of_documents_to_human_string(number_of_documents):
-    return "{:,}".format(number_of_documents)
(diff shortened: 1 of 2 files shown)


esrally/utils/convert.py
```diff
@@ -25,6 +25,10 @@ def bytes_to_human_string(b):
     return "%d bytes" % b
 
 
+def number_of_documents_to_human_string(number_of_documents):
```
danielmitterdorfer: I think this method name is too special as it can be used to format any number. Can you please rename it to `number_to_human_string`?

akhil-rane: Fixed it

akhil-rane:
Thanks @danielmitterdorfer

PR #587, opened 2018-10-18
Fix typos and inconsistencies in help documentation

msufa:
Start each description with an uppercase letter and end with a period.

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -28,7 +28,7 @@ def create_arg_parser():
         preserve_install = False
 
     parser = argparse.ArgumentParser(prog=PROGRAM_NAME,
-                                     description=BANNER + "\n\n You know for benchmarking Elasticsearch.",
+                                     description=BANNER + "\n\n You Know, for Benchmarking Elasticsearch.",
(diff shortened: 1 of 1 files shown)


esrally/rally.py
```diff
@@ -143,147 +143,147 @@ def positive_number(v):
...
             default="")
         p.add_argument(
             "--plugin-params",
-            help="define a comma-separated list of key:value pairs that are injected verbatim to all plugins as variables",
+            help="Define a comma-separated list of key:value pairs that are injected verbatim to all plugins as variables.",
             default=""
         )
         p.add_argument(
             "--target-hosts",
-            help="define a comma-separated list of host:port pairs which should be targeted iff using the pipeline 'benchmark-only' "
```
danielmitterdorfer: fyi: "iff" is not necessarily a typo. It is short-hand for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if) but I admit it's not widely known. However, let's keep your change here.

msufa: I did not know that, thanks for the explanation @danielmitterdorfer 

Anyway, I feel that for the vast majority of people looking at the docs the fixed spelling is going to look much less suspicious :)

Issue #439, opened 2018-03-13
Manage development dependencies in setup.py


danielmitterdorfer:
Currently maintainers need to install development dependencies manually. However, we can use the `extras_require` functionality to declare optional dependencies for development in `setup.py` and simplify development setup for new maintainers (see e.g. https://github.com/elastic/elasticsearch-py/blob/master/setup.py#L66-L68 for an example; use `pip install -e .[develop]` to install).

danielmitterdorfer:
Note that when addressing this we should also cleanup the workaround implemented in https://github.com/elastic/rally/pull/599.

PR #599, opened 2018-11-14
Simplify development setup

danielmitterdorfer:
With this commit we amend the `make install` command to also include
development dependencies so developers don't need to perform this step
manually when setting up Rally. However, this is only a workaround and
should be properly addressed by #439.

Relates #439


danielmitterdorfer:
I am personally happy with a Docker-less workflow but I don't see a reason why we should not introduce `docker-docs`. :)

PR #612, opened 2018-11-30
Extract hits either as number or structured object

### LINKED ISSUE #611: Extract total hits as either structured object or number
With https://github.com/elastic/elasticsearch/pull/35849, Elasticsearch will return total hits as an object, e.g.:

```json
{
  "hits": {
    "total": {
      "value": 100,
      "relation": "eq"
    }
  }
}
```

but so far, Rally assumes the following structure:

```json
{
  "hits": {
    "total": 100
  }
}
```

Rally needs to be able to extract the information about the total hits in both scenarios (hits as int and as structured object). For backwards compatibility we should store meta-data in Rally in two properties: 

* `hits` (as is) which contains the actual value (extracted either from `hits.total` or `hits.total.value`)
* `hits_relation` (new) which contains either `"eq"` if `hits.total` is a number or the value of `hits.total.relation`.

Implementation note: This requires changes in `esrally.driver.runner.Query#request_body_query()`.

danielmitterdorfer:
With this commit we add the ability to extract metadata about the hit
count of a search request regardless whether Elasticsearch returns it as
a number (versions before ES 7.0) or as a structured object (as of ES
7.0). We also introduce a new metadata field `hits_relation` which
indicates whether the hit count by Elasticsearch is accurate or an
indication of a lower bound.

Relates elastic/elasticsearch#35849
Closes #611

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -551,6 +551,7 @@ class Query(Runner):
                   Always 1 for normal queries and the number of retrieved pages for scroll queries.
     * ``unit``: The unit in which to interpret ``weight``. Always "ops".
     * ``hits``: Total number of hits for this operation.
+    * ``hits_relation``: whether ``hits`` is accurate (``eq``) or a lower bound of the actual hit count (``gte``).
     * ``timed_out``: Whether the search has timed out. For scroll queries, this flag is ``True`` if the flag was ``True`` for any of the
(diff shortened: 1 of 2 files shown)


ebadyano:
lgtm

danielmitterdorfer:
Thank you both! I just merged it. :)

PR #617, opened 2018-12-18
Fix conflicting pipelines and distribution version

dliappis:
Users frequently combine --distribution-version together with non
compatible pipelines.

Exit with a descriptive error message and don't allow choosing an
incompatible pipeline together with with `--distribution-version`.

### DIFF
--- a/esrally/racecontrol.py
+++ b/esrally/racecontrol.py
@@ -368,7 +368,9 @@ def list_pipelines():
 def run(cfg):
     logger = logging.getLogger(__name__)
     name = cfg.opts("race", "pipeline")
+
     if len(name) == 0:
+        # assume from-distribution pipeline if distribution.version has been specified and --pipeline cli arg not set
         if cfg.exists("mechanic", "distribution.version"):
             name = "from-distribution"
         else:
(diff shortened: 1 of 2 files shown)


esrally/racecontrol.py
```diff
@@ -368,14 +368,24 @@ def list_pipelines():
...
         cfg.add(config.Scope.applicationOverride, "race", "pipeline", name)
     else:
+        if (cfg.exists("mechanic", "distribution.version") and
+                name in ["from-sources-complete", "from-sources-skip-build", "benchmark-only"]):
+            raise exceptions.SystemSetupError(
+                "--distribution-version can only be used together with pipeline from-distribution, """
+                "but you specified {}.\n"
+                "If you intend to benchmark an externally provisioned cluster, don't specify --distribution-version otherwise\n"
+                "please read the docs for from-distribution pipeline at "
+                "https://esrally.readthedocs.io/en/stable/pipelines.html#from-distribution".format(name))
```
danielmitterdorfer: Can you please use the `DOC_LINK` constant here? https://github.com/elastic/rally/blob/be6336c04498e60d8ede05a21280704fa1072cc1/esrally/__init__.py#L11-L14

dliappis: Thanks addressed in https://github.com/elastic/rally/pull/617/commits/c4b144fee48a2707f44fcfd4c4843f701b7f3ac2.

Also realized that `del p` that we used to have to remove the additional test-unit pipeline wasn't working and now that I added a new unit test `test_conflicting_pipeline_and_distribution_version` that precedes alphabetically `test_finds_available_pipelines`, the latter fails. Fixed in the same commit.

PR #625, opened 2019-01-08
Allow to specify a team revision

danielmitterdorfer:
With this commit we add the ability to override Rally's logic to match
rally-teams with an Elasticsearch version. This allows us to benchmark
arbitrary historic commits of Elasticsearch even if incompatible
configuration changes have been introduced meanwhile.


danielmitterdorfer:
Thanks for the quick review!

PR #627, opened 2019-01-09
Warn about skewed results when using node-stats telemetry device

### LINKED ISSUE #608: Warn when enabling node-stats
When `docs` / `completion` is requested via node stats (which is done by the node stats telemetry device) it will trigger additional refreshes in Elasticsearch which will skew results and highly likely make benchmark results invalid. We should document that and maybe even emit a warning when the node stats telemetry device is turned on.

dliappis:
With this PR when executing Rally with the node-stats telemetry device enabled a warning gets printed:

```

[WARNING] You have enabled the node-stats telemetry device, but requests to the _nodes/stats Elasticsearch endpoint
          trigger additional refreshes and WILL SKEW results.
          If you still require this, consider reducing the interval.
```

Also adjust docs.

Closes #608

### DIFF
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -443,14 +443,19 @@ class CcrStatsRecorder:
 
 
 class NodeStats(TelemetryDevice):
+    """
+    Gathers different node stats.
+    """
+
     internal = False
     command = "node-stats"
     human_name = "Node Stats"
     help = "Regularly samples node stats"
+    warning = """You have enabled the node-stats telemetry device, but requests to the _nodes/stats Elasticsearch endpoint
+          trigger additional refreshes and WILL SKEW results.
+          If you still require this, consider reducing the interval.
+    """
 
-    """
-    Gathers different node stats.
-    """
     def __init__(self, telemetry_params, clients, metrics_store):
         super().__init__()
         self.telemetry_params = telemetry_params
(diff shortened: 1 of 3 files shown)


esrally/mechanic/telemetry.py
```diff
@@ -443,14 +443,19 @@ def record_stats_per_index(self, name, stats):
...
+
     internal = False
     command = "node-stats"
     human_name = "Node Stats"
     help = "Regularly samples node stats"
-
-    """
-    Gathers different node stats.
+    warning = """You have enabled the node-stats telemetry device, but requests to the _nodes/stats Elasticsearch endpoint
+          trigger additional refreshes and WILL SKEW results.
```
danielmitterdorfer: Not sure if we need to upper-case?

esrally/mechanic/telemetry.py
```diff
@@ -443,14 +443,19 @@ def record_stats_per_index(self, name, stats):
...
+          If you still require this, consider reducing the interval.
```
danielmitterdorfer: This implies that a reduced interval might be ok but it still might skew results.

esrally/mechanic/telemetry.py
dliappis: Just wanted to really ensure this is visible (because there may be other info/warn messages).

esrally/mechanic/telemetry.py
dliappis: Agreed. Less is more in this case, will remove.

dliappis: Addressed in https://github.com/elastic/rally/pull/627/commits/698621aaea06c113bc8015a14f505d678f8c3669

PR #630, opened 2019-01-22
Improve error message on missing repo directory

danielmitterdorfer:
So far we only check whether repo is a valid working copy and raise an
error if it is not. With this commit we add a further check if the
directory exists at all (because that case needs to be fixed differently
by the user).


danielmitterdorfer:

PR #634, opened 2019-01-24
Fix typo

breml:


danielmitterdorfer:
Good catch! Thanks for providing a fix. :)

PR #638, opened 2019-01-29
Correct recorder-based sampling interval

danielmitterdorfer:
With this commit we remove the call to `time.sleep` from the node stats and CCR
stats recorder's `record` method. Thes calls are redundant because the sampler
thread that manages the recorder already calls `time.sleep`. This means that the
effective sample rate is effectively half of the specified one.


danielmitterdorfer:
Thanks, merged!

PR #642, opened 2019-02-06
Allow passing any request parameter to the create index API

dliappis:
https://github.com/elastic/rally-tracks/pull/64 requires passing a
parameter in the create-index runner, however, the underlying ES Python
client method[1] doesn't support it. Additionally our docs[2] hint
that this is possible, but it isn't.
This leads to errors like: 
`create() got an unexpected keyword argument 'include_type_name'`.

This isn't caught by unit tests as the ES client is mocked and an
integration test is needed (TODO in a separate PR).

Directly call es.transport.perform_request to create an index, allowing
passing of any arbitrary parameter (such as `include_type_name`) to
the create index Elasticsearch API.

[1] https://elasticsearch-py.readthedocs.io/en/master/api.html?highlight=indices%20create#elasticsearch.client.IndicesClient.create
[2] https://esrally.readthedocs.io/en/stable/track.html#create-index

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -783,7 +783,12 @@ class CreateIndex(Runner):
         indices = mandatory(params, "indices", self)
         request_params = params.get("request-params", {})
         for index, body in indices:
-            es.indices.create(index=index, body=body, **request_params)
+            # We don't use es.indices.create() because it doesn't support params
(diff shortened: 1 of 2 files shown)


dliappis:
We'll need to add an integration test to catch what the unit test didn't catch here (i.e. passing parameters for create-index), but I thought it can be done in a separate PR.

PR #644, opened 2019-02-12
Allow collection of jvm gc section in node-stats telemetry device

dliappis:
We already collect the mem section under node-stats/jvm but GC stats
can also be very useful for understanding the jvm memory behavior during
a benchmark.

Add `node-stats-include-gc` telemetry parameter (enabled by default) and
add it to docs.

### DIFF
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -84,6 +84,8 @@ The node-stats telemetry device regularly calls the `cluster node-stats API <htt
 * Indices stats (key ``indices`` in the node-stats API)
 * Thread pool stats (key ``jvm.thread_pool`` in the node-stats API)
 * JVM buffer pool stats (key ``jvm.buffer_pools`` in the node-stats API)
+* JVM gc stats (key ``jvm.gc`` in the node-stats API)
+* JVM mem stats (key ``jvm.mem`` in the node-stats API)
 * Circuit breaker stats (key ``breakers`` in the node-stats API)
 * Network-related stats (key ``transport`` in the node-stats API)
 * Process cpu stats (key ``process.cpu`` in the node-stats API)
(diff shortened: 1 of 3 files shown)


dliappis:
I tested this using

```
esrally --configuration-name=metricstore --runtime-jdk=8 --track=geonames --challenge=append-no-conflicts-index-only --telemetry=node-stats --telemetry-params="node-stats-include-thread-pools:false,node-stats-include-buffer-pools:false,node-stats-include-breakers:false,node-stats-include-mem:false,node-stats-include-gc:true,node-stats-include-network:false,node-stats-include-process:false" --pipeline=from-sources-complete
```

and an example node-stats doc looks like:

```
{
  "_index": "rally-metrics-2019-02",
  "_type": "metrics",
  "_id": "9TlU4WgBStDFtUU6amBA",
  "_version": 1,
  "_score": null,
  "_source": {
    "name": "node-stats",
    "jvm_gc_collectors_young_collection_count": 1457,
    "jvm_gc_collectors_young_collection_time_in_millis": 14798,
    "jvm_gc_collectors_old_collection_count": 81,
    "jvm_gc_collectors_old_collection_time_in_millis": 3323,
    "@timestamp": 1549968631332,
    "relative-time": 4259307,
    "trial-id": "81f7b3c4-f935-4ff5-bd1f-a2aa220f3cb9",
    "trial-timestamp": "20190212T104315Z",
    "environment": "local",
    "track": "geonames",
    "lap": 1,
    "challenge": "append-no-conflicts-index-only",
    "car": "defaults",
    "meta": {
      ...
}
```

dliappis:
> I wonder whether we really need to add a new flag for this or whether we should just start capturing it by default?

It will be captured by default, but I thought that being able to control it via `node-stats-include-gc` as we can already do with `node-stats-include-mem` can be beneficial for users who want "less polluted/more targeted " benchmarks.

danielmitterdorfer:
Ok, then let's keep it as you've implemented it.

dliappis:
Thanks for the quick review.

Issue #645, opened 2019-02-12
Make MetricsStore multi cluster aware


dliappis:
node-stats which in turn relies ExternalEnvironmentInfo/ClusterEnvironmentInfo currently can't push the same meta information to additional clusters as it does for the "default" cluster, e.g. things like the ones shown below are missing:

"default" cluster:

```
 "meta": {
      "cpu_logical_cores": 16,
      "node_name": "aws-dl-es-cluster-b-0",
      "cluster": "default",
      "jvm_vendor": "Oracle Corporation",
      "jvm_version": "1.8.0_191",
      "attribute_xpack.installed": "true",
      "distribution_version": "6.7.0",
      "host_name": "10.20.1.31",
      "tag_x-pack-security-enabled": "true",
      "os_name": "Linux",
      "tag_track": "http_logs",
      "source_revision": "808db1f",
      "os_version": "4.4.0-1061-aws"
    },
```

vs another cluster:

```
"meta": {
      "attribute_xpack.installed": "true",
      "distribution_version": "6.7.0",
      "tag_x-pack-security-enabled": "true",
      "cluster": "cluster_a",
      "tag_track": "http_logs",
      "source_revision": "808db1f"
    },
```

This ticket tracks adding multi cluster support to all affected components, including the MetricsStore.

PR #646, opened 2019-02-12
Add node_name in node-stats docs for ...

dliappis:
... non-default clusters

Until #645 gets resolved, add node name meta information for node stats
collected from non default clusters.


dliappis:
Thanks for the review.

PR #647, opened 2019-02-14
Make types optional

danielmitterdorfer:
With this commit we make types an optional concept in Rally as types
have been removed from the Elasticsearch APIs with 7.0.0. It is still
possible to use types for older versions of Elasticsearch though.


danielmitterdorfer:
Thanks both for your quick review!

PR #649, opened 2019-02-19
Pass request-params as is in supported operations

### LINKED ISSUE #302: Allow to pass any request parameters for ES queries
With #290 we have allowed to pass any request parameter that is known to the Elasticsearch Python client. However, in some cases (e.g. benchmarking a custom plugin which defines new request parameters) we want to pass any request parameter, not just the ones known by the Elasticsearch Python client.

Background: https://discuss.elastic.co/t/query-parameters-for-operation-type-search/93478

As there is a workaround at the moment by implementing a custom runner we just put this in the backlog.

Implementation hint: This has to work for normal queries and scroll queries (Rally differentiates between the two for implementation reasons).

dliappis:
For all operations the `request-params` property, allow passing any
parameter without requiring that its explicitly supported by the
respective Elasticsearch Py API.

Closes #302

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -7,7 +7,7 @@ Migrating to Rally 1.1.0
 ``request-params`` in operations are passed as is and not serialized
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-With Rally 1.1.0 any operations supporting the option ``request-params`` property will pass the structure as is without attempting to serialize values.
+With Rally 1.1.0 any operations supporting the optional ``request-params`` property will pass the structure as is without attempting to serialize values.
 Until now, ``request-params`` relied on parameters being supported by the Elasticsearch Python client API calls. This means that for example boolean type parameters
 should be specified as strings i.e. ``"true"`` or ``"false"`` rather than ``true/false``.
 
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -5,7 +5,7 @@ Migrating to Rally 1.1.0
 ------------------------
 
 ``request-params`` in operations are passed as is and not serialized
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 With Rally 1.1.0 any operations supporting the option ``request-params`` property will pass the structure as is without attempting to serialize values.
 Until now, ``request-params`` relied on parameters being supported by the Elasticsearch Python client API calls. This means that for example boolean type parameters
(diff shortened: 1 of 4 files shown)


docs/migrate.rst
```diff
@@ -1,6 +1,45 @@
 Migration Guide
 ===============
 
+Migrating to Rally 1.1.0
+------------------------
+
+``request-params`` in operations are passed as is and not serialized
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Rally 1.1.0 any operations supporting the option ``request-params`` property will pass the structure as is without attempting to serialize values.
```
danielmitterdorfer: Remove "option"? (or did you mean to say "option**al**"?)

docs/track.rst
```diff
@@ -369,7 +369,11 @@ With the operation type ``search`` you can execute `request body searches <http:
...
 * ``cache`` (optional): Whether to use the query request cache. By default, Rally will define no value thus the default depends on the benchmark candidate settings and Elasticsearch version.
-* ``request-params`` (optional): A structure containing arbitrary request parameters. The supported parameters names are documented in the `Python ES client API docs <http://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch.search>`_. Parameters that are implicitly set by Rally (e.g. `body` or `request_cache`) are not supported (i.e. you should not try to set them and if so expect unspecified behavior).
+* ``request-params`` (optional): A structure containing arbitrary request parameters. The supported parameters names are documented in the `Search URI Request docs <https://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html#_parameters_3>`_.
+
+  NOTE:
```
danielmitterdorfer: Should we use the RST formatting approach for notes here? I.e.:

```
.. note::
```

dliappis:
Note: this PR requires some changes in the [tracks repo](https://github.com/elastic/rally-tracks) as well as elastic/rally-eventdata-track to ensure there are no uses of json booleans with request-params.

docs/migrate.rst
dliappis: Precisely, it was a typo. Fixed in a0fe4763761593a9ca29defcfeb073ba9d2a54a8

docs/track.rst
dliappis: I didn't elect to use `.. note::` because I was unsure if it can be indented (the note applies for `request-params` only). It appears this is possible, so I addressed it in 70639ce6d4c73c4357c8f3ebc8e4cccc31ef5de6

esrally/driver/runner.py
```diff
@@ -593,7 +593,7 @@ def __call__(self, es, params):
     def request_body_query(self, es, params):
         request_params = params.get("request-params", {})
         if "cache" in params:
-            request_params["request_cache"] = params["cache"]
+            request_params["request_cache"] = str(params["cache"]).lower() if isinstance(params["cache"], bool) else params["cache"]
```
danielmitterdorfer: simplify the right hand side to `str(params["cache"]).lower()`?

dliappis: Addressed in b0d6c4192f88ffaf32a8e5842d5d7568cba3ecfb

Issue #605, opened 2018-11-26
Update rally-* index templates for ES 7.0


danielmitterdorfer:
Rally defines default index templates for the index patterns `rally-metrics-*`, `rally-races-*` and `rally-results-*` in the resources directory of the source tree. When using an ES 6.x metrics store we get the following warnings:

```
Deprecated field [template] used, replaced by [index_patterns]
```

```
[_all] is deprecated in 6.0+ and will be removed in 7.0. As a replacement, you can use [copy_to] on mapping fields to create your own catch all field.
```

```
[_default_] mapping is deprecated since it is not useful anymore now that indexes cannot have more than one type
```

To be compatible with 7.0, we should change the mappings that Rally provides out of the box.

We should also switch the (sole) type for all indices to `_doc` as recommended in the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.5/removal-of-types.html).

synhershko:
There you go: https://github.com/elastic/rally/pull/648 :)

ebadyano:
Mapping got fixed in #648. I tested that changes work with older version ES 6.3.2. We need extra changes to metrics.py to be compatible with 7.0. @danielmitterdorfer is working on PR for that.

PR #654, opened 2019-02-26
Allow to use ES 7.x as metrics store

danielmitterdorfer:
With this commit we implement the changes necessary to talk to 7.x
metrics stores while retaining compatibility with older versions.

Relates #605


danielmitterdorfer:
Thanks!

PR #655, opened 2019-02-26
Add compatibility layer for existing metrics store

danielmitterdorfer:
With this commit we add a compatibility layer for existing metrics
stores that ensures Rally will not create incompatible mappings. Without
this change, Elasticsearch will complain that an existing metrics, races
or results index contains multiple types. By detecting this condition
upfront and instead writing to an index with the suffix `.new` we avoid
this condition. Note that when a new index is created (i.e. after a
month), it picks up the new mapping and is thus not prone to this issue.

Relates #605


danielmitterdorfer:

PR #668, opened 2019-03-18
Chart generator segment memory and new track combinations

dliappis:
Support new format of track combinations file and include new segment memory visualizations.

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -1334,7 +1334,7 @@ def generate_index_ops(chart_type, race_configs, environment, logger):
     idx_race_configs = list(filter(lambda c: "indexing" in c.charts, race_configs))
     for race_conf in idx_race_configs:
         logger.debug("Gen index visualization for race config with name:[%s] / label:[%s] / flavor: [%s] / license: [%s]",
-                     race_conf.name, race_conf.label, race_conf.flavor, race_conf.lic)
+                     race_conf.name, race_conf.label, race_conf.flavor, race_conf.es_license)
     charts = []
 
     if idx_race_configs:
@@ -1470,12 +1470,12 @@ def generate_dashboard(chart_type, environment, track, charts, flavor=None):
 
 
 class RaceConfig:
-    def __init__(self, track, cfg=None, flavor=None, lic=None, challenge=None, car=None, node_count=None, charts=None):
+    def __init__(self, track, cfg=None, flavor=None, es_license=None, challenge=None, car=None, node_count=None, charts=None):
         self.track = track
         if cfg:
             self.configuration = cfg
             self.configuration["flavor"] = flavor
-            self.configuration["lic"] = lic
+            self.configuration["es_license"] = es_license
         else:
             self.configuration = {
                 "charts": charts,
@@ -1493,8 +1493,8 @@ class RaceConfig:
         return self.configuration.get("flavor")
 
     @property
-    def lic(self):
-        return self.configuration.get("lic")
+    def es_license(self):
+        return self.configuration.get("es_license")
 
     @property
     def label(self):
@@ -1544,7 +1544,7 @@ def load_race_configs(cfg, chart_type, chart_spec_path=None):
     def add_configs(race_configs_per_lic, flavor_name="oss", lic="oss"):
         configs_per_lic = []
         for race_config in race_configs_per_lic:
-            configs_per_lic.append(RaceConfig(track=t, cfg=race_config, flavor=flavor_name, lic=lic))
+            configs_per_lic.append(RaceConfig(track=t, cfg=race_config, flavor=flavor_name, es_license=lic))
         return configs_per_lic
 
     def add_race_configs(license_configs, flavor_name):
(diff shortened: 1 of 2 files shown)


esrally/chart_generator.py
```diff
@@ -51,11 +53,18 @@ def index_label(race_config):
     return label
 
 
-def format_title(environment, track_name, suffix=None):
+def format_title(chart_type, environment, track_name, suffix=None, flavor=None):
```
danielmitterdorfer: Nit: I think the suffix should always be the last parameter.

esrally/chart_generator.py
```diff
@@ -51,11 +53,18 @@ def index_label(race_config):
...
+    title = "{}-{}".format(environment, track_name)
+
+    # as we generate two sets of Time Series/Nightly charts per flavor, the titles needs to be different.
+    if chart_type == TimeSeriesCharts:
```
danielmitterdorfer: This smells as if it is now in the wrong place if we need to do a type test. IMHO we should implement this now as a method on the respective chart classes.

esrally/chart_generator.py
```diff
@@ -755,6 +769,9 @@ def index(environment, race_configs, title):
             }
         }
 
+    def __repr__(self):
```
danielmitterdorfer: Why is this needed?

esrally/chart_generator.py
```diff
@@ -1187,6 +1319,9 @@ def index(environment, race_configs, title):
...
```
danielmitterdorfer: Why is this needed?

esrally/chart_generator.py
```diff
@@ -1395,28 +1596,66 @@ def load_race_configs(cfg):
...
+        structures.append(dashboard)
+
+    return structures
+
+
+def gen_charts_from_track_combinations(race_configs, chart_type, env, logger):
+    structures = []
+    for flavor, race_configs_per_flavor in race_configs.items():
+        for race_configs_per_track in race_configs_per_flavor:
+            logger.debug("Generating charts for race_configs with name:[{}]/flavor:[{}]".format(race_configs_per_track[0].name, flavor))
```
danielmitterdorfer: Our convention is to use plain logger strings for log statements instead of the format API.

esrally/chart_generator.py
```diff
@@ -1195,21 +1330,40 @@ def load_track(cfg, name=None):
     return track.load_track(cfg)
 
 
-def generate_index_ops(chart_type, race_configs, environment):
+def generate_index_ops(chart_type, race_configs, environment, logger):
     idx_race_configs = list(filter(lambda c: "indexing" in c.charts, race_configs))
+    for race_conf in idx_race_configs:
+        logger.debug("Gen index visualization for race config with name:[{}] / label:[{}] / flavor: {} / license: {}".format(
```
danielmitterdorfer: Our convention is to use plain logger strings for log statements instead of the format API.

esrally/chart_generator.py
```diff
@@ -1195,21 +1330,40 @@ def load_track(cfg, name=None):
...
-        return [chart_type.index(environment, idx_race_configs, title)]
-    else:
-        return []
+        title = format_title(chart_type, environment, race_configs[0].track, "indexing-throughput", flavor=race_configs[0].flavor)
+        charts = [chart_type.index(environment, idx_race_configs, title)]
+    return charts
 
 
-def filter_string(environment, race_config):
+def filter_string_time_series(environment, race_config):
```
danielmitterdorfer: This should now be a method on the respective classes if we need to do type tests.

esrally/chart_generator.py
```diff
@@ -1320,6 +1488,14 @@ def __init__(self, track, cfg=None, challenge=None, car=None, node_count=None, c
     def name(self):
         return self.configuration.get("name")
 
+    @property
+    def flavor(self):
+        return self.configuration.get("flavor")
+
+    @property
+    def lic(self):
```
danielmitterdorfer: What is `lic`? License? Is there a reason for the shortcut?

dliappis: Yes, it's license, but `license` is a built-in in Python: https://docs.python.org/3/library/builtins.html

danielmitterdorfer: 🤦‍♂️ - fair enough but does a property name really clash with a builtin? I mean, it's always used in the context of its enclosing object.

dliappis: It produces noise on the IDE at well and that was an additional motivation behind it. For variables at least it seems to be a [strong antipattern](https://stackoverflow.com/questions/16523789/naming-conflict-with-built-in-function) to shadow a built-in and @property in Python makes it act as an instance variable so didn't want to do this.

How about `eslicense`? Naming's not my strongest suit and yet it's hitting me all the time :)

danielmitterdorfer: Maybe `es_license` then?

esrally/chart_generator.py
dliappis: The problem originally stemed from calling the static method `index` from within the floating function `generate_index_ops`:

https://github.com/elastic/rally/blob/c331b00faae9f544c7323ee2adeff5ea2f6390d8/esrally/chart_generator.py#L1341

I've refactored things in 5a3bedabf588681fce7aa0bb8d18789396447df9

Issue #567, opened 2018-09-05
Ability to specify request timeout per operation


cdahlqvist:
Currently Rally supports overriding the request timeout for all operations through custom client options. For complex tracks that mix different types of operations, it would be useful to be able to specify the request timeout per operations. This would make it possible to e.g. give a forcemerge operation a long timeout while keeping the default short for indexing and query requests.

danielmitterdorfer:
Implementation note: This is the [`request_timeout` parameter](https://elasticsearch-py.readthedocs.io/en/master/api.html#timeout) in the Elasticsearch client.

hub-cap:
closed by https://github.com/elastic/rally/pull/1070

PR #669, opened 2019-03-19
Allow to override request timeout for force-merge

danielmitterdorfer:
With this commit we add the ability to override the global request
timeout for the `force-merge` operation. This relates to #567 but at the
moment we only implement it for `force-merge` and due to the fact that
we cannot solve this in a generic fashion for that operation anyway we
the current implementation is likely to stay even when we implement
full support in #567. We intentionally add this new parameter without
documentation so we are free to change the parameter name later on.

Relates #567


danielmitterdorfer:
Thanks for the review!

PR #670, opened 2019-03-20
Pass JAVA12_HOME in integration tests

danielmitterdorfer:
With this commit we pass the environment variable JAVA12_HOME to Rally's
integration tests.


danielmitterdorfer:
Thanks!

PR #671, opened 2019-03-29
Add ability to download from S3 buckets

danielmitterdorfer:
With this commit we add the ability to download from S3 buckets. A
download URL must start  with the scheme `s3://` to be recognized as S3
bucket. Provided that S3 access is properly configured Rally can also
handle downloads from private S3 buckets.

### DIFF
--- a/esrally/utils/net.py
+++ b/esrally/utils/net.py
@@ -65,39 +65,84 @@ class Progress:
         self.p.finish()
 
 
+# This function is not meant to be called externally and only exists for unit tests
+def _download_from_s3_bucket(bucket_name, bucket_path, local_path, expected_size_in_bytes=None, progress_indicator=None):
+    # lazily initialize S3 support - we might not need it
+    import boto3.s3.transfer
+
+    class S3ProgressAdapter:
+        def __init__(self, size, progress):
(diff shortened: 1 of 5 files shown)


esrally/utils/net.py
```diff
@@ -65,39 +65,84 @@ def finish(self):
...
+    # we need to remove the leading "/"
+    bucket_path = bucket_and_path[bucket_end_index + 1:]
+
+    logger.info("Downloading from S3 bucket [%s] and path [%s] to [%s].", bucket, bucket_path, local_path)
+    _download_from_s3_bucket(bucket, bucket_path, local_path, expected_size_in_bytes, progress_indicator)
+
+    return expected_size_in_bytes
+
+
+def download_http(url, local_path, expected_size_in_bytes=None, progress_indicator=None):
```
ebadyano: does it also make sense to add a unit test for `download_http`? I couldn't find if we have a test for download().. although it is a well executed code path so not sure if we need it.

danielmitterdorfer: Tbh, I attempted to do it but I ended up mocking so much code that I had the impression it's not really worth it.

PR #673, opened 2019-04-01
Honor runtime JDK in provisioner

danielmitterdorfer:
We currently rely on `java` being available on the PATH when
provisioning an Elasticsearch node but instead we should be consistent
and always use the specified runtime JDK when provisoning a node. With
this commit we determine the correct runtime JDK and set it as
`JAVA_HOME` when provisioning a node.

Relates elastic/rally-teams#24

### DIFF
--- a/esrally/mechanic/java_resolver.py
+++ b/esrally/mechanic/java_resolver.py
@@ -1,0 +1,43 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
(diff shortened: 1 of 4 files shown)


danielmitterdorfer:
Note that this change is backwards-compatible and can be merged independently of elastic/rally-teams#24.

danielmitterdorfer:
Steps to reproduce the problem:

```
unset JAVA_HOME
esrally --test-mode --client-options="use_ssl:true,verify_certs:false,basic_auth_user:'rally',basic_auth_password:'rally-password'" --car="4gheap,x-pack-security" --revision=latest
```

danielmitterdorfer:
Thanks for the review! :)

PR #674, opened 2019-04-03
[Experimental] Capture peak usage of JVM mem pools

danielmitterdorfer:
With this commit we store peak usage of all JVM memory pool at the end
of a benchmark. We will start with this limited scope to get more
insights on the usefulness of this feature and might expand this later:

* Capture peak usage also at other times during the benchmark
* Store peak usage also per race or show it in the summary report

We consider this feature experimental and as such it is intentionally
undocumented. It may change in incompatible ways or disappear at any
point.


danielmitterdorfer:
Thank you for the review! :)

PR #675, opened 2019-04-08
Fix chart generator segment memory bug

dliappis:


esrally/chart_generator.py
```diff
@@ -1026,7 +1026,7 @@ def segment_memory(title, environment, race_config):
                         "seperate_axis": 1,
                         "split_mode": "filters",
                         "stacked": "none",
-                        "filter": "environment:nightly AND track:geonames",
+                        "filter": "environment:{} AND track:geonames".format(environment),
```
danielmitterdorfer: Shouldn't the track be parameterized as well?

dliappis: Great catch! Fixed in https://github.com/elastic/rally/pull/675/commits/30dc641e4aac86e9b6eff43df9e376d62d350e81

PR #677, opened 2019-04-10
Allow shell redirects if possible

danielmitterdorfer:
With this commit we allow Rally to write to stdout even if not attached
to a terminal (e.g. when a shell redirect is in use). This allows users
to use shell redirects e.g. to write Rally's output to a file. Note that
this will only work if the Rally daemon is not started prior to Rally
because in that case the corresponding actors will be spawned by the
actor system that has already detached itself from the terminal and
there is no (robust) way to inject the current stdout file descriptor to
those processes.

### DIFF
--- a/esrally/utils/console.py
+++ b/esrally/utils/console.py
@@ -142,6 +142,8 @@ def _do_print(msg, end, flush):
             except BaseException:
                 # force-close quietly
                 pass
+        else:
+            raise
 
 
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
Thanks! :)

danielmitterdorfer:
I have removed the milestone now as we have reverted it due to troublesome behavior in some environments.

PR #678, opened 2019-04-10
Upgrade Jinja to 2.10.1

danielmitterdorfer:


danielmitterdorfer:
Yes, I also tested that locally. Thanks for the review! :)

PR #679, opened 2019-04-10
Skip Python install if already existing

danielmitterdorfer:
With this commit we tell pyenv to skip installing an already existing
Python version. This speeds up initial setup in such cases.


danielmitterdorfer:
Thanks for the review!

PR #682, opened 2019-04-16
Provide a platform-independent Rally binary

danielmitterdorfer:
With this commit we configure pip in the offline install script to use
the source package of MarkupSafe to avoid platform-dependent offline
packages.


danielmitterdorfer:
For more background info for the review, please see the [`pip download` docs](https://pip.pypa.io/en/stable/reference/pip_download/).

danielmitterdorfer:
Thanks for your review! :)

PR #683, opened 2019-04-18
Store mean for response-related metrics

danielmitterdorfer:
With this commit we calculate and store the mean of response-related
metrics (i.e. throughput, service time and latency). This can be useful
because mean sample values tend to be normally distributed even if the
population distribution (the underlying distribution that is represented
by each individual mean) is not. This is also known as the central limit
theorem. This is useful because certain statistical tests, like the
t-test, assume normally distributed values.

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -812,6 +812,20 @@ class MetricsStore:
         percentiles = self.get_percentiles(name, task, operation_type, sample_type, lap, percentiles=[median])
         return percentiles[median] if percentiles else None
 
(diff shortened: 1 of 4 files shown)


ebadyano:
Do we also want to display it in the report or is it enough to just store it in the metric store?

danielmitterdorfer:
> Do we also want to display it in the report or is it enough to just store it in the metric store?

For now it's intentionally only stored but not shown in the summary report. First, we'd make the summary report even larger and second this is primarily meant for users that want to run statistical significance tests (which I'd expect to use Elasticsearch as a metrics store already). If we decide to display it in the summary report, I suggest we do it in a follow-up PR.

PR #686, opened 2019-04-26
Honor track-params in chart-generator

dliappis:
Ensure that track-params (which may influence the name of a task like in
 [1]) get passed to track loader ensuring that task filters, such as
 ones used by TimeSeries charts for indexing ops, use the
 right name.

 [1] https://github.com/elastic/rally-tracks/blob/9733f6addb8dc1a5148b4e18562bf64f244dd3d6/http_logs/challenges/default.json#L201

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -1364,13 +1364,23 @@ class TimeSeriesCharts:
         }
 
 
-def load_track(cfg, name=None, track_params={}):
-    # hack to make this work with multiple tracks (Rally core is usually not meant to be used this way)
-    if name:
-        cfg.add(config.Scope.applicationOverride, "track", "track.name", name)
-    # another hack to ensure any track-params in the race config are used by Rally's track loader
-    cfg.add(config.Scope.applicationOverride, "track", "params", track_params)
-    return track.load_track(cfg)
+class RaceConfigTrack:
+    def __init__(self, cfg, name=None):
+        self.cached_track = self.load_track(cfg, name)
+
+    def load_track(self, cfg, name=None, params={}):
+        # hack to make this work with multiple tracks (Rally core is usually not meant to be used this way)
+        if name:
+            cfg.add(config.Scope.applicationOverride, "track", "track.name", name)
+        # another hack to ensure any track-params in the race config are used by Rally's track loader
+        cfg.add(config.Scope.applicationOverride, "track", "params", params)
+        return track.load_track(cfg)
+
+    def get_track(self, cfg, name=None, params={}):
+        if params:
+            return self.load_track(cfg, name, params)
+        # if no params specified, return the initially cached, (non-parametrized) track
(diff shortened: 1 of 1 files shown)


esrally/chart_generator.py
```diff
@@ -1564,7 +1566,12 @@ def load_race_configs(cfg, chart_type, chart_spec_path=None):
     def add_configs(race_configs_per_lic, flavor_name="oss", lic="oss"):
         configs_per_lic = []
         for race_config in race_configs_per_lic:
-            configs_per_lic.append(RaceConfig(track=t, cfg=race_config, flavor=flavor_name, es_license=lic))
+            configs_per_lic.append(
+                RaceConfig(track=load_track(cfg, item["track"], race_config.get("track-params", {})),
```
danielmitterdorfer: I think you need to do the same when loading a single track here: https://github.com/elastic/rally/blob/ae3e14ac40548029c5fc8ed54a29c6b3a7c5ddbb/esrally/chart_generator.py#L1609

(assuming users can pass `--track-params` when invoking the chart generator which should be supported in that case)

Also we're loading the very same track repeatedly (which is quite time-consuming already) and I wonder whether we should avoid this and use a cached version if a race configuration doesn't use track parameters (i.e. the dictionary is empty). This would speed up the standard case significantly IMHO.

dliappis: I address both comments in https://github.com/elastic/rally/pull/686/commits/4a93a9f7053422950d44f1a4929a0891a8cbdc26, could you take another pass please?

esrally/chart_generator.py
```diff
@@ -1612,7 +1623,7 @@ def add_race_configs(license_configs, flavor_name):
             car_name = car_names[0]
         race_configs = [
             [
-                RaceConfig(track=load_track(cfg),
+                RaceConfig(track=track.load_track(cfg),
```
danielmitterdorfer: Shouldn't this be `race_config_track.load_track`?

esrally/chart_generator.py
```diff
@@ -1364,13 +1364,23 @@ def index(environment, race_configs, title):
...
-def load_track(cfg, name=None, track_params={}):
-    # hack to make this work with multiple tracks (Rally core is usually not meant to be used this way)
-    if name:
-        cfg.add(config.Scope.applicationOverride, "track", "track.name", name)
-    # another hack to ensure any track-params in the race config are used by Rally's track loader
-    cfg.add(config.Scope.applicationOverride, "track", "params", track_params)
-    return track.load_track(cfg)
+class RaceConfigTrack:
+    def __init__(self, cfg, name=None):
+        self.cached_track = self.load_track(cfg, name)
```
danielmitterdorfer: By loading the parameterless version eagerly we would load it unneccesarily when there are only parameterized versions of track but that's probably ok.

PR #693, opened 2019-05-22
Assume UTC timezone if not specified

### LINKED ISSUE #692: Assertion failure in TemplateRenderTests.test_render_simple_template
Expected: All tests should pass.  This one doesn't.

**Steps to reproduce**:
(0. have a system timezone UTC-7 or further west.  Tested in UTC-8)
1. make test

```
E                 {
E       -             "key": 5864,
E       ?                       ^
E       +             "key": 5863,
E       ?                       ^
E                     "key2": "static value"
E                 }
tests/track/loader_test.py:706: AssertionError
```

drawlerr:
Python naive timestamp delta assumes local timezone, which makes template rendering inconsistent between timezones.

Closes #692.

### DIFF
--- a/esrally/time.py
+++ b/esrally/time.py
@@ -16,7 +16,7 @@
 # under the License.
 
 import time
-from datetime import datetime
+from datetime import datetime, timezone
 
 
 def to_epoch_millis(t):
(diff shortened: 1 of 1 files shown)


esrally/time.py
```diff
@@ -51,14 +51,17 @@ def sleep(seconds):
     time.sleep(seconds)
 
 
-def _to_datetime(val, date_format=None):
+def _to_datetime(val, date_format=None, default_tz=timezone.utc):
```
danielmitterdorfer: As this is a module-internal function I think it would make sense to expose this parameter also in the function `days_ago` which is using it.

PR #696, opened 2019-05-23
Remove CPU sampling

danielmitterdorfer:
With this commit we remove the ability to sample CPU usage. For
coarse-grained analysis we instead suggest to use system monitoring
tools (e.g. Metricbeat) and for fine-grained analysis a profiler is a
better choice.

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -1,6 +1,15 @@
 Migration Guide
 ===============
 
+Migrating to Rally 1.2.0
+------------------------
+
+CPU usage is not measured anymore
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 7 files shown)


docs/migrate.rst
```diff
@@ -1,6 +1,15 @@
 Migration Guide
 ===============
 
+Migrating to Rally 1.2.0
+------------------------
+
+CPU usage is not measured anymore
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With Rally 1.2.0, CPU usage will neither be measured nor reported. We suggest to use system monitoring tools like ``vmstat`` or `Metricbeat <https://www.elastic.co/downloads/beats/metricbeat>`_ to measure CPU usage instead.
```
dliappis: nit: I'd rather mention `mpstat` or `sar` instead of `vmstat` for CPU. Taking a quick glance of where tot CPU time gets spent with `vmstat` is fine too of course.

danielmitterdorfer: addressed in https://github.com/elastic/rally/pull/696/commits/59fa4fbc8c910447734ac4ff47350916f483051a

PR #698, opened 2019-05-24
Enter degraded mode on bootstrap failure

danielmitterdorfer:
In some cases the internal actor system might not be able to open a
server socket. We detect this condition and fallback to a degraded mode
which uses queues for communication between actors.


danielmitterdorfer:
Thanks! :)

PR #699, opened 2019-05-29
Fix home directory paths in provisioner_test.

drawlerr:


tests/mechanic/provisioner_test.py
```diff
@@ -40,11 +43,11 @@ def null_apply_config(source_root_path, target_root_path, config_vars):
         team.Car(
             names="unit-test-car",
             root_path=None,
-            config_paths=["~/.rally/benchmarks/teams/default/my-car"],
+            config_paths=[f"{HOME_DIR}/.rally/benchmarks/teams/default/my-car"],
             variables={"heap": "4g"}),
             java_home="/usr/local/javas/java8",
             node_name="rally-node-0",
-            node_root_dir="~/.rally/benchmarks/races/unittest",
+            node_root_dir=f"{HOME_DIR}/.rally/benchmarks/races/unittest",
```
dliappis: Python f-Strings are really great, but unfortunately we perform integration tests across 3.5, 3.6 and 3.7 and f-Strings became available only starting with Python 3.6.

To catch these it's recommended to run `make it` to run the integration test-suite.

tests/mechanic/provisioner_test.py
```diff
@@ -18,12 +18,15 @@
...
 import tempfile
 import unittest.mock as mock
+from pathlib import Path
 from unittest import TestCase
 
 from esrally.mechanic import provisioner, team
 from esrally.utils import convert
 from esrally import exceptions
 
+HOME_DIR = str(Path.home())
```
dliappis: While the `Path` module is really great with its object oriented interface, in Rally we strive for consistency and typically use `os.path.expanduser("~")` for grabbing home. Example: https://github.com/elastic/rally/blob/cc689dfdd30dca5626847cf939018992e9745e58/esrally/log.py#L42

PR #700, opened 2019-06-03
Allow stdout print when running in Docker

dliappis:
Our current stdout.isatty() check doesn't allow Rally display normal
progress messages when running in Docker, as typically the container
isn't run with tty enabled.

Workaround this, by expecting an environment variable
RALLY_RUNNING_IN_DOCKER set to True.

### DIFF
--- a/esrally/racecontrol.py
+++ b/esrally/racecontrol.py
@@ -17,10 +17,11 @@
 
 import collections
 import logging
+import os
 import sys
-
 import tabulate
 import thespian.actors
+import urllib
 
 from esrally import actor, config, exceptions, track, driver, mechanic, reporter, metrics, time, DOC_LINK, PROGRAM_NAME
 from esrally.utils import console, convert, opts
@@ -386,9 +387,14 @@ def run(cfg):
     logger = logging.getLogger(__name__)
     name = cfg.opts("race", "pipeline")
 
-    if console.RALLY_RUNNING_IN_DOCKER:
+    if os.environ.get("RALLY_RUNNING_IN_DOCKER", "").upper() == "TRUE":
         # in this case only benchmarking remote Elasticsearch clusters makes sense
-        cfg.add(config.Scope.applicationOverride, "race", "pipeline", "benchmark-only")
+        if name != "benchmark-only":
+            raise exceptions.SystemSetupError(
+                "Only the [benchmark-only] pipeline is supported by the Rally Docker image.\n"
+                "Add --pipeline=benchmark-only in your Rally arguments and try again.\n"
+                "For more details read the docs for the benchmark-only pipeline in {}\n".format(
+                    urllib.parse.urljoin(DOC_LINK, "pipelines.html#benchmark-only")))
     elif len(name) == 0:
         # assume from-distribution pipeline if distribution.version has been specified and --pipeline cli arg not set
         if cfg.exists("mechanic", "distribution.version"):
--- a/esrally/racecontrol.py
+++ b/esrally/racecontrol.py
@@ -387,15 +387,7 @@ def run(cfg):
     logger = logging.getLogger(__name__)
     name = cfg.opts("race", "pipeline")
 
-    if os.environ.get("RALLY_RUNNING_IN_DOCKER", "").upper() == "TRUE":
-        # in this case only benchmarking remote Elasticsearch clusters makes sense
-        if name != "benchmark-only":
-            raise exceptions.SystemSetupError(
-                "Only the [benchmark-only] pipeline is supported by the Rally Docker image.\n"
-                "Add --pipeline=benchmark-only in your Rally arguments and try again.\n"
-                "For more details read the docs for the benchmark-only pipeline in {}\n".format(
-                    urllib.parse.urljoin(DOC_LINK, "pipelines.html#benchmark-only")))
-    elif len(name) == 0:
+    if len(name) == 0:
         # assume from-distribution pipeline if distribution.version has been specified and --pipeline cli arg not set
         if cfg.exists("mechanic", "distribution.version"):
             name = "from-distribution"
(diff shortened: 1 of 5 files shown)


tests/utils/console_test.py
```diff
@@ -0,0 +1,162 @@
...
+from unittest import TestCase
+from esrally.utils import console
+
+
+class ConsoleFunctionTests(TestCase):
+    @mock.patch("os.environ.get")
+    def test_global_rally_running_in_docker_is_false(self, patched_env_get):
+        patched_env_get.return_value = random.choice(["false", "False", "FALSE"])
+
+        _ = console.init()
```
danielmitterdorfer: Maybe just call the function without assigning anything?

tests/utils/console_test.py
```diff
@@ -0,0 +1,162 @@
...
+import unittest.mock as mock
+
+from unittest import TestCase
+from esrally.utils import console
+
+
+class ConsoleFunctionTests(TestCase):
+    @mock.patch("os.environ.get")
+    def test_global_rally_running_in_docker_is_false(self, patched_env_get):
+        patched_env_get.return_value = random.choice(["false", "False", "FALSE"])
```
danielmitterdorfer: Maybe also add `None` as an option?

tests/utils/console_test.py
```diff
@@ -0,0 +1,162 @@
...
+
+    @mock.patch("os.environ.get")
+    def test_global_rally_running_in_docker_is_true(self, patched_env_get):
+        patched_env_get.return_value = random.choice(["true", "True", "TRUE"])
+
+        _ = console.init()
+        self.assertEqual(True, console.RALLY_RUNNING_IN_DOCKER)
+
+    @mock.patch("sys.stdout.isatty")
+    @mock.patch("builtins.print")
```
danielmitterdorfer: I wonder if we should define defaults in `console` and allow to override them instead of monkey-patching? E.g. allow to pass the `print` implementation or just have a global variable and change it in the test?

tests/utils/console_test.py
```diff
@@ -0,0 +1,162 @@
...
+    def test_println_randomized_dockertrue_or_istty_and_isnotquiet(self, patched_print, patched_isatty):
+        console.QUIET = False
```
danielmitterdorfer: Should we always call `console.init()` as the first step in these tests?

tests/utils/console_test.py
```diff
@@ -0,0 +1,162 @@
...
+        message = "Unit test message"
+        width = random.randint(20, 140)
+        progress_reporter = console.CmdLineProgressReporter(width=width)
+        progress_reporter.print(message=message, progress=".")
+        patched_print.assert_has_calls([
+            mock.call(" " * width, end=""),
+            mock.call("\x1b[{}D{}{}.".format(width, message, " "*(width-len(message)-1)), end="")
+        ])
+
+    @mock.patch("sys.stdout.flush")
```
danielmitterdorfer: There should be no need to mock `flush` because it is not called. Or are you doing this for consistency? Should we then check that `flush` is not called?

dliappis:
@danielmitterdorfer Thanks for the comments! I believe I addressed all of them, could you take another pass please?

esrally/racecontrol.py
```diff
@@ -386,9 +387,14 @@ def run(cfg):
...
-    if console.RALLY_RUNNING_IN_DOCKER:
+    if os.environ.get("RALLY_RUNNING_IN_DOCKER", "").upper() == "TRUE":
         # in this case only benchmarking remote Elasticsearch clusters makes sense
-        cfg.add(config.Scope.applicationOverride, "race", "pipeline", "benchmark-only")
+        if name != "benchmark-only":
+            raise exceptions.SystemSetupError(
+                "Only the [benchmark-only] pipeline is supported by the Rally Docker image.\n"
+                "Add --pipeline=benchmark-only in your Rally arguments and try again.\n"
+                "For more details read the docs for the benchmark-only pipeline in {}\n".format(
+                    urllib.parse.urljoin(DOC_LINK, "pipelines.html#benchmark-only")))
```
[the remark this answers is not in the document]
dliappis: The nice property of `urljoin` is that it manages trailing slashes, as required, by itself.
With the standard way we've been doing things, i.e. string join, the resulting URL has two double forward slashes (due to https://github.com/elastic/rally/blob/646cc1e0f3edd06742e027461e90459b0a56005f/esrally/__init__.py#L29-L31).

For example the resulting URL in https://github.com/elastic/rally/blob/646cc1e0f3edd06742e027461e90459b0a56005f/esrally/racecontrol.py#L405 produces something like `...//pipelines.html#from-distribution`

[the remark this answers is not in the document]
danielmitterdorfer: I am not 100% convinced we need this but instead we should fix the formatting of the other link as we have full control of the prefix? In any case I think we should either offer a consistent helper function across the code base (I have no good answer where this would belong; `console`, `convert` and `io` are possible candidates but none of them is a good fit IMHO) or refrain from providing doc links in exception messages?

esrally/utils/console.py
```diff
@@ -151,12 +147,15 @@ def progress(width=90):
 class CmdLineProgressReporter:
     """
     CmdLineProgressReporter supports displaying an updating progress indication together with an information message.
+
+    :param custom_print: allow use of a different print method to assist with patching in unittests
     """
 
-    def __init__(self, width, plain_output=False):
+    def __init__(self, width, plain_output=False, custom_print=print):
```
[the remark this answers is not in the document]
dliappis: Addressed in https://github.com/elastic/rally/pull/700/commits/5767cac206a6a623a4d2c1ec04e2e843319967ef

esrally/racecontrol.py
```diff
@@ -386,9 +387,14 @@ def run(cfg):
     logger = logging.getLogger(__name__)
     name = cfg.opts("race", "pipeline")
 
-    if console.RALLY_RUNNING_IN_DOCKER:
+    if os.environ.get("RALLY_RUNNING_IN_DOCKER", "").upper() == "TRUE":
```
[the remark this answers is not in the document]
dliappis: Addressed in https://github.com/elastic/rally/pull/700/commits/5767cac206a6a623a4d2c1ec04e2e843319967ef

esrally/racecontrol.py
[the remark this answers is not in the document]
dliappis: As discussed we'll tackle the rest of the places where this needs to be adjusted in a separate PR.

esrally/utils/console.py
```diff
@@ -145,12 +147,15 @@ def progress(width=90):
...
+
+    :param custom_print: allow use of a different print method to assist with patching in unittests
     """
 
-    def __init__(self, width, plain_output=False):
+    def __init__(self, width, plain_output=False, printer=print):
         self._width = width
         self._first_print = True
         self._plain_output = plain_output
+        self._custom_print = printer
```
danielmitterdorfer: Nit: How about we also call the field `_printer`?

dliappis: Fixed in https://github.com/elastic/rally/pull/700/commits/c45a2b92d1e800b0ee7b46e7bd27341048ee56d3

dliappis:
@danielmitterdorfer Thanks for the review!

Issue #697, opened 2019-05-23
Allow to manage Elasticsearch nodes separately from benchmarking


danielmitterdorfer:
Currently Rally can be used as a load generator and to manage Elasticsearch instances (building a distribution from sources, installing and configuring Elasticsearch as well as launching / stopping it). It is possible to use Rally as a standalone load generator (`--pipeline=benchmark-only`) but it is not possible to use Rally only to manage Elasticsearch nodes. In order to do that, we should allow to call the provisioning and launch phases as separate subcommands in Rally. We should only support a local mode though, i.e. it is not possible to only setup a cluster from the coordinator node but users must invoke Rally instead on each of the target nodes. As a corollary, each command  applies to only one node.

We will add the following subcommands:

* `provision`: Provisions a single node. The artefact might be built from sources or a distribution that has been downloaded.
* `start`: Starts an already provisioned node.
* `stop`: Stops an already started node.

**Implementation note**: None of these commands should rely on the actor system and instead run the commands in the main Rally process.

### Provisioning

This covers all steps from retrieving the binary (either download or build it via Gradle) until a fully configured node is installed. Note that there are essentially no new steps, it's just that we expose this now as a dedicated subcommand. One tricky aspect is that we are currently able to create a unique node name by increasing a counter (`rally-node-{N}`), see also: 

https://github.com/elastic/rally/blob/f575790902afd18e9cdd17028102724994142d35/esrally/mechanic/mechanic.py#L178-L186)

This will not be possible anymore with the `provision` subcommand as it does not have a global view of the cluster but rather a per-node view.

**Post condition**

After the provisioner has (successfully) run the following conditions are met:

* An Elasticsearch node is installed in a well-known directory `~/.rally/benchmarks/nodes/$ID/install` where `$ID` is a globally unique id generated by the provisioner.
* A meta-data file has been written to `~/.rally/benchmarks/nodes/$ID/provisioner.json` which contains the necessary data that needs to be exchanged between the provisioner and the launcher (see `NodeConfiguration` for data that are exchanged between those two).
* `$ID` is written to stdout. This id can then be used later on to start and stop this node.

**Note**: It might make sense that we specify all telemetry devices as part of the provisioning process regardless when they get attached (some are only attached on launch) and exchange the necessary information via the provisioner metadata.

### Start

[...]

PR #701, opened 2019-06-03
Implement ES daemon-mode in process launcher

### LINKED ISSUE #718: sysstats.py crashed if no valid brand string in cpuinfo on an eMAG aarch64 system
**Rally version** (get with `esrally --version`):
1.1.0

**Invoked command**:
esrally --distribution-version=6.0.0

**Configuration file (located in `~/.rally/rally.ini`))**:
all default
datastore.type = in-memory

**JVM version**:
jdk1.8, the minor version does not matter

**OS version**:
CentOS 7.5

It is a very straightforward problem, the [py-cpuinfo](https://github.com/workhorsy/py-cpuinfo) tool cannot get a valid brand string from various system info or utilities on my aarch64 testing system. In theory, sysstats.py should handle this as a "Unknown" cpu model, at least. However, as "brand" is not a valid key in cpuinfo.get_cpu_info(), the test subsequently crashed and quit. A strict check as below has to be done.
 
https://github.com/elastic/rally/blob/4c160c8fbe7382f7db2af1d9c74c176f009029e5/esrally/utils/sysstats.py#L48
```
ci=cpuinfo.get_cpu_info()
return ci["brand"] if (cpuinfo_available and "brand" in ci) else "Unknown"
```
This issue was exposed and the quick fix has been verified fine on an Ampere Computing eMAG aarch64 system.

drawlerr:
Launch Elasticsearch as a daemon.

Relates to #697 
Closes #718


esrally/mechanic/launcher.py
```diff
@@ -376,43 +445,56 @@ def _start_process(self, env, node_name, binary_path):
...
+            cmd.extend(["-d", "-p", "pid"])
+            proc = subprocess.Popen(cmd,
+                                    close_fds=True)
+            ret = proc.wait()
+            if ret != 0:
+                msg = "Daemon startup failed with exit code[{}]".format(ret)
+                logging.error(msg)
+                raise exceptions.LaunchError(msg)
+
+            es_log_file = _get_es_log_filename(binary_path)
```
danielmitterdorfer: We already have a check in place that waits for the cluster to be ready (in `ClusterLauncher`) so I think we can consider a node to be "up" when we detect that the PID file has been written and remove the log file parsing entirely.

esrally/mechanic/launcher.py
```diff
@@ -297,10 +364,12 @@ def __init__(self, cfg, metrics_store, races_root_dir, clock=time.Clock):
         self._clock = clock
         self.races_root_dir = races_root_dir
         self.keep_running = self.cfg.opts("mechanic", "keep.running")
+        self.daemon = self.cfg.opts("mechanic", "daemon", default_value=False, mandatory=False)
```
danielmitterdorfer: That's definitely a good idea for testing in the WIP PR but I think we can fully switch to daemonized startup once this work is finished. While it is not mentioned in the [docs for Elasticsearch 1.7](https://www.elastic.co/guide/en/elasticsearch/reference/1.7/setup.html), daemonized startup and writing to a PID file is supported since then:

```
daniel@io:distributions/elasticsearch-1.7.6 $ ./bin/elasticsearch --help
Usage: ./bin/elasticsearch [-vdh] [-p pidfile] [-D prop] [-X prop]
Start elasticsearch.
    -d            daemonize (run in background)
    -p pidfile    write PID to <pidfile>
    -h
    --help        print command line options
    -v            print elasticsearch version, then exit
    -D prop       set JAVA system property
    -X prop       set non-standard JAVA system property
   --prop=val
   --prop val     set elasticsearch property (i.e. -Des.<prop>=<val>)
```

In other words: We can rely that this works consistently across all versions.

tests/mechanic/launcher_test.py
```diff
@@ -93,6 +99,116 @@ def info(self, *args, **kwargs):
...
 
 
+logging.basicConfig(level=logging.DEBUG)
+HOME_DIR = os.path.expanduser("~")
+
+
+def create_config():
+    cfg = config.Config()
+    # collect some mandatory config here
+    cfg.add(config.Scope.application, "node", "root.dir", HOME_DIR + "/.rally/benchmarks")
```
danielmitterdorfer: We'd usually use `os.path.join(HOME_DIR, ".rally", "benchmarks")` in such cases.

tests/mechanic/launcher_test.py
```diff
@@ -93,6 +99,116 @@ def info(self, *args, **kwargs):
...
+    return team.load_car(HOME_DIR + "/.rally/benchmarks/teams/default",
+                         ["defaults"],
+                         None)
+
+
+def create_provisioner(car, ver):
+    installer = provisioner.ElasticsearchInstaller(car=car,
+                                                   java_home=guess_java_home(),
+                                                   node_name="rally-node-0",
+                                                   node_root_dir=HOME_DIR + "/.rally/benchmarks/races/unittest",
```
danielmitterdorfer: Again, we'd use `os.path.join` here.

tests/mechanic/launcher_test.py
```diff
@@ -93,6 +99,116 @@ def info(self, *args, **kwargs):
...
+                                                   http_port=9200)
+    p = provisioner.BareProvisioner(cluster_settings={"indices.query.bool.max_clause_count": 50000},
+                                    es_installer=installer,
+                                    plugin_installers=[],
+                                    preserve=True,
+                                    distribution_version=ver)
+    return p
+
+
+class ProcessLauncherTests(TestCase):
```
danielmitterdorfer: I think testing this gets much simpler once we only rely on the PID. Then we can just mock process startup and mock the check for a PID file. Once we expose a `provision` and `launch` subcommand this should also become way simpler to add this as an integration test in `integration-test.sh`. I'd imagine an integration test along those lines:

```
PROVISIONER_ID=$(esrally provision --distribution-version=6.8.0 --car=4gheap)
esrally start --provisioner-id=${PROVISIONER_ID}
# maybe do a check here that the node is actually reachable via HTTP
esrally stop --provisioner-id=${PROVISIONER_ID}
```

esrally/mechanic/launcher.py
drawlerr: My intent was actually to allow the original "inproc" codepath to still be used, if desired.  But I don't know if that will have any value.  What do you think?

danielmitterdorfer: IMHO we don't need to keep the other approach as well.

esrally/mechanic/launcher.py
```diff
@@ -232,9 +164,10 @@ def start(self, node_configurations):
             nodes.append(cluster.Node(p, host_name, node_name, t))
         return nodes
 
-    def _start_process(self, cmd, node_name):
-        return _start(subprocess.Popen(shlex.split(cmd),
-                                       stdout=subprocess.PIPE, stderr=subprocess.STDOUT, stdin=subprocess.DEVNULL), node_name)
+    def _start_process(self, cmd):
+        subprocess.Popen(shlex.split(cmd), close_fds=True)
+        #TODO:  wait_for_docker_pidfile?
```
danielmitterdorfer: @dliappis can we assume that if `docker-compose` returns (with exit code `0`), Elasticsearch is about to startup? (i.e. no need to wait for the PID file)?

esrally/mechanic/launcher.py
```diff
@@ -281,11 +216,29 @@ def start(self, node_configurations=None):
...
         pass
 
 
-class InProcessLauncher:
+def _get_pid_from_file(filename):
+    with open(filename, "r") as f:
+        return int(f.read())
+
+
+def wait_for_pidfile(pidfilename, timeout=10):
```
danielmitterdorfer: We have seen situations where the JVM was slow to startup due to fragmented memory on the system so I suggest to raise this timeout to 30 seconds or maybe even 60.

setup.py
```diff
@@ -70,7 +70,9 @@ def str_from_file(name):
     #   botocore: Apache 2.0
     #   jmespath: MIT
     #   s3transfer: Apache 2.0
-    "boto3==1.9.120"
+    "boto3==1.9.120",
+    # License: MIT
+    "pyyaml==5.1"
```
danielmitterdorfer: I think this dependency is not needed anymore?

esrally/mechanic/launcher.py
dliappis: I am afraid it's a "it depends" answer :)

With docker-compose it's possible to have healthchecks and depending on how they are defined we can safely assume ES is actually listening; this is done for example in https://github.com/elastic/rally/blob/69cb36d65f35833632dbddabbb043f387039f8bf/docker/docker-compose-tests.yml#L26-L30 and the other service, which can be a sidecar, [depends on it](https://github.com/elastic/rally/blob/69cb36d65f35833632dbddabbb043f387039f8bf/docker/docker-compose-tests.yml#L11-L13).

danielmitterdorfer: @dliappis I think this is entirely in our control as the only way to start the Docker image is via https://github.com/elastic/rally/blob/master/esrally/resources/docker-compose.yml - So we cannot assume it is already listening (which is ok) but I wonder whether we can assume that the Elasticsearch process has been started already?

esrally/mechanic/launcher.py
```diff
@@ -335,8 +290,8 @@ def _start_node(self, node_configuration, node_count_on_host):
         t = telemetry.Telemetry(enabled_devices, devices=node_telemetry)
         env = self._prepare_env(car, node_name, java_home, t)
         t.on_pre_node_start(node_name)
-        node_process = self._start_process(env, node_name, binary_path)
-        node = cluster.Node(node_process, host_name, node_name, t)
+        node_pid = self._start_process(env, node_name, binary_path)
```
danielmitterdorfer: The parameters do not match the signature of `_start_process`. I noticed this because Rally failed when I attempted to test this locally with:

```
esrally --test-mode --distribution-version=1.7.6 --on-error=abort --challenge=append-no-conflicts-index-only
```

esrally/mechanic/launcher.py
```diff
@@ -371,48 +326,47 @@ def _set_env(self, env, k, v, separator=' '):
...
-    def _start_process(self, env, node_name, binary_path):
+    @staticmethod
+    def _start_process(binary_path):
         if os.geteuid() == 0:
             raise exceptions.LaunchError("Cannot launch Elasticsearch as root. Please run Rally as a non-root user.")
         os.chdir(binary_path)
         cmd = ["bin/elasticsearch"]
-        return _start(subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, stdin=subprocess.DEVNULL, env=env), node_name)
+        cmd.extend(["-d", "-p", "pid"])
+        proc = subprocess.Popen(cmd,
```
danielmitterdorfer: At this time we still need to pass the environment otherwise telemetry devices that rely on `ES_JAVA_OPTS` will not work anymore.

esrally/mechanic/launcher.py
```diff
@@ -327,13 +321,14 @@ def _set_env(self, env, k, v, separator=' '):
...
     @staticmethod
-    def _start_process(binary_path):
+    def _start_process(binary_path, env):
         if os.geteuid() == 0:
             raise exceptions.LaunchError("Cannot launch Elasticsearch as root. Please run Rally as a non-root user.")
         os.chdir(binary_path)
         cmd = ["bin/elasticsearch"]
         cmd.extend(["-d", "-p", "pid"])
         proc = subprocess.Popen(cmd,
+                                env,
```
danielmitterdorfer: You need to use a keyword-arg here (i.e. `env=env`) otherwise it treats this argument as positional and at this position it expects the `bufsize` argument. You can check this e.g. by running:

```
esrally --test-mode --distribution-version=1.7.6 --on-error=abort --challenge=append-no-conflicts-index-only
```

This fails with:

```
[ERROR] Cannot race. bufsize must be an integer
```

esrally/mechanic/launcher.py
```diff
@@ -232,9 +164,9 @@ def start(self, node_configurations):
             nodes.append(cluster.Node(p, host_name, node_name, t))
         return nodes
 
-    def _start_process(self, cmd, node_name):
-        return _start(subprocess.Popen(shlex.split(cmd),
-                                       stdout=subprocess.PIPE, stderr=subprocess.STDOUT, stdin=subprocess.DEVNULL), node_name)
+    def _start_process(self, cmd):
+        subprocess.Popen(shlex.split(cmd), close_fds=True)
```
dliappis: This starts a child process and given that we start `docker-compose` without `-d` in line `156` this means that when the specific Rally command is over then docker-compose will be stopped too.

esrally/mechanic/launcher.py
```diff
@@ -216,14 +148,14 @@ def __init__(self, cfg, metrics_store):
...
         for node_configuration in node_configurations:
-
             node_name = node_configuration.node_name
             host_name = node_configuration.ip
             binary_path = node_configuration.binary_path
             self.binary_paths[node_name] = binary_path
 
-            p = self._start_process(cmd="docker-compose -f %s up" % binary_path, node_name=node_name)
-            # only support a subset of telemetry for Docker hosts (specifically, we do not allow users to enable any devices)
+            p = self._start_process(cmd="docker-compose -f %s up" % binary_path)
```
dliappis: I suggest we modify the `docker-compose.yml` file (btw let's rename it to `.yml.j2`) to have a healthcheck like:

```
    healthcheck:
      test: nc -z 127.0.0.1 {{http_port}}
      interval: 5s
      timeout: 2s
      retries: 10
```

Then we can start in daemonized mode with `-d`  **and** immediately after poll the output of `docker ps -a --filter "id=$(docker-compose -f %s ps -q)" --filter "status=running" --filter "health=healthy" -q` for say 1 minute:

- if result is still an empty line after 1 minute, it failed.
- if successfully up it will return the container id (like: `29f8d70b94dc`), terminate the loop early and return success.

esrally/utils/sysstats.py
```diff
@@ -45,7 +45,8 @@ def cpu_model():
     """
     :return: The CPU model name.
     """
-    return cpuinfo.get_cpu_info()["brand"] if cpuinfo_available else "Unknown"
+    cpu_info = cpuinfo.get_cpu_info()
```
danielmitterdorfer: The idea behind this single line statement was that we are lenient whether the `cpuinfo` package is supported on the target platform (see comment at the top in `sysstats`). If we execute `cpuinfo.get_cpu_info()` unconditionally we'd raise an error on those platforms.

esrally/mechanic/launcher.py
drawlerr: I implemented the rough equivalent with the Docker Python SDK.  What do you think?

esrally/mechanic/launcher.py
```diff
@@ -122,83 +127,32 @@ def stop(self, c):
...
+
+def _wait_for_healthy_running_container(container_id, timeout=60):
+    import docker
+    cli = docker.api.APIClient()
+
+    endtime = _time() + timeout
+    while _time() < endtime:
+        containers = cli.containers(quiet=True, all=True, filters={"id": container_id,
+                                                                   "health": "healthy",
+                                                                   "status": "running"})
```
dliappis: So if I understand correctly this is the only place where we use the imported docker library.

There is no question that this is cleaner from a code PoV, however, this adds a dependency that apart from needing to be listed (incl. its license, apache2) in `setup.py` it pulls a number of transitive dependencies (see below.

Ideally we want to keep the number of dependencies to the minimum and some of the dependent libraries are also used by Rally, hence have the potential for version conflicts in the future.

Given all this, does it make sense to just rely on explicit docker cli invocation here and not rely on the library? Did you hit any issue that I am overlooking that made you steer towards the library?


```
$ pip3 install docker
Collecting docker
  Using cached https://files.pythonhosted.org/packages/91/93/310fe092039f6b0759a1f8524e9e2c56f8012804fa2a8da4e4289bb74d7c/docker-4.0.1-py2.py3-none-any.whl
Requirement already satisfied: six>=1.4.0 in ./.venv/lib/python3.6/site-packages (from docker) (1.12.0)
Requirement already satisfied: requests!=2.18.0,>=2.14.2 in ./.venv/lib/python3.6/site-packages (from docker) (2.21.0)
Collecting websocket-client>=0.32.0 (from docker)
  Using cached https://files.pythonhosted.org/packages/29/19/44753eab1fdb50770ac69605527e8859468f3c0fd7dc5a76dd9c4dbd7906/websocket_client-0.56.0-py2.py3-none-any.whl
Requirement already satisfied: idna<2.9,>=2.5 in ./.venv/lib/python3.6/site-packages (from requests!=2.18.0,>=2.14.2->docker) (2.8)
Requirement already satisfied: urllib3<1.25,>=1.21.1 in ./.venv/lib/python3.6/site-packages (from requests!=2.18.0,>=2.14.2->docker) (1.22)
Requirement already satisfied: certifi>=2017.4.17 in ./.venv/lib/python3.6/site-packages (from requests!=2.18.0,>=2.14.2->docker) (2018.11.29)
Requirement already satisfied: chardet<3.1.0,>=3.0.2 in ./.venv/lib/python3.6/site-packages (from requests!=2.18.0,>=2.14.2->docker) (3.0.4)
Installing collected packages: websocket-client, docker
Successfully installed docker-4.0.1 websocket-client-0.56.0
```

drawlerr: I only used the SDK due to a general preference for APIs over CLIs in my code, but it does not make much sense here.

esrally/mechanic/launcher.py
```diff
@@ -122,83 +127,32 @@ def stop(self, c):
...
-        else:
-            msg += " The process seems to be still running with PID [%s]." % process.pid
-        log.error(msg)
-        raise exceptions.LaunchError(msg)
+    output = subprocess.check_output(args=shlex.split(compose_ps_cmd))
+    return output.decode("utf-8").rstrip()
+
+
+def _wait_for_healthy_running_container(container_id, timeout=60):
+    import docker
```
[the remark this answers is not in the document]
drawlerr: I didn't originally put it in setup.py because I didn't want to impose the docker SDK requirement on people who weren't using Docker functionality.
However, since I am removing usage of the Docker SDK entirely, I will mark this as "resolved"

esrally/mechanic/launcher.py
```diff
@@ -243,7 +199,7 @@ def stop(self, nodes):
             self.logger.info("Stopping Docker container")
             for node in nodes:
                 node.telemetry.detach_from_node(node, running=True)
-                process.run_subprocess_with_logging("docker-compose -f %s down" % self.binary_paths[node.node_name])
+                process.run_subprocess_with_logging(_get_docker_compose_cmd(self.binary_paths[self.node_name], "down"))
```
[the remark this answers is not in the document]
drawlerr: This was a mistake that I corrected shortly after testing, but it seems I forgot to push the fix.

dliappis:
This needs to be re-opened as nightlies fail with:

```
esrally.exceptions.RallyError: (Daemon startup failed with exit code[1], 'Traceback (most recent call last):\n  File "/var/lib/jenkins/src/rally/esrally/mechanic/mechanic.py", line 576, in receiveMsg_StartNodes\n    nodes = self.mechanic.start_engine()\n  File "/var/lib
/jenkins/src/rally/esrally/mechanic/mechanic.py", line 712, in start_engine\n    self.nodes = self.launcher.start(node_configs)\n  File "/var/lib/jenkins/src/rally/esrally/mechanic/launcher.py", line 282, in start\n    return [self._start_node(node_configuration, node_c
ount_on_host) for node_configuration in node_configurations]\n  File "/var/lib/jenkins/src/rally/esrally/mechanic/launcher.py", line 282, in <listcomp>\n    return [self._start_node(node_configuration, node_count_on_host) for node_configuration in node_configurations]\n
  File "/var/lib/jenkins/src/rally/esrally/mechanic/launcher.py", line 311, in _start_node\n    node_pid = self._start_process(binary_path, env)\n  File "/var/lib/jenkins/src/rally/esrally/mechanic/launcher.py", line 361, in _start_process\n    raise exceptions.LaunchEr
ror(msg)\nesrally.exceptions.LaunchError: (\'Daemon startup failed with exit code[1]\', None)\n')
```

PR #702, opened 2019-06-04
Add Rally Docker image to release process

### LINKED ISSUE #607: Provide a Rally docker image
In some environments it is troublesome to install all of Rally's dependencies. We should therefore provide a Docker image for Rally during the release process. One aspect that we can also address in a later release are use cases which involve the Rally daemon, like load test drivers that are distributed.

dliappis:
and also add integration tests for the dockerized version of Rally.

~Opening this as draft until we can finalize some details.~

Closes #607


docker/Dockerfiles/Dockerfile-dev
```diff
@@ -0,0 +1,77 @@
+##############################################################
+# Build stage 0 `builder`:
+# Install Rally from source inside a virtualenv
+################################################################################
```
danielmitterdorfer: Very minor nit: The `#` lines have a different length

docker/Dockerfiles/Dockerfile-release
```diff
@@ -0,0 +1,42 @@
+FROM python:3.7.3-slim
+ARG RALLY_VERSION
+ARG RALLY_LICENSE
+
+ENV RALLY_RUNNING_IN_DOCKER True
+
+RUN apt-get -y update && \
+    apt-get install -y curl git gcc && \
+    apt-get -y upgrade && \
```
danielmitterdorfer: Could we decrease the size of the image a bit further by cleaning up any downloaded files that `apt-get` has cached locally?

docker/bin/entrypoint.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+set -Eeo pipefail
+
+if [[ $1 == *"bash"* || $1 == *"sh"* ]]; then
+    : # noop
+elif [[ "$*" == *"--help"* || "$*" == *"list "* ]]; then
+    set -- esrally "$@" # allow --help, --list commands to be passed as is
```
danielmitterdorfer: Did you mean to write `list` instead of `--list`?  Also, why does this need special treatment? What about other subcommands like `compare`?

integration-test.sh
```diff
@@ -395,6 +395,54 @@ function test_proxy_connection {
     restore_rally_log
 }
 
+function docker_compose_dev {
+    if [[ "$1" == "up" ]]; then
+        docker-compose -f docker/docker-compose-tests.yml up --abort-on-container-exit
```
danielmitterdorfer: Should we mention `docker` and `docker-compose` as [prerequisites for development in our docs](https://esrally.readthedocs.io/en/stable/developing.html#prerequisites)?

integration-test.sh
```diff
@@ -395,6 +395,54 @@ function test_proxy_connection {
...
+    elif [[ "$1" == "down" ]]; then
+        docker-compose -f docker/docker-compose-tests.yml down -v
+    fi
```
danielmitterdorfer: Do you think it makes sense to add an `else` branch that raises an error?

integration-test.sh
```diff
@@ -395,6 +395,54 @@ function test_proxy_connection {
...
+}
+
+function test_docker_image {
+    if ! docker_is_running; then
```
danielmitterdorfer: I wonder whether we should turn this into a hard requirement now,.

integration-test.sh
```diff
@@ -395,6 +395,54 @@ function test_proxy_connection {
...
+
+    # Build the docker image
+    docker build -t elastic/rally:${RALLY_VERSION} --build-arg RALLY_VERSION --build-arg RALLY_LICENSE -f docker/Dockerfiles/Dockerfile-dev $PWD
+
+    export TEST_COMMAND="--pipeline=benchmark-only --test-mode --track=geonames --challenge=append-no-conflicts-index-only --target-hosts=es01:9200"
+    info "Testing Rally docker image using parameters: ${TEST_COMMAND}"
+    docker_compose_dev up
+    docker_compose_dev down
+
+    # --list should work
```
danielmitterdorfer: Nit: `list`, not `--list`

docker/docker-compose-tests.yml
```diff
@@ -0,0 +1,38 @@
+version: '2.2'
+services:
+  rally:
+    image: elastic/rally:${RALLY_VERSION}
+    container_name: rally
+    command: ${TEST_COMMAND}
```
danielmitterdorfer: I assume this is blocking and we won't run into a race condition in our tests where we basically do a `docker-compose up` followed by a `docker-compose down`?

release-docker.sh
```diff
@@ -0,0 +1,127 @@
...
+
+function docker_compose {
+    if [[ "$1" == "up" ]]; then
+        docker-compose -f docker/docker-compose-tests.yml up --abort-on-container-exit
+    elif [[ "$1" == "down" ]]; then
+        docker-compose -f docker/docker-compose-tests.yml down -v
+    fi
+}
+
+function test_docker_image {
```
danielmitterdorfer: Could we delegate this testing maybe to `integration-test.sh`, e.g. by calling it with a parameter? I think it would be ok to assume that this workflow is only called by a Makefile target.

release-docker.sh
```diff
@@ -0,0 +1,127 @@
...
+    docker_compose up
+    docker_compose down
+
+
+    unset TEST_COMMAND
+}
+
+function push_failed {
+    echo "Error while pushing Docker image. Did you \`docker login\`?"
+
```
danielmitterdorfer: Nit: Trailing new line

docker/Dockerfiles/Dockerfile-release
dliappis: As mentioned in https://docs.docker.com/develop/develop-images/#run (need to scroll further down to APT):

> Official Debian and Ubuntu images automatically run apt-get clean, so explicit invocation is not required.

The image we are basing on (python:3.7.3-slim) is based on the official debian stretch (https://github.com/docker-library/python/blob/4df129649329186fb42c9c33215b41d82bc0e58e/3.7/stretch/slim/Dockerfile#L7) so what we are doing here (i.e. `rm -rf /varl/ib/apt/lists/*`) should be enough.

docker/bin/entrypoint.sh
dliappis: Good catch left over from previous logic. We don't need special handling for parameters any more, only the "syntactic sugar" of prefixing `esrally` if not specified. Will address.

docker/docker-compose-tests.yml
dliappis: Correct, docker-compose will return control only when all services are up and satisfy the declared dependencies.

dliappis:
@danielmitterdorfer I addressed the ^^ comments and also added documentation (both regular and release).

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
...
+    -------------------------------
+
+
+As a next step, we assume that Elasticsearch is running on ``es01:9200`` and is accessible from the host where you are running the Rally Docker image.
+Run the ``nyc_taxis`` track in ``test-mode`` using::
+
+    $ docker run elastic/rally --track=nyc_taxis --test-mode --pipeline=benchmark-only --target-hosts=es01:9200
+
+
+:NOTE: We didn't need to explicitly specify ``esrally`` as we'd normally do in a normal CLI invocation; the entrypoint in the Docker image does this automatically.
```
danielmitterdorfer: This is rendered differently than in the rest of our docs. We should use this here:

```rst
.. note::
    We didn't need to explicitly specify ``esrally`` as we'd normally do in a normal CLI invocation; the entrypoint in the Docker image does this automatically.
```

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
...
+    myrally/benchmarks/races/
+    └── 2019-06-05-14-03-44
+        └── race.json
+
+    1 directory, 1 file
+
+
+In case you forgot to bind mount a directory, the Rally Docker image will create an `anonymous volume <https://success.docker.com/article/different-types-of-volumes>`_ for ``/rally/.rally`` to ensure logs and results get persisted even after the container has terminated.
+
+For example, after executing our earlier quickstart example ``docker run elastic/rally --track=nyc_taxis --test-mode --pipeline=benchmark-only --target-hosts=es01:9200``, ``docker volume ls`` shows a volume:::
```
danielmitterdorfer: Nit: This line ends with three colons but it only requires two.

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
...
+
+Specifics about the image
+-------------------------
+
+Rally runs as user ``1000`` and its files are installed with uid:gid ``1000:0`` (to support `Arbitrary User IDs <https://docs.openshift.com/enterprise/3.1/creating_images/guidelines.html>`_).
+
+Extending the Docker image
+--------------------------
+
+Creating your own customized Docker image is simple. You can start with a ``Dockerfile`` as simple as::
```
danielmitterdorfer: We should probably rephrase this to avoid calling it "simple"? ([rationale](https://developers.google.com/style/tone)). How about this?

> You can also create your own customized Docker image on top of the existing one. The example below shows how to get started::

docs/install.rst
```diff
@@ -66,6 +66,11 @@ We recommend that you use `Homebrew <https://brew.sh/>`_::
 
     brew install python3
 
+**Docker**
```
danielmitterdorfer: This is in the wrong section (prerequisites)? I think this should go in the section `Installing Rally` just below `PyEnv Install`?

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
+Running Rally with Docker
+=========================
+
+Rally is also available as a `Docker image <https://hub.docker.com/r/elastic/rally>`_.
```
danielmitterdorfer: Remove the `also`?

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
...
+
+.. _docker_limitations:
+
+Limitations
+-----------
+
+The following Rally functionality isn't supported or isn't recommended when using the Docker image:
```
danielmitterdorfer: I think we can scratch the part `or isn't recommended` as we *actively* prohibit this.

docs/install.rst
dliappis: Right thanks for spotting. Fixed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
...
+    Configuration successfully written to /rally/.rally/rally.ini. Happy benchmarking!
+
+    More info about Rally:
+
+    * Type esrally --help
+    * Read the documentation at https://esrally.readthedocs.io/en/latest/
+    * Ask a question on the forum at https://discuss.elastic.co/c/elasticsearch/rally
+
+    # now run our benchmark
+    docker run --rm -v $PWD/tmpdir:/rally/.rally elastic/rally --track=nyc_taxis --test-mode --pipeline=benchmark-only --target-hosts=es01:9200
```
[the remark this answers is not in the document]
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
```diff
@@ -0,0 +1,162 @@
...
+
+If your local bind mount doesn't contain a ``rally.ini`` the container will create one for you during the first run.
+
+Example::
+
+    mkdir myrally
+    sudo chgrp 0 myrally
+
+    # First run will generate the rally.ini
+    docker run --rm -v $PWD/tmpdir:/rally/.rally elastic/rally --track=nyc_taxis --test-mode --pipeline=benchmark-only --target-hosts=es01:9200
```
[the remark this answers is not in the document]
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/af1a12402329af7bf27343bc33d63d3fef7abd4b

docs/docker.rst
dliappis: Addressed in https://github.com/elastic/rally/pull/702/commits/f2e6db89dbd9cde0414581dbfe4a83a808013aa9

dliappis:
@danielmitterdorfer I addressed your comments. If there are no other commends I think we are good to merge?

PR #705, opened 2019-06-05
Provide default for datastore.secure in all cases

danielmitterdorfer:
With this commit we choose `False` as default value for the
configuration property `datastore.secure` even when an in-memory metrics
store is used. Contrary to the previous default (`""` (empty string)),
this avoids errors when the user manually changes their configuration
file to use an Elasticsearch metrics store but only provides host and
port.


danielmitterdorfer:
Thanks for the review! :)

PR #706, opened 2019-06-07
Register pyenv Python versions

danielmitterdorfer:
With this commit we ensure that make file targets work with all
supported Python versions by registering them in a project-specific file
that `pyenv` will read during the build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,8 @@ prereq: make-requirements.txt
 	pyenv install --skip-existing $(PY37)
 #	pyenv global system $(PY34) $(PY35) $(PY36) $(PY37)
 	pyenv global system $(PY35) $(PY36) $(PY37)
+	@# Ensure all Python versions are registered for this project
+	@eval "$$(pyenv init -)" && pyenv versions --bare > .python-version;
 	-@ printf $(PYENV_PREREQ_HELP)
 
 venv-create:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -36,6 +36,8 @@ prereq: make-requirements.txt
 	pyenv install --skip-existing $(PY37)
 #	pyenv global system $(PY34) $(PY35) $(PY36) $(PY37)
 	pyenv global system $(PY35) $(PY36) $(PY37)
+	@# Ensure all Python versions are registered for this project
+	@eval "$$(pyenv init -)" && pyenv versions --bare > .python-version;
```
dliappis: As discussed makes sense to just use the already present `make-requirements.txt` with something like `awk -F'=' '{print $2}' make-requirements.txt >.python-version`

Makefile
```diff
@@ -37,7 +37,7 @@ prereq: make-requirements.txt
 #	pyenv global system $(PY34) $(PY35) $(PY36) $(PY37)
 	pyenv global system $(PY35) $(PY36) $(PY37)
 	@# Ensure all Python versions are registered for this project
-	@eval "$$(pyenv init -)" && pyenv versions --bare > .python-version;
+	@awk -F'=' '{print $$2 > ".python-version"}' make-requirements.txt
```
dliappis: This works, but IMHO it is a bit strange to do a file redirection _inside_ awk for this use case;
would it be simpler to
`@awk -F'=' '{print $$2}' make-requirements.txt >.python-version`

i.e. redirect the final result of awk to `.python-version`?

danielmitterdorfer: I've change it now.

danielmitterdorfer:
Thanks for the review! :)

PR #707, opened 2019-06-07
Upgrade pip for virtualenv

danielmitterdorfer:
With this commit we upgrade pip to the latest version in the `make
install` target (used only for development). This avoids warnings about
an outdated version and let's us also use newer pip features in the
future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,7 @@ check-venv:
 	fi
 
 install: venv-create
+    . $(VENV_ACTIVATE_FILE); pip install --upgrade pip
 	. $(VENV_ACTIVATE_FILE); pip3 install -e .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -58,6 +58,7 @@ check-venv:
 	fi
 
 install: venv-create
+    . $(VENV_ACTIVATE_FILE); pip install --upgrade pip
```
dliappis: The indentation looks a bit odd. Tabs are needed for `make` targets, and I think this line doesn't have a tab character like the rest. Let's replace it with a tab.

danielmitterdorfer: Done

danielmitterdorfer:
Thanks for the review! :)

PR #708, opened 2019-06-07
Align default request param extraction for queries

danielmitterdorfer:
With this commit we align the extraction of default request parameters
for the `query` runner which comes in two flavors: request body search
and scroll search.


danielmitterdorfer:
Thanks for the review! :)

PR #709, opened 2019-06-07
Drop support for Python 3.4

danielmitterdorfer:
Python 3.4 has reached [end of
life](https://www.python.org/dev/peps/pep-0429/#release-schedule).
Consequently, we're dropping support in Rally as well and raise the
minimum supported version to Python 3.5.


docs/install.rst
```diff
@@ -27,32 +27,9 @@ Python
...
-.. note::
-
-    You will need to enable `EPEL <https://fedoraproject.org/wiki/EPEL>`_ before.
-
-*Tested on CentOS Linux release 7.4.1708 (Core).*
-
-::
-
-    sudo yum install -y gcc python34.x86_64 python34-devel.x86_64 python34-pip.noarch
+Please refer to the `installation instructions for Python 3.5 in the Red Hat Software Collections <https://www.softwarecollections.org/en/scls/rhscl/rh-python35/>`_.
```
dliappis: Makes sense to recommend scl for the CentOS platforms.

danielmitterdorfer:
Thanks for the review!

PR #710, opened 2019-06-07
Don't pass request_cache by default

### LINKED ISSUE #596: Search request_cache set to false by default
**Rally version** (get with `esrally --version`): 1.0.1

I am running benchmarks against master and I see that operation-type search sets `request_cache` set to `false` by default, though the docs say:

> cache (optional): Whether to use the query request cache. By default, Rally will define no value thus the default depends on the benchmark candidate settings and Elasticsearch version.

Based on the above, I would expect rally not to set the value, given that I am not providing it in my operation config, maybe though it is a good idea to disable the cache by default, so
I am not sure whether this is a bug or a docs problem.

danielmitterdorfer:
With this commit Rally will only pass the `request_cache` query
parameter to Elasticsearch iff the `cache` property is defined in the
track. Otherwise, it will adhere to the default that has been specified
in the index settings.

Closes #596

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -670,8 +670,9 @@ class Query(Runner):
 
     def _default_request_params(self, params):
         request_params = params.get("request-params", {})
-        if "cache" in params:
-            request_params["request_cache"] = str(params["cache"]).lower()
+        cache = params.get("cache")
+        if cache is not None:
+            request_params["request_cache"] = str(cache).lower()
         return request_params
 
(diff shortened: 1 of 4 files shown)


danielmitterdorfer:
I've marked this change as breaking as Rally will behave differently now in case `cache` is not specified in the track. To ensure that the standard benchmark suite still behaves the same, this requires https://github.com/elastic/rally-tracks/pull/78 to be merged (and backported) first.

danielmitterdorfer:
Thanks for your reviews! :)

PR #711, opened 2019-06-13
Change telemetry devices to rely on jvm.config instead of ES_JAVA_OPTS

ebadyano:
Move telemetry devices that currently rely on ES_JAVA_OPTS from the
launcher to the provisioner and instead persist the necessary
information in config/jvm.options

### DIFF
--- a/esrally/mechanic/provisioner.py
+++ b/esrally/mechanic/provisioner.py
@@ -247,7 +247,6 @@ class BareProvisioner:
         
         java_opts = self._prepare_java_opts()
         if java_opts:
-            print("here not empty")
             provisioner_vars["additional_java_settings"] = java_opts
         
 
--- a/esrally/mechanic/provisioner.py
+++ b/esrally/mechanic/provisioner.py
@@ -24,7 +24,7 @@ import jinja2
 
 from esrally import exceptions
 from esrally.mechanic import team, java_resolver, telemetry
-from esrally.utils import io, process, versions
+from esrally.utils import io, process, versions, jvm
 
 
 def local_provisioner(cfg, car, plugins, cluster_settings, all_node_ips, target_root, node_id):
@@ -166,10 +166,11 @@ class BareProvisioner:
         self.distribution_version = distribution_version
         self.apply_config = apply_config
         self.telemetry = telemetry
+        self.logger = logging.getLogger(__name__)
 
     def prepare(self, binary):
         if not self.preserve:
-            logging.getLogger(__name__).info("Rally will delete the benchmark candidate after the benchmark")
+            self.logger.info("Rally will delete the benchmark candidate after the benchmark")
         self.es_installer.install(binary["elasticsearch"])
         # we need to immediately delete it as plugins may copy their configuration during installation.
         self.es_installer.delete_pre_bundled_configuration()
@@ -197,6 +198,20 @@ class BareProvisioner:
 
     def cleanup(self):
         self.es_installer.cleanup(self.preserve)
+        
+    def _prepare_java_opts(self):
+        java_opts = []
+        if self.telemetry is not None:
+            java_opts.extend(self.telemetry.instrument_candidate_java_opts(self.es_installer.car, self.es_installer.node_name))
+        
+        exit_on_oome_flag = "-XX:+ExitOnOutOfMemoryError"
+        if jvm.supports_option(self.es_installer.java_home, exit_on_oome_flag):
+            self.logger.info("Setting [%s] to detect out of memory errors during the benchmark.", exit_on_oome_flag)
+            java_opts.append(exit_on_oome_flag)
+        else:
+            self.logger.info("JVM does not support [%s]. A JDK upgrade is recommended.", exit_on_oome_flag)
+        
+        return java_opts
 
     def _provisioner_variables(self):
         plugin_variables = {}
@@ -229,8 +244,13 @@ class BareProvisioner:
         provisioner_vars.update(self.es_installer.variables)
         provisioner_vars.update(plugin_variables)
         provisioner_vars["cluster_settings"] = cluster_settings
-        if self.telemetry is not None:
-            provisioner_vars["additional_java_settings"] = self.telemetry.instrument_candidate_java_opts(self.es_installer.car, self.es_installer.node_name)
+        
+        java_opts = self._prepare_java_opts()
+        if java_opts:
+            print("here not empty")
+            provisioner_vars["additional_java_settings"] = java_opts
+        
+
 
         return provisioner_vars
 
--- a/esrally/mechanic/provisioner.py
+++ b/esrally/mechanic/provisioner.py
@@ -39,7 +39,7 @@ def local_provisioner(cfg, car, plugins, cluster_settings, all_node_ips, target_
 
     _, java_home = java_resolver.java_home(car, cfg)
     
-    node_telemetry_dir = "%s/telemetry" % node_root_dir
+    node_telemetry_dir = os.path.join(node_root_dir, "telemetry")
     java_major_version, java_home = java_resolver.java_home(car, cfg)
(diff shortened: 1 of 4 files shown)


ebadyano:
Depends on https://github.com/elastic/rally-teams/pull/28

esrally/mechanic/provisioner.py
```diff
@@ -38,11 +38,22 @@ def local_provisioner(cfg, car, plugins, cluster_settings, all_node_ips, target_
     node_root_dir = "%s/%s" % (target_root, node_name)
 
     _, java_home = java_resolver.java_home(car, cfg)
+    
+    node_telemetry_dir = "%s/telemetry" % node_root_dir
```
danielmitterdorfer: We should replace this with `os.path.join(node_root_dir, "telemetry")`.

esrally/mechanic/provisioner.py
```diff
@@ -165,6 +177,9 @@ def prepare(self, binary):
         # determine after installation because some variables will depend on the install directory
         target_root_path = self.es_installer.es_home_path
         provisioner_vars = self._provisioner_variables()
+        
+        # add java options for telemetry devices
+        provisioner_vars.update({"additional_java_settings" : self.telemetry.instrument_candidate_env(self.es_installer.car, self.es_installer.node_name)})
```
danielmitterdorfer: We could move this to `prepare()`? Also, a simpler option would be to write it like:

```python
provisioner_vars["additional_java_settings"] = self.telemetry.instrument_candidate_env(self.es_installer.car, self.es_installer.node_name)
```

esrally/mechanic/telemetry.py
```diff
@@ -47,16 +47,12 @@ def __init__(self, enabled_devices=None, devices=None):
         self.devices = devices
 
     def instrument_candidate_env(self, car, candidate_id):
```
danielmitterdorfer: I wonder whether we should rename this method because we now provide only Java options.

esrally/mechanic/telemetry.py
```diff
@@ -47,16 +47,12 @@ def __init__(self, enabled_devices=None, devices=None):
...
-        opts = {}
+        opts = []
         for device in self.devices:
             if self._enabled(device):
                 additional_opts = device.instrument_env(car, candidate_id)
```
danielmitterdorfer: Similarly we should rename `instrument_env` of the individual telemetry devices.

esrally/mechanic/telemetry.py
```diff
@@ -190,7 +186,7 @@ def instrument_env(self, car, candidate_id):
         java_opts = self.java_opts(log_file)
 
         self.logger.info("jfr: Adding JVM arguments: [%s].", java_opts)
-        return {"ES_JAVA_OPTS": java_opts}
+        return java_opts.split()
```
danielmitterdorfer: I think we just need to return `java_opts` here? A Python list does not provide a `split` method?

esrally/mechanic/telemetry.py
```diff
@@ -232,8 +228,8 @@ def instrument_env(self, car, candidate_id):
         io.ensure_dir(self.log_root)
         log_file = "%s/%s-%s.jit.log" % (self.log_root, car.safe_name, candidate_id)
         console.info("%s: Writing JIT compiler log to [%s]" % (self.human_name, log_file), logger=self.logger)
-        return {"ES_JAVA_OPTS": "-XX:+UnlockDiagnosticVMOptions -XX:+TraceClassLoading -XX:+LogCompilation "
-                                "-XX:LogFile=%s -XX:+PrintAssembly" % log_file}
+        return ["-XX:+UnlockDiagnosticVMOptions", "-XX:+TraceClassLoading", "-XX:+LogCompilation",
+                                "-XX:LogFile=%s", "-XX:+PrintAssembly" % log_file ]
```
danielmitterdorfer: I think this should be:

```python
"-XX:LogFile={}".format(log_file), "-XX:+PrintAssembly"
```

esrally/mechanic/telemetry.py
```diff
@@ -255,12 +251,12 @@ def instrument_env(self, car, candidate_id):
 
     def java_opts(self, log_file):
         if self.java_major_version < 9:
-            return {"ES_JAVA_OPTS": "-Xloggc:%s -XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+PrintGCTimeStamps "
-                                    "-XX:+PrintGCApplicationStoppedTime -XX:+PrintGCApplicationConcurrentTime "
-                                    "-XX:+PrintTenuringDistribution" % log_file}
+            return ["-Xloggc:%s", "-XX:+PrintGCDetails", "-XX:+PrintGCDateStamps", "-XX:+PrintGCTimeStamps",
```
danielmitterdorfer: Similarly `"-Xloggc:{}".format(log_file)`

esrally/mechanic/telemetry.py
```diff
@@ -255,12 +251,12 @@ def instrument_env(self, car, candidate_id):
...
+                                    "-XX:+PrintGCApplicationStoppedTime", "-XX:+PrintGCApplicationConcurrentTime",
+                                    "-XX:+PrintTenuringDistribution" % log_file]
         else:
             # see https://docs.oracle.com/javase/9/tools/java.htm#JSWOR-GUID-BE93ABDC-999C-4CB5-A88B-1994AAAC74D5
-            return {"ES_JAVA_OPTS": "-Xlog:gc*=info,safepoint=info,age*=trace:file=%s:utctime,uptimemillis,level,tags:filecount=0" % log_file}
+            return ["-Xlog:gc*=info,safepoint=info,age*=trace:file=%s:utctime,uptimemillis,level,tags:filecount=0" % log_file]
```
danielmitterdorfer: For newer code we should use `#format()` instead of `%` unless it is a hot code path.

esrally/mechanic/provisioner.py
ebadyano: Sorry, not sure what you mean, this snippet is already in prepare()..

ebadyano: Should i move it to _provisioner_variables?

esrally/mechanic/telemetry.py
ebadyano: @danielmitterdorfer  For jfr version of this method, I kept java_opts as string and then split it into list by white space, the code seemed simpler that way. But I can change it to use list from the beginning. What do you think?

ebadyano: actually nevermind, it seems to work fine if i keep it as a string and return java_opts..

esrally/mechanic/provisioner.py
danielmitterdorfer: Sorry, I meant `self._provisioner_variables()`, yes. :)

esrally/mechanic/telemetry.py
ebadyano: java_opts = self.java_opts(log_file) for FlightRecorder still returns string, and here I split it into list. Do you think I should change java_opts to return list instead to make it more clear?

danielmitterdorfer: Yes, I think we should always return lists.

ebadyano:
@danielmitterdorfer Should I also add support for telemetry java options in docker provisioner in this PR or a different PR?

danielmitterdorfer:
IMHO this change should be consistent so you'd need to adapt the Docker provisioner as well. As I will start my vacation at the end of the day it would be best if @dliappis could pick this up as a reviewer.

dliappis:
@ebadyano I will do a review as well and test this a bit, so let's not merge yet please until I've done that too.

ebadyano:
Thank you for the review @dliappis . I update the change according to your comments.

PR #713, opened 2019-06-17
Check tags in track and team repos

danielmitterdorfer:
With this commit Rally does not only check for branch matches but also
for matching tags in track and team repos. This allows to "freeze"
support for older versions of Elasticsearch, i.e. once we drop support
for Elasticsearch 1.x, we can tag the branch `1` in rally-teams and
rally-tracks as `v1` (using a `v` prefix as convention for tags). Rally
can then still pick up configurations for these older versions that are
out of maintenance.


danielmitterdorfer:

PR #714, opened 2019-06-17
Document which ES versions are supported by Rally

danielmitterdorfer:
With this commit we add a dedicated page to our documentation that
explains which Elasticsearch versions can be benchmarked with which
Rally version.


danielmitterdorfer:
Thanks for the review!

PR #716, opened 2019-06-17
Drop support for Elasticsearch 1.x

### LINKED ISSUE #715: Drop support for Elasticsearch 1.x
In order to reduce maintenance effort, we should stop supporting outdated versions of Elasticsearch. Elasticsearch 1.7 is end of life since January 2017 so we will drop support for it. We will take a couple of precautionary measures though:

* We will give users a clear indication what Rally version supports which Elasticsearch versions (see #714)
* While we will remove the respective branches also from rally-teams and rally-tracks, we will add a tag to effectively "freeze" the latest state. Rally will bring support to detect these tags and use them (see #713). This will allow users to switch to an older Rally version and still benchmark old Elasticsearch versions. 

Additional tasks:

* [ ] Drop support in Rally: https://github.com/elastic/rally/pull/716
* [ ] Drop support in rally-tracks: https://github.com/elastic/rally-tracks/issues/11
* [ ] Drop support in rally-teams

The latter two items have to be implemented by removing the branch `1` in the respective repos and adding a corresponding `v1` tag. Contrary to Rally, these changes are live "immediately" and thus would affect users. Therefore, we should aim for a grace period and remove the branches late in the release cycle for Rally 1.3.0.

danielmitterdorfer:
With this commit we remove support in Rally to benchmark Elasticsearch
1.x.

Closes #715

### DIFF
--- a/docs/versions.rst
+++ b/docs/versions.rst
@@ -4,7 +4,7 @@ Elasticsearch Version Support
 Minimum supported version
 =========================
 
-Rally |release| can benchmark Elasticsearch 1.7.0 and above.
+Rally |release| can benchmark Elasticsearch 2.0.0 and above.
 
 End-of-life Policy
 ==================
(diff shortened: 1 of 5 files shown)


danielmitterdorfer:
Thanks for the review! :)

PR #720, opened 2019-07-01
Change filestore to be indexed by unique ID

drawlerr:
Races in the filestore will now be stored by their Trial ID, not their timestamp.

Relates #697

### DIFF
--- a/docs/tournament.rst
+++ b/docs/tournament.rst
@@ -25,16 +25,16 @@ After we've run both races, we want to know about the performance impact. With R
                     /____/
     Recent races:
 
-    Race Timestamp    Track    Track Parameters   Challenge            Car       User Tag
-    ----------------  -------  ------------------ -------------------  --------  ------------------------------
-    20160518T122341Z  pmc                         append-no-conflicts  defaults  intention:reduce_alloc_1234
-    20160518T112057Z  pmc                         append-no-conflicts  defaults  intention:baseline_github_1234
-    20160518T101957Z  pmc                         append-no-conflicts  defaults
+    Race ID                               Race Timestamp    Track    Track Parameters    Challenge            Car       User Tags
+    ------------------------------------  ----------------  -------  ------------------  ------------------   --------  ------------------------------
+    beb154e4-0a05-4f45-ad9f-e34f9a9e51f7  20160518T122341Z  pmc                          append-no-conflicts  defaults  intention:reduce_alloc_1234
+    0bfd4542-3821-4c79-81a2-0858636068ce  20160518T112057Z  pmc                          append-no-conflicts  defaults  intention:baseline_github_1234
+    0cfb3576-3025-4c17-b672-d6c9e811b93e  20160518T101957Z  pmc                          append-no-conflicts  defaults
 
 
 We can see that the user tag helps us to recognize races. We want to compare the two most recent races and have to provide the two race timestamps in the next step::
 
-    dm@io:~ $ esrally compare --baseline=20160518T112057Z --contender=20160518T112341Z
+    dm@io:~ $ esrally compare --baseline=beb154e4-0a05-4f45-ad9f-e34f9a9e51f7 --contender=0bfd4542-3821-4c79-81a2-0858636068ce
 
         ____        ____
        / __ \____ _/ / /_  __
--- a/docs/tournament.rst
+++ b/docs/tournament.rst
@@ -32,9 +32,9 @@ After we've run both races, we want to know about the performance impact. With R
     0cfb3576-3025-4c17-b672-d6c9e811b93e  20160518T101957Z  pmc                          append-no-conflicts  defaults
 
 
-We can see that the user tag helps us to recognize races. We want to compare the two most recent races and have to provide the two race timestamps in the next step::
+We can see that the user tag helps us to recognize races. We want to compare the two most recent races and have to provide the two race IDs in the next step::
 
-    dm@io:~ $ esrally compare --baseline=beb154e4-0a05-4f45-ad9f-e34f9a9e51f7 --contender=0bfd4542-3821-4c79-81a2-0858636068ce
+    dm@io:~ $ esrally compare --baseline=0bfd4542-3821-4c79-81a2-0858636068ce --contender=beb154e4-0a05-4f45-ad9f-e34f9a9e51f7
 
         ____        ____
        / __ \____ _/ / /_  __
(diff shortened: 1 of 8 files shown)


esrally/metrics.py
```diff
@@ -1396,8 +1396,8 @@ def __init__(self, es_store, es_results_store, file_store):
         self.es_results_store = es_results_store
         self.file_store = file_store
 
-    def find_by_timestamp(self, timestamp):
-        return self.es_store.find_by_timestamp(timestamp)
+    def find_by_trial_id(self, trial_id):
```
dliappis: Do we need this commit? My understanding is that when it comes to storing things in the `rally-races-*` index, we store both trial-id and trial-timestamp. Additionally `rally list races` still shows the timestamp (as well as `night-rally-admin list races`) so I thought the intent of https://github.com/elastic/rally/issues/697 is just about the installation directory:

>  Switch from installation directories based on the race timestamp to installation directories based on unique ids (~/.rally/benchmarks/nodes/$ID)

drawlerr: paths.race_root was modified to use trial_id instead of timestamp, which broke the previous implementation of find_by_timestamp for FileRaceStore.  Strictly speaking, there's no reason the EsRaceStore couldn't keep its find_by_timestamp functionality (as it doesn't use this function), but I thought it was cleaner to make it consistent with FileRaceStore and the abstract RaceStore.

dliappis:
I took an initial pass and have the following top level comments:

1. stop command doesn't seem to wipe away unneeded files (see https://github.com/elastic/rally/issues/697#issue-447596336 #stop)

    e.g. after `./rally stop --trial-id 9cbdf669-d2c4-424e-8a55-df4ee1d17ce0`

    I still see:
    ```
    $ du -sh ~/.rally/benchmarks/races/9cbdf669-d2c4-424e-8a55-df4ee1d17ce0/rally-node-0/*
    0B	/Users/dl/.rally/benchmarks/races/9cbdf669-d2c4-424e-8a55-df4ee1d17ce0/rally-node-0/heapdump
    423M	/Users/dl/.rally/benchmarks/races/9cbdf669-d2c4-424e-8a55-df4ee1d17ce0/rally-node-0/install
    48K	/Users/dl/.rally/benchmarks/races/9cbdf669-d2c4-424e-8a55-df4ee1d17ce0/rally-node-0/logs
    ```

2. Does `--node-ip` get used for [all_node_ids](https://github.com/elastic/rally/pull/722/files#diff-b307cecc08029f8c173e5b171522b463R51)?

    To setup a cluster we need `all_node_ips` to contain all IP addresses as it's used by the [elasticsearch.yml](https://github.com/elastic/rally-teams/blob/b5c15819f41dc7471610a04d74487636183b73eb/cars/v1/vanilla/templates/config/elasticsearch.yml#L72-L74) template for discovery.

3. I know this is what [#697](https://github.com/elastic/rally/issues/697) specifies and this PR implements it well, but the trial-id needs to be collected by some orchestrator (shell script, ansible play ...) and currently it's contained between the Rally banner and the success message:

    ```
    $ ./rally provision --node-ip 127.0.0.1 --docker --distribution-version=7.2.0

        ____        ____
    / __ \____ _/ / /_  __
    / /_/ / __ `/ / / / / /
    / _, _/ /_/ / / / /_/ /
    /_/ |_|\__,_/_/_/\__, /
                    /____/

    c9a4f171-48b4-48ad-8f4e-28c9ad262922

    -------------------------------
    [INFO] SUCCESS (took 1 seconds)
    -------------------------------```
    ```

    So the process invoking `provision` will have a hard time matching the id; should this be something like `trial-id: ...`? I was thinking of a simple json structure but again the orchestrator could have trouble skipping noise and multiline matching the json structure.
    Alternatively we could skip printing the banner and the final info message, but perhaps it's an unnecessary complication. @danielmitterdorfer WDYT?

dliappis:
@drawlerr https://github.com/elastic/rally/pull/720/commits/c8f0d829472683fdffeb9bdfdf6c150dcd02757e and https://github.com/elastic/rally/pull/720/commits/4131665f702209deb1dc485932145ebcaaf438c0 look good to me, thanks! The [tournament docs](https://esrally.readthedocs.io/en/stable/tournament.html) still need updating.

esrally/reporter.py
```diff
@@ -60,15 +60,15 @@ def summarize(race, cfg, lap=None):
...
-    baseline_ts = cfg.opts("reporting", "baseline.timestamp")
-    contender_ts = cfg.opts("reporting", "contender.timestamp")
+    baseline_id = cfg.opts("reporting", "baseline.id")
+    contender_id = cfg.opts("reporting", "contender.id")
 
-    if not baseline_ts or not contender_ts:
+    if not baseline_id or not contender_id:
         raise exceptions.SystemSetupError("compare needs baseline and a contender")
     race_store = metrics.race_store(cfg)
     ComparisonReporter(cfg).report(
```
danielmitterdorfer: I think in the output we should also show the `Race ID` (in addition to `Race timestamp`). Here is an example output (of the current state on this branch):

```
Comparing baseline
  Race timestamp: 2019-07-16 07:59:59
  Challenge: append-no-conflicts-index-only
  Car: defaults+basic-license

with contender
  Race timestamp: 2019-07-16 08:01:16
  Challenge: append-no-conflicts-index-only
  Car: defaults+basic-license
```

IMHO we should add in the first line also the Race ID` (as this is what the user has provided on the command line)

docs/tournament.rst
```diff
@@ -25,16 +25,16 @@ After we've run both races, we want to know about the performance impact. With R
...
-    20160518T112057Z  pmc                         append-no-conflicts  defaults  intention:baseline_github_1234
-    20160518T101957Z  pmc                         append-no-conflicts  defaults
+    Race ID                               Race Timestamp    Track    Track Parameters    Challenge            Car       User Tags
+    ------------------------------------  ----------------  -------  ------------------  ------------------   --------  ------------------------------
+    beb154e4-0a05-4f45-ad9f-e34f9a9e51f7  20160518T122341Z  pmc                          append-no-conflicts  defaults  intention:reduce_alloc_1234
+    0bfd4542-3821-4c79-81a2-0858636068ce  20160518T112057Z  pmc                          append-no-conflicts  defaults  intention:baseline_github_1234
+    0cfb3576-3025-4c17-b672-d6c9e811b93e  20160518T101957Z  pmc                          append-no-conflicts  defaults
 
 
 We can see that the user tag helps us to recognize races. We want to compare the two most recent races and have to provide the two race timestamps in the next step::
```
danielmitterdorfer: The sentence is incorrect: We are now providing race ids, not race timestamps.

PR #721, opened 2019-07-03
Update DiskIo telemetry device to persist the counters

ebadyano:
Ensure that DiskIo telemetry does not rely on Rally being a parent
process of Elasticsearch and persists the disk counters at the beginning
of a benchmark and can read it again afterwards.

Relates to https://github.com/elastic/rally/issues/697

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -295,7 +295,7 @@ class ProcessLauncher:
         enabled_devices = self.cfg.opts("mechanic", "telemetry.devices")
         telemetry_params = self.cfg.opts("mechanic", "telemetry.params")
         node_telemetry = [
-            telemetry.DiskIo(self.metrics_store, node_count_on_host, node_telemetry_dir, car, node_name),
+            telemetry.DiskIo(self.metrics_store, node_count_on_host, node_telemetry_dir, node_name),
             telemetry.NodeEnvironmentInfo(self.metrics_store),
             telemetry.IndexSize(data_paths, self.metrics_store),
             telemetry.MergeParts(self.metrics_store, node_configuration.log_path),
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -295,7 +295,7 @@ class ProcessLauncher:
         enabled_devices = self.cfg.opts("mechanic", "telemetry.devices")
         telemetry_params = self.cfg.opts("mechanic", "telemetry.params")
         node_telemetry = [
-            telemetry.DiskIo(self.metrics_store, node_count_on_host),
+            telemetry.DiskIo(self.metrics_store, node_count_on_host, node_telemetry_dir, car, node_name),
             telemetry.NodeEnvironmentInfo(self.metrics_store),
             telemetry.IndexSize(data_paths, self.metrics_store),
             telemetry.MergeParts(self.metrics_store, node_configuration.log_path),
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -776,10 +776,11 @@ class MergeParts(InternalTelemetryDevice):
 
 
 class DiskIo(InternalTelemetryDevice):
+    human_name = "Disk IO"
     """
     Gathers disk I/O stats.
     """
-    def __init__(self, metrics_store, node_count_on_host):
+    def __init__(self, metrics_store, node_count_on_host, log_root, car, node_name):
         super().__init__()
         self.metrics_store = metrics_store
         self.node_count_on_host = node_count_on_host
@@ -787,6 +788,9 @@ class DiskIo(InternalTelemetryDevice):
         self.process = None
         self.disk_start = None
         self.process_start = None
+        self.car = car
+        self.node_name = node_name
+        self.log_root = log_root
 
     def attach_to_node(self, node):
         self.node = node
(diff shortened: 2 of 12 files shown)


ebadyano:
Accidentally clicked before marking it a draft PR, so adding `WIP` to the title.

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +799,53 @@ def attach_to_node(self, node):
     def on_benchmark_start(self):
         if self.process is not None:
             self.process_start = sysstats.process_io_counters(self.process)
+            read_bytes = 0
+            write_bytes = 0
+            io.ensure_dir(self.log_root)
+            log_file = "%s/%s-%s.io" % (self.log_root, self.car.safe_name, self.node.node_name)
```
dliappis: This for example produces something like:

```
.rally/benchmarks/races/2019-07-04-15-05-34/rally-node-0/telemetry/defaults-rally-node-0.io
```

I don't think `self.car.safe_name` (in this case `defaults`) provides value in the filename, in which case you don't even need to pass car in the constructor. WDYT?

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +799,53 @@ def attach_to_node(self, node):
...
                 try:
                     self.disk_start = sysstats.disk_io_counters()
+                    read_bytes = self.disk_start.read_bytes
+                    write_bytes = self.disk_start.write_bytes
                     self.logger.warning("Process I/O counters are not supported on this platform. Falling back to less accurate disk "
                                         "I/O counters.")
                 except RuntimeError:
                     self.logger.exception("Could not determine I/O stats at benchmark start.")
+            with open(log_file, "wt", encoding="utf-8") as f:
+                diskio_str = "%d %d" % (read_bytes, write_bytes)
```
dliappis: Have you considered writing to a dict and dumping it as json instead? What you have here works perfectly, but I wonder if standardizing to json for all structures we have in Rally makes sense.

So here we'd have `{"read_bytes": read_bytes, "write_bytes": write_bytes}` and then on the next line it could be `json.dump(diskio_str, f))`

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +799,53 @@ def attach_to_node(self, node):
...
 
     def on_benchmark_stop(self):
         if self.process is not None:
+            process_end = sysstats.process_io_counters(self.process)
+            disk_end = sysstats.disk_io_counters()
+            io.ensure_dir(self.log_root)
+            log_file = "%s/%s-%s.io" % (self.log_root, self.car.safe_name, self.node.node_name)
+            io_str = ""
+            with open(log_file, "rt", encoding="utf-8") as f:
+                io_str = f.read()
```
dliappis: See my earlier suggestion whether we could [json.load](https://docs.python.org/3.6/library/json.html#json.load) instead and use it as a dictionary without the need to split.

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +799,53 @@ def attach_to_node(self, node):
...
             # noinspection PyBroadException
             try:
                 # we have process-based disk counters, no need to worry how many nodes are on this host
-                if self.process_start:
-                    process_end = sysstats.process_io_counters(self.process)
-                    read_bytes = process_end.read_bytes - self.process_start.read_bytes
-                    write_bytes = process_end.write_bytes - self.process_start.write_bytes
-                elif self.disk_start:
+                if process_end:
+                    read_bytes = process_end.read_bytes - int(io_bytes[0])
```
dliappis: The json approach, as a benefit, can ensure type safety in a type unsafe language and you can get rid of the int conversions.

esrally/mechanic/launcher.py
```diff
@@ -295,7 +295,7 @@ def _start_node(self, node_configuration, node_count_on_host):
         enabled_devices = self.cfg.opts("mechanic", "telemetry.devices")
         telemetry_params = self.cfg.opts("mechanic", "telemetry.params")
         node_telemetry = [
-            telemetry.DiskIo(self.metrics_store, node_count_on_host),
+            telemetry.DiskIo(self.metrics_store, node_count_on_host, node_telemetry_dir, car, node_name),
```
dliappis: See my question below whether `car` is actually needed.

esrally/mechanic/telemetry.py
ebadyano: I agree, but I was trying to be consistent with what we do fo gcstats and jfr and other telemetry files. So not sure whether to go for simplicity or consistency. What's your preference?

esrally/mechanic/telemetry.py
ebadyano: I like your json idea, I will change it to use that. Thanks!

ebadyano:
@dliappis I wonder if I do need to do something about self.metrics_store as well. It's initialized in the launcher, but by the time stop/start is executed we don't have it anymore..

dliappis:
> @dliappis I wonder if I do need to do something about self.metrics_store as well. It's initialized in the launcher, but by the time stop/start is executed we don't have it anymore..

@ebadyano this needs to be looked together with https://github.com/elastic/rally/pull/722; as per https://github.com/elastic/rally/issues/697#issue-447596336 the race-id will be provided to the start/stop commands and each should still initialize the metric stores used for recoding telemetry and other metric records so in theory if all this initialization is done in #722 the code here should work as is.

esrally/mechanic/telemetry.py
danielmitterdorfer: Contrary to other telemetry devices this one is not meant to be inspected by the user. Instead this is a temporary state file that should also be removed when the benchmark ends (i.e. in `on_benchmark_stop`). Hence, I'd go for simplicity here.

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +800,52 @@ def attach_to_node(self, node):
     def on_benchmark_start(self):
         if self.process is not None:
             self.process_start = sysstats.process_io_counters(self.process)
+            read_bytes = 0
+            write_bytes = 0
+            io.ensure_dir(self.log_root)
+            log_file = "%s/%s-%s.io" % (self.log_root, self.car.safe_name, self.node.node_name)
+            console.info("%s: Writing start I/O stats to [%s]" % (self.human_name, log_file), logger=self.logger)
```
danielmitterdorfer: As this is only a temporary file that is necessary due to the way the `start` and `stop` subcommands will work, we should not inform the user about it.

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +800,52 @@ def attach_to_node(self, node):
...
             else:
                 try:
                     self.disk_start = sysstats.disk_io_counters()
+                    read_bytes = self.disk_start.read_bytes
+                    write_bytes = self.disk_start.write_bytes
                     self.logger.warning("Process I/O counters are not supported on this platform. Falling back to less accurate disk "
                                         "I/O counters.")
                 except RuntimeError:
                     self.logger.exception("Could not determine I/O stats at benchmark start.")
+            diskio_str = {"read_bytes": read_bytes, "write_bytes": write_bytes}
```
danielmitterdorfer: Should we avoid the `_str` suffix? Also, I think we could just inline the `dict` declaration and get rid of the variable completely.

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +800,52 @@ def attach_to_node(self, node):
...
+            with open(log_file, "wt", encoding="utf-8") as f:
+                json.dump(diskio_str, f)
 
     def on_benchmark_stop(self):
         if self.process is not None:
+            process_end = sysstats.process_io_counters(self.process)
+            disk_end = sysstats.disk_io_counters()
+            io.ensure_dir(self.log_root)
+            log_file = "%s/%s-%s.io" % (self.log_root, self.car.safe_name, self.node.node_name)
```
danielmitterdorfer: Can we use `os.path.join` instead of `/` here?

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +800,52 @@ def attach_to_node(self, node):
...
+            io_str = ""
```
danielmitterdorfer: Leftover?

esrally/mechanic/telemetry.py
```diff
@@ -795,35 +800,52 @@ def attach_to_node(self, node):
...
             # noinspection PyBroadException
             try:
                 # we have process-based disk counters, no need to worry how many nodes are on this host
-                if self.process_start:
-                    process_end = sysstats.process_io_counters(self.process)
-                    read_bytes = process_end.read_bytes - self.process_start.read_bytes
-                    write_bytes = process_end.write_bytes - self.process_start.write_bytes
-                elif self.disk_start:
+                if process_end:
+                    read_bytes = process_end.read_bytes - io_bytes['read_bytes']
```
danielmitterdorfer: We should use double quotes for string constants by default (there are cases where it makes sense to use single quotes, e.g. when there are a lot of double quotes used in the string constant, it is easier to use single quotes but here this is not the case).

esrally/mechanic/telemetry.py
```diff
@@ -776,17 +777,21 @@ def _store_merge_times(self, merge_times):
 
 
 class DiskIo(InternalTelemetryDevice):
+    human_name = "Disk IO"
```
danielmitterdorfer: when we get rid of the console message (see my comment below), this will be unused and we can remove it as well.

esrally/mechanic/telemetry.py
```diff
@@ -776,17 +777,21 @@ def _store_merge_times(self, merge_times):
 
 
 class DiskIo(InternalTelemetryDevice):
```
danielmitterdorfer: Unfortunately, we don't have a test yet for this class but I think it would be good to have one now. For the test I suggest you set `log_root` to the temporary directory (see `tempfile.mkdtemp()`).

esrally/mechanic/telemetry.py
ebadyano: fixed

ebadyano:
@danielmitterdorfer Thank you for the review, I addressed your comments. Could you please take another look? Thanks!

PR #723, opened 2019-07-08
Always exit on OOME

danielmitterdorfer:
With this commit we add the JVM flag `ExitOnOutOfMemoryError`
unconditionally when Elasticsearch is configured by Rally. Previously we
had to check whether the JVM in use supports this flag because we
supported Elasticsearch 1.x which can be run with Java 1.7. As the JVM flag
has only been introduced with Java 8, we had a check in place. Now that
we have dropped support for Elasticsearch 1.x (in #716), we can safely
assume that the JVM supports this flag and unconditionally set it.

Relates #715

### DIFF
--- a/esrally/mechanic/provisioner.py
+++ b/esrally/mechanic/provisioner.py
@@ -199,17 +199,10 @@ class BareProvisioner:
         self.es_installer.cleanup(self.preserve)
         
     def _prepare_java_opts(self):
-        java_opts = []
+        # To detect out of memory errors during the benchmark
+        java_opts = ["-XX:+ExitOnOutOfMemoryError"]
         if self.telemetry is not None:
             java_opts.extend(self.telemetry.instrument_candidate_java_opts(self.es_installer.car, self.es_installer.node_name))
-        
-        exit_on_oome_flag = "-XX:+ExitOnOutOfMemoryError"
-        if jvm.supports_option(self.es_installer.java_home, exit_on_oome_flag):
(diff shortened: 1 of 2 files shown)


esrally/mechanic/provisioner.py
```diff
@@ -199,17 +199,10 @@ def cleanup(self):
...
         if self.telemetry is not None:
             java_opts.extend(self.telemetry.instrument_candidate_java_opts(self.es_installer.car, self.es_installer.node_name))
-        
-        exit_on_oome_flag = "-XX:+ExitOnOutOfMemoryError"
-        if jvm.supports_option(self.es_installer.java_home, exit_on_oome_flag):
-            self.logger.info("Setting [%s] to detect out of memory errors during the benchmark.", exit_on_oome_flag)
-            java_opts.append(exit_on_oome_flag)
-        else:
-            self.logger.info("JVM does not support [%s]. A JDK upgrade is recommended.", exit_on_oome_flag)
-
```
drawlerr: Looks like +ExitOnOutOfMemoryError was added in 7u101, 8u92.  We are removing this check because we have no intention of supporting such old JVMs?

danielmitterdorfer: Correct. We're dropping support for Elasticsearch 1.x in #716 and for 2.x we need at least Java 8. I think we can safely assume that users run 8u92 or higher.

drawlerr: Sounds good to me 👍

danielmitterdorfer:
Thanks!

Issue #664, opened 2019-03-07
Capture team and track revision in metrics metadata


danielmitterdorfer:
We already store some version info in metrics metadata:

* Elasticsearch version + revision
* JDK version
* kernel version
* Rally version

However, we do not store the track and team repository revision and we should add this as well.

**Implementation notes**

* It is possible for users to specify `--track-path` / `--team-path`. In those cases we should not track any revision because we are only referring to a plain file
* Similarly, if users are using `--pipeline=benchmark-only`, Rally does not manage the cluster thus th team-revision is not applicable.
* It will be a bit tricky to get the revisions:
    * For teams, the because is checked out on the target nodes where Elasticsearch is deployed but I think we should derive this info in `MechanicActor#receiveMsg_StartEngine()` when we are about to load the team. We then need to communicate this information back to the benchmark coordinator actor (`BenchmarkActor`) in `MechanicActor#on_cluster_started()` as an additional property in the `EngineStarted` message.
    * For tracks, we load the track in `BenchmarkActor#setup()` and can derive the revision there.

PR #725, opened 2019-07-10
Capture team and track revisions in metrics metadata

ebadyano:
Still missing track revision info.

Closes https://github.com/elastic/rally/issues/664

### DIFF
--- a/esrally/mechanic/mechanic.py
+++ b/esrally/mechanic/mechanic.py
@@ -23,7 +23,7 @@ from collections import defaultdict
 import thespian.actors
 
 from esrally import actor, client, paths, config, metrics, exceptions
-from esrally.utils import net, console, git
+from esrally.utils import net, console
 from esrally.mechanic import supplier, provisioner, launcher, team
 
 
@@ -303,10 +303,8 @@ class MechanicActor(actor.RallyActor):
         self.metrics_store = cls(self.cfg)
         self.metrics_store.open(ctx=msg.open_metrics_context)
         name = self.cfg.opts("race", "pipeline")
-        team_path = self.cfg.opts("mechanic", "team.path", mandatory=False)
-        if not (name == "benchmark-only" or team_path):
-            self.team_revision = git.head_revision(team.team_path(self.cfg))
         self.car, _ = load_team(self.cfg, msg.external)
+        self.team_revision = self.cfg.opts("mechanic", "repository.revision")
 
         # In our startup procedure we first create all mechanics. Only if this succeeds we'll continue.
         hosts = self.cfg.opts("client", "hosts").default
--- a/esrally/mechanic/mechanic.py
+++ b/esrally/mechanic/mechanic.py
@@ -425,6 +425,7 @@ class MechanicActor(actor.RallyActor):
     def on_cluster_started(self):
         # We don't need to store the original node meta info when the node started up (NodeStarted message) because we actually gather it
         # in ``on_all_nodes_started`` via the ``ClusterLauncher``.
+        self.cluster.team_revision = self.team_revision
         self.send(self.race_control,
                   EngineStarted(ClusterMetaInfo([NodeMetaInfo(n) for n in self.cluster.nodes],
                                                 self.cluster.source_revision,
--- a/esrally/mechanic/mechanic.py
+++ b/esrally/mechanic/mechanic.py
@@ -23,7 +23,7 @@ from collections import defaultdict
 import thespian.actors
 
 from esrally import actor, client, paths, config, metrics, exceptions
-from esrally.utils import net, console
+from esrally.utils import net, console, git
 from esrally.mechanic import supplier, provisioner, launcher, team
 
 
@@ -45,11 +45,12 @@ def download(cfg):
 ##############################
 
 class ClusterMetaInfo:
-    def __init__(self, nodes, revision, distribution_version, distribution_flavor):
+    def __init__(self, nodes, revision, distribution_version, distribution_flavor, team_revision):
         self.nodes = nodes
         self.revision = revision
         self.distribution_version = distribution_version
         self.distribution_flavor = distribution_flavor
+        self.team_revision = team_revision
 
     def as_dict(self):
         return {
@@ -57,7 +58,8 @@ class ClusterMetaInfo:
             "node-count": len(self.nodes),
             "revision": self.revision,
             "distribution-version": self.distribution_version,
-            "distribution-flavor": self.distribution_flavor
+            "distribution-flavor": self.distribution_flavor,
+            "team-revision": self.team_revision
         }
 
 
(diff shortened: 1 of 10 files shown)


esrally/racecontrol.py
```diff
@@ -106,6 +106,7 @@ def __init__(self):
         self.start_sender = None
         self.mechanic = None
         self.main_driver = None
+        self.track_revision= None
```
dliappis: PEP8: whitespace missing before `=`

esrally/racecontrol.py
```diff
@@ -206,7 +207,10 @@ def setup(self, msg, sender):
                 raise exceptions.SystemSetupError("A distribution version is required. Please specify it with --distribution-version.")
             self.logger.info("Automatically derived distribution version [%s]", distribution_version)
             self.cfg.add(config.Scope.benchmark, "mechanic", "distribution.version", distribution_version)
-
+        
+        name = self.cfg.opts("race", "pipeline")
+        if not (name == "benchmark-only" or track.is_simple_track_mode(self.cfg)):
+            self.track_revision = git.head_revision(track.track_path(self.cfg))
```
dliappis: IMHO invoking `git.head_revision` i.e. an FS+external command operation against a local directory seems out of place for this file.

As we are anyway supporting the `--track-revision` command line argument, we could always populate it for the `cfg` object with the exact value, if unspecified, in a separate module. In [loader.py/GitTrackRepository](https://github.com/elastic/rally/blob/9b9c7e03b79a11414740fa8f543b7a58b162149b/esrally/track/loader.py#L188-L206) and [SimpleTrackRepository](https://github.com/elastic/rally/blob/9b9c7e03b79a11414740fa8f543b7a58b162149b/esrally/track/loader.py#L219) we are doing a bunch of low-level stuff, so this could be a starting point?

Note that we are passing the [cfg object](https://github.com/elastic/rally/blob/9b9c7e03b79a11414740fa8f543b7a58b162149b/esrally/racecontrol.py#L231) gets passed to the actor system so the information should be accessible by actors.

esrally/mechanic/mechanic.py
```diff
@@ -290,6 +296,10 @@ def receiveMsg_StartEngine(self, msg, sender):
         cls = metrics.metrics_store_class(self.cfg)
         self.metrics_store = cls(self.cfg)
         self.metrics_store.open(ctx=msg.open_metrics_context)
+        name = self.cfg.opts("race", "pipeline")
+        team_path = self.cfg.opts("mechanic", "team.path", mandatory=False)
+        if not (name == "benchmark-only" or team_path):
+            self.team_revision = git.head_revision(team.team_path(self.cfg))
```
dliappis: Similar to my comment below about track-parameters, could we have a cfg argument for `team_revision` -- which could even be specified via cli arguments -- and initialize it, if unset, e.g. in [team_path](https://github.com/elastic/rally/blob/9b9c7e03b79a11414740fa8f543b7a58b162149b/esrally/mechanic/team.py#L128-L147)? The information could then be retrieved from the cfg object here.

esrally/metrics.py
```diff
@@ -1325,6 +1328,8 @@ def to_result_dicts(self):
         if plugins:
             result_template["plugins"] = list(plugins)
 
+        if self.track_revision:
+            result_template["track-revision"] = self.track_revision
```
dliappis: We'll need to adjust tests in [metrics_test.py](https://github.com/elastic/rally/blob/9b9c7e03b79a11414740fa8f543b7a58b162149b/tests/metrics_test.py#L911-L1087) to check this too, right?

dliappis:
re: abstracting the `git` operations I read #664 that this PR references and observed there is some guidance there about where each value (esp. for teams) could be retrieved:

> For teams, the because is checked out on the target nodes where Elasticsearch is deployed but I think we should derive this info in MechanicActor#receiveMsg_StartEngine() when we are about to load the team. We then need to communicate this information back to the benchmark coordinator actor (BenchmarkActor) in MechanicActor#on_cluster_started() as an additional property in the EngineStarted message.

> For tracks, we load the track in BenchmarkActor#setup() and can derive the revision there.

esrally/racecontrol.py
```diff
@@ -24,7 +24,7 @@
 import urllib
 
 from esrally import actor, config, exceptions, track, driver, mechanic, reporter, metrics, time, DOC_LINK, PROGRAM_NAME
-from esrally.utils import console, convert, opts
+from esrally.utils import console, convert, opts, git
```
dliappis: `git` is not needed anymore, right?

esrally/track/loader.py
```diff
@@ -204,6 +211,8 @@ def __init__(self, cfg, fetch, update, repo_class=repo.RallyRepository):
                 self.repo.checkout(repo_revision)
             else:
                 self.repo.update(distribution_version)
+                track_revision = git.head_revision(self.track_dir(self.track_name))
```
dliappis: Same comment here as earlier, if you move this inside `RallyRepository` you won't have to directly use `git` here, but just get the info from the instance attribute.

esrally/metrics.py
```diff
@@ -1203,7 +1203,7 @@ def format_dict(d):
         console.println("No recent races found.")
 
 
-def create_race(cfg, track, challenge):
+def create_race(cfg, track, challenge, track_revision=None):
```
dliappis: What about the team_revision? I didn't see it getting stored in metrics.

tests/metrics_test.py
```diff
@@ -830,6 +830,7 @@ def test_store_race(self):
                             pipeline="from-sources", user_tags={"os": "Linux"}, track=t, track_params={"shard-count": 3},
                             challenge=t.default_challenge, car="defaults", car_params={"heap_size": "512mb"}, plugin_params=None,
                             total_laps=12,
+                            track_revision="abc1",
```
dliappis: We should also support and test `team_revision`, right?

esrally/mechanic/team.py
```diff
@@ -144,6 +144,8 @@ def team_path(cfg):
             current_team_repo.checkout(repo_revision)
         else:
             current_team_repo.update(distribution_version)
+            team_revision = git.head_revision(current_team_repo.repo_dir)
```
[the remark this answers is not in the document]
ebadyano: Makes sense, thank you for the detailed explanation.

esrally/metrics.py
ebadyano: It's stored as part of ClusterMetaInfo.

dliappis: I did a test run with the latest commit in this PR against an Elasticsearch metrics store and didn't see the team revision recorded.
e.g.:

image

Btw the particular class [ClusterMetaInfo](https://github.com/elastic/rally/blob/c65654f8bf635de1162f4d97cc153ae7296e539e/esrally/mechanic/mechanic.py#L47) has a comment before indicating it's used internally.

tests/metrics_test.py
dliappis: As mentioned in the [above comment](https://github.com/elastic/rally/pull/725#discussion_r310100420) I think we need to test for `team_revision` to ensure things work as they are supposed?

ebadyano:
@dliappis Thank you for all the reviews!

PR #727, opened 2019-07-12
Allow function-level invocation of integration-test.sh.

drawlerr:
Use the first argument as the name of a function in integration-test.sh to call directly, in order to:
 - speed up testing during development
 - generalize release-docker.sh's invocation

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -543,12 +543,14 @@ function main {
 
 check_prerequisites
 
-# allow invocation from release-docker.sh
-if [[ $1 == "test-docker-release" ]]; then
-    test_docker_release_image
+trap "tear_down" EXIT
+
+# if argument is the name of a function, set up and call it
+if declare -f "$1" > /dev/null
+then
+    set_up
+    $1
     exit
 fi
-
-trap "tear_down" EXIT
 
 main
--- a/release-docker.sh
+++ b/release-docker.sh
@@ -46,7 +46,7 @@ echo "======================================================="
 echo "Testing Docker image for Rally release $RALLY_VERSION  "
 echo "======================================================="
 
-./integration-test.sh test-docker-release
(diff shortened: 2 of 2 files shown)


integration-test.sh
```diff
@@ -543,12 +543,14 @@ function main {
...
-# allow invocation from release-docker.sh
-if [[ $1 == "test-docker-release" ]]; then
-    test_docker_release_image
+trap "tear_down" EXIT
+
+# if argument is the name of a function, set up and call it
+if declare -f "$1" > /dev/null
+then
+    set_up
+    $1
```
dliappis: Since there are various helper functions in this script and maybe it's time consuming to scan for what is a real test, does it make sense that we store a whitelist with the functions we are allowed to call, currently: `test_configure,test_list,test_distribution_fails_with_wrong_track_params,test_download,test_sources,test_distributions,test_benchmark_only,test_docker_dev_image,test_docker_release_image`? We could also rely on the convention, like Python's testunit, that all callable functions start with `test_`.

If we do this, we could simplify the code in [run_test()](https://github.com/elastic/rally/blob/9b9c7e03b79a11414740fa8f543b7a58b162149b/integration-test.sh#L493-L512) to just loop through the whitelist.

danielmitterdorfer:
@drawlerr do you think we can move forward with this PR based on the review from @dliappis?

dliappis:
Just to clarify; the above [comment](https://github.com/elastic/rally/pull/727#discussion_r303412984) was an idea for an enhancement that could also slightly simplify the code base. If the implementation is too complex, IMHO, it can be part of a separate PR.

integration-test.sh
drawlerr: Personally, it doesn't feel to me like the protection we get (prevent users from calling utility functions individually?) is worth the additional maintenance burden, so I will merge as is.

PR #728, opened 2019-07-16
Allow to set distribution version as parameter

danielmitterdorfer:
With this commit we allow (again) to set the distribution version as a
command line parameter even in cases where it does not seem to be useful
initially, like the `benchmark-only` pipeline. However, if Rally fails
to determine the cluster version (for whatever reason) it can be useful
to be able to override this on the command line. Similarly, it can be
useful to override this when benchmarking a source build that is *not*
based on master but rather of an older branch. In these cases, Rally
would still take the right decisions with some guidance by the user.


danielmitterdorfer:
Thanks! :)

PR #722, opened 2019-07-04
Provision, start, stop subcommands

drawlerr:
Implement provision, start and stop subcommands for independently managing the benchmarking candidate.

Relates #697
Closes #733

PR #731, opened 2019-07-17
Update DiskIo telemetry device to persist the counters

ebadyano:
Ensure that DiskIo telemetry does not rely on Rally being a parent
process of Elasticsearch and persists the disk counters at the beginning
of a benchmark and can read it again afterwards.

Relates to #697

### DIFF
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -779,12 +779,12 @@ class DiskIo(InternalTelemetryDevice):
         super().__init__()
         self.metrics_store = metrics_store
         self.node_count_on_host = node_count_on_host
+        self.log_root = log_root
+        self.node_name = node_name
         self.node = None
         self.process = None
         self.disk_start = None
         self.process_start = None
-        self.node_name = node_name
-        self.log_root = log_root
 
     def attach_to_node(self, node):
         self.node = node
@@ -796,7 +796,7 @@ class DiskIo(InternalTelemetryDevice):
             read_bytes = 0
             write_bytes = 0
             io.ensure_dir(self.log_root)
-            tmp_io_file = os.path.join(self.log_root, "%s.io" % self.node_name)
+            tmp_io_file = os.path.join(self.log_root, "{}.io".format(self.node_name))
             if self.process_start:
                 read_bytes = self.process_start.read_bytes
                 write_bytes = self.process_start.write_bytes
@@ -814,29 +814,29 @@ class DiskIo(InternalTelemetryDevice):
                 json.dump({"pid": self.node.pid, "read_bytes": read_bytes, "write_bytes": write_bytes}, f)
 
     def on_benchmark_stop(self):
-            io.ensure_dir(self.log_root)
-            tmp_io_file = os.path.join(self.log_root, "%s.io" % self.node_name)
-            with open(tmp_io_file, "rt", encoding="utf-8") as f:
-                io_bytes = json.load(f)
-            os.remove(tmp_io_file)
-            self.process = sysstats.setup_process_stats(io_bytes["pid"])
-            process_end = sysstats.process_io_counters(self.process)
-            disk_end = sysstats.disk_io_counters()
             # Be aware the semantics of write counts etc. are different for disk and process statistics.
             # Thus we're conservative and only report I/O bytes now.
             # noinspection PyBroadException
             try:
+                io.ensure_dir(self.log_root)
+                tmp_io_file = os.path.join(self.log_root, "{}.io".format(self.node_name))
+                with open(tmp_io_file, "rt", encoding="utf-8") as f:
+                    io_stats = json.load(f)
+                os.remove(tmp_io_file)
+                self.process = sysstats.setup_process_stats(io_stats["pid"])
+                process_end = sysstats.process_io_counters(self.process)
+                disk_end = sysstats.disk_io_counters()
                 # we have process-based disk counters, no need to worry how many nodes are on this host
                 if process_end:
-                    read_bytes = process_end.read_bytes - io_bytes["read_bytes"]
-                    write_bytes = process_end.write_bytes - io_bytes["write_bytes"]
+                    read_bytes = process_end.read_bytes - io_stats["read_bytes"]
+                    write_bytes = process_end.write_bytes - io_stats["write_bytes"]
                 elif disk_end:
                     if self.node_count_on_host > 1:
                         self.logger.info("There are [%d] nodes on this host and Rally fell back to disk I/O counters. Attributing [1/%d] "
                                          "of total I/O to [%s].", self.node_count_on_host, self.node_count_on_host, self.node_name)
 
-                    read_bytes = (disk_end.read_bytes - io_bytes['read_bytes']) // self.node_count_on_host
-                    write_bytes = (disk_end.write_bytes - io_bytes['write_bytes']) // self.node_count_on_host
+                    read_bytes = (disk_end.read_bytes - io_stats['read_bytes']) // self.node_count_on_host
+                    write_bytes = (disk_end.write_bytes - io_stats['write_bytes']) // self.node_count_on_host
                 else:
                     raise RuntimeError("Neither process nor disk I/O counters are available")
 
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -771,10 +771,11 @@ class MergeParts(InternalTelemetryDevice):
 
 
 class DiskIo(InternalTelemetryDevice):
+    human_name = "Disk IO"
     """
     Gathers disk I/O stats.
     """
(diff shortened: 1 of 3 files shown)


ebadyano:
Re-opening https://github.com/elastic/rally/pull/721 here as the repository state got messed up.

tests/mechanic/telemetry_test.py
```diff
@@ -2149,22 +2148,25 @@ class DiskIoTests(TestCase):
...
-        device = telemetry.DiskIo(metrics_store, 1, tmp_dir, "rally0")
+        device = telemetry.DiskIo(metrics_store, node_count_on_host=1, log_root=tmp_dir, node_name="rally0")
         t = telemetry.Telemetry(enabled_devices=[], devices=[device])
         node = cluster.Node(pid=None, host_name="localhost", node_name="rally0", telemetry=t)
         t.attach_to_node(node)
         t.on_benchmark_start()
-        t.on_benchmark_stop()
         t.detach_from_node(node, running=True)
         t.detach_from_node(node, running=False)
+        node2 = cluster.Node(pid=None, host_name="localhost", node_name="rally0", telemetry=t)
```
danielmitterdorfer: What I meant in https://github.com/elastic/rally/pull/721#discussion_r303944880 is that we recreate the telemetry device, not the node. This ensures that we don't rely on any state that is tied to that instance.

tests/mechanic/telemetry_test.py
```diff
@@ -2177,25 +2179,28 @@ def test_diskio_process_io_counters(self, metrics_store_node_count, process_io_c
...
+        t.on_benchmark_stop()
+        t.detach_from_node(node2, running=True)
+        t.detach_from_node(node2, running=False)
 
+        # expected result is 1 byte because there are two nodes on the machine. Result is calculated with total_bytes / node_count
```
danielmitterdorfer: Can you please wrap that line at 120 characters?

tests/mechanic/telemetry_test.py
```diff
@@ -2177,25 +2179,28 @@ def test_diskio_process_io_counters(self, metrics_store_node_count, process_io_c
...
-        device = telemetry.DiskIo(metrics_store, 2, tmp_dir, "rally0")
+        device = telemetry.DiskIo(metrics_store, node_count_on_host=2, log_root=tmp_dir, node_name="rally0")
         t = telemetry.Telemetry(enabled_devices=[], devices=[device])
         node = cluster.Node(pid=None, host_name="localhost", node_name="rally0", telemetry=t)
         t.attach_to_node(node)
         t.on_benchmark_start()
-        t.on_benchmark_stop()
         t.detach_from_node(node, running=True)
         t.detach_from_node(node, running=False)
+        node2 = cluster.Node(pid=None, host_name="localhost", node_name="rally0", telemetry=t)
```
danielmitterdorfer: Similarly to the test above we should create a new instance of `DiskIo` instead of the node.

esrally/mechanic/telemetry.py
```diff
@@ -774,14 +775,16 @@ class DiskIo(InternalTelemetryDevice):
     """
     Gathers disk I/O stats.
     """
-    def __init__(self, metrics_store, node_count_on_host):
+    def __init__(self, metrics_store, node_count_on_host, log_root, node_name):
```
danielmitterdorfer: The `DockerLauncher` calls this constructor as well. Can you please update the constructor invocation accordingly and double-check you have adapted all call-sites?

esrally/mechanic/telemetry.py
```diff
@@ -774,14 +775,16 @@ class DiskIo(InternalTelemetryDevice):
...
         super().__init__()
         self.metrics_store = metrics_store
         self.node_count_on_host = node_count_on_host
         self.node = None
         self.process = None
         self.disk_start = None
         self.process_start = None
+        self.node_name = node_name
```
danielmitterdorfer: Nit: Can we we match the order in the constructor's argument list here?

esrally/mechanic/telemetry.py
```diff
@@ -790,40 +793,55 @@ def attach_to_node(self, node):
     def on_benchmark_start(self):
         if self.process is not None:
             self.process_start = sysstats.process_io_counters(self.process)
+            read_bytes = 0
+            write_bytes = 0
+            io.ensure_dir(self.log_root)
+            tmp_io_file = os.path.join(self.log_root, "%s.io" % self.node_name)
```
danielmitterdorfer: Nit: As per convention we prefer the `format` method to `%` formatting in code that is not performance-critical, i.e. you'd need to change this to `"{}.io".format(self.node_name))` (similarly in `on_benchmark_stop`)

esrally/mechanic/telemetry.py
```diff
@@ -790,40 +793,55 @@ def attach_to_node(self, node):
...
                     self.logger.exception("Could not determine I/O stats at benchmark start.")
+            with open(tmp_io_file, "wt", encoding="utf-8") as f:
+                json.dump({"pid": self.node.pid, "read_bytes": read_bytes, "write_bytes": write_bytes}, f)
 
     def on_benchmark_stop(self):
-        if self.process is not None:
+            io.ensure_dir(self.log_root)
+            tmp_io_file = os.path.join(self.log_root, "%s.io" % self.node_name)
+            with open(tmp_io_file, "rt", encoding="utf-8") as f:
+                io_bytes = json.load(f)
```
danielmitterdorfer: As you now also store the PID, the name `io_bytes` does not seem appropriate anymore? How about `io_stats`?

esrally/mechanic/telemetry.py
ebadyano: As per offline discussion, I fixed the constructor call. But uncovered that `attach_to_node` isn't called for telemetry devices in DockerLauncher and so DIskIo isn't working properly: opened https://github.com/elastic/rally/issues/733 to fix it separately.

danielmitterdorfer: Thank you.

tests/mechanic/telemetry_test.py
```diff
@@ -2140,6 +2142,71 @@ def test_resilient_if_error_response(self):
...
+        device = telemetry.DiskIo(metrics_store, node_count_on_host=2, log_root=tmp_dir, node_name="rally0")
+        t = telemetry.Telemetry(enabled_devices=[], devices=[device])
+        node = cluster.Node(pid=None, host_name="localhost", node_name="rally0", telemetry=t)
+        t.attach_to_node(node)
+        t.on_benchmark_start()
+        device2 = telemetry.DiskIo(metrics_store, node_count_on_host=2, log_root=tmp_dir, node_name="rally0")
+        t2 = telemetry.Telemetry(enabled_devices=[], devices=[device2])
+        t2.on_benchmark_stop()
+        t.detach_from_node(node, running=True)
+        t.detach_from_node(node, running=False)
```
danielmitterdorfer: Both `detach_from_node` calls should be on `t2`, not on `t`?

tests/mechanic/telemetry_test.py
```diff
@@ -2140,6 +2142,71 @@ def test_resilient_if_error_response(self):
...
+        device = telemetry.DiskIo(metrics_store, node_count_on_host=1, log_root=tmp_dir, node_name="rally0")
+        t = telemetry.Telemetry(enabled_devices=[], devices=[device])
+        node = cluster.Node(pid=None, host_name="localhost", node_name="rally0", telemetry=t)
+        t.attach_to_node(node)
+        t.on_benchmark_start()
+        device2 = telemetry.DiskIo(metrics_store, node_count_on_host=1, log_root=tmp_dir, node_name="rally0")
+        t2 = telemetry.Telemetry(enabled_devices=[], devices=[device2])
+        t2.on_benchmark_stop()
+        t.detach_from_node(node, running=True)
+        t.detach_from_node(node, running=False)
```
danielmitterdorfer: Both `detach_from_node` calls should be on `t2`, not on `t`?

tests/mechanic/telemetry_test.py
ebadyano: since we call `t.on_benchmark_start()` only on `t` does it make sense to call detach only on `t`?

danielmitterdorfer: No, because you need to consider that we test here the behavior with #722:

* In the first invocation (the new `start` subcommand), the telemetry device will be instantiated and we attach the telemetry device. After that the process *terminates* and `t` will be gone.
* In the second invocation (the new `stop` subcommand), a new instance of the telemetry device will be created and we detach it from the node.

Hence, we need to call the start-related methods on one instance and the stop-related ones on the other to simulate this behavior and to test that we do not rely on any state of the previous instance (`t` in that case).

Makes sense?

ebadyano: Yes, thank you for explaining. I'll add the fix.

PR #735, opened 2019-07-22
Assume string type for params by default

danielmitterdorfer:
With this commit we allow users to skip enclosing quotes for params
(like `track-params` or `team-params`) when the value is clearly a
string (i.e. it cannot be coerced to a number or a boolean value). It is
still possible to surround a parameter with single-quotes to enforce
treatment as a string.


danielmitterdorfer:
Thanks for the review!

PR #737, opened 2019-07-31
Allow to attach telemetry devices without reprovisioning

ebadyano:
By using ES_JAVA_OPTS we can provision a node, run a benchmark, and then “dynamically” (i.e. without reprovisioning) start the node again with telemetry attached.

Relates to #697
Relates to #711

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -348,7 +348,7 @@ class ProcessLauncher:
         if v is not None:
             if k not in env:
                 env[k] = v
-            elif (prepend == True):
+            elif prepend:
                     env[k] = v + separator + env[k]
             else:
                     env[k] = env[k] + separator + v
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -332,7 +332,7 @@ class ProcessLauncher:
         env = {}
         env.update(os.environ)
         env.update(car.env)
-        self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep)
+        self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep, prepend=True)
         # Don't merge here!
         env["JAVA_HOME"] = java_home
         env["ES_JAVA_OPTS"] = "-XX:+ExitOnOutOfMemoryError"
@@ -344,12 +344,15 @@ class ProcessLauncher:
         self.logger.debug("env for [%s]: %s", node_name, str(env))
         return env
 
-    def _set_env(self, env, k, v, separator=' '):
+    def _set_env(self, env, k, v, separator=' ', prepend=False):
         if v is not None:
             if k not in env:
                 env[k] = v
             else:  # merge
-                env[k] = v + separator + env[k]
+                if (prepend == True):
+                    env[k] = v + separator + env[k]
+                else:
+                    env[k] = env[k] + separator + v
 
     @staticmethod
     def _start_process(binary_path, env):
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -348,10 +348,9 @@ class ProcessLauncher:
         if v is not None:
             if k not in env:
                 env[k] = v
-            else:  # merge
-                if (prepend == True):
+            elif (prepend == True):
                     env[k] = v + separator + env[k]
-                else:
+            else:
                     env[k] = env[k] + separator + v
 
     @staticmethod
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -299,10 +299,7 @@ class ProcessLauncher:
         car = node_configuration.car
         binary_path = node_configuration.binary_path
         data_paths = node_configuration.data_paths
-        #node_telemetry_dir = "%s/telemetry" % node_configuration.node_root_path
         node_telemetry_dir = os.path.join(node_configuration.node_root_path, "telemetry")
-
-        print("here in oy %s" % node_telemetry_dir)
 
         java_major_version, java_home = java_resolver.java_home(car, self.cfg)
 
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -349,7 +349,7 @@ class ProcessLauncher:
             if k not in env:
                 env[k] = v
             else:  # merge
-                env[k] = env[k] + separator + v
+                env[k] = v + separator + env[k]
 
     @staticmethod
     def _start_process(binary_path, env):
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -299,7 +299,11 @@ class ProcessLauncher:
         car = node_configuration.car
         binary_path = node_configuration.binary_path
         data_paths = node_configuration.data_paths
-        node_telemetry_dir = "%s/telemetry" % node_configuration.node_root_path
+        #node_telemetry_dir = "%s/telemetry" % node_configuration.node_root_path
+        node_telemetry_dir = os.path.join(node_configuration.node_root_path, "telemetry")
+
+        print("here in oy %s" % node_telemetry_dir)
+
         java_major_version, java_home = java_resolver.java_home(car, self.cfg)
 
         self.logger.info("Starting node [%s] based on car [%s].", node_name, car)
@@ -307,6 +311,9 @@ class ProcessLauncher:
         enabled_devices = self.cfg.opts("mechanic", "telemetry.devices")
         telemetry_params = self.cfg.opts("mechanic", "telemetry.params")
         node_telemetry = [
+            telemetry.FlightRecorder(telemetry_params, node_telemetry_dir, java_major_version),
+            telemetry.JitCompiler(node_telemetry_dir),
+            telemetry.Gc(node_telemetry_dir, java_major_version),
             telemetry.DiskIo(self.metrics_store, node_count_on_host, node_telemetry_dir, node_name),
             telemetry.NodeEnvironmentInfo(self.metrics_store),
             telemetry.IndexSize(data_paths, self.metrics_store),
@@ -331,6 +338,11 @@ class ProcessLauncher:
         self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep)
         # Don't merge here!
         env["JAVA_HOME"] = java_home
+        env["ES_JAVA_OPTS"] = "-XX:+ExitOnOutOfMemoryError"
+        
+        # we just blindly trust telemetry here...
+        for v in t.instrument_candidate_java_opts(car, node_name):
+            self._set_env(env, "ES_JAVA_OPTS", v)
 
         self.logger.debug("env for [%s]: %s", node_name, str(env))
         return env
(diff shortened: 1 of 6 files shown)


esrally/mechanic/launcher.py
```diff
@@ -340,7 +352,7 @@ def _set_env(self, env, k, v, separator=' '):
             if k not in env:
                 env[k] = v
             else:  # merge
-                env[k] = v + separator + env[k]
+                env[k] = env[k] + separator + v
```
danielmitterdorfer: Can you please explain why you've changed the order here?

tests/mechanic/launcher_test.py
```diff
@@ -119,6 +119,7 @@ def wait(self):
 
 
 class ProcessLauncherTests(TestCase):
+    @mock.patch('os.path.join', return_value="/telemetry")
```
danielmitterdorfer: Why do we need to mock this? IMHO we shouldn't have mocked quite a few of the dependencies of this test case to begin with and specifically not `esrally.mechanic.provisioner.NodeConfiguration`. 

I think we need to clean this up in a separate PR. @drawlerr can you please work on a new PR (based on master) and rewrite this test case to mock only dependencies that have side effects? (as a guideline: things in `os` are likely ok, things like `esrally.config.Config` or `esrally.metrics.MetricsStore` should not be mocked).

esrally/mechanic/launcher.py
```diff
@@ -299,14 +299,21 @@ def _start_node(self, node_configuration, node_count_on_host):
         car = node_configuration.car
         binary_path = node_configuration.binary_path
         data_paths = node_configuration.data_paths
-        node_telemetry_dir = "%s/telemetry" % node_configuration.node_root_path
+        #node_telemetry_dir = "%s/telemetry" % node_configuration.node_root_path
+        node_telemetry_dir = os.path.join(node_configuration.node_root_path, "telemetry")
+
+        print("here in oy %s" % node_telemetry_dir)
```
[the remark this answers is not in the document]
ebadyano: oh, thanks for catching

esrally/mechanic/launcher.py
ebadyano: For some reason JFR telemetry would not start and rally would never return with pid if the the options didn't start with "-XX:+UnlockDiagnosticVMOptions" That's how it was originally, but now that we changed the telemetry to return list instead of string the order got reversed when we were adding them to the env[k]. So I reversed again so it's in the original order. I could change the order in telemetry instead if you prefer.

tests/mechanic/launcher_test.py
ebadyano: I modified the launcher to use os.path instead of concatenating strings. If I don't update the test I get the following error. Not sure how to get away without mocking it..
```
________________ ProcessLauncherTests.test_daemon_start_stop __________________

args = (<tests.mechanic.launcher_test.ProcessLauncherTests testMethod=test_daemon_start_stop>, <MagicMock name='Process' id='...ion' id='4573833480'>, <MagicMock name='MetricsStore' id='4573862824'>, <MagicMock name='Config' id='4577452888'>, ...)
keywargs = {}
extra_args = [<MagicMock name='Process' id='4574868424'>, <MagicMock name='wait_for_pidfile' id='4578813376'>, <MagicMock name='Nod...tricsStore' id='4573862824'>, <MagicMock name='Config' id='4577452888'>, <MagicMock name='chdir' id='4578781056'>, ...]
entered_patchers = [<unittest.mock._patch object at 0x10d987438>, <unittest.mock._patch object at 0x10d987ef0>, <unittest.mock._patch obj...bject at 0x10d987a58>, <unittest.mock._patch object at 0x10d987320>, <unittest.mock._patch object at 0x10d987ac8>, ...]
exc_info = (<class 'TypeError'>, TypeError("test_daemon_start_stop() missing 1 required positional argument: 'path'"), <traceback object at 0x110efe5c8>)
patching = <unittest.mock._patch object at 0x10d987438>
arg = <MagicMock name='kill' id='4578728480'>

    @wraps(func)
    def patched(*args, **keywargs):
        extra_args = []
        entered_patchers = []

        exc_info = tuple()
        try:
            for patching in patched.patchings:
                arg = patching.__enter__()
                entered_patchers.append(patching)
                if patching.attribute_name is not None:
                    keywargs.update(arg)
                elif patching.new is DEFAULT:
                    extra_args.append(arg)

            args += tuple(extra_args)
>           return func(*args, **keywargs)
E           TypeError: test_daemon_start_stop() missing 1 required positional argument: 'path'

/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/unittest/mock.py:1191: TypeError
```

esrally/mechanic/launcher.py
danielmitterdorfer: I see. That's very subtle but it makes sense that we preserve the order in the merge routine (i.e. as you've implemented it now). Can you please add a test so we ensure that `-XX:+UnlockDiagnosticVMOptions` is the first option when we activate multiple telemetry devices (you can just use the full list of JVM args in the assertion)?

tests/mechanic/launcher_test.py
danielmitterdorfer: You also need to remove `path` from the argument list of the test. Otherwise the test framework cannot call `test_daemon_start_stop`.

ebadyano: right, sorry if if remove both mock and path in the arguments I get this:
```
    @mock.patch('os.kill')
    @mock.patch('subprocess.Popen',new=MockPopen)
    @mock.patch('esrally.mechanic.java_resolver.java_home', return_value=(12, "/java_home/"))
    @mock.patch('esrally.utils.jvm.supports_option', return_value=True)
    @mock.patch('os.chdir')
    @mock.patch('esrally.config.Config')
    @mock.patch('esrally.metrics.MetricsStore')
    @mock.patch('esrally.mechanic.provisioner.NodeConfiguration')
    @mock.patch('esrally.mechanic.launcher.wait_for_pidfile', return_value=MOCK_PID_VALUE)
    @mock.patch('psutil.Process')
    def test_daemon_start_stop(self, process, wait_for_pidfile, node_config, ms, cfg, chdir, supports, java_home, kill):
        proc_launcher = launcher.ProcessLauncher(cfg, ms, paths.races_root(cfg))

>       nodes = proc_launcher.start([node_config])

tests/mechanic/launcher_test.py:135:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
esrally/mechanic/launcher.py:294: in start
    return [self._start_node(node_configuration, node_count_on_host) for node_configuration in node_configurations]
esrally/mechanic/launcher.py:294: in <listcomp>
    return [self._start_node(node_configuration, node_count_on_host) for node_configuration in node_configurations]
esrally/mechanic/launcher.py:302: in _start_node
    node_telemetry_dir = os.path.join(node_configuration.node_root_path, "telemetry")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

a = <MagicMock name='NodeConfiguration.node_root_path' id='4598523160'>
p = ('telemetry',)

    def join(a, *p):
        """Join two or more pathname components, inserting '/' as needed.
        If any component is an absolute path, all previous path components
        will be discarded.  An empty last part will result in a path that
        ends with a separator."""
>       a = os.fspath(a)
E       TypeError: expected str, bytes or os.PathLike object, not MagicMock

/usr/local/Cellar/python/3.7.1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/posixpath.py:80: TypeError
```

esrally/mechanic/launcher.py
ebadyano: Did a bit more testing, so it seems that order of options for other telemetry devices doesn't matter. It's that specifically for jfr that `-XX:+UnlockDiagnosticVMOptions` has to come before  `-XX:+DebugNonSafepoints` but other telemetry can be before or after jfr options. So maybe modifying order in telemery for jfr makes more sense than in `_set_env`.. WDYT?

danielmitterdorfer: Yes, your proposal makes sense to me.

tests/mechanic/launcher_test.py
danielmitterdorfer: As discussed in another channel this is the result of mocking other dependencies so it is ok if you keep the mock on `os.path.join` here as this will be cleaned up in a dedicated PR.

esrally/mechanic/launcher.py
```diff
@@ -352,7 +349,7 @@ def _set_env(self, env, k, v, separator=' '):
             if k not in env:
                 env[k] = v
             else:  # merge
-                env[k] = env[k] + separator + v
+                env[k] = v + separator + env[k]
```
danielmitterdorfer: I tested this with

```
esrally --test-mode --distribution-version=2.4.6 --on-error=abort --challenge=append-no-conflicts-index-only --telemetry=jfr,gc
```

and indeed you have to preserve the order of the JVM arguments otherwise you'll get an error that the JVM process could not be created. Can you please:

* Change it to your previous implementation so that the order is preserved
* Add a test in ProcessLauncher that adds a telemetry device and checks `ES_JAVA_OPTS`?

esrally/mechanic/telemetry.py
```diff
@@ -191,7 +191,7 @@ def instrument_java_opts(self, car, candidate_id):
 
     def java_opts(self, log_file):
         recording_template = self.telemetry_params.get("recording-template")
-        java_opts = ["-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints"]
+        java_opts = ["-XX:+DebugNonSafepoints", "-XX:+UnlockDiagnosticVMOptions"]
```
danielmitterdorfer: If the order is preserved (see my comment above) also this needs to be reversed again (`-XX:+UnlockDiagnosticVMOptions` must be specified before `-XX:+DebugNonSafepoints`).

tests/mechanic/launcher_test.py
```diff
@@ -139,6 +140,25 @@ def test_daemon_start_stop(self, process, wait_for_pidfile, node_config, ms, cfg
         proc_launcher.keep_running = False
         proc_launcher.stop(nodes)
         self.assertTrue(kill.called)
+             
+    @mock.patch('os.path.join', return_value="/java_home/bin")
```
danielmitterdorfer: This should not be necessary and this test should work without mocking `os.path.join`.

ebadyano:
@danielmitterdorfer Thank you for the review! I addressed your latest comments.

esrally/mechanic/launcher.py
```diff
@@ -348,10 +348,9 @@ def _set_env(self, env, k, v, separator=' ', prepend=False):
         if v is not None:
             if k not in env:
                 env[k] = v
-            else:  # merge
-                if (prepend == True):
+            elif (prepend == True):
```
danielmitterdorfer: While this will work correctly, idiomatic usage is `elif prepend` instead of `elif (prepend == True)`. Can you please change that?

PR #739, opened 2019-08-08
Remove lap feature and all references.

drawlerr:
This feature complicates development and can be done with user tags, so is unnecessary.


esrally/reporter.py
```diff
@@ -79,7 +79,7 @@ def print_header(message):
     print_internal(console.format.bold(message))
 
 
-def write_single_report(report_file, report_format, cwd, headers, data_plain, data_rich, write_header=True):
+def write_single_report(report_file, report_format, cwd, headers, data_plain, data_rich):
```
danielmitterdorfer: There is another callsite in `ComparisonReporter` that still passes this parameter.

docs/migrate.rst
```diff
@@ -10,18 +10,15 @@ This means that on disk, a given race will be found at ``benchmarks/races/62d1e9
 
 Laps feature removed
 ^^^^^^^^^^^^^^^^^^^^
-The --laps parameter and corresponding multi-run trial functionality has been removed from execution and reporting.
-If you need lap functionality, the following shell script can be used instead:
+The ```--laps``` parameter and corresponding multi-run trial functionality has been removed from execution and reporting.
```
danielmitterdorfer: This needs to be enclosed in double-backticks, not triple-backticks.

PR #743, opened 2019-08-14
Change MergeParts to use detach_from_node.

drawlerr:


drawlerr:
I've merged the ProcessLaunchTests fixes  (#746) for the test failure into this PR and #744.

danielmitterdorfer:
Thanks for the new PR (#746). However, to avoid confusion we should not merge unmerged PRs into other PRs. First of all, it makes it more complicated for the reviewer and secondly we might need to iterate on the other PR until we can get it in. Can you thus please revert merging #746 here?

danielmitterdorfer:
I had a look at the diff and it seems to be quite off, e.g. I see changes from #739 here. Can you please double-check your commits here and cleanup the history?

dliappis:
@drawlerr `test_daemon_start_stop` fails for me as well. See:

https://gist.github.com/dliappis/bb3bd1fa20046e2176539a35d1853f19

danielmitterdorfer:
> `test_daemon_start_stop` fails for me as well.

Given that we've recently raised https://github.com/elastic/rally/issues/754 should we close this PR instead?

drawlerr:
As per #754, no point adding minor changes to a soon-to-be-removed device.

PR #744, opened 2019-08-14
Change DiskIO callbacks to use attach/detach

drawlerr:


drawlerr:
I've merged the ProcessLaunchTests fixes  (#746) for the test failure into this PR and #744.

danielmitterdorfer:
Thanks for the new PR (#746). However, to avoid confusion we should not merge unmerged PRs into other PRs. First of all, it makes it more complicated for the reviewer and secondly we might need to iterate on the other PR until we can get it in. Can you thus please revert merging #746 here?

esrally/mechanic/telemetry.py
```diff
@@ -789,8 +789,6 @@ def __init__(self, metrics_store, node_count_on_host, log_root, node_name):
     def attach_to_node(self, node):
         self.node = node
         self.process = sysstats.setup_process_stats(node.pid)
-
-    def on_benchmark_start(self):
         if self.process is not None:
```
danielmitterdorfer: You can get rid of this check as it is impossible that `self.process` is `None` at this point.

danielmitterdorfer:
Now that #746 is closed, can you please also merge master into your branch? (I did that locally and I noticed that there are small changes that you need to address afterwards).

PR #746, opened 2019-08-16
Fixup ProcessLauncherTests issues

drawlerr:
The start_stop test in ProcessLauncherTests uses a MagicMock object for Process, which is incompatible with telemetry operations.
- Replace Process MagicMock with a minimal hand-mocked class.  
 - Also, remove Config + MetricsStore mocks while we're at it.

Relates to PRs #743, #744

### DIFF
--- a/tests/mechanic/launcher_test.py
+++ b/tests/mechanic/launcher_test.py
@@ -14,10 +14,13 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import tempfile
 from datetime import datetime
 import io
 import uuid
 from unittest import TestCase, mock
+
+import psutil
 
 from esrally import config, exceptions, paths
 from esrally.mechanic import launcher
@@ -115,6 +118,21 @@ class MockPopen:
         return 0
 
 
+class MockProcess:
+    def __init__(self, pid):
+        self.pid = pid
+        self.killed = False
+
+    def name(self):
+        return "p{pid}".format(pid=self.pid)
+
+    def wait(self, timeout=None):
+        raise psutil.TimeoutExpired(timeout)
+
+    def kill(self):
+        self.killed = True
+
+
 def get_metrics_store(cfg):
     cfg.add(config.Scope.application, "track", "params", None)
     cfg.add(config.Scope.application, "system", "env.name", None)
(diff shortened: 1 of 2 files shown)


tests/mechanic/launcher_test.py
```diff
@@ -115,23 +118,60 @@ def wait(self):
...
-    @mock.patch('esrally.metrics.MetricsStore')
-    @mock.patch('esrally.mechanic.provisioner.NodeConfiguration')
     @mock.patch('esrally.mechanic.launcher.wait_for_pidfile', return_value=MOCK_PID_VALUE)
-    @mock.patch('psutil.Process')
-    def test_daemon_start_stop(self, process, wait_for_pidfile, node_config, ms, cfg, chdir, supports, java_home, kill):
+    @mock.patch('psutil.Process', new=MockProcess)
+    def test_daemon_start_stop(self, wait_for_pidfile, chdir, get_size, supports, java_home, kill):
+        cfg = config.Config()
+        cfg.add(config.Scope.application, "node", "root.dir", "test")
+        cfg.add(config.Scope.application, "mechanic", "keep.running", False)
```
danielmitterdorfer: I think we can then remove this in line 179: `proc_launcher.keep_running = False` (I cannot comment on the line that needs to be removed thus I do it here)

tests/mechanic/launcher_test.py
```diff
@@ -115,23 +118,60 @@ def wait(self):
...
+        cfg.add(config.Scope.application, "mechanic", "telemetry.devices", [])
+        cfg.add(config.Scope.application, "mechanic", "telemetry.params", None)
+
+        ms = get_metrics_store(cfg)
         proc_launcher = launcher.ProcessLauncher(cfg, ms, paths.races_root(cfg))
 
+        tmpdir = tempfile.mkdtemp()
```
danielmitterdorfer: Is this really needed? As far as I can see all file system operations are mocked so we can just provide a fake path as string here?

tests/mechanic/launcher_test.py
```diff
@@ -115,23 +118,60 @@ def wait(self):
...
+    def wait(self, timeout=None):
+        raise psutil.TimeoutExpired(timeout)
+
+    def kill(self):
+        self.killed = True
+
+
+def get_metrics_store(cfg):
+    cfg.add(config.Scope.application, "track", "params", None)
+    cfg.add(config.Scope.application, "system", "env.name", None)
```
danielmitterdorfer: IMHO such a function should be free of (globally visible) side effects and we should not modify `cfg` here. I'd be ok if we make track params optional  (with a default value of `{}`) in the constructor of `MetricsStore` but I think all call sites must provide `env.name` when calling this function.

drawlerr: Normally, I agree completely!  I compromised here for convenience to reduce the number of changed lines necessary when changing from MockMetricsStore to a real one.  I like your suggestions though, so I'll incorporate them.

PR #747, opened 2019-08-20
Log git output

### LINKED ISSUE #745: git interferes with Rally stdout even with --quiet
For example when using Rally installed [for development](https://esrally.readthedocs.io/en/stable/developing.html#installation-instructions-for-development) and it's lagging behind the latest commit, trying:

`esrally download --quiet --distribution-version=7.3.0`

breaks the json output by printing:

`b\'Fast-forwarded 7 to origin/7.\\n{\\n  "elasticsearch": "/Users/dl/.rally/benchmarks/distributions/elasticsearch-7.2.0-linux-x86_64.tar.gz"\\n}\\n\'`

danielmitterdorfer:
With this commit we send git output to logs (except when cloning a repo
initially as it might require the user to enter credentials). This
avoids that git output interferes with Rally's. This is needed e.g. when
Rally's `download` subcommand is invoked where we usually expect only a
JSON fragment as output.

Closes #745

### DIFF
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -55,23 +55,20 @@ def clone(src, remote):
 
 @probed
 def fetch(src, remote="origin"):
-    # Don't swallow output but silence git at least a bit... (--quiet)
-    if process.run_subprocess(
-            "git -C {0} fetch --prune --tags --quiet {1}".format(src, remote)):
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Thanks for the review to both of you! :)

Issue #740, opened 2019-08-12
track-revision is not allowed together with track-repository


dliappis:
#### Add rally-eventdata-track to rally.ini:

```
...
[tracks]
default.url = https://github.com/elastic/rally-tracks
eventdata.url = https://github.com/elastic/rally-eventdata-track
...
```

#### Execute:

```
esrally --track=eventdata --track-repository=eventdata --track-revision=ede1773912b47594af758f1d41865793db70c8b5 --challenge=bulk-update --distribution-version=6.8.2
```

### Expected results
Race should run

### Actual result
```
esrally: error: argument --track-revision: not allowed with argument --track-repository
```

PR #751, opened 2019-08-23
Option track-revision should work with track-repository

ebadyano:
Both options track-repository and track-revision should work together but not with track-path option.

Fixes: https://github.com/elastic/rally/issues/740

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -191,7 +191,7 @@ def create_arg_parser():
         track_source_group.add_argument(
             "--track-path",
             help="Define the path to a track.")
-        track_source_group.add_argument(
+        p.add_argument(
             "--track-revision",
             help="Define a specific revision in the track repository that Rally should use.",
             default=None)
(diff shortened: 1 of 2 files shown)


esrally/rally.py
```diff
@@ -191,7 +191,7 @@ def positive_number(v):
         track_source_group.add_argument(
             "--track-path",
             help="Define the path to a track.")
-        track_source_group.add_argument(
+        p.add_argument(
```
dliappis: Could we add an integration test?

We could use the [list command](https://github.com/elastic/rally/blob/966891a38397227300d55ac32581e4b1d9d53d85/integration-test.sh#L219-L233): https://github.com/elastic/rally/blob/966891a38397227300d55ac32581e4b1d9d53d85/integration-test.sh#L219-L233

and, in order to avoid having to add the `eventdata` repo in the rally.ini used by the integration script (which is randomized btw), we could just rely on the default repository with something like:

```
   info "test list can use track revision together with track repository"
    esrally list tracks --configuration-name="${cfg}" --track-repository=default --track-revision=4080dc9850d07e23b6fc7cfcdc7cf57b14e5168d
}
```

PR #758, opened 2019-08-30
Improve error message on SSL errors

### LINKED ISSUE #757: Provide more clues to what's causing the error SSL: WRONG_VERSION_NUMBER
**Rally version** (get with `esrally --version`): 1.2.1

**Invoked command**: 

`esrally --offline --target-hosts=https://<my_ip>:9200 --pipeline=benchmark-only --client-options="timeout:120,use_ssl:true,verify_certs --report-file=test-1.csv --track-path /mnt/rally/benchmarks/tracks/default/http_logs/`

**OS version**:

```
cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=19.04
DISTRIB_CODENAME=disco
DISTRIB_DESCRIPTION="Ubuntu 19.04"
```

When using SSL options against a non-ssl endpoint, the error message is a bit cryptic as to what has failed.

```
[ERROR] Cannot race. Error in race control (ConnectionError([SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1056)) caused by: SSLError([SSL: WRONG_VERSION_NUMBER] wrong version number (_ssl.c:1056)))
```

**Steps to reproduce**:

I saw this issue using `--client-options="timeout:120,use_ssl:true,verify_certs:false` against an endpoint with  `--target-hosts=https://whateverhost` 

Setting `use_ssl` to false and leaving verify_certs:false still saw the error occurring.
I had to remove the https and use http instead to have it correctly connect to the endpoint.

**Provide logs (if relevant)**:

**Describe the feature**:

A list of errors with possible causes/resolutions in a troubleshooting guide within the Rally docs is probably enough to steer users to the fix as I don't think this would be all that common.

Additionally, a more informative error message might give users the right hint faster (`"is this a https endpoint?"`)

danielmitterdorfer:
With this commit we check for SSL-related errors when waiting for the
REST layer to be available and provide a better error message when the
user attempts to connect via SSL to a http endpoint.

Closes #757

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -25,7 +25,7 @@ from time import monotonic as _time
 
 from esrally import config, time, exceptions, client
 from esrally.mechanic import telemetry, cluster, java_resolver
-from esrally.utils import process, jvm
+from esrally.utils import process
 
 
 def wait_for_rest_layer(es, max_attempts=20):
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Thanks for the review! :)

Issue #665, opened 2019-03-07
Show track and team revision when listing races


danielmitterdorfer:
After implementing #664 we will have the necessary meta-data to show them also in the output of `esrally list races`. However, we should think whether we want to expose this information all the time or whether we hide it by default and only expose it if the user specifies an additional command line flag (e.g. `--verbose`).

PR #759, opened 2019-08-30
Show track and team revision when listing races

ebadyano:
Closes https://github.com/elastic/rally/issues/665

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -1244,7 +1244,6 @@ class Race:
             "cluster": self.cluster.as_dict(),
             "results": self.results.as_dict()
         }
-
         if self.track_revision:
             d["track-revision"] = self.track_revision
         if not self.challenge.auto_generated:
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -1314,7 +1314,9 @@ class Race:
         else:
             user_tags = {}
 
-        team_revision = d.get("cluster").get("team-revision")
+        team_revision = None
+        if d.get("cluster"):
+            team_revision = d.get("cluster").get("team-revision")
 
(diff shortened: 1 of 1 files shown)


esrally/metrics.py
```diff
@@ -1314,11 +1314,13 @@ def from_dict(cls, d):
         else:
             user_tags = {}
 
-        # Don't restore a few properties like cluster because they (a) cannot be reconstructed easily without knowledge of other modules
+        team_revision = d.get("cluster").get("team-revision")
```
danielmitterdorfer: If there is no `cluster` property this will fail because `#get()` will return `None`.

esrally/metrics.py
```diff
@@ -1160,11 +1160,11 @@ def format_dict(d):
     races = []
     for race in race_store(cfg).list():
         races.append([race.trial_id, time.to_iso8601(race.trial_timestamp), race.track, format_dict(race.track_params), race.challenge_name, race.car_name,
-                      format_dict(race.user_tags)])
+                      format_dict(race.user_tags), race.track_revision, race.cluster.get("team-revision")])
```
danielmitterdorfer: fyi: I thought whether we might want to change the order here and group track-related properties and team-related properties but those two properties are so special that it's probably ok to leave them at the end.

esrally/metrics.py
ebadyano: thank you for  catching this!

esrally/metrics.py
```diff
@@ -1314,7 +1314,9 @@ def from_dict(cls, d):
         else:
             user_tags = {}
 
-        team_revision = d.get("cluster").get("team-revision")
+        team_revision = None
+        if d.get("cluster"):
```
danielmitterdorfer: I think more idiomatic is `if "cluster" in d:`

PR #760, opened 2019-09-02
Improve logging of schedules

danielmitterdorfer:
With this commit we add more detailed info logging to schedules
indicating how long / often a schedule is expected to run and when it
has terminated.


danielmitterdorfer:
Thanks for the review! :)

PR #761, opened 2019-09-03
Allow to retry until success

danielmitterdorfer:
With this commit we add a new parameter `retry-until-success` to all
retryable operations (e.g. cluster health) that will allow to retry
those operations until they report success. Note that this is actually
just a convenience property that will increase the number of retries to
`sys.maxsize`.


danielmitterdorfer:
Thanks for the review!

PR #762, opened 2019-09-03
Check that origin points to main repo for release

### LINKED ISSUE #717: Ensure origin is elastic/rally before release
We should add a check that the origin remote points to elastic/rally before allowing a release, as scripts implicitly assume this is the case.

danielmitterdorfer:
With this commit we add a new pre-release check that ensures that the
git remote named `origin` points to Rally's main repository instead of a
fork. This is necessary for releases because later during the release
process, changes will be pushed automatically to the main repository
(such as the release tags).

Closes #717

### DIFF
--- a/release-checks.sh
+++ b/release-checks.sh
@@ -51,7 +51,7 @@ then
 fi
 
 ORIGIN_URL=$(git remote get-url --push origin)
-if [[ $(echo ${ORIGIN_URL} | grep -c "elastic/rally") == "0" ]]
+if [[ ${ORIGIN_URL} != *"elastic/rally"* ]]
 then
     echo "Error: the git remote [origin] does not point to Rally's main repo at elastic/rally but to [${ORIGIN_URL}]."
     exit 1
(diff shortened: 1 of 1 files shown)


release-checks.sh
```diff
@@ -50,6 +50,13 @@ then
     exit 1
 fi
 
+ORIGIN_URL=$(git remote get-url --push origin)
+if [[ $(echo ${ORIGIN_URL} | grep -c "elastic/rally") == "0" ]]
```
dliappis: How about something simpler here like `if [[ $ORIGIN_URL != *"elastic/rally"* ]]`?

danielmitterdorfer: Thanks for the suggestion. I've pushed this change in https://github.com/elastic/rally/pull/762/commits/ae88a21d00435f61fcaca162436a467966cea4e5.

danielmitterdorfer:
Thank you both for the reviews! :)

Issue #754, opened 2019-08-26
Remove MergeParts internal telemetry device


dliappis:
The [Merge Parts](https://www.google.com/url?q=https://esrally.readthedocs.io/en/stable/summary_report.html?highlight%3Dmerge%2520parts%23merge-time-x&sa=D&ust=1566811434528000&usg=AFQjCNFXi9UUPNNXNUS9YQ0OZB2iqpaqjg) telemetry device -- which relies on a specific logging configuration via [rally-teams](https://github.com/elastic/rally-teams/blob/c91cf7085edbdb812b0abf9094c53598bd4990a5/cars/v1/vanilla/templates/config/log4j2.properties#L64-L72) allowing Lucene index writer trace logging -- is present for historical reasons and can be removed.

This ticket tracks:

- [ ] Removal of [MergeParts](https://github.com/elastic/rally/blob/966891a38397227300d55ac32581e4b1d9d53d85/esrally/mechanic/telemetry.py#L715-L771) telemetry device and all references to it in the code base (e.g. in `reporter.py`, `launcher.py`, `telemetry_test.py` and possibly other places).
- [ ] Remove references in [docs](https://github.com/elastic/rally/blob/master/docs/summary_report.rst#merge-time-x)

dliappis:
**IMPORTANT NOTE**: This ticket needs to be dealt with after https://github.com/elastic/rally/pull/743 has been merged.

PR #764, opened 2019-09-05
Remove MergeParts internal telemetry device

ebadyano:
Remove the Merge Parts telemetry device, which would display different merge times as reported by Lucene only when Lucene index writer trace logging was enabled (`--car-params="verbose_iw_logging_enabled:true"`). The following metric keys will not be gathered anymore: `merge_parts_total_time_*`  where * is one of `postings, stored fields, doc values, norms, vectors, points`

Closes https://github.com/elastic/rally/issues/754

### DIFF
--- a/esrally/mechanic/telemetry.py
+++ b/esrally/mechanic/telemetry.py
@@ -18,7 +18,6 @@
 import collections
 import logging
 import os
-import re
 import signal
 import subprocess
 import tabulate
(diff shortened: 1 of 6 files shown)


esrally/mechanic/telemetry.py
```diff
@@ -712,65 +712,6 @@ def attach_to_node(self, node):
         self.metrics_store.put_value_node_level(node.node_name, "node_startup_time", self.timer.total_time(), "s")
 
 
-class MergeParts(InternalTelemetryDevice):
-    """
-    Gathers merge parts time statistics. Note that you need to run a track setup which logs these data.
-    """
-    MERGE_TIME_LINE = re.compile(r": (\d+) msec to merge ([a-z ]+) \[(\d+) docs\]")
```
danielmitterdorfer: This results in (at least) one new unused import `re` in this module. Can you please check unused imports and remove them?

PR #765, opened 2019-09-10
Start release process creating NOTICE

danielmitterdorfer:
With this commit we move the creation of the NOTICE file to the
beginning of the release process. As this depends on pulling files from
repositories that are not under our control, it can fail. By doing this
as the first step, we can correct any errors before other tasks have
modified state.


danielmitterdorfer:
Thanks for the review! :)

PR #766, opened 2019-09-10
Ensure tox environment is clean

danielmitterdorfer:
With this commit we force-clean the environment used by tox in order to
ensure dependencies are always up-to-date as tox / pip will not upgrade
dependencies automatically.


danielmitterdorfer:
Thanks for the review! :)

PR #767, opened 2019-09-10
Remove merge times from command line report

danielmitterdorfer:
With this commit we remove all merge time related metrics from the
command line report as they are not measured anymore since #764.

Relates #764


danielmitterdorfer:
Thanks for the review! :)

PR #768, opened 2019-09-10
Honor ingest-percentage for bulks

danielmitterdorfer:
With this commit we honor the parameter `ingest-percentage` again for bulk
requests. Before #763, the method `size()` was used to determine upfront when a
parameter source is exhausted. When we changed the API then in #763 the
parameter source's progress indication was instead done with the
`percent_completed` property but this is only informational and the parameter
source is required to raise `StopIteration` to indicate completion. This means
that `ingest-percentage` was not considered anymore for bulks but instead we
always read the input file(s) completely and then terminated with
`StopIteration`.

Relates #763

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -554,9 +554,11 @@ class PartitionBulkIndexParamSource:
         self.ingest_percentage = ingest_percentage
         self.id_conflicts = id_conflicts
         self.pipeline = pipeline
+        # this is only intended for unit-testing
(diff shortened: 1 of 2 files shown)


dliappis:
For the record, this was observed with a reproduction like:

```
./rally --distribution-version=7.3.0 --track=geonames --challenge=append-no-conflicts-index-only --track-params='ingest_percentage:50'
```

where progress for `index-append` would exceed 100%.

danielmitterdorfer:
Thanks for the review!

PR #773, opened 2019-09-13
Store race files always as race.json

danielmitterdorfer:
With this commit we store the race file always as `race.json`.
Previously Rally had a logic built-in to create a unique file name if
there was a potential for duplicate race files (with the same race
timestamp) but as we're using the race id now for the base directory
this is not possible anymore.

Relates #720

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -18,6 +18,7 @@
 import collections
 import logging
 import math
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:
Thanks for the review! :)

PR #774, opened 2019-09-13
Raise an error if race is not found by id

danielmitterdorfer:
Previously Rally was lenient when a user was searching for a race by id
but it was not found. This leads to follow-up errors that are hard to
understand. With this commit we raise an error instead if a race is not
found by its id.


danielmitterdorfer:
Thanks for the review! :)

PR #776, opened 2019-09-27
Add heapdump telemetry device

danielmitterdorfer:
With this commit we add a new telemetry device `heapdump` that allows
users to capture a heapdump right before a node is shutdown. This can be
useful together with `--on-error=abort` where we then want to know the
state of the heap at this point.


esrally/telemetry.py
```diff
@@ -243,6 +244,25 @@ def java_opts(self, log_file):
...
+    help = "Captures a heap dump."
+
+    def __init__(self, log_root):
+        super().__init__()
+        self.log_root = log_root
+
+    def detach_from_node(self, node, running):
+        if running:
+            heap_dump_file = os.path.join(self.log_root, "heap_at_exit_{}.hprof".format(node.pid))
+            console.info("{}: Writing heap dump to [{}}]".format(self.human_name, heap_dump_file), logger=self.logger)
```
dliappis: This will error out as `[{}]]` needs to be `[{}]`. (Tests also fail)

danielmitterdorfer: I've addressed it now in https://github.com/elastic/rally/pull/776/commits/e44cbfeddab4d4a8ad41e074c834851cbb65bac4. Can you please have another look?

Issue #777, opened 2019-09-27
Remove deprecated properties trial-id and trial-timestamp


danielmitterdorfer:
With #778 we deprecate `trial-id` and `trial-timestamp` (which are written to `rally-*` indices in the metrics store) with Rally 1.4.0  in favor of `race-id` and `race-timestamp`. There is a BWC layer that allows users to still continue to use these properties but we should clean this up (see `TODO #777` comments in the code; in addition all mappings need to be cleaned up).

PR #778, opened 2019-09-27
Expose race-id as command line parameter

danielmitterdorfer:
With this commit we expose the race id as command line parameter. As the
internal name "trial id" and the related name "trial timestamp" are
inconsistent with this terminology we rename them to "race id" and "race
timestamp" respectively and add a backwards-compatibility layer that
will be removed again in #777.

### DIFF
--- a/esrally/resources/metrics-template.json
+++ b/esrally/resources/metrics-template.json
@@ -42,6 +42,18 @@
             }
           }
         },
+        "race-id": {
+          "type": "keyword"
+        },
+        "race-timestamp": {
+          "type": "date",
+          "format": "basic_date_time_no_millis",
+          "fields": {
+            "raw": {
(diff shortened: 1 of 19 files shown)


esrally/resources/metrics-template.json
```diff
@@ -42,6 +42,18 @@
             }
           }
         },
+        "race-id": {
```
ebadyano: Just to confirm, are we keeping trial-id  and trial-timestamp sections above for backwards compatibility?

danielmitterdorfer: Correct. This will be removed with https://github.com/elastic/rally/issues/777.

danielmitterdorfer:
Thanks for the review!

PR #781, opened 2019-10-01
Add support for OSNAME and ARCH variables in dist repo URLs.

drawlerr:
Since adding native libraries, ES download URLs now come in platform + arch-specific form.  

On the current branch 7 for config in rally_teams, we hardcode the platform and arch in the URL.  This works fine for the nightly machines but not elsewhere.  So, making these vars template parameters.

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -528,15 +528,14 @@ class DistributionRepository:
         return self._substitute_vars(url_template)
 
     def _substitute_vars(self, s):
-
         substitutions = {
             "{{VERSION}}": self.version,
             "{{OSNAME}}": sysstats.os_name().lower(),
             "{{ARCH}}": sysstats.cpu_arch().lower()
         }
         r = s
-        for k in substitutions.keys():
-            r = r.replace(k, substitutions[k])
+        for key, replacement in substitutions.items():
+            r = r.replace(key, replacement)
         return r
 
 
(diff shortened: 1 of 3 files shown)


esrally/mechanic/supplier.py
```diff
@@ -525,7 +525,20 @@ def _url_for(self, user_defined_key, default_key, mandatory=True):
                 raise exceptions.SystemSetupError("Neither config key [{}] nor [{}] is defined.".format(user_defined_key, default_key))
             else:
                 return None
-        return url_template.replace("{{VERSION}}", self.version)
+        return self._substitute_vars(url_template)
+
+    def _substitute_vars(self, s):
```
danielmitterdorfer: Nice refactoring!

esrally/utils/sysstats.py
```diff
@@ -53,6 +53,16 @@ def cpu_model():
     return "Unknown"
 
 
+def cpu_arch():
```
danielmitterdorfer: How about we implement this as below? This has the advantage that it is part of the Python standard library.

```python
def cpu_arch():
    return platform.uname().machine
```

esrally/mechanic/supplier.py
```diff
@@ -525,7 +525,20 @@ def _url_for(self, user_defined_key, default_key, mandatory=True):
...
+
+        substitutions = {
+            "{{VERSION}}": self.version,
+            "{{OSNAME}}": sysstats.os_name().lower(),
+            "{{ARCH}}": sysstats.cpu_arch().lower()
+        }
+        r = s
+        for k in substitutions.keys():
```
danielmitterdorfer: We could replace this with:

```python
for key, replacement in substitutions.items():
    r = r.replace(key, replacement)
```

This extracts the dictionary key-value pair already in the iterator and we don't need to lookup the value separately.

PR #783, opened 2019-10-02
Upgrade pytest to 5.2.0

danielmitterdorfer:
With this commit we upgrade pytest to 5.2.0 and also upgrade related
dependencies. This version make pytest compatible with recent versions
of its dependency `attrs`.


danielmitterdorfer:
Thanks for the review! :)

PR #784, opened 2019-10-08
Attach telemetry device on Docker launch

danielmitterdorfer:
With this commit we ensure that telemetry devices are attached also when
Elasticsearch is launched via Docker. We also add an integration test to
test the respective pipeline.


danielmitterdorfer:
Thanks for the review!

PR #789, opened 2019-10-10
Let the runner determine progress

danielmitterdorfer:
With this commit we add the ability to have runners determine progress.
Usually the runner is just responsible for executing requests and the
parameter source is determining completion. However, there are cases,
like snapshot/restore or replication where we need to determine progress
based on a request against Elasticsearch.


esrally/driver/runner.py
```diff
@@ -123,54 +123,96 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return False
 
 
-class SingleClusterDelegatingRunner(Runner):
-    def __init__(self, runnable, name, context_manager_enabled=False):
+def _single_cluster_runner(runnable, name, context_manager_enabled=False):
+    # only pass the default ES client
+    delegate = DelegatingRunner(runnable, name, lambda es: es["default"], context_manager_enabled)
```
dliappis: Clever idea!

tests/driver/driver_test.py
```diff
@@ -677,6 +699,55 @@ def test_execute_schedule_in_throughput_mode(self, es):
...
+        test_track = track.Track(name="unittest", description="unittest track",
+                                 indices=None,
+                                 challenges=None)
+
+        task = track.Task("time-based", track.Operation("time-based", operation_type="unit-test-recovery", params={
+            "indices-to-restore": "*",
+            # The runner will determine progress
+            "size": None
+        },
+                                                        param_source="driver-test-param-source"),
```
dliappis: nit: indentation seems a bit wonky here

danielmitterdorfer: Thanks! Addressed in https://github.com/elastic/rally/pull/789/commits/b1c5b724e8f15c85272d727a59de156c551c18f2.

PR #793, opened 2019-10-14
Add ability to restore from a snapshot

### LINKED ISSUE #341: Add ability to restore from a snapshot
Currently we always need to bulk-index data to get data into Elasticsearch but sometimes it would be better to just provide a snapshot and use that.

## Amended ticket with more information

_Amended on: 20191010_

In this ticket we want to tackle two things and implement two runners:

1. Register a repository

    Needs to support all settings per repository [type](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html). For cloud based types (e.g. `azure`, `gcs`, `s3`) the important ones are:
    - `bucket`
    - `base_path`
    - `client`
    - `max_snapshot_bytes_per_sec`
    - `max_restore_bytes_per_sec`
    - `readonly`

2. Restore snapshot

    This runner will [restore](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-snapshots.html#restore-snapshot) from any registered repository which may be specified with 1.
    Needs to support specifying the body of the request where things like `include_global_state` can be defined.

    Implementation details:
        1. Would be ideal if all kinds of retrying could be disabled for the elasticsearch-py call
        2. Would make sense to have runner level user specified absolute timeout? I am thinking of this in case it's hard to find via subsequent ES API calls if there was a problem with restoration.
        3. After the initial restore call, should monitor progress and return when done. A good API to do this is `_cat/recovery?v\&format=json\&filter_path=bytes_percent`
        4. Progress is optional but nice to have (can be tackled after https://github.com/elastic/rally/pull/789 is in)

danielmitterdorfer:
With this commit we introduce new operations in Rally that allow Rally
to restore data from a snapshot.

Closes #341

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -805,6 +805,54 @@ With the operation ``sleep`` you can sleep for a certain duration to ensure no r
 
 This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
 
+delete-snapshot-repository
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With the operation ``delete-snapshot-repository`` you can delete an existing snapshot repository. The ``delete-snapshot-repository`` operation supports the following parameter:
+
+* ``repository`` (mandatory): The name of the snapshot repository to delete.
+
+This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+create-snapshot-repository
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With the operation ``create-snapshot-repository`` you can create a new snapshot repository. The ``create-snapshot-repository`` operation supports the following parameters:
+
+* ``repository`` (mandatory): The name of the snapshot repository to create.
+* ``body`` (mandatory): The body of the create snapshot repository request.
+* ``request-params`` (optional): A structure containing HTTP request parameters.
+
+This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+restore-snapshot
+~~~~~~~~~~~~~~~~
+
+With the operation ``restore-snapshot`` you can restore a snapshot from an already created snapshot repository. The ``restore-snapshot`` operation supports the following parameters:
+
+* ``repository`` (mandatory): The name of the snapshot repository to use. This snapshot repository must exist prior to calling ``restore-snapshot``.
+* ``snapshot`` (mandatory): The name of the snapshot to restore.
+* ``wait-for-completion`` (optional, defaults to ``False``): Whether this call should return immediately or block until the snapshot is restored.
+* ``request-params`` (optional): A structure containing HTTP request parameters.
+
+.. note::
+    In order to ensure that the track execution only continues after a snapshot has been restored, set ``wait-for-completion`` to ``true``. In that case you should also increase the request timeout. In the example below we set it to 7200 seconds (or 2 hours)::
+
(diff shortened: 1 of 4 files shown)


tests/driver/runner_test.py
```diff
@@ -1976,6 +1976,140 @@ def test_sleep(self, sleep, es):
...
+                            "index": {
+                                "size": {
+                                    "total": "175.4mb",
+                                    "total_in_bytes": 179063092,
+                                    "recovered": "165.7mb",
+                                    "recovered_in_bytes": 168891939,
+                                },
+                            }
+                        }
+
```
dliappis: nit: empty line here

tests/driver/runner_test.py
```diff
@@ -1976,6 +1976,140 @@ def test_sleep(self, sleep, es):
...
+                    "shards": [
+                        {
+                            "id": 0,
+                            "index": {
+                                "size": {
+                                    "total": "75.4mb",
+                                    "total_in_bytes": 79063092,
+                                    "recovered": "65.7mb",
+                                    "recovered_in_bytes": 68891939,
+                                },
```
dliappis: nit: dangling comma

tests/driver/runner_test.py
```diff
@@ -1976,6 +1976,140 @@ def test_sleep(self, sleep, es):
...
+                        },
+                        {
+                            "id": 1,
+                            "index": {
+                                "size": {
+                                    "total": "175.4mb",
+                                    "total_in_bytes": 179063092,
+                                    "recovered": "165.7mb",
+                                    "recovered_in_bytes": 168891939,
+                                },
```
dliappis: nit: dangling comma

docs/track.rst
```diff
@@ -805,6 +805,54 @@ With the operation ``sleep`` you can sleep for a certain duration to ensure no r
...
+
+With the operation ``restore-snapshot`` you can restore a snapshot from an already created snapshot repository. The ``restore-snapshot`` operation supports the following parameters:
+
+* ``repository`` (mandatory): The name of the snapshot repository to use. This snapshot repository must exist prior to calling ``restore-snapshot``.
+* ``snapshot`` (mandatory): The name of the snapshot to restore.
+* ``wait-for-completion`` (optional, defaults to ``False``): Whether this call should return immediately or block until the snapshot is restored.
+* ``request-params`` (optional): A structure containing HTTP request parameters.
+
+.. note::
+    In order to ensure that the track execution only continues after a snapshot has been restored, set ``wait-for-completion`` to ``true``. In that case you should also increase the request timeout. In the example below we set it to 7200 seconds (or 2 hours)::
```
dliappis: How about joining the two sentences like ...

> In order to ensure that the track execution only continues after a snapshot has been restored, set ``wait-for-completion`` to ``true`` **and** increase the request timeout.

docs/track.rst
```diff
@@ -805,6 +805,54 @@ With the operation ``sleep`` you can sleep for a certain duration to ensure no r
...
+
+        "request-params": {
+            "request_timeout": 7200
+        }
+
+    However, this might not work if a proxy is in between the client and Elasticsearch and that proxy has a shorter request timeout configured than the client. In that case, keep the default value for ``wait-for-completion`` and instead add a ``wait-for-recovery`` runner in the next step. This has the additional advantage that you'll get a progress report while the snapshot is being restored.
```
dliappis: Somehow `that` looks out of place to me when `this` is used in every other sentence and everything looks very specific. How about: `In this case, keep the default value for ``wait-for-completion`` and instead add a ``wait-for-recovery`` runner in the next step.`

PR #800, opened 2019-10-17
Properly wait for recovery to finish

### LINKED ISSUE #796: IndicesRecovery runner runs with 1 iteration and ends abruptly
The new `IndicesRecovery` runner introduced in #793 (used by the [wait-for-recovery](https://esrally.readthedocs.io/en/latest/track.html#wait-for-recovery) operation) seems to use an iteration-count based schedule with 1 iteration. As a result it runs once, (calculates [percent_completed](https://github.com/elastic/rally/blob/a05207f7d63be6decdcc071905ce1e69c91b8b0f/esrally/driver/runner.py#L1197-L1198) correctly) and ends due to the schedule. Logs show:

> Creating iteration-count based schedule with [deterministic] distribution for [wait-for-recovery] with [0] warmup iterations and [1] iterations.

My understanding is that since we don't specify `iterations` or `warmup_iterations`, based on [#requires_time_period_schedule](https://github.com/elastic/rally/blob/a05207f7d63be6decdcc071905ce1e69c91b8b0f/esrally/driver/driver.py#L1398-L1405) which then sets [infinite=True](https://github.com/elastic/rally/blob/a05207f7d63be6decdcc071905ce1e69c91b8b0f/esrally/driver/driver.py#L1516-L1517) forcing 1 iteration in [#schedule_for](https://github.com/elastic/rally/blob/a05207f7d63be6decdcc071905ce1e69c91b8b0f/esrally/driver/driver.py#L1386-L1388).

Was it meant to be ran with a custom schedule?

Related: #789 and #763

danielmitterdorfer:
With this commit we ensure that a proper schedule is chosen when waiting
for shard recovery to finish. We also ensure that a short time-span where no
recoveries are active (but more might continue soon) is not mistakenly
treated as a condition where all recoveries have already finished.

Closes #796

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -849,7 +849,25 @@ This is an administrative operation. Metrics are not reported by default. Report
 wait-for-recovery
 ~~~~~~~~~~~~~~~~~
 
-With the operation ``wait-for-recovery`` you can wait until an ongoing index recovery finishes. The ``wait-for-recovery`` operation does not support any parameters.
+With the operation ``wait-for-recovery`` you can wait until an ongoing shard recovery finishes. The ``wait-for-recovery`` operation supports the following parameters:
+
+* ``completion-recheck-attempts`` (optional, defaults to 3): It might be possible that the `index recovery API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-recovery.html>`_ reports that there are no active shard recoveries when a new one might be scheduled shortly afterwards. Therefore, this operation will check several times whether there are still no active recoveries. In between those attempts, it will wait for a time period specified by ``completion-recheck-wait-period``.
+* ``completion-recheck-wait-period`` (optional, defaults to 2 seconds): Time in seconds to wait in between consecutive attempts.
+
+.. warning::
+
+    By default this operation will run unthrottled (like any other) but you should limit the number of calls by specifying the ``target-throughput`` property on the corresponding task::
+
+        {
+          "operation": {
+            "operation-type": "wait-for-recovery",
+            "completion-recheck-attempts": 2,
+            "completion-recheck-wait-period": 5
+          },
+          "target-throughput": 10
+        }
+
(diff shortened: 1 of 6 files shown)


docs/track.rst
```diff
@@ -849,7 +849,25 @@ This is an administrative operation. Metrics are not reported by default. Report
 wait-for-recovery
 ~~~~~~~~~~~~~~~~~
 
-With the operation ``wait-for-recovery`` you can wait until an ongoing index recovery finishes. The ``wait-for-recovery`` operation does not support any parameters.
+With the operation ``wait-for-recovery`` you can wait until an ongoing shard recovery finishes. The ``wait-for-recovery`` operation supports the following parameters:
+
+* ``completion-recheck-attempts`` (optional, defaults to 3): It might be possible that the `index recovery API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-recovery.html>`_ reports that there are no active shard recoveries when a new one might be scheduled shortly afterwards. Therefore, this operation will check several times whether there are still no active recoveries. In between those attempts, it will wait for a time period specified by ``completion-recheck-wait-period``.
```
dliappis: s/It might/It's

tests/driver/driver_test.py
```diff
@@ -578,6 +598,23 @@ def test_finite_schedule_with_progress_indication(self):
...
+
+        invocations = driver.schedule_for(self.test_track, task, 0)
+
+        self.assert_schedule([
+            (0.0, metrics.SampleType.Normal, None, {"body": ["a"]}),
+            (1.0, metrics.SampleType.Normal, None, {"body": ["a"]}),
+            (2.0, metrics.SampleType.Normal, None, {"body": ["a"]}),
+            (3.0, metrics.SampleType.Normal, None, {"body": ["a"]}),
+            (4.0, metrics.SampleType.Normal, None, {"body": ["a"]}),
+        ], invocations, infinite_schedule=True)
```
dliappis: I think that when we merge this to current master it's going to fail due to https://github.com/elastic/rally/pull/801/files

danielmitterdorfer: Good catch! I've merged master and changed the test now in https://github.com/elastic/rally/pull/800/commits/78235be4769abbe88f0f456efcd7dcf81404defb.

PR #804, opened 2019-10-28
Allow multiple nodes per IP

danielmitterdorfer:
With this commit we expose a new provisioner variable `all_node_names`
which consists of all Elasticsearch node names in the cluster. This can
be used to set `cluster.initial_master_nodes` in order to ensure the
discovery process takes all nodes into account even if they are started
on the same host. Previously we have used the IP address for this
purpose but this is ambiguous when multiple nodes are started on the
same host.

### DIFF
--- a/esrally/mechanic/mechanic.py
+++ b/esrally/mechanic/mechanic.py
@@ -58,19 +58,20 @@ class StartEngine:
         self.port = port
         self.node_id = node_id
 
-    def for_nodes(self, all_node_ips=None, ip=None, port=None, node_ids=None):
+    def for_nodes(self, all_node_ips=None, all_node_ids=None, ip=None, port=None, node_ids=None):
         """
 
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:
Note: This requires https://github.com/elastic/rally-teams/pull/34 to work as intended.

ebadyano:
Should we add anything to documentation at all to mention that it is possible?

danielmitterdorfer:
> Should we add anything to the documentation at all to mention that starting with multiple nodes on same ip address is possible now?

Actually this was always meant to be possible and in fact multiple nodes also start up on a single host. The problem was that the nodes did not form a cluster due to misconfigured discovery settings in that case.

ebadyano:
makes sense, thank you!

danielmitterdorfer:
Thanks for your review! :)

PR #806, opened 2019-10-31
Upgrade jsonschema dependency

danielmitterdorfer:
So far Rally uses version 2.5.1 of `jsonschema` which is more than four
years old by now and is also not officially compatible with Python 3.7.
With this commit we upgrade to the currently latest version 3.1.1 which
is compatible with Python 3.5 - 3.7.


danielmitterdorfer:
Thanks!

PR #807, opened 2019-10-31
Upgrade tabulate dependency

danielmitterdorfer:
So far Rally uses version 0.8.1 of `tabulate` which is not fully
compatible with Python 3.7 (see
https://bitbucket.org/astanin/python-tabulate/commits/3bfe29401e0f32ba4cff374d234ba37331a472a1).
Compatibility issues have been fixed in the most recent version 0.8.5
hence we upgrade to it.


danielmitterdorfer:
Thanks!

PR #814, opened 2019-11-12
Fix error handling for automatically derived version

### LINKED ISSUE #813: Version check error message is misleading
At https://github.com/elastic/rally/blob/master/esrally/racecontrol.py#L200, we notify the user that a distribution version must be provided when we have failed to get the version by connecting to the cluster.  
If Rally cannot connect to the cluster with the information provided, supplying a version will not help the situation.  We should have a more helpful error message in this situation.

drawlerr:
Let the source of the problem get to the user.

Closes #813

### DIFF
--- a/esrally/racecontrol.py
+++ b/esrally/racecontrol.py
@@ -195,7 +195,11 @@ class BenchmarkActor(actor.RallyActor):
         # cases (external pipeline and user did not specify the distribution version) where we need to derive it ourselves. For source
         # builds we always assume "master"
         if not msg.sources and not self.cfg.exists("mechanic", "distribution.version"):
-            distribution_version = mechanic.cluster_distribution_version(self.cfg)
+            import elasticsearch
+            try:
+                distribution_version = mechanic.cluster_distribution_version(self.cfg)
+            except elasticsearch.ConnectionError:
+                raise exceptions.SystemSetupError("Could not connect to cluster at specified address, please check your connection.")
             self.logger.info("Automatically derived distribution version [%s]", distribution_version)
             self.cfg.add(config.Scope.benchmark, "mechanic", "distribution.version", distribution_version)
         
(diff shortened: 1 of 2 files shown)


ebadyano:
@drawlerr Thank you for opening a PR! With your change we now get this error:
```
[ERROR] Cannot race. Error in race control (ConnectionError(<urllib3.connection.HTTPConnection object at 0x105b30f60>: Failed to establish a new connection: [Errno 61] Connection refused) caused by: NewConnectionError(<urllib3.connection.HTTPConnection object at 0x105b30f60>: Failed to establish a new connection: [Errno 61] Connection refused))
```
Do we want to add something more user friendly, like: `Could not connect to cluster at specified address, please check your connection.` Or something along those lines?

esrally/racecontrol.py
```diff
@@ -195,9 +195,12 @@ def setup(self, msg, sender):
         # cases (external pipeline and user did not specify the distribution version) where we need to derive it ourselves. For source
         # builds we always assume "master"
         if not msg.sources and not self.cfg.exists("mechanic", "distribution.version"):
-            distribution_version = mechanic.cluster_distribution_version(self.cfg)
-            if not distribution_version:
-                raise exceptions.SystemSetupError("A distribution version is required. Please specify it with --distribution-version.")
+            import elasticsearch
+            try:
+                distribution_version = mechanic.cluster_distribution_version(self.cfg)
+            except elasticsearch.ConnectionError as e:
```
danielmitterdorfer: Is there a reason we are only catching connection errors? Does it make sense to catch the base type `elasticsearch.ElasticsearchException` instead? Also, we are leaking implementation details (usage of the Elasticsearch Python client) and I wonder whether we should push down that exception handling code into the callee `mechanic#cluster_distribution_version()`?

drawlerr: Pushing the exception handling down brings us back to the situation we had before this change.
I think catching anything besides ConnectionError here risks another misleading error message.  I've decided "friendliness" isn't worth it here and reverted to the "unfriendly" version.

danielmitterdorfer: > Pushing the exception handling down brings us back to the situation we had before this change.

Not really: Previously we were catching any exception and returning `None`. Now we are catching an exception and reraise a friendlier one so this is definitely a change in behavior.  Anyway, I'm ok with your change in https://github.com/elastic/rally/pull/814/commits/61889c6c90938187371b9e50c2c728171450a930 so I'll approve.

PR #816, opened 2019-11-15
Harmonize local pyenv versions with CI ones

dliappis:
Use the same Python patch versions for 3.5/3.6/3.7 as the ones used
in CI.


dliappis:
NOTE: After merging, `make prereq` will be required for local development.

PR #817, opened 2019-11-15
Fix mistake in wait-for-recovery docs

dliappis:
To run an operation every 10s we specify target-interval rather than
target-throughput.

Relates #800


dliappis:
Thanks for the quick review!

PR #818, opened 2019-11-18
Store build.log in log directory

danielmitterdorfer:
With this commit we move Rally's build log from a race-specific
directory to its log directory. This file captures Gradle's output when
users are benchmarking a source build of Elasticsearch. By moving this
file to the log directory it is better visible to users (already in the
same location as other log files).


danielmitterdorfer:
Marked as breaking as the build log file moves and thus it may have an impact on users.

docs/configuration.rst
```diff
@@ -135,6 +135,18 @@ The log file will not be rotated automatically as this is problematic due to Ral
             notifempty              # don't attempt to rotate empty ones
     }
 
+If you are benchmarking source builds of Elasticsearch, Rally will write the log output of the build to ``~/.rally/logs/build.log``. Similarly to the configuration for ``rally.log``, you can use the following ``logrotate`` configuration as a starting point::
```
dliappis: When reading this first time, it may not be entirely clear what the logrotate config is for, I suggest we rephrase:

> you can use the following ``logrotate`` configuration as a starting point::

to

> you can use the following ``logrotate`` configuration to restrict the volume of the log file::

docs/configuration.rst
```diff
@@ -135,6 +135,18 @@ The log file will not be rotated automatically as this is problematic due to Ral
...
+
+    /home/user/.rally/logs/build.log {
+            daily                   # rotate daily
```
dliappis: In-line comments don't work. They should be in a separate line.

(From [man logrotate](https://linux.die.net/man/8/logrotate)

> Note that comments may appear anywhere in the config file as long as the first non-whitespace character on the line is a #.
)

We need to update also the example for `rally.log`.

docs/configuration.rst
danielmitterdorfer: I thought that can be inferred based on context (i.e. reading the paragraph above about `rally.log`).

docs/configuration.rst
danielmitterdorfer: addressed in https://github.com/elastic/rally/pull/818/commits/dd68f38f65058f1e9029ce8ade1e0d7071075de6.

danielmitterdorfer:
Thanks!

PR #819, opened 2019-11-18
Reduce API surface of loader

danielmitterdorfer:
With this commit we reduce the number of functions that are exported by
the track loader to other modules.


danielmitterdorfer:
Thanks for the review! :)

PR #820, opened 2019-11-18
Upgrade psutil to work with newer linux kernels

hub-cap:
The 4.18 kernel introduced a number of new fields into the disk io stats
proc fileystem, and running tests on new versions of linux is causing
failures because psutil cannot interpret the new fields. This fix
updates the version such that it works w/ newer kernels.

Relates giampaolo/psutil#1354

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,7 @@ install_requires = [
     # transitive dependency urllib3: MIT
     "elasticsearch==7.0.5",
     # License: BSD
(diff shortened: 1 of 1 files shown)


hub-cap:
As this is my first PR, I know very little about proper labeling procedures, as well as what I would do to test this change to ensure that the upgrade did not cause a significant change in benchmarks.

PR #821, opened 2019-11-19
Move logrotate comments to separate lines

dliappis:
because logrotate doesn't support in-line comments.

Relates #818


dliappis:
Will go ahead and merge now as #818 depends on it, thanks all for the reviews.

PR #825, opened 2019-11-20
Change remaining it tests to port 19200

### LINKED ISSUE #824: Rally docker tests use port 9200
The rally docker tests use the default elasticsearch port instead of a random or known nonstandard port. this should be updated so that it does not use 9200 as that can collide with existing es instances locally.

hub-cap:
The docker tests were using port 9200 because target_hosts was not
specified. Additionally the tests that were using compose were also
leaking a 9200 port on to a default bridge network that other containers
might be using, which causes a collision and failure. This commit fixes
both of these problems.

Closes #824

### DIFF
--- a/docker/docker-compose-tests.yml
+++ b/docker/docker-compose-tests.yml
@@ -16,15 +16,16 @@ services:
     container_name: es01
     environment:
       - discovery.type=single-node
+      - http.port=19200
       - "ES_JAVA_OPTS=-Xms1g -Xmx1g"
     volumes:
       - esdata1:/usr/share/elasticsearch/data
     ports:
-      - 19200:9200
+      - 19200:19200
     networks:
       - esnet
     healthcheck:
-          test: curl -f http://localhost:9200
+          test: curl -f http://localhost:19200
           interval: 5s
           timeout: 2s
           retries: 10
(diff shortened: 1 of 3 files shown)


docker/docker-compose-tests.yml
```diff
@@ -16,15 +16,16 @@ services:
     container_name: es01
     environment:
       - discovery.type=single-node
+      - http.port=19200
```
dliappis: Just curious, was there a need to change the where Elasticsearch listens inside the container?

hub-cap: I put a bit in the comment. We use a default bridge which other containers can share. i was thinking about it more on my walk this morning and i think we should just use our own bridge network for testing so this does not happen.

dliappis: Oh I missed that. I think we can also use our own bridge network e.g. as seen in the Elasticsearch Docker image [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html#docker-compose-file). Either way I am fine.

hub-cap: yea I decided this was a better solution, just in case we add more nodes eventually, we dont want to have to continue to choose ports-o-fun. im running the it tests now on the change :)

PR #826, opened 2019-11-21
Add note about mustache template

### LINKED ISSUE #823: Document how to escape Mustache expressions in tracks
If a user wants to use Mustache expresssions in their Rally tracks (e.g. because they want to [use Watcher](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/actions-email.html) or [search templates](https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-template.html)), they run into trouble, because each Rally track is actually a Jinja template and thus this expression will be treated as Jinja template expression. Therefore, they need to be [escaped](https://jinja.palletsprojects.com/en/2.10.x/templates/#escaping) using `{% raw %}...{% endraw %}` but this is not obvious. We should therefore add an example to our track reference.

hub-cap:
closes #823


hub-cap:
image

docs/adding_tracks.rst
```diff
@@ -626,6 +626,9 @@ Template Language
 
 Rally uses `Jinja2 <http://jinja.pocoo.org/docs/dev/>`_ as a template language so you can use Jinja2 expressions in track files.
 
+.. note::
```
dliappis: Reading the ticket (and thinking about the maximum visibility and benefit) I think we want to have a small example besides an operation.

What if we place a second example inside the [put pipeline operation](https://esrally.readthedocs.io/en/stable/track.html#put-pipeline) doc page showing a similar example to what we have in the Elasticsearch docs for [handling failure in pipelines](https://www.elastic.co/guide/en/elasticsearch/reference/current/handling-failure-in-pipelines.html).
We could raise the attention with a note that "if you need to use Elasticsearch Mustache `{{ }}` syntax you must escape it properly" (and link the note you added in this PR), for example:

```
{
  "name": "define-ip-geocoder",
  "operation-type": "put-pipeline",
  "id": "ip-geocoder",
  "body": {
    "description": "Extracts location information from the client IP.",
    "processors": [
      {
        "geoip": {
          "field": "clientip",
          "properties": [
            "city_name",
            "country_iso_code",
            "country_name",
            "location"
          ]
        }
      }
    ],
    "on_failure": [
      {
        "set" : {
          "field" : "_index",
          {% raw %}
          "value" : "failed-{{ _index }}"
          {% endraw %}
        }
      }        
    ]
}
```

^^ For the sake of not repeating things we could even just show `"processors": [ ... ],` as used in the Elasticsearch docs page.

docs/adding_tracks.rst
```diff
@@ -626,6 +626,9 @@ Template Language
...
+
+    Elasticsearch utilizes Mustache formatting in a few places, notably in `Search templates <https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-template.html>`_ and `Watcher templates <https://www.elastic.co/guide/en/elasticsearch/reference/7.4/actions-email.html>`_. If you are using Mustache in your rally templates you must `escape them properly <https://jinja.palletsprojects.com/en/2.10.x/templates/#escaping>`_.
```
dliappis: nit: we use `Rally` (capitalized) consistently everywhere.

docs/adding_tracks.rst
hub-cap: @dliappis I decided to just add to the existing example, and added a link from the adding tracks page.

docs/adding_tracks.rst
```diff
@@ -626,6 +626,7 @@ Template Language
 
 Rally uses `Jinja2 <http://jinja.pocoo.org/docs/dev/>`_ as a template language so you can use Jinja2 expressions in track files.
 
+Elasticsearch utilizes Mustache formatting in a few places, notably in `Search templates <https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-template.html>`_ and `Watcher templates <https://www.elastic.co/guide/en/elasticsearch/reference/7.4/actions-email.html>`_. If you are using Mustache in your Rally templates you must `escape them properly <https://jinja.palletsprojects.com/en/2.10.x/templates/#escaping>`_. See :ref:`put_pipeline` for an example.
```
danielmitterdorfer: Any reason you've removed the `.. note::` that you've added in the previous commit?

danielmitterdorfer: Nit: "Search templates" -> "search templates"

docs/track.rst
```diff
@@ -427,6 +429,16 @@ Example::
               ]
             }
           }
+        ],
+        "on-failure": [
```
danielmitterdorfer: There is a syntax error in the pipeline. The element is called `on_failure`instead of `on-failure` (see [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/handling-failure-in-pipelines.html)).

docs/track.rst
```diff
@@ -427,6 +429,16 @@ Example::
...
+          {
+            "set": {
+              "field": "_index",
+              {% raw %}
+              "value": "failed-{{ _index }}"
```
danielmitterdorfer: I think this is pretty complex to grasp. We should add a note about the `raw` / `endraw` tags below the example and we should also link to the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/handling-failure-in-pipelines.html) so users can lookup what this is actually doing.

docs/adding_tracks.rst
hub-cap: ya i removed it cuz it got big enough for a full paragraph and just looked like a huge note w/ a bunch of links, rather than some concise text. I can change it back tho if youd like

docs/track.rst
hub-cap: so are you saying to revert it, or to just add links to show what its doing better?

docs/track.rst
```diff
@@ -447,6 +447,8 @@ This example requires that the ``ingest-geoip`` Elasticsearch plugin is installe
 
 This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
 
+Please see `the pipeline documentation <https://www.elastic.co/guide/en/elasticsearch/reference/current/handling-failure-in-pipelines.html>`_ for details on handling failures in pipelines.
```
danielmitterdorfer: Maybe move this paragraph so it is directly below the example? Also, some introductory text would be nice; something along the lines of: 

> In the following example we setup a pipeline that ... . Note that we need to use `raw` / `endraw` texts to ensure that Rally passes the Mustache template as is and does not attempt to resolve it itself ...

(and then link back to your newly added note about templating Rally?)

docs/adding_tracks.rst
```diff
@@ -626,6 +626,7 @@ Template Language
 
 Rally uses `Jinja2 <http://jinja.pocoo.org/docs/dev/>`_ as a template language so you can use Jinja2 expressions in track files.
 
+Elasticsearch utilizes Mustache formatting in a few places, notably in `search templates <https://www.elastic.co/guide/en/elasticsearch/reference/7.4/search-template.html>`_ and `Watcher templates <https://www.elastic.co/guide/en/elasticsearch/reference/7.4/actions-email.html>`_. If you are using Mustache in your Rally templates you must `escape them properly <https://jinja.palletsprojects.com/en/2.10.x/templates/#escaping>`_. See :ref:`put_pipeline` for an example.
```
danielmitterdorfer: Nit: If you are using Mustache in your Rally **tracks**?

docs/track.rst
hub-cap: done

docs/track.rst
```diff
@@ -409,6 +409,8 @@ With the operation-type ``put-pipeline`` you can execute the `put pipeline API <
 * `id` (mandatory): Pipeline id
 * `body` (mandatory): Pipeline definition
 
+In this example we setup a pipeline that adds location information to a ingested document as well as a pipeline failure block to change the index in which the document was supposed to be written. Note that we need to use the ``raw`` and ``endraw`` blocks to ensure that Rally does not attempt to resolve the Mustache template. See :ref:`template_language` for more information.
```
danielmitterdorfer: nit:
 
```
:ref:`template_language`
```

is rendered as:

```
Template Language
```

(i.e. upper-case)

How about we spell it out explicitly?

```
See the documentation about Rally's :ref:`template language <template_language>` for more information.
```

hub-cap: ohhh i did not know this was possible. TIL!

PR #828, opened 2019-11-26
Remove auto-upgrade for old log config

danielmitterdorfer:
With this commit we remove the automatic upgrade of an old log
configuration that has been shipped with Rally 1.0.0. This version has
been shipped in July 2018 and we expect all users have upgraded to a
newer version of Rally by now.


danielmitterdorfer:
Thanks for the quick review!

PR #829, opened 2019-11-26
Reduce usage of platform-specific code

danielmitterdorfer:
With this commit we improve portability by relying less on
platform-dependent code. We also enhance terminal output on Windows by
avoiding any escape sequences.

### DIFF
--- a/esrally/log.py
+++ b/esrally/log.py
@@ -57,7 +57,11 @@ def install_default_log_config():
         source_path = io.normalize_path(os.path.join(os.path.dirname(__file__), "resources", "logging.json"))
         with open(log_config, "w", encoding="UTF-8") as target:
             with open(source_path, "r", encoding="UTF-8") as src:
-                contents = src.read().replace("${LOG_PATH}", paths.logs())
+                # Ensure we have a trailing path separator as after LOG_PATH there will only be the file name
+                log_path = os.path.join(paths.logs(), "")
+                # the logging path might contain backslashes that we need to escape
+                log_path = io.escape_path(log_path)
+                contents = src.read().replace("${LOG_PATH}", log_path)
                 target.write(contents)
     io.ensure_dir(paths.logs())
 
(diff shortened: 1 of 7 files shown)


esrally/log.py
```diff
@@ -57,7 +57,11 @@ def install_default_log_config():
         source_path = io.normalize_path(os.path.join(os.path.dirname(__file__), "resources", "logging.json"))
         with open(log_config, "w", encoding="UTF-8") as target:
             with open(source_path, "r", encoding="UTF-8") as src:
-                contents = src.read().replace("${LOG_PATH}", paths.logs())
+                # Ensure we have a trailing path separator as after LOG_PATH there will only be the file name
+                log_path = os.path.join(paths.logs(), "")
+                # the logging path might contain backslashes that we need to escape
+                log_path = io.escape_path(log_path)
```
dliappis: I wonder if we can avoid some of the special handling in the future by relying on the methods in the [pathlib](https://docs.python.org/3/library/pathlib.html#module-pathlib) module.

esrally/resources/logging.json
```diff
@@ -20,14 +20,14 @@
   "handlers": {
     "rally_log_handler": {
       "class": "logging.handlers.WatchedFileHandler",
-      "filename": "${LOG_PATH}/rally.log",
+      "filename": "${LOG_PATH}rally.log",
```
dliappis: Ugh it's unfortunate it has to look like that, but as we are currently consistently using the `os` module I don't see a workaround.

esrally/log.py
danielmitterdorfer: I agree that this is unfortunate but the problem is that this string ends up in JSON. On Windows, a path ends up as the JSON string `C:\some\path` and when this string is read, the `\s` and `\p` are interpreted as escape sequences. This situation is similar to regex literals but for these we can use raw strings (literal prefixed by `r`) to avoid that.

Issue #736, opened 2019-07-22
Investigate moving integration tests to a more solid foundation


danielmitterdorfer:
We currently use a [shell script for our integration tests](https://github.com/elastic/rally/blob/b0c7b9378cc9c890c2973ea19df6797317e45eb3/integration-test.sh). This has several limitiations:

* We need to resort to adhoc solutions (e.g. we need Perl or require Docker for search / replace)
* Limited ability for assertions
* We cannot report test failures properly
* Hard to extend

We should instead investigate how we can properly support integration tests with Python (e.g. via py.test). 

Requirements:

* Should avoid all of the shorcomings above
* We should be able to run unit tests and integration tests separately. Ideally we can also clearly see in the source code structure which tests are unit tests and which ones are integration tests (separate root directories?)
* We should not rely on external tools such as Perl (note: Docker is required for Docker tests). Note: In order to reduce scope, we might want to port the existing tests first 1:1 and then improve them in follow-up PRs

danielmitterdorfer:
Initially addressed in https://github.com/elastic/rally/pull/919. We will track the actual migration in https://github.com/elastic/rally/issues/975.

PR #830, opened 2019-11-27
Manage Elasticsearch nodes with dedicated subcommands

### LINKED ISSUE #697: Allow to manage Elasticsearch nodes separately from benchmarking

danielmitterdorfer:
With this commit we introduce three new subcommands to Rally:

* `install`: To install a single Elasticsearch node locally
* `start`: To start an Elasticsearch node that has been previously installed
* `stop`: To stop a running Elasticsearch node

To run a benchmark, users first issue `install`, followed by `start` on all
nodes. Afterwards, the benchmark is run using the `benchmark-only` pipeline.
Finally, the `stop` command is invoked on all nodes to shutdown the cluster.

To ensure that system metrics are stored consistently (i.e. they contain the
same metadata like race id and race timestamp), we expose the race id as a
command line parameter and defer writing any system metrics until the `stop`
command is invoked. We attempt to read race metadata from the Elasticsearch
metrics store for that race id which have been written earlier by the benchmark
and merge the metadata when we write the system metrics.

The current implementation is considered a new experimental addition to the
existing mechanism to manage clusters with the intention to eventually replace
it. The command line interface is specific to Zen discovery and subject to
change as we learn more about its use.

Closes #697


danielmitterdorfer:
To help with the review, here are some test commands:

```
# benchmark a single-node cluster
esrally install --quiet --distribution-version=7.4.2 --build-type=tar --node-name="rally-node-0" --master-nodes="rally-node-0" --network-host="127.0.0.1" --seed-hosts="127.0.0.1:39300"

# TODO: Capture the installation id from the previous command
esrally start --installation-id=c2e6d5fb-0405-4f39-9213-cf8799a760cc --race-id=b1228394-998f-413d-b454-adef7e0f2a7b

esrally --pipeline=benchmark-only --target-host=127.0.0.1:39200 --track=geonames --challenge=append-no-conflicts-index-only --on-error=abort --race-id=b1228394-998f-413d-b454-adef7e0f2a7b

esrally stop --installation-id=c2e6d5fb-0405-4f39-9213-cf8799a760cc --race-id=b1228394-998f-413d-b454-adef7e0f2a7b
```

To test a Docker container (see `--build-type=docker`):

```
esrally install --quiet --distribution-version=7.4.2 --build-type=docker --node-name="rally-node-0" --master-nodes="rally-node-0" --network-host="127.0.0.1" --seed-hosts="127.0.0.1:39300"
```

All other commands (`start`, `stop`) are identical to the scenario above. The Docker support in Rally has (and always had) several restrictions, e.g. you can only spin up a single node. This has not changed with this PR.

It is also possible to use user-tags. Just specify them when running the benchmark:

```

esrally --pipeline=benchmark-only --target-host=127.0.0.1:39200 --track=geonames --challenge=append-no-conflicts-index-only --on-error=abort --race-id=de604d0d-926c-4764-bbef-d8bc564755ae --user-tag="with-user-tags:true"
```

This should also show up on system metrics.

To benchmark a source build, you need to modify the install command (use `--skip-build` to skip the build):

```
esrally install --revision=latest --node-name="rally-node-0" --network-host="127.0.0.1" --master-nodes="rally-node-0" --seed-hosts="127.0.0.1:39300"
```

For a multi-node cluster, install multiple nodes but reference the other node(s) as seed hosts:

```
esrally install --distribution-version=7.4.2 --node-name="rally-node-0" --network-host="127.0.0.1" --master-nodes="rally-node-0,rally-node-1" --seed-hosts="127.0.0.1:39300,127.0.0.1:39301"

esrally install --distribution-version=7.4.2 --node-name="rally-node-1" --network-host="127.0.0.1" --master-nodes="rally-node-0,rally-node-1" --seed-hosts="127.0.0.1:39300,127.0.0.1:39301"
```

Note that we are always using the default http port (which is 39200). To customize it, specify e.g. `--http-port=19200`. Note that the transport port will always be 100 ports above the http port.

integration-test.sh
```diff
@@ -470,13 +476,20 @@ function test_docker_dev_image {
 function test_node_management_commands {
     local cfg
     local dist
+    local build_type
     random_configuration cfg
     random_distribution dist
+    random_build_type build_type
+
+    # for Docker we force the most recent distribution as we don't have Docker images for all versions that are tested
+    if [[ "$build_type" == "docker" ]]; then
```
danielmitterdorfer: While we kill Rally processes in `kill_related_es_processes` we don't do the same for any running Docker images. I think that we might need to stop Docker containers as well but I wonder whether we can find a robust approach of finding the correct image.

dliappis: This is possible. What we need to do is:

1. Change `docker-compose.yml.j2` to further specify a label like so:

```
diff --git a/esrally/resources/docker-compose.yml.j2 b/esrally/resources/docker-compose.yml.j2
index c5bc10a..19e72e9 100644
--- a/esrally/resources/docker-compose.yml.j2
+++ b/esrally/resources/docker-compose.yml.j2
@@ -4,6 +4,8 @@ services:
     cap_add:
       - IPC_LOCK
     image: "{{docker_image}}:{{es_version}}"
+    labels:
+      io.rally.description: "Label to help identify Elasticsearch containers launched by Rally"
     {%- if docker_cpu_count is defined %}
     cpu_count: {{docker_cpu_count}}
     {%- endif %}
```

2. Use the label to retrieve the ID of container(s) launched by Rally using a filter like `docker ps --filter "label=io.rally.description" --format "{{.ID}}"` (if >1 they'll be newline separated).
    This will match all nodes matching the specific label key. To more targeted, we can also match the value which could be optionally specified via a j2 variable in the compose file, but this functionality would require an optional argument for the start subcommand (e.g. `--docker-label="..."`) to override the j2 variable.

integration-test.sh
```diff
@@ -467,6 +473,41 @@ function test_docker_dev_image {
...
+      dist="${DISTRIBUTIONS[${#DISTRIBUTIONS[@]}-1]}"
+    fi
+
+    info "test install [--configuration-name=${cfg}] [--build-type=${build_type}]"
+    kill_rally_processes
+
+    raw_install_id=$(esrally install --quiet --configuration-name="${cfg}" --distribution-version="${dist}" --build-type="${build_type}" --node-name="rally-node-0" --master-nodes="rally-node-0" --network-host="127.0.0.1" --http-port=39200 --seed-hosts="127.0.0.1:39300")
+    install_id=$(echo "${raw_install_id}" | grep installation-id | cut -d '"' -f4)
```
dliappis: Don't want to rely on the presence of `jq` I presume?

integration-test.sh
danielmitterdorfer: Thanks! I'll implement this in the integration test. I think it's fine to use the less targeted approach and rely on a static label in `docker-compose.yml.j2`.

dliappis: I agree.

esrally/rally.py
```diff
@@ -171,6 +171,135 @@ def positive_number(v):
...
+        "--telemetry-params",
+        help="Define a comma-separated list of key:value pairs that are injected verbatim to the telemetry devices as parameters.",
+        default=""
+    )
+
+    stop_parser = subparsers.add_parser("stop", help="Stops an Elasticsearch node locally")
+    stop_parser.add_argument(
+        "--installation-id",
+        required=True,
+        help="The id of the installation to start",
```
drawlerr: s/start/stop

esrally/mechanic/launcher.py
```diff
@@ -70,44 +40,67 @@ def start(self, node_configurations):
...
+                return
+            time.sleep(0.5)
+        msg = "No healthy running container after {} seconds!".format(timeout)
+        logging.error(msg)
+        raise exceptions.LaunchError(msg)
+
+    def stop(self, nodes, metrics_store):
+        self.logger.info("Shutting down [%d] nodes running in Docker on this host.", len(nodes))
+        for node in nodes:
+            # readd meta-data - we already did this on startup but in case dedicated subcommands are used for
```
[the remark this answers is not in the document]
danielmitterdorfer: I really meant this as: "add the meta-data *again*". But this was only valid in an earlier commit so I'll remove the comment entirely.

esrally/mechanic/launcher.py
```diff
@@ -207,42 +190,44 @@ def _set_env(self, env, k, v, separator=' ', prepend=False):
...
-        if self.keep_running:
-            self.logger.info("Keeping [%d] nodes on this host running.", len(nodes))
-        else:
-            self.logger.info("Shutting down [%d] nodes on this host.", len(nodes))
+    def stop(self, nodes, metrics_store):
+        self.logger.info("Shutting down [%d] nodes on this host.", len(nodes))
         for node in nodes:
             proc = psutil.Process(pid=node.pid)
             node_name = node.node_name
+            # readd meta-data - we already did this on startup but in case dedicated subcommands are used for
```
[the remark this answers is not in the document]
danielmitterdorfer: I really meant this as: "add the meta-data *again*". But this was only valid in an earlier commit so I'll remove the comment entirely.

esrally/mechanic/mechanic.py
```diff
@@ -37,6 +39,133 @@ def download(cfg):
...
+    root_path = paths.install_root(cfg)
+    # avoid double-launching
+    try:
+        _load_node_file(root_path)
+        install_id = cfg.opts("system", "install.id")
+        race_id = cfg.opts("system", "race.id")
+        raise exceptions.SystemSetupError("A node with this installation id is already running. Please stop it first "
+                                          "with {} stop --installation-id={} --race-id={}"
+                                          .format(PROGRAM_NAME, install_id, race_id))
+    except FileNotFoundError:
```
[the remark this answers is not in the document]
danielmitterdorfer: Thanks! I was not aware and have changed the code accordingly. I've also moved the comment.

esrally/rally.py
```diff
@@ -171,6 +171,135 @@ def positive_number(v):
...
+        # the default will be dynamically derived by racecontrol based on the presence / absence of other command line options
+        default="")
+    stop_parser.add_argument(
+        "--race-id",
+        required=True,
+        help="Define a unique id for this race.",
+        default="")
```
drawlerr: Presumably, the "stop" command shouldn't be caring about races as it is just stopping nodes?

esrally/rally.py
```diff
@@ -345,20 +474,15 @@ def positive_number(v):
             help=argparse.SUPPRESS,
             type=lambda s: datetime.datetime.strptime(s, "%Y-%m-%d %H:%M:%S"),
             default=None)
-        # keeps the cluster running after the benchmark, only relevant if Rally provisions the cluster
-        p.add_argument(
-            "--keep-cluster-running",
```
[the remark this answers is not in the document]
danielmitterdorfer: restored

esrally/rally.py
```diff
@@ -598,21 +730,15 @@ def main():
     cfg.add(config.Scope.applicationOverride, "mechanic", "car.plugins", opts.csv_to_list(args.elasticsearch_plugins))
     cfg.add(config.Scope.applicationOverride, "mechanic", "car.params", opts.to_dict(args.car_params))
     cfg.add(config.Scope.applicationOverride, "mechanic", "plugin.params", opts.to_dict(args.plugin_params))
-    if args.keep_cluster_running:
```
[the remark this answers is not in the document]
danielmitterdorfer: restored

integration-test.sh
danielmitterdorfer: Yes, this was my intention. My hope is that with https://github.com/elastic/rally/issues/736 we'll extract this in Python anyway.

esrally/rally.py
danielmitterdorfer: Good catch!

esrally/rally.py
danielmitterdorfer: Upon stopping a node, Rally needs to store metrics and aggregate them. To have consistent meta-data across all nodes involved in a benchmark, we need the race id to tie them together and that's why this command line parameter is needed. We could probably store it when a node is started but I've opted for this simpler approach here. In fact, we could even skip it upon node startup because we only store metrics when a node is stopped but this felt like leaking implementation details into the command line interface which I wanted to avoid.

drawlerr: I like the approach of storing the active race ID.  Any chance we could include it in this PR?

danielmitterdorfer: Good idea, it makes the interface definitely simpler and easier to use. I've implemented this in https://github.com/elastic/rally/pull/830/commits/dbbbdb40e4d3fe8c172c75bc626256d13205d210.

integration-test.sh
danielmitterdorfer: Added in https://github.com/elastic/rally/pull/830/commits/9ef5660278c22ebdb88587e410dabf13b98711d9.

danielmitterdorfer:
@dliappis, @drawlerr thanks for your feedback. I've addressed it now. Can you please have another look?

data-params-hot.json
```diff
@@ -0,0 +1,9 @@
+{
```
dliappis: This file was added by accident?

data-params-warm.json
```diff
@@ -0,0 +1,9 @@
+{
```
dliappis: This file was added by accident?

master-params.json
```diff
@@ -0,0 +1,8 @@
+{
```
dliappis: This file was added by accident?

data-params-hot.json
danielmitterdorfer: Indeed! :)

danielmitterdorfer: removed in https://github.com/elastic/rally/pull/830/commits/947c11016ee26f62abea2d84f686bee9eaebc3d2

data-params-warm.json
danielmitterdorfer: removed in https://github.com/elastic/rally/pull/830/commits/947c11016ee26f62abea2d84f686bee9eaebc3d2

master-params.json
danielmitterdorfer: removed in https://github.com/elastic/rally/pull/830/commits/947c11016ee26f62abea2d84f686bee9eaebc3d2

danielmitterdorfer:
Thanks to both of you for the review! :)

PR #831, opened 2019-11-29
Expose API for cluster settings

danielmitterdorfer:
With this commit we introduce a new `put-settings` operation that can be
used to update cluster settings via the REST API. We also deprecate the
track property `cluster-settings` which had a similar purpose but the
cluster settings ended up in `elasticsearch.yml` instead of being
updated via an API. This is now tricky as we will move away from an
integrated cluster management (see also #830) and we should instead add
settings that need to be persistent in `elasticsearch.yml` via
`--car-params` and settings that are per track via the cluster settings
API.

Relates elastic/rally-tracks#93

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -4,6 +4,11 @@ Migration Guide
 Migrating to Rally 1.4.0
 ------------------------
 
+cluster-settings is deprecated in favor of the put-settings operation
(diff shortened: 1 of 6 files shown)


danielmitterdorfer:
Thanks for the review!

PR #832, opened 2019-11-29
Only keep the most recent build log

danielmitterdorfer:
With this commit we continously override a potentially existing
Elasticsearch build log. Keeping a history of build logs is not
necessary as it is only relevant to investigate on failures. Thus a
history only adds complexity (e.g. log rotation setup) for little value.

Relates #818


danielmitterdorfer:
Thanks for the review! :)

PR #836, opened 2019-12-04
Upgrade boto3

danielmitterdorfer:
With this commit we update the boto3 dependency to the latest released
version so Rally includes the latest changes there.


danielmitterdorfer:
Thanks for the quick review! :)

PR #840, opened 2019-12-16
Add make target to serve docs locally

danielmitterdorfer:
With this commit we add the make target `serve-docs` which starts a web
server locally so docs changes can be easily checked.


Makefile
```diff
@@ -92,6 +92,9 @@ lint: check-venv
 docs: check-venv
 	cd docs && $(MAKE) html
 
+serve-docs: docs
+	@cd docs/_build/html && python3 -m http.server
```
hub-cap: Since the `docs/Makefile` already has a `make serve`, it might make sense to do `        cd docs && $(MAKE) serve` instead.

hub-cap:
@elasticmachine run tests

Makefile
danielmitterdorfer: Good catch. Let's use that.

PR #841, opened 2019-12-16
Store Disk I/O metrics if available

danielmitterdorfer:
With this commit we split the check for disk I/O metrics for reads and
writes. This avoids the case when one of these metrics is either `None`
or `0`, the other metric is still added to the metrics store (this
scenario might happen if data is only written but never read so the read
counter is zero). We also change the check to test for `None` explicitly
because we even want to add metrics to the metrics store if the value is
zero.


danielmitterdorfer:
Thanks for the review! :)

Issue #496, opened 2018-05-11
Add task exclude filter


danielmitterdorfer:
Rally allows to include tasks with `--include-tasks` but has no exclude filter which would be handy to exclude only single operations but include the rest.

PR #844, opened 2019-12-16
Add task exclude filter

ebadyano:
Closes https://github.com/elastic/rally/issues/496

### DIFF
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -305,7 +305,7 @@ To specify different workloads in the same track you can use so-called challenge
 
 When should you use challenges? Challenges are useful when you want to run completely different workloads based on the same track but for the majority of cases you should get away without using challenges:
 
-* To run only a subset of the tasks, you can use :ref:`task filtering <clr_include_tasks>`, e.g. ``--include-tasks="create-index,bulk"`` will only run these two tasks in the track above or ``--exclude-tasks="bulk"`` will run all tasks except for `bulk`
+* To run only a subset of the tasks, you can use :ref:`task filtering <clr_include_tasks>`, e.g. ``--include-tasks="create-index,bulk"`` will only run these two tasks in the track above or ``--exclude-tasks="bulk"`` will run all tasks except for ``bulk``.
 * To vary parameters, e.g. the number of clients, you can use :ref:`track parameters <clr_track_params>`
 
 Structuring your track
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -305,7 +305,7 @@ To specify different workloads in the same track you can use so-called challenge
 
 When should you use challenges? Challenges are useful when you want to run completely different workloads based on the same track but for the majority of cases you should get away without using challenges:
 
-* To run only a subset of the tasks, you can use :ref:`task filtering <clr_include_tasks>`, e.g. ``--include-tasks="create-index,bulk"`` will only run these two tasks in the track above.
+* To run only a subset of the tasks, you can use :ref:`task filtering <clr_include_tasks>`, e.g. ``--include-tasks="create-index,bulk"`` will only run these two tasks in the track above or ``--exclude-tasks="bulk"`` will run all tasks except for `bulk`
 * To vary parameters, e.g. the number of clients, you can use :ref:`track parameters <clr_track_params>`
 
 Structuring your track
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -290,14 +290,14 @@ You can use ``--include-tasks`` to specify a comma-separated list of tasks that 
 ``exclude-tasks``
 ~~~~~~~~~~~~~~~~~
 
-Similarly to :ref:`include-tasks <clr_include_tasks>` when challenge consists of one or more tasks you might be interested in excluding a single operations but include the rest.
+Similarly to :ref:`include-tasks <clr_include_tasks>` when a challenge consists of one or more tasks you might be interested in excluding a single operations but include the rest.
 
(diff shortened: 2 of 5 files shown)


tests/track/loader_test.py
```diff
@@ -1357,6 +1357,84 @@ def test_filters_tasks(self):
...
+                }
+            ]
+        }
+        reader = loader.TrackSpecificationReader()
+        full_track = reader("unittest", track_specification, "/mappings")
+        self.assertEqual(4, len(full_track.challenges[0].schedule))
+
+        filtered = loader.filter_tasks(full_track, [track.TaskNameFilter("index-3")], exclude=True)
+
+        schedule = filtered.challenges[0].schedule
```
ebadyano: If we exclude one of the parallel tasks then nothing in that group runs.. should I correct the behaviour for that or does it make sense like that?

tests/track/loader_test.py
```diff
@@ -1257,21 +1257,21 @@ def test_sets_absolute_path(self, path_exists):
 
 class TrackFilterTests(TestCase):
```
danielmitterdorfer: Can you please change the test case method names? They sometimes reference the term "included" which is outdated now.

tests/track/loader_test.py
danielmitterdorfer: I think if we exclude a single task in a parallel element, only that single one should be excluded but all others should still be included.

ebadyano: Fixed.

danielmitterdorfer: I had another look and I meant something else than what is implemented at the moment. I have visualized the challenge `index-and-query-logs-fixed-daily-volume` from the eventdata track here which contains such a parallel element (I've simplified it to only simulate a single day):

```
1. measure-maximum-utilization (8 clients)
2. delete-measurement-index
3. 5 parallel tasks (12 clients):
	3.1 bulk-index-logs-100%-utilization (8 clients)
	3.2 current-kibana-traffic-country-dashboard_60m-querying-100%-utilization
	3.3 current-kibana-discover_30m-querying-100%-utilization
	3.4 current-kibana-traffic-dashboard_30m-querying-100%-utilization
	3.5 current-kibana-content_issues-dashboard_30m-querying-100%-utilization
```

If we specify `--exclude-tasks="type:search"`, I expect the following schedule:

```
1. measure-maximum-utilization (8 clients)
2. delete-measurement-index
3. 1 parallel task (8 clients):
	3.1 bulk-index-logs-100%-utilization (8 clients)
```

but instead we get:

```
1. measure-maximum-utilization (8 clients)
2. delete-measurement-index
3. 5 parallel tasks (12 clients):
	3.1 bulk-index-logs-100%-utilization (8 clients)
	3.2 current-kibana-traffic-country-dashboard_60m-querying-100%-utilization
	3.3 current-kibana-discover_30m-querying-100%-utilization
	3.4 current-kibana-traffic-dashboard_30m-querying-100%-utilization
	3.5 current-kibana-content_issues-dashboard_30m-querying-100%-utilization
```

Can you please have another look?

docs/command_line_reference.rst
```diff
@@ -287,6 +287,19 @@ You can use ``--include-tasks`` to specify a comma-separated list of tasks that
 * Execute only tasks of type ``search``: ``--include-tasks="type:search"``
 * You can also mix and match: ``--include-tasks="index,type:search"``
 
+``exclude-tasks``
+~~~~~~~~~~~~~~~~~
+
+Similarly to :ref:`include-tasks <clr_include_tasks>` when challenge consists of one or more tasks you might be interested in excluding a single operations but include the rest.
```
danielmitterdorfer: nit: when *a* challenge.

docs/adding_tracks.rst
```diff
@@ -305,7 +305,7 @@ To specify different workloads in the same track you can use so-called challenge
 
 When should you use challenges? Challenges are useful when you want to run completely different workloads based on the same track but for the majority of cases you should get away without using challenges:
 
-* To run only a subset of the tasks, you can use :ref:`task filtering <clr_include_tasks>`, e.g. ``--include-tasks="create-index,bulk"`` will only run these two tasks in the track above.
+* To run only a subset of the tasks, you can use :ref:`task filtering <clr_include_tasks>`, e.g. ``--include-tasks="create-index,bulk"`` will only run these two tasks in the track above or ``--exclude-tasks="bulk"`` will run all tasks except for `bulk`
```
danielmitterdorfer: Nit: missing period.

docs/command_line_reference.rst
```diff
@@ -287,6 +287,19 @@ You can use ``--include-tasks`` to specify a comma-separated list of tasks that
...
+
+You can use ``--exclude-tasks`` to specify a comma-separated list of tasks that you want to skip. Each item in the list defines either the name of a task or the operation type of a task. Only the tasks that did not match will be executed.
+
+**Examples**:
+
+* Do not execute any tasks with the name ``index`` and ``term``: ``--exclude-tasks="index,term"``
```
danielmitterdorfer: Suggestion: Use "Skip" instead of "Do not execute"?

tests/track/loader_test.py
danielmitterdorfer: fyi, I've created this output with a new `info` subcommand that I've added to Rally. It already includes support for your change but as it is not merged yet I've only pushed it as draft PR https://github.com/elastic/rally/pull/850.

danielmitterdorfer: Scratch that. As discussed offline, the operation type is `kibana` not `search`. So the filter is actually working fine. Sorry for the confusion. :)

docs/adding_tracks.rst
danielmitterdorfer: You need to use double-backticks so it is 
```
``bulk``
```

instead of

```
`bulk`
```

Issue #838, opened 2019-12-12
Fix linting issues reported by `make precommit`


dliappis:
Now that we got infrastructure to enforce coding conventions in #794 we need to start coming up with PRs to fix issues. This issue tracks the effort and should be only closed when `make precommit` succeeds with no errors reported.

 - [ ] W0707 (raise-missing-from) - 439
 - [ ] C4001 - 283
 - [ ] W0613 - 158
 - [x] C0330 - 98 @ebadyano #1164
 - [ ] W0621 - 95
 - [x] C0415 - 56 @danielmitterdorfer #1131
 - [x] W0612 - 36 @hub-cap #1076
 - [x] W0221 - 30 @hub-cap #1081 
 - [x] C0301 - 30 @hub-cap #855
 - [x] E1120 - 27 @hub-cap #1084 
 - [x] W0212 - 25 @hub-cap #1083 
 - [x] C0303 - 21 @hub-cap #854
 - [ ] C0302 - 12
 - [x] W0404 - 10 @hub-cap #1080
 - [x] W0106 - 9 @hub-cap #1082 
 - [x] C4002 - 9 @hub-cap #1079
 - [x] R1722 - 5 @hub-cap #1077
 - [x] R0904 - 5 @danielmitterdorfer #1127
 - [x] W0311 - 4 @ebadyano #1021
 - [x] W0107 - 4 @hub-cap #1073 
 - [x] W0105 - 4 @danielmitterdorfer #1039
 - [x] W0235 - 3 @hub-cap #1072
 - [x] W0104 - 3 @danielmitterdorfer #1039
 - [x] R1707 - 3 @danielmitterdorfer #1038
 - [x] R1704 - 3 @danielmitterdorfer #1010
 - [x] C0121 - 3 @ebadyano #1021
 - [x] E1128 (assignment-from-none) - 2 @danielmitterdorfer #1113
 - [x] W1201 - 2 @drawlerr #869
 - [x] W0631 - 2 @drawlerr #868 
 - [x] W0143 - 2 @drawlerr #867
 - [x] W0108 - 2 @hub-cap #849 
 - [x] W0102 - 2 @hub-cap #848 
 - [x] C0305 - 2 @hub-cap #847
 - [x] W1202 - 1  through e0601: @drawlerr #845 
 - [x] W0706 - 1
 - [x] W0702 - 1
 - [x] W0201 - 1
 - [x] W0127 - 1
 - [x] R1723 - 1
 - [x] R1714 - 1
[...]

PR #845, opened 2019-12-17
Cleanup Pylint singleton issues

drawlerr:
There are a number of 1-off pylint warnings that I've fixed and bundled in a single PR for simplicity.
Addressed:
 -W1202
 -W0706
 -W0702
 -W0127
 -R1723
 -R1714
 -R1706
 -R1702
 -E0911
 -E0601

### DIFF
--- a/esrally/actor.py
+++ b/esrally/actor.py
@@ -224,7 +224,7 @@ def bootstrap_actor_system(try_join=False, prefer_local_only=False, local_ip=Non
                 coordinator_ip = None
                 local_ip = None
         else:
-            if system_base != "multiprocTCPBase" and system_base != "multiprocUDPBase":
+            if system_base not in ('multiprocTCPBase', 'multiprocUDPBase'):
                 raise exceptions.SystemSetupError("Rally requires a network-capable system base but got [%s]." % system_base)
             if not coordinator_ip:
                 raise exceptions.SystemSetupError("coordinator IP is required")
--- a/esrally/actor.py
+++ b/esrally/actor.py
@@ -224,7 +224,7 @@ def bootstrap_actor_system(try_join=False, prefer_local_only=False, local_ip=Non
                 coordinator_ip = None
                 local_ip = None
         else:
-            if system_base not in ('multiprocTCPBase', 'multiprocUDPBase'):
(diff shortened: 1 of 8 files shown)


esrally/actor.py
```diff
@@ -224,7 +224,7 @@ def bootstrap_actor_system(try_join=False, prefer_local_only=False, local_ip=Non
                 coordinator_ip = None
                 local_ip = None
         else:
-            if system_base != "multiprocTCPBase" and system_base != "multiprocUDPBase":
+            if system_base not in ('multiprocTCPBase', 'multiprocUDPBase'):
```
danielmitterdorfer: Can we please keep the double quotes?

esrally/mechanic/supplier.py
```diff
@@ -144,7 +144,7 @@ def _supply_requirements(sources, distribution, build, plugins, revisions, distr
                 # * --pipeline=from-sources-skip-build --distribution-version=X.Y.Z where the plugin should not be built but ES should be
                 #   a distributed version.
                 # * --distribution-version=X.Y.Z --revision="my-plugin:abcdef" where the plugin should be built from sources.
-                plugin_needs_build = (sources and build) or distribution
+                plugin_needs_build = build if sources else distribution
```
danielmitterdorfer: Personally I find the previous styler easier to read and I vote for keeping this and ignoring pylint's suggestion.

esrally/track/track.py
```diff
@@ -450,63 +450,64 @@ def admin_op(self):
     @classmethod
     def from_hyphenated_string(cls, v):
```
danielmitterdorfer: I think `R0911` is unjustified for a method like this. I vote to ignore it here and keep the old style.

danielmitterdorfer:
@drawlerr can you please:

* Reference this PR in all issues mentioned in the checklist in #838 so we know what's fixed by which PR? You should do this already now before the PR is merged.
* Add a `Relates #838` in your merge commit?

esrally/actor.py
drawlerr: Oops, I was just taking the Pylint suggestion as-is.  Changing back to double quotes.

PR #847, opened 2019-12-17
Fix lint issue C0305

hub-cap:
Relates #838


hub-cap:
@elasticmachine run tests

PR #850, opened 2019-12-18
Allow to show track details

danielmitterdorfer:
With this commit we introduce a new `info` subcommand to Rally that
allows users to print details about a track on the command line. This
can be useful to see which challenges exist and which tasks are executed
in each of them.

### DIFF
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -202,6 +202,35 @@ The new track appears when you run ``esrally list tracks --track-path=~/rally-tr
     ----------  ----------------------------- -----------  ---------------  -----------------
     tutorial    Tutorial benchmark for Rally      11658903  N/A              1.4 GB
 
+You can also show details about your track with ``esrally info --track-path=~/rally-tracks/tutorial``::
+
+    dm@io:~ $ esrally info --track-path=~/rally-tracks/tutorial
+
+        ____        ____
+       / __ \____ _/ / /_  __
+      / /_/ / __ `/ / / / / /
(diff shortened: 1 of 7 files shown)


docs/adding_tracks.rst
```diff
@@ -202,6 +202,35 @@ The new track appears when you run ``esrally list tracks --track-path=~/rally-tr
     ----------  ----------------------------- -----------  ---------------  -----------------
     tutorial    Tutorial benchmark for Rally      11658903  N/A              1.4 GB
 
+You can also show details about your track with ``esrally info --track-path=~/rally-tracks/tutorial``::
+
+    dm@io:~ $ esrally info --track-path=~/rally-tracks/tutorial
```
ebadyano: should we remove `dm@io $` portion?

danielmitterdorfer: This was kind of intentional to make it clear that this is a prompt.

ebadyano:
Very cool feature, thank you! 

Do you think we want to add any tests? The feature itself is kind of like for testing, but maybe we want to add some to make sure some changes in the future don't break it?

danielmitterdorfer:
> Do you think we want to add any tests?

Thanks for the review. As discussed on another channel, let's add some integration tests for starters.

danielmitterdorfer:
@ebadyano I've added an integration test now in https://github.com/elastic/rally/pull/850/commits/2b0cb8e74e574bd31c009e10d500ad44005c168f. Can you please have another look?

danielmitterdorfer:
Thanks for the review! :)

PR #853, opened 2020-01-08
Allow to use the bundled JDK in Elasticsearch

danielmitterdorfer:
With this commit we add special handling for Elasticsearch versions that
bundle the JDK when the command line parameter `--runtime-jdk=bundled`
is set. In that case, Rally will use the bundled JDK (i.e. not set
`JAVA_HOME` when starting Elasticsearch).

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -510,6 +510,8 @@ Example::
    # Force to run with JDK 7
    esrally --distribution-version=2.4.0 --runtime-jdk=7
 
+There is also limited support to specify the JDK that is bundled with Elasticsearch with the special value ``bundled``. At the moment this only works on Linux. The JDK is bundled from Elasticsearch 7.0.0 onwards.
+
 .. _clr_revision:
 
 ``revision``
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -510,7 +510,7 @@ Example::
    # Force to run with JDK 7
    esrally --distribution-version=2.4.0 --runtime-jdk=7
 
-There is also limited support to specify the JDK that is bundled with Elasticsearch with the special value ``bundled``. At the moment this only works on Linux. The JDK is bundled from Elasticsearch 7.0.0 onwards.
+There is also limited support to specify the JDK that is bundled with Elasticsearch with the special value ``bundled``. At the moment this only works on Linux. The `JDK is bundled from Elasticsearch 7.0.0 onwards <https://www.elastic.co/guide/en/elasticsearch/reference/current/release-highlights-7.0.0.html#_bundle_jdk_in_elasticsearch_distribution>`_.
 
 .. _clr_revision:
 
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -362,6 +362,21 @@ Example::
 
    esrally --team-path=~/Projects/es-teams
 
+``target-os``
+~~~~~~~~~~~~~
+
+Specifies the name of the target operating system for which an artifact should be downloaded. By default this value is automatically derived based on the operating system Rally is run. This command line flag is only applicable to the ``download`` subcommand and allows to download an artifact for a different operating system. Example::
+
+    esrally download --distribution-version=7.5.1 --target-os=linux
+
+``target-arch``
+~~~~~~~~~~~~~~~
+
+Specifies the name of the target CPU architecture for which an artifact should be downloaded. By default this value is automatically derived based on the CPU architecture Rally is run. This command line flag is only applicable to the ``download`` subcommand and allows to download an artifact for a different CPU architecture. Example::
+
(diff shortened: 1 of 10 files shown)


docs/command_line_reference.rst
```diff
@@ -510,6 +510,8 @@ Example::
    # Force to run with JDK 7
    esrally --distribution-version=2.4.0 --runtime-jdk=7
 
+There is also limited support to specify the JDK that is bundled with Elasticsearch with the special value ``bundled``. At the moment this only works on Linux. The JDK is bundled from Elasticsearch 7.0.0 onwards.
```
dliappis: Does it make sense to link: https://www.elastic.co/blog/elasticsearch-7-0-0-released re: `The JDK is bundled from Elasticsearch 7.0.0 onwards.`?

tests/mechanic/java_resolver_test.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_resolves_java_home_for_bundled_jdk(self):
+
+        cfg = config.Config()
+        cfg.add(config.Scope.application, "mechanic", "runtime.jdk", "bundled")
+
+        major, java_home = java_resolver.java_home("12,11,10,9,8", cfg)
+
+        # assumes most recent JDK
+        self.assertEqual(major, 12)
+        # does not extract JAVA_HOME for the bundled JDK
```
dliappis: nit: does not **set** JAVA_HOME?

docs/command_line_reference.rst
danielmitterdorfer: I'd consider a reference documentation more stable than a blog and would thus rather avoid pointing to a blog post?

tests/mechanic/java_resolver_test.py
danielmitterdorfer: Good point. I'll push a change

docs/command_line_reference.rst
dliappis: Maybe this one then? https://www.elastic.co/guide/en/elasticsearch/reference/current/release-highlights-7.0.0.html#_bundle_jdk_in_elasticsearch_distribution

danielmitterdorfer: Addessed in https://github.com/elastic/rally/pull/853/commits/66624207845b8b00a0de4de4c1ebc8ccfd03ddbc

dliappis: Thanks!

danielmitterdorfer:
Together with https://github.com/elastic/rally-teams/pull/38, Rally can now:

* Use system specific artifacts (automatically)
* Users can override the default behavior for the `download` subcommand with the new command line flags `--target-os` and `--target-arch`.

@dliappis can you please have another look?

As we already rely on new properties that will be introduced in https://github.com/elastic/rally-teams/pull/38, that PR needs to be merged first.

esrally/mechanic/supplier.py
```diff
@@ -174,6 +184,30 @@ def _src_dir(cfg, mandatory=True):
...
+        else:
+            self.os = sysstats.os_name().lower()
+        if arch is not None:
+            self.arch = arch
+        else:
+            self.arch = sysstats.cpu_arch().lower()
+
+    def render(self, template):
+        substitutions = {
+            "{{VERSION}}": self.version,
```
dliappis: Is there a reason -- apart from simplicity -- we can't use j2 template rendering to substitute those (and at the same time be more lenient with spaces like `{{ VERSION }}` and allow filters)? Many other files in rally-teams are treated as j2 hence the question.

esrally/rally.py
```diff
@@ -859,6 +870,9 @@ def main():
             console.println("--target-hosts and --client-options must define the same keys for multi cluster setups.")
             exit(1)
     # split by component?
+    if sub_command == "download":
```
dliappis: Don't we need to support `--target-arch` and `--target-os` also for the the `install` subcommand?

danielmitterdorfer:
@elasticmachine test this

danielmitterdorfer:

esrally/rally.py
danielmitterdorfer: I would like to restrict this to the `download` subcommand for the moment because we should not support installing an artifact for a different platform IMHO as the successive startup is likely to fail in some scenarios and I also don't see why one would use this approach.

esrally/mechanic/supplier.py
danielmitterdorfer: For other templates (like `elasticsearch.yml` or `jvm.options`) I think Jinja makes sense because their purpose is to serve as templates. `config.ini` is a high-level file meant to control how Rally behaves and I would not expect much templating happening here. As discussed elsewhere, I'll raise a separate enhancement issue.

danielmitterdorfer: I've raised https://github.com/elastic/rally/issues/856.

PR #857, opened 2020-01-09
Disable all pylint checks that we want to fix

hub-cap:
This commit will allow us to make progress on the pylint issues we would
like to fix, while allowing users to commit new code that may fail some
of these. What ends up happening is that new code can be merged by other
users, and if precommit is not run in CI, we might accidentally allow a
pylint violation to sneak back in. Removing the pylint check in the
branch/PR that is removing these will allow CI to fail the check so the
user can get feedback if another person has merged code in that violates
the check in the branch/PR.

Relates #838

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -165,7 +165,39 @@ disable=print-statement,
         no-member,
         too-many-instance-attributes,
         too-many-statements,
-        inconsistent-return-statements
+        inconsistent-return-statements,
+        C0121,
+        C0301,
+        C0302,
+        C0303,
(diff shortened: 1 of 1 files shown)


hub-cap:
Hi, I know its preferred to use the new labels instead of the obscure CXXXX labels, but this is the quickest way to match up what a user might be fixing against #838. New PRs that fix a #838 pylint warning should also remove that from the pylint file so that CI can catch it.

I will be updating CI once this is merged to run precommit, which passes on this branch.

PR #860, opened 2020-01-10
More resilient node shutdown

danielmitterdorfer:
With this commit we make the shutdown procedure more resilient by
handling the case that the Elasticsearch process has already terminated
and moving steps during shutdown with the `stop` subcommand so related
errors happen in related places in the code and also as early as
possible.

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -206,6 +206,7 @@ class ProcessLauncher:
             self.logger.info("Keeping [%d] nodes on this host running.", len(nodes))
         else:
             self.logger.info("Shutting down [%d] nodes on this host.", len(nodes))
+        stopped_nodes = []
         for node in nodes:
             node_name = node.node_name
             telemetry.add_metadata_for_node(metrics_store, node_name, node.host_name)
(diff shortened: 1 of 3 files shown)


esrally/mechanic/launcher.py
```diff
@@ -208,26 +208,33 @@ def stop(self, nodes, metrics_store):
...
-                    self.logger.warning("No process found with PID [%s] for node [%s]", proc.pid, node_name)
-                except psutil.TimeoutExpired:
-                    self.logger.info("kill -KILL node [%s]", node_name)
+                if es:
+                    stop_watch = self._clock.stop_watch()
+                    stop_watch.start()
                     try:
-                        # kill -9
-                        proc.kill()
+                        os.kill(es.pid, signal.SIGTERM)
```
dliappis: Doesn't `es.terminate()` kill via SIGTERM? ref: https://psutil.readthedocs.io/en/latest/#psutil.Process.terminate

danielmitterdorfer: I was only moving this code but have not touched it otherwise. However, I agree this change makes sense. I'll push this in a separate commit.

danielmitterdorfer: I've pushed https://github.com/elastic/rally/pull/860/commits/32100f528aac4dc20a88f53a4ca568053d8e4555

danielmitterdorfer:
@dliappis can you please have another look?

PR #864, opened 2020-01-10
Remove test step from precommit

hub-cap:


hub-cap:
will do next week. got sucked into a local installation rabbit hole today :)

Issue #835, opened 2019-12-03
Update force-merge operation to only force-merge track specific indices


rohitnair:
It looks like the existing force-merge operation by default performs a force-merge on all indices by passing "_all" as the index parameter. It might be better to scope this to only the indices created by the current rally track, or allow the user to specify a specific index to force-merge, similar to how other operations like delete-index currently work. 

It seems like it should be a fairly minor change, and I can submit a pull request for this if this sounds good.

danielmitterdorfer:
Agreed, the behavior you're describing makes sense. Thanks for offering to work on a PR. Just some hints for the PR: To make this work, you'd need to add a new "parameter source" for force merge that derives the indices by itself. You can look at `SearchParamSource` for an example:

https://github.com/elastic/rally/blob/47e3d025a4c2ddffdfd52c3c8e039eab0c073621/esrally/track/params.py#L345-L350

The corresponding parameter source needs to be registered so it can be used for the force-merge operations, see here:

https://github.com/elastic/rally/blob/47e3d025a4c2ddffdfd52c3c8e039eab0c073621/esrally/track/params.py#L918-L924

Finally, the force merge runner needs to use the `index` parameter from the provided `params` dictionary:

https://github.com/elastic/rally/blob/47e3d025a4c2ddffdfd52c3c8e039eab0c073621/esrally/driver/runner.py#L570-L581

rohitnair:
Thanks for the hints. I've created a pull request for this - https://github.com/elastic/rally/pull/865

PR #865, opened 2020-01-10
Updating force merge operation to force merge only track indices by default

rohitnair:
This change fixes https://github.com/elastic/rally/issues/835 

I have run the tests and ensured they pass. Also tested locally to ensure it does what we expect it to do.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -344,7 +344,6 @@ force-merge
 With the operation type ``force-merge`` you can call the `force merge API <http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html>`_. On older versions of Elasticsearch (prior to 2.1), Rally will use the ``optimize API`` instead. It supports the following parameters:
 
 * ``index`` (optional, defaults to the indices defined in the ``indices`` section or ``_all`` if no indices are defined there): The name of the index that should be force-merged.
-
 * ``max-num-segments`` (optional)  The number of segments the index should be merged into. Defaults to simply checking if a merge needs to execute, and if so, executes it.
 
 This is an administrative operation. Metrics are not reported by default. If reporting is forced by setting ``include-in-reporting`` to ``true``, then throughput is reported as the number of completed force-merge operations per second.
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -341,7 +341,9 @@ Throughput will be reported as number of indexed documents per second.
 force-merge
 ~~~~~~~~~~~
 
-With the operation type ``force-merge`` you can call the `force merge API <http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html>`_. On older versions of Elasticsearch (prior to 2.1), Rally will use the ``optimize API`` instead. It supports the following parameter:
+With the operation type ``force-merge`` you can call the `force merge API <http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html>`_. On older versions of Elasticsearch (prior to 2.1), Rally will use the ``optimize API`` instead. It supports the following parameters:
+
+* ``index`` (optional, defaults to the indices defined in the ``indices`` section or ``_all`` if no indices are defined there): The name of the index that should be force-merged.
 
 * ``max-num-segments`` (optional)  The number of segments the index should be merged into. Defaults to simply checking if a merge needs to execute, and if so, executes it.
 
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -580,7 +580,7 @@ class PartitionBulkIndexParamSource:
         return self.current_bulk / self.total_bulks
 
 
-class ForceMergeParamSouce:
+class ForceMergeParamSouce(ParamSource):
     def __init__(self, track, params, **kwargs):
         if len(track.indices) > 0:
             default_index = ','.join(map(str, track.indices))
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -580,18 +580,23 @@ class PartitionBulkIndexParamSource:
         return self.current_bulk / self.total_bulks
 
 
-class ForceMergeParamSouce(ParamSource):
+class ForceMergeParamSource(ParamSource):
     def __init__(self, track, params, **kwargs):
+        super().__init__(track, params, **kwargs)
         if len(track.indices) > 0:
             default_index = ','.join(map(str, track.indices))
         else:
             default_index = "_all"
 
         self._index_name = params.get("index", default_index)
+        self._max_num_segments = params.get("max-num-segments")
+        self._request_timeout = params.get("request-timeout")
 
     def params(self):
         return {
-            "index": self._index_name
+            "index": self._index_name,
+            "max-num-segments": self._max_num_segments,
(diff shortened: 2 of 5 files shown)


esrally/track/params.py
```diff
@@ -580,6 +580,21 @@ def percent_completed(self):
         return self.current_bulk / self.total_bulks
 
 
+class ForceMergeParamSouce:
```
danielmitterdorfer: Can you please document the new parameter in https://github.com/elastic/rally/blob/master/docs/track.rst#force-merge?

IMHO it should be the first parameter in the list and the docs should say something like:

```
* ``index`` (optional, defaults to the indices defined in the ``indices`` section or ``_all`` if no indices are defined there): The name of the index that should be force-merged.
```

esrally/track/params.py
```diff
@@ -580,6 +580,21 @@ def percent_completed(self):
...
+    def __init__(self, track, params, **kwargs):
```
danielmitterdorfer: Can you please add tests in `tests/track/params_test.py`? You can have a look at `DeleteIndexParamSourceTests` for inspiration. IMHO it should check all three cases how the  `index` parameter can be set (via the track, fallback via `_all` and explicitly set via the `index` parameter on the track).

danielmitterdorfer:

esrally/track/params.py
danielmitterdorfer: Furthermore, can you please derive this class from the class `ParamSource`? All Rally-internal parameter sources are derived from it and it provides a default implementation for required methods. Currently, the following invocation:

```
esrally --test-mode --distribution-version=7.4.2 --challenge=append-no-conflicts-index-only --on-error=abort
```

is failing with:

```
[ERROR] Cannot race. Error in load generator ('ForceMergeParamSouce' object has no attribute 'partition')
```

This can be solved by deriving your class from `ParamSource` as I've outlined above.

rohitnair: Updated the docs.

esrally/track/params.py
rohitnair: Thanks for the pointer, added tests to cover the three cases as you mentioned.

esrally/track/params.py
rohitnair: Done.

rohitnair:
Thanks for taking a look @danielmitterdorfer . I've addressed your comments and pushed an update. Let me know if you have any other feedback.

danielmitterdorfer:

danielmitterdorfer:
Thanks for iterating! I'm gonna have another look soon. Meanwhile I've triggered a fresh CI build for your changes.

esrally/track/params.py
danielmitterdorfer: `make precommit` is failing now with:

```

08:13:11 + make precommit
08:13:19 ************* Module esrally.track.params
08:13:19 esrally/track/params.py:584:4: W0231: __init__ method from base class 'ParamSource' is not called (super-init-not-called)
```

can you please add the following in the first line of the constructor:

```python
super().__init__(track, params, **kwargs)
```

docs/track.rst
```diff
@@ -341,7 +341,9 @@ Throughput will be reported as number of indexed documents per second.
 force-merge
 ~~~~~~~~~~~
 
-With the operation type ``force-merge`` you can call the `force merge API <http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html>`_. On older versions of Elasticsearch (prior to 2.1), Rally will use the ``optimize API`` instead. It supports the following parameter:
+With the operation type ``force-merge`` you can call the `force merge API <http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html>`_. On older versions of Elasticsearch (prior to 2.1), Rally will use the ``optimize API`` instead. It supports the following parameters:
+
```
danielmitterdorfer: Nit: Remove this empty line.

tests/track/params_test.py
```diff
@@ -1687,3 +1687,32 @@ def test_replaces_body_params(self):
...
+        self.assertEqual("index2", p["index"])
+
+    def test_default_force_merge_index(self):
+        source = params.ForceMergeParamSouce(track.Track(name="unit-test"), params={})
+
+        p = source.params()
+
+        self.assertEqual("_all", p["index"])
+
+
```
danielmitterdorfer: `make precommit` is failing with:

```
08:13:19 ************* Module tests.track.params_test
08:13:19 tests/track/params_test.py:1718:0: C0305: Trailing newlines (trailing-newlines)
```

This is because you added two trailing lines but only one is allowed. Can you please remove this last line?

esrally/track/params.py
```diff
@@ -580,7 +580,7 @@ def percent_completed(self):
         return self.current_bulk / self.total_bulks
 
 
-class ForceMergeParamSouce:
+class ForceMergeParamSouce(ParamSource):
```
danielmitterdorfer: I noticed there is a typo: It says `ForceMergeParamSouce` but it should be `ForceMergeParamSource` (the "r" is missing in the word "source").

esrally/track/params.py
```diff
@@ -580,6 +580,21 @@ def percent_completed(self):
...
 
 
+class ForceMergeParamSouce(ParamSource):
+    def __init__(self, track, params, **kwargs):
+        if len(track.indices) > 0:
+            default_index = ','.join(map(str, track.indices))
+        else:
+            default_index = "_all"
+
+        self._index_name = params.get("index", default_index)
```
danielmitterdorfer: I noticed that we'd lose support for the other two parameters `max-num-segments` and `request-timeout` (see the `ForceMerge` runner class). Could you please add the following lines here in the constructor?

```python
        self._max_num_segments = params.get("max-num-segments")
        self._request_timeout = params.get("request-timeout")
```

and the `params` method needs to change to:

```python
    def params(self):
        return {
            "index": self._index_name,
            "max-num-segments": self._max_num_segments,
            "request-timeout": self._request_timeout
        }
```

you might wonder why this needs to be changed now. Previously there was a generic parameter source assigned  for the `ForceMerge` runner which would pass any parameters that are defined on the operation but now that we have a custom parameter source we need to be explicit about it.

elasticcla:
Hi @rohitnair, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/rally/pull/865.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

docs/track.rst
rohitnair: Done.

tests/track/params_test.py
rohitnair: Fixed this, and made sure make precommit passes.

esrally/track/params.py
rohitnair: 🤦‍♂ sorry about that.

esrally/track/params.py
rohitnair: Thanks for explaining, makes sense. I've updated it as per above, and also added a test that validates all the params.

danielmitterdorfer:

PR #867, opened 2020-01-13
Fix W0143

drawlerr:
Open false positive, see https://github.com/PyCQA/pylint/issues/2306
Using workaround.

Relates #838.


danielmitterdorfer:
@drawlerr is it possible to selectively disable pylint for these statements instead of implementing a workaround?

drawlerr:
> @drawlerr is it possible to selectively disable pylint for these statements instead of implementing a workaround?
I believe so.  Since the bug may be fixed at some point in the future, I suppose the annotation is a bit more clear and easily removed when that happens.  I'll change it.

PR #868, opened 2020-01-14
Fix W0631

drawlerr:
It would appear that a refactoring operation resulted in some methods referencing variables from their calling function that were not passed as arguments.  Python allows this, but for sanity's sake it's best if function arguments are a bit more explicit.

Relates #838.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -194,7 +194,6 @@ disable=print-statement,
         W0612,
         W0613,
         W0621,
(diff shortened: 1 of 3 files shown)


hub-cap:
Dear github, why dont you let me approve this pull request? What did I do to u? Something I have done is not allowing me to LGTM this in a review, so LGTM in a comment.

Issue #597, opened 2018-11-09
Improve error message when no index has been specified in search operation with multiple indices


dliappis:
As discussed in [this discuss topic](https://discuss.elastic.co/t/error-cannot-race-error-in-track-preparator-index-is-mandatory-none/156013), when having multiple indices in the track, if user omits `index` in the `search` operation, the following error message gets produced:

`[ERROR] Cannot race. Error in track preparator (("'index' is mandatory", None))`

This doesn't help identify which operation required the index property and needs to be improved.

emilmirzayev:
I guess, this was the [discussion topic](https://discuss.elastic.co/t/error-cannot-race-error-in-track-preparator-index-is-mandatory-none/156013) you meant. It was my question.

dliappis:
Ugh thanks @emilmirzayev, I updated the link.

PR #872, opened 2020-01-15
Extend error msg to specify operation name for missing index

ebadyano:
Closes https://github.com/elastic/rally/issues/597

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -314,7 +314,7 @@ def operation_parameters(t, op):
     if op.param_source:
         return params.param_source_for_name(op.param_source, t, op.params)
     else:
-        return params.param_source_for_operation(op.type, t, op.params)
+        return params.param_source_for_operation(op.type, t, op.params, op.name)
 
 
 def used_corpora(t, cfg):
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -310,11 +310,12 @@ class SimpleTrackRepository:
         return self._track_file
 
 
-def operation_parameters(t, op):
+def operation_parameters(t, task):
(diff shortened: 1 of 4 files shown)


esrally/track/params.py
```diff
@@ -365,7 +365,7 @@ def __init__(self, track, params, **kwargs):
         }
 
         if not index_name:
-            raise exceptions.InvalidSyntax("'index' is mandatory")
+            raise exceptions.InvalidSyntax("'index' is mandatory and is missing for operation '%s'" % kwargs.get("operation_name"))
```
danielmitterdorfer: As per convention we should use the `format` method instead of % formatting.

esrally/track/loader.py
```diff
@@ -314,7 +314,7 @@ def operation_parameters(t, op):
     if op.param_source:
         return params.param_source_for_name(op.param_source, t, op.params)
     else:
-        return params.param_source_for_operation(op.type, t, op.params)
+        return params.param_source_for_operation(op.type, t, op.params, op.name)
```
danielmitterdorfer: Each operation will have a name but if we specify operations inline with a task, it's the operation type's name. How about we instead use the task's name here which is always recognizable by the user (see attached [patch](https://github.com/elastic/rally/files/4070975/task_name.patch.txt) for a sketch).

danielmitterdorfer:
@ebadyano seems `make precommit` has failed on the PR build:

```

15:01:15 + make precommit
15:01:23 ************* Module tests.track.params_test
15:01:23 tests/track/params_test.py:1606:0: C0303: Trailing whitespace (trailing-whitespace)
```

Can you please have a look?

PR #873, opened 2020-01-16
Prevent installation with unsupported Python

danielmitterdorfer:
With this commit we specify `python_requires` in our setup file which
generates the necessary metadata so that pip checks whether the user's
Python version is within the allowed range and aborts the installation
if the requirement is not met. This feature requires pip 9.0.0 or
better.


danielmitterdorfer:
> an alternative to the 999 patchset version would be >=3.5, <3.9 I believe this would work, but it does not matter much to me.

Indeed it does but we also generate the (human-readable) list of supported versions from that list and that's the main reason we base `python_requires` on an inclusive range.

danielmitterdorfer:
@hub-cap I've implemented a different approach now in https://github.com/elastic/rally/pull/873/commits/bb4027c88ea1b3834dc8dbf5fb73173cdf636fd8 which defines the version range as `>=3.5,<3.9`. Can you please take another look?

PR #874, opened 2020-01-17
Store system metrics if race metadata are present

danielmitterdorfer:
With this commit we avoid interacting in any way with the metrics store
if we could not retrieve any race metadata. This situation can happen
when we have never started a benchmark, the provided race id does not
match or we use an in-memory metrics store instead of Elasticsearch. In
all these cases we cleanly shutdown the node instead of raising errors.

Relates #860 
Relates #830


danielmitterdorfer:
Thanks for your review!

PR #879, opened 2020-01-27
Don't read from stdout/stderr on ES startup

danielmitterdorfer:
With this commit we avoid reading from stdout/stderr which might have
been closed by the application and rely solely on the process' return
code.

Relates elastic/elasticsearch#50259

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -16,6 +16,8 @@
 # under the License.
 import logging
 import os
+import shlex
+import subprocess
 
 import psutil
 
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:

esrally/mechanic/launcher.py
```diff
@@ -186,14 +188,28 @@ def _set_env(self, env, k, v, separator=' ', prepend=False):
             else:
                     env[k] = env[k] + separator + v
 
+    @staticmethod
+    def _run_subprocess(command_line, env):
+        command_line_args = shlex.split(command_line)
+
+        # pylint: disable=subprocess-popen-preexec-fn
```
drawlerr: I know we didn't comment this last time when this was added, but when going back over this code I completely forgot about why we have a preexec_fn.  Seems an uncommonly used function that is prone to trouble as in the Pylint warning.  Perhaps we can briefly comment why it is necessary for detached processes?

drawlerr: Alternatively, we can "upgrade" to using the `start_new_session` argument described in https://docs.python.org/3.2/library/subprocess.html

danielmitterdorfer: Thanks for the feedback! As this is currently blocking our nightly benchmarks I'll merge this PR as is but this is definitely a good idea for a for a future PR. I'll keep this on my radar! :)

danielmitterdorfer: I've pushed https://github.com/elastic/rally/pull/884.

PR #882, opened 2020-01-29
Make REST API check stricter

danielmitterdorfer:
So far we have used the info API to determine whether the REST API of
Elasticsearch is available. However, we might get lucky that a quorum
(but not all) of the target hosts are available yet. While certain nodes
would then respond to HTTP requests, others might not which can lead to
situations where the REST API check succeeds but we run into connection
errors later on (because we hit a different host from the connection
pool).

With this commit we make this check stricter by using the cluster health
API and blocking until  at least the number of target hosts in the
cluster is available.

### DIFF
--- a/esrally/client.py
+++ b/esrally/client.py
@@ -134,23 +134,34 @@ def wait_for_rest_layer(es, max_attempts=20):
 
     :param es: Elasticsearch client to use for connecting.
     :param max_attempts: The maximum number of attempts to check whether the REST API is available.
-    :return: True iff Elasticsearch is available.
+    :return: True iff Elasticsearch's REST API is available.
     """
+    # assume that at least the hosts that we expect to contact should be available. Note that this is not 100%
+    # bullet-proof as a cluster could have e.g. dedicated masters which are not contained in our list of target hosts
+    # but this is still better than just checking for any random node's REST API being reachable.
+    expected_node_count = len(es.transport.hosts)
+    logger = logging.getLogger(__name__)
     for attempt in range(max_attempts):
+        logger.debug("REST API is available after %s attempts", attempt)
         import elasticsearch
         try:
-            es.info()
+            # see also WaitForHttpResource in Elasticsearch tests. Contrary to the ES tests we consider the API also
+            # available when the cluster status is RED (as long as all required nodes are present)
+            es.cluster.health(wait_for_nodes=">={}".format(expected_node_count))
+            logger.info("REST API is available for >= [%s] nodes after [%s] attempts.", expected_node_count, attempt)
             return True
         except elasticsearch.ConnectionError as e:
             if "SSL: UNKNOWN_PROTOCOL" in str(e):
                 raise exceptions.SystemSetupError("Could not connect to cluster via https. Is this an https endpoint?", e)
             else:
-                time.sleep(1)
+                logger.debug("Got connection error on attempt [%s]. Sleeping...", attempt)
+                time.sleep(3)
         except elasticsearch.TransportError as e:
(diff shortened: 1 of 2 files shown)


esrally/client.py
```diff
@@ -134,23 +134,34 @@ def wait_for_rest_layer(es, max_attempts=20):
 
     :param es: Elasticsearch client to use for connecting.
     :param max_attempts: The maximum number of attempts to check whether the REST API is available.
-    :return: True iff Elasticsearch is available.
+    :return: True iff Elasticsearch's REST API is available.
     """
+    # assume that at least the hosts that we expect to contact should be available. Note that this is not 100%
+    # bullet-proof as a cluster could have e.g. dedicated masters which are not contained in our list of target hosts
+    # but this is still better than just checking for any random node's REST API being reachable.
```
dliappis: This is a very useful comment to have.

Question: would it be dangerous to trigger a sniff of the eligible http hosts (e.g. via a helper method in our EsClientFactory invoking [#elasticsearch.Transport#sniff_hosts()](https://elasticsearch-py.readthedocs.io/en/master/connection.html#elasticsearch.Transport.sniff_hosts))? I was thinking if we explicitly ask for a fresh list of hosts before the check, then no unavailable hosts should be reachable. Then the same call could be invoked before the load driver starts. The caveat with this approach would be that it could potentially override the explicitly list provided by `--target-hosts`. Thoughts?

esrally/client.py
```diff
@@ -134,23 +134,34 @@ def wait_for_rest_layer(es, max_attempts=20):
...
+                time.sleep(3)
         except elasticsearch.TransportError as e:
-            if e.status_code == 503:
-                time.sleep(1)
-            elif e.status_code == 401:
-                time.sleep(1)
+            # cluster block, x-pack not initialized yet, our wait condition is not reached
+            if e.status_code in (503, 401, 408):
+                logger.debug("Got status code [%s] on attempt [%s]. Sleeping...", e.status_code, attempt)
+                time.sleep(3)
```
dliappis: With the default `max_attempts=20` the worst case scenario would mean waiting 3*20=60s plus whatever time spent executing the 20 API calls. Given the (potential) difference in performance between different hosts building from source, should we increase this e.g. to 6 (2mins in total)?

danielmitterdorfer: I'm fine increasing this although I'd opt for more retries instead of a larger sleep period.

dliappis: More retries is fine by me too.

esrally/client.py
danielmitterdorfer: I'd try not to build any smartness into this? Without reading all of the involved code I don't think we can reason what nodes will be returned on cluster bootstrap (let's assume not all nodes are up yet or not all of them might have opened the HTTP port). I was even considering exposing an explicit command line parameter but thought that this would be a good compromise.

esrally/client.py
danielmitterdorfer: I've increased the number of retries to 40 now in https://github.com/elastic/rally/pull/882/commits/cbf6decc1dd68f2a5ff8da3d32e18e9279ce1961.

esrally/client.py
dliappis: Agreed. Let's keep things simple here.

For the record I had a look what the elasticsearch py client does when sniff gets invoked [here](https://github.com/elastic/elasticsearch-py/blob/a9cdb247001c40e94d916f58c14f87e4b281867c/elasticsearch/transport.py#L198-L238) and it collects a list of eligible http hosts via `/_nodes/_all/http`.

PR #883, opened 2020-01-29
Stop all locally running nodes

danielmitterdorfer:
When starting multiple nodes on one machine, Rally only stopped the
first node. With this commit we actually stop all nodes.

### DIFF
--- a/tests/mechanic/launcher_test.py
+++ b/tests/mechanic/launcher_test.py
@@ -174,9 +174,11 @@ class ProcessLauncherTests(TestCase):
         ms = get_metrics_store(cfg)
         proc_launcher = launcher.ProcessLauncher(cfg)
 
-        node_configs = [NodeConfiguration(build_type="tar", car_env={}, car_runtime_jdks="12,11", ip="127.0.0.1",
-                                          node_name="testnode-{}".format(n), node_root_path="/tmp", binary_path="/tmp",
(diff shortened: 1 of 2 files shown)


tests/mechanic/launcher_test.py
```diff
@@ -174,12 +174,12 @@ def test_daemon_start_stop(self, wait_for_pidfile, chdir, get_size, supports, ja
         ms = get_metrics_store(cfg)
         proc_launcher = launcher.ProcessLauncher(cfg)
 
-        node_config = NodeConfiguration(build_type="tar", car_env={}, car_runtime_jdks="12,11", ip="127.0.0.1",
-                                        node_name="testnode", node_root_path="/tmp", binary_path="/tmp",
-                                        data_paths="/tmp")
+        node_configs = [NodeConfiguration(build_type="tar", car_env={}, car_runtime_jdks="12,11", ip="127.0.0.1",
+                                          node_name="testnode-{}".format(n), node_root_path="/tmp", binary_path="/tmp",
+                                          data_paths="/tmp") for n in range(2)]
```
drawlerr: I think it would improve readability if we used a regular for-loop here instead of a list comprehension.

danielmitterdorfer: Thanks! I changed it in https://github.com/elastic/rally/pull/883/commits/51c0df6046f26167f4664b71f76b85832ae70615.

PR #884, opened 2020-01-29
Detach ES using Python standard mechanism

danielmitterdorfer:
With this commit we use the parameter `start_new_session` to detach
Elasticsearch from its parent process group (Rally) instead of the
problematic (and low-level) `preexec_fn`. Under the hood this will use
the `setsid` system call instead of `setpgrp`.

See also:

* https://linux.die.net/man/2/setsid
* https://linux.die.net/man/2/setpgrp

Relates #859

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -192,12 +192,11 @@ class ProcessLauncher:
     def _run_subprocess(command_line, env):
         command_line_args = shlex.split(command_line)
 
-        # pylint: disable=subprocess-popen-preexec-fn
         with subprocess.Popen(command_line_args,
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
The CI build has failed with:

```
16:08:53 2020-01-29 16:08:49,724 ActorAddr-(T|:35637)/PID:13692 esrally.actor ERROR Load generator [3] has exited prematurely. Aborting benchmark.
```

(but only for the Python 3.5 build). Unfortunately we don't have more information. This seems unrelated to this change here though and it seems transient. I'll trigger another build.

danielmitterdorfer:

danielmitterdorfer:
Thanks for the review! :)

PR #885, opened 2020-01-31
Fix issue with fetching test-mode -1k document corpora.

drawlerr:
In test-mode, we download a document corpus with "-1k" added to the filename, that is not specifically described in `track.json`.  This is problematic because, the "expected bytes" of the file is None in this situation, breaking download progress reporting when we divide by None.
To work around this, I've made the net.Progress callback resilient to this situation and added a test.

### DIFF
--- a/tests/utils/net_test.py
+++ b/tests/utils/net_test.py
@@ -14,7 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-
 import random
 import unittest.mock as mock
 from unittest import TestCase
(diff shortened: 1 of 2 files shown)


tests/utils/net_test.py
```diff
@@ -33,3 +32,15 @@ def test_download_from_s3_bucket(self, download):
...
+        progress = net.Progress("test")
+        mock_progress = mock.Mock()
+        progress.p = mock_progress
+        progress(42, 100)
+        assert mock_progress.print.called
+        mock_progress.reset_mock()
+        progress(42, None)
+        assert mock_progress.print.called
+
+
```
danielmitterdorfer: `make precommit` has failed with:

```

03:59:48 ************* Module tests.utils.net_test
03:59:48 tests/utils/net_test.py:46:0: C0305: Trailing newlines (trailing-newlines)
```

can you please have a look?

drawlerr: https://github.com/elastic/rally/pull/885/commits/5a64bfe1e131336b8782dc55ac1e0b5baa145521

PR #890, opened 2020-02-11
Speed up client-side bulk-handling

danielmitterdorfer:
With this commit we speed up preparing bulk requests from data files by
implementing several optimizations:

* Bulk data are passed as a string instead of a list to the runner. This
avoids the cost of converting the raw list to a string in the Python
Elasticsearch client.
* Lines are read in bulk from the data source instead of line by line.
This avoids many method calls.
* We provide a special implementation for the common case (ids are
autogenerated, no conflicts) to make the hot code path as simple as
possible.

This commit also adds a microbenchmark that measures the speedup. The
following table shows a comparison of the throughput of the bulk reader
for various bulk sizes:

| Bulk Size | master [ops/s] | This PR [ops/s] | Speedup |
|-----------|----------------|-----------------|---------|
| 100       | 14829          | 92395           | 6.23    |
| 1000      | 1448           | 10953           | 7.56    |
| 10000     | 148            | 1100            | 7.43    |
| 100000    | 15             | 107             | 7.13    |

All data have been measured using Python 3.8 on Linux.

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -571,7 +571,7 @@ class PartitionBulkIndexParamSource:
         # self.internal_params always reads all files. This is necessary to ensure we terminate early in case
         # the user has specified ingest percentage.
         if self.current_bulk == self.total_bulks:
-            raise StopIteration
+            raise StopIteration()
         self.current_bulk += 1
         return next(self.internal_params)
 
@@ -647,13 +647,12 @@ def create_default_reader(docs, offset, num_lines, num_docs, batch_size, bulk_si
     source = Slice(io.FileSource, offset, num_lines)
 
     if docs.includes_action_and_meta_data:
-        am_handler = SourceActionMetaData(source)
+        return SourceOnlyIndexDataReader(docs.document_file, batch_size, bulk_size, source, docs.target_index, docs.target_type)
     else:
         am_handler = GenerateActionMetaData(docs.target_index, docs.target_type,
                                             build_conflicting_ids(id_conflicts, num_docs, offset), conflict_probability,
                                             on_conflict, recency)
-
-    return IndexDataReader(docs.document_file, batch_size, bulk_size, source, am_handler, docs.target_index, docs.target_type)
+        return MetadataIndexDataReader(docs.document_file, batch_size, bulk_size, source, am_handler, docs.target_index, docs.target_type)
 
 
 def create_readers(num_clients, client_index, corpora, batch_size, bulk_size, id_conflicts, conflict_probability, on_conflict, recency,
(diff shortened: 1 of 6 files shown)


esrally/track/params.py
```diff
@@ -901,38 +898,86 @@ def __next__(self):
...
+            self.read_bulk = self._read_bulk_regular
+        return self
+
+    def _read_bulk_fast(self):
+        """
+        Special-case implementation for bulk data files where the action and meta-data line is always identical.
+        """
         current_bulk = []
-        for action_metadata_item, document in zip(self.action_metadata, self.file_source):
+        # hoist
```
dliappis: Firstly I had to look this up as my understanding this is mostly JS terminology whereas we in Python we talk about block scoping.

But what does this comment refer at example? Which variable, in which block, is getting hoisted?

esrally/track/params.py
```diff
@@ -758,15 +757,15 @@ class GenerateActionMetaData:
...
                                             (index_name, type_name, "%s")
-            self.meta_data_index_no_id = '{"index": {"_index": "%s", "_type": "%s"}}' % (index_name, type_name)
+            self.meta_data_index_no_id = '{"index": {"_index": "%s", "_type": "%s"}}\n' % (index_name, type_name)
         else:
-            self.meta_data_index_with_id = '{"index": {"_index": "%s", "_id": "%s"}}' % (index_name, "%s")
-            self.meta_data_update_with_id = '{"update": {"_index": "%s", "_id": "%s"}}' % (index_name, "%s")
-            self.meta_data_index_no_id = '{"index": {"_index": "%s"}}' % index_name
+            self.meta_data_index_with_id = '{"index": {"_index": "%s", "_id": "%s"}}\n' % (index_name, "%s")
+            self.meta_data_update_with_id = '{"update": {"_index": "%s", "_id": "%s"}}\n' % (index_name, "%s")
+            self.meta_data_index_no_id = '{"index": {"_index": "%s"}}\n' % index_name
```
dliappis: nit: this might become something to fix when we enable [C4001 in pylintrc](https://github.com/elastic/rally/issues/838#issuecomment-564943721); should we `# pylint: disable=invalid-string-quote` right after `__init__` (I found a ref [here](https://stackoverflow.com/a/49392177/12047396)).

esrally/track/params.py
danielmitterdorfer: I know that term from JVM optimizations. In any case we ensure that the field access is turned into a local variable access because it is used in the loop on the hot code path and this is what I was referring to here.

esrally/track/params.py
danielmitterdorfer: How about we deal with this when we enable `C4001`?

esrally/track/params.py
dliappis: I see. For a person like me it doesn't add any additional info (seems obvious from the implementation) but if it's valuable to someone else, that's fine.

esrally/track/params.py
dliappis: That's fine.

esrally/track/params.py
danielmitterdorfer: I did not mean to confuse with the comment. Thanks for the feedback. Let's remove the comment then if it is not useful.

danielmitterdorfer:
Thanks for your review!

PR #899, opened 2020-02-14
Use zeroes instead of whitespaces as padding bytes

### LINKED ISSUE #896: Use zeroes instead of whitespaces as padding bytes
When generating ids for conflicts, rally uses spaces as padding bytes. I rarely see spaces in ids, so I think this is less realistic than using zeroes? Furthermore Elasticsearch has some optimizations for ids that look like base64 or numeric ids, which get defeated by the usage of spaces.

drawlerr:
It is more realistic to use 0-padding than whitespace padding in IDs, and this is the format Elasticsearch prefers anyways.

Applying @jpountz 's patch for ID generation, fixing associated tests.

Closes #896

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -1,6 +1,16 @@
 Migration Guide
 ===============
 
+Migrating to Rally 1.4.1
+------------------------
+
+Document IDs are now padded with 0 instead of spaces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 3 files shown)


drawlerr:
Arrg.  Seems we've managed to coax out an unrelated race condition in Rally during the integration test.  I've written this up in #900, but for now I'm just going to run tests again as we shouldn't be unlucky enough to hit it twice in a row:

@elasticsearchmachine  run tests

danielmitterdorfer:

jpountz:
Woohoo

Issue #892, opened 2020-02-12
More descriptive output of `esrallyd status`


Battleroid:
Would it be possible to have the output of `esrallyd status` be more descriptive?

Right now it appears to just check if a port is open, it'd be helpful if it could describe the actors associated with the rally cluster.

e.g. associated IPs of load drivers and the coordinator for example.

danielmitterdorfer:
The status command is really meant as a high-level info. If you are interested in more details I suggest you start a [Thespian shell](https://thespianpy.com/doc/in_depth.html#hH-058d8939-b973-4270-975b-3afd9c607176). Here is an example on an idle system but where the actor system is currently running:

```
python3 -m thespian.shell
Thespian Actor shell.  Type help or '?' to list commands.'

thespian> start multiprocTCPBase
Starting multiprocTCPBase ActorSystem
Capabilities: {}
Started multiprocTCPBase ActorSystem
thespian> address localhost 1900
Args is: {'port': '1900', 'ipaddr': 'localhost'}
Actor Address 0:  ActorAddr-(T|:1900)
thespian> status
Requesting status from Actor (or Admin) @ ActorAddr-(T|:1900) (#0)
Status of ActorSystem @ ActorAddr-(T|192.168.14.2:1900) [#1]:
  |Capabilities[9]:
                               ip: 192.168.14.2
          Convention Address.IPv4: 192.168.14.2:1900
              Thespian Generation: (3, 9)
         Thespian Watch Supported: True
                   Python Version: (3, 5, 2, 'final', 0)
        Thespian ActorSystem Name: multiprocTCPBase
     Thespian ActorSystem Version: 2
                 Thespian Version: 1581669778176
                      coordinator: True
  |Convention Leader: ActorAddr-(T|192.168.14.2:1900) [#1]
  |Convention Attendees [3]:
    @ ActorAddr-(T|192.168.14.4:1900) [#2]: Expires_in_0:21:41.056599
    @ ActorAddr-(T|192.168.14.3:1900) [#3]: Expires_in_0:21:41.030934
    @ ActorAddr-(T|192.168.14.5:1900) [#4]: Expires_in_0:21:41.391251
  |Primary Actors [0]:
  |Rate Governer: Rate limit: 4480 messages/sec (currently low with 1077 ticks)
  |Pending Messages [0]:
  |Received Messages [0]:
  |Pending Wakeups [0]:
  |Pending Address Resolution [0]:
  |>        1077 - Actor.Message Send.Transmit Started
[...]

PR #902, opened 2020-02-17
Add example how to determine actor system status

danielmitterdorfer:
With this commit we add an example to the docs that shows how to invoke
the Thespian shell to retrieve the status of the locally running actor
system.

Relates #892


danielmitterdorfer:
The PR build has timed out.

PR #903, opened 2020-02-17
Show more user-friendly track loader errors

danielmitterdorfer:
With this commit we add more details when Rally cannot load a track. In
all cases we include the path to the completely rendered track file. We
also detect syntax errors separately and provide the specific context in
which the error happens.

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -834,11 +834,11 @@ class TrackFileReader:
         """
 
         self.logger.info("Reading track specification file [%s].", track_spec_file)
+        # render the track to a temporary file instead of dumping it into the logs. It is easier to check for error messages
+        # involving lines numbers and it also does not bloat Rally's log file so much.
+        tmp = tempfile.NamedTemporaryFile(delete=False, suffix=".json")
         try:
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
Here is an example of the new output for a missing comma:

```
[ERROR] Cannot list. Could not load '/Users/daniel/.rally/benchmarks/tracks/default/geonames/track.json': Expecting ',' delimiter: line 46 column 7 (char 1066). Lines containing the error:

    {
      "name": "default",
      "operation-type": "search"
      "body": {
------^ Error is here
        "query": {
          "match_all": {}

The complete track has been written to '/var/folders/pg/gbmwwwms3n50fttdmnqd4sl40000gn/T/tmpy3b8nifs.json' for diagnosis.
```

another example with a missing double-quote at the end of a string:

```
[ERROR] Cannot list. Could not load '/Users/daniel/.rally/benchmarks/tracks/default/geonames/track.json': Invalid control character at: line 45 column 33 (char 1059). Lines containing the error:

    },
    {
      "name": "default",
      "operation-type": "search,
--------------------------------^ Error is here
      "body": {
        "query": {

The complete track has been written to '/var/folders/pg/gbmwwwms3n50fttdmnqd4sl40000gn/T/tmpf33rckho.json' for diagnosis.
```

PR #904, opened 2020-02-17
Eliminate deprecation warnings in Python 3.7+

danielmitterdorfer:
With this commit we upgrade Jinja2 from version 2.10.1 to 2.10.3. Prior
to that version, Jinja2 has used API that has been deprecated in Python
3.7 and thus emitted warnings.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -60,7 +60,7 @@ install_requires = [
     "jsonschema==3.1.1",
     # License: BSD
     # transitive dependency Markupsafe: BSD
-    "Jinja2==2.10.1",
+    "Jinja2==2.10.3",
     # License: MIT
     "thespian==3.9.3",
     # recommended library for thespian to identify actors more easily with `ps`


danielmitterdorfer:
for reference, these are the warnings we see when running `make test`:

```
=============================== warnings summary ===============================
.venv/lib/python3.8/site-packages/jinja2/utils.py:485
  /Users/daniel/Projects/rally/.venv/lib/python3.8/site-packages/jinja2/utils.py:485: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import MutableMapping

.venv/lib/python3.8/site-packages/jinja2/runtime.py:318
  /Users/daniel/Projects/rally/.venv/lib/python3.8/site-packages/jinja2/runtime.py:318: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import Mapping

-- Docs: https://docs.pytest.org/en/latest/warnings.html
======================= 694 passed, 2 warnings in 7.30s ========================
```

PR #914, opened 2020-02-21
Improve release docs and IT prechecks

drawlerr:
Failing integration tests delayed a recent release.
Note that IT should be passing as a precondition to release, as well as add a precondition to verify it will be possible to run the Docker portion instead of failing at the very end.

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -32,7 +32,7 @@ readonly ES_METRICS_STORE_TRANSPORT_PORT="63200"
 readonly ES_ARTIFACT_PATH="elasticsearch-${ES_METRICS_STORE_VERSION}"
 readonly ES_ARTIFACT="${ES_ARTIFACT_PATH}.tar.gz"
 readonly MIN_CURL_VERSION=(7 12 3)
-readonly MIN_DOCKER_MEM_GB=6.0
+readonly MIN_DOCKER_MEM_GB=6
 readonly RALLY_LOG="${HOME}/.rally/logs/rally.log"
 readonly RALLY_LOG_BACKUP="${HOME}/.rally/logs/rally.log.it.bak"
 
@@ -44,9 +44,8 @@ TEST_SUCCESS=0
 function check_prerequisites {
     exit_if_docker_not_running
 
-    DOCKER_MEM=$(docker info | grep Memory | awk '{print ($3 + 0)}')
-    HAS_ENOUGH_MEM=$(echo "$DOCKER_MEM > $MIN_DOCKER_MEM_GB" | bc)
-    if [[ ${HAS_ENOUGH_MEM} -eq 0 ]]; then
+    DOCKER_MEM=$(docker info | grep Memory | awk '{print int($3 + 0)}')
+    if [[ ${DOCKER_MEM} -lt ${MIN_DOCKER_MEM_GB} ]]; then
         echo "Error: Docker is not configured with enough memory."
         echo "Please increase memory available to Docker to at least ${MIN_DOCKER_MEM_GB} GB"
         exit 1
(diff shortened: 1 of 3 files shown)


release-checks.sh
```diff
@@ -21,6 +21,9 @@
 set -eu
 RELEASE_VERSION=$1
 
+# Minimum Docker memory required to run Docker integration tests for release
+MIN_DOCKER_MEM=6.0
```
danielmitterdorfer: Can we please indicate the unit of this value somehow (either in the comment or - IMHO even better - in the variable name)?

release-checks.sh
```diff
@@ -57,6 +60,16 @@ then
     exit 1
 fi
 
+if [[ $(uname) == "Darwin" ]]; then
+    DOCKER_MEM=$(docker info | grep Memory | awk '{print ($3 + 0)}')
+    HAS_ENOUGH_MEM=$(echo "$DOCKER_MEM > $MIN_DOCKER_MEM" | bc)
+    if [[ ${HAS_ENOUGH_MEM} -eq 0 ]]; then
+        echo "Error: Docker is not configured with enough memory."
+        echo "Please increase memory available to Docker to at least ${MIN_DOCKER_MEM} gb"
```
danielmitterdorfer: Can we change this to the correct unit symbol, which is `GB` (upper-case)?

release-checks.sh
```diff
@@ -57,6 +60,16 @@ then
     exit 1
 fi
 
+if [[ $(uname) == "Darwin" ]]; then
```
danielmitterdorfer: I think this check should be move to precondition checks for the integration tests (because this is where this matters as far as I understand) but not to the precondition checks of the release?

release-checks.sh
drawlerr: https://github.com/elastic/rally/pull/914/commits/601eb0ebb08c31769662843d1cd5b3b9adde2c9f

release-checks.sh
drawlerr: https://github.com/elastic/rally/pull/914/commits/601eb0ebb08c31769662843d1cd5b3b9adde2c9f

release-checks.sh
drawlerr: https://github.com/elastic/rally/pull/914/commits/601eb0ebb08c31769662843d1cd5b3b9adde2c9f

drawlerr:
IT fails due to rally-tracks change that has been reverted.  Let's try again.

@elasticmachine run tests

integration-test.sh
```diff
@@ -43,6 +44,13 @@ TEST_SUCCESS=0
 function check_prerequisites {
     exit_if_docker_not_running
 
+    DOCKER_MEM=$(docker info | grep Memory | awk '{print int($3 + 0)}')
```
danielmitterdorfer: If I specify exactly 6GB in the Docker Desktop UI (on Mac) this check fails. 

The reason is that memory with `docker info` is reported in `GiB` instead of GB. I played around a bit with `docker info` and you can actually force it to show the raw number in bytes with `docker info --format="{{.MemTotal}}"`. You can also use the `expr` utility to convert the GB value specified by `MIN_DOCKER_MEM_GB` to bytes with `expr $MIN_DOCKER_MEM_GB \* 1024 \* 1024 \* 1024` and base the entire calculation on bytes.

One confusing aspect that I noticed is that I get `6237241344` (bytes) as output from `docker info` when specifying 6GB in the UI but 6GB converted to bytes is 6000000000 bytes so I assume their UI should actually label this as GiB which results in 6442450944 bytes and I further assume that this is the actual value used for the underlying VM but it reserves ~195MB for the OS and only provides 6237241344 bytes to Docker. I wonder whether we should allow for a small margin of - say 5% - in the actual check to account for that (even better would be to get the actual value specified in the UI of course)?

drawlerr: I think it is less important to check on the UI values, compared to the actual value available to the VM as this is what will break integration testing.  The UI inconsistency is unfortunate, but ultimately out of scope here.

That said, the 6 GB minimum isn't as exact as the check would make it seem, but it works and I don't see much advantage to a margin here.

danielmitterdorfer: Given the target audience that's fine for me.

PR #918, opened 2020-02-25
Gather system metrics on Docker stop if possible

danielmitterdorfer:
With this commit we check whether the stop subcommand could determine
details about the current race and only attempt to store system metrics
if this is the case. If Docker is used with an in-memory metrics store
for example we have no means of retrieving the race and in this case we
need to avoid storing anything.


danielmitterdorfer:

PR #919, opened 2020-02-26
Run integration tests with pytest

danielmitterdorfer:
With this commit we add a proof-of-concept implementation of how
integration tests could look like with py.test instead of shell scripts.
For the proof of concept we migrate the integration tests for the list
subcommands and source builds.

It also spins up an Elasticsearch metrics store and checks that any
preconditions (basically Docker being present and up and running) are
met. Finally, it also integrates into CI by writing JUnit XML files for
test results.

Relates #736


danielmitterdorfer:
Unfortunately we don't see any Rally output in the PR checks. Therefore I have raised https://github.com/elastic/rally/pull/923.

it/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def stop(self):
+        self.cluster.stop()
+
+
+def install_integration_test_config():
+    def copy_config(name):
+        f = ConfigFile(name)
+        io.ensure_dir(f.rally_home)
+        with open(f.target_path, "w", encoding="UTF-8") as target:
```
dliappis: Would it make sense to just read src into `contents` with something:

```
with open(f.source_path, "r", encoding="UTF-8") as src:
    contents = src.read()
```

and then replace using `Template`'s native capabilities, to avoid pure string based replacements, i.e.

```
from string import Template
with open(f.target_path, "w", encoding="UTF-8") as target:
    target.write(Template(contents).substitute(USER_HOME=f.user_name)
```

[AFAIK](https://docs.python.org/3/library/string.html#template-strings) `${identifier}`-like template strings are supported out of the box.

it/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+        c = socket.socket()
+        connect_result = c.connect_ex(("127.0.0.1", port_number))
+        # noinspection PyBroadException
+        try:
+            if connect_result == errno.ECONNREFUSED:
+                c.close()
+                return
+            else:
+                c.close()
+                time.sleep(0.5)
```
dliappis: Not directly related to this PR, but going forwards should we start embracing the https://github.com/jd/tenacity library, which seems to be becoming the standard in retrying in Python? It offers a lot flexibility in what to retry with boolean operations.

it/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+        raise AssertionError("Docker is required to run integration tests.")
+    if process.run_subprocess_with_logging("docker ps") != 0:
+        raise AssertionError("Docker daemon must be up and running to run integration tests.")
+    if process.run_subprocess_with_logging("docker-compose --help") != 0:
+        raise AssertionError("Docker Compose is required to run integration tests.")
+
+
+class ConfigFile:
+    def __init__(self, config_name):
+        self.user_home = os.path.expanduser("~")
```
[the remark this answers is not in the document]
danielmitterdorfer: I'd tend to be consistent across our code base and keep this idiom.

it/__init__.py
```diff
@@ -0,0 +1,187 @@
...
+                time.sleep(0.5)
+        except Exception:
+            pass
+
+
+def check_prerequisites():
+    if process.run_subprocess_with_logging("docker") != 0:
+        raise AssertionError("Docker is required to run integration tests.")
+    if process.run_subprocess_with_logging("docker ps") != 0:
+        raise AssertionError("Docker daemon must be up and running to run integration tests.")
```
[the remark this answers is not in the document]
danielmitterdorfer: Good point. I combined the checks in https://github.com/elastic/rally/pull/919/commits/7786bdc366728f49f7287fe4ff19f27fc3ded850.

it/__init__.py
danielmitterdorfer: Thanks! Much better. Addressed in https://github.com/elastic/rally/pull/919/commits/eb03450a0fa37fe5ad06ea2225736d6fedd8e32d.

it/__init__.py
danielmitterdorfer: I am reluctant adding additional dependencies for such small pieces of functionality but I added a two minute timeout in https://github.com/elastic/rally/pull/919/commits/3537d0bb657f94ea211ab4a05e3029cb8923b3f4.

danielmitterdorfer:
Thanks for your reviews!

PR #925, opened 2020-03-03
Allow to specify an indices stats condition

danielmitterdorfer:
With this commit we expand the existing `indices-stats` operation and
allow to pass a condition based on a path and an expected value. This
has a variety of use cases:

* Verify that the number of documents in an index matches the expected
number of documents
* Wait until an operation has finished (e.g. no ongoing segment merges)
by specifying additional retry properties.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -32,7 +32,7 @@ __RUNNERS = {}
 def register_default_runners():
     register_runner(track.OperationType.Bulk.name, BulkIndex())
     register_runner(track.OperationType.ForceMerge.name, ForceMerge())
-    register_runner(track.OperationType.IndicesStats.name, IndicesStats())
+    register_runner(track.OperationType.IndicesStats.name, Retry(IndicesStats()))
     register_runner(track.OperationType.NodesStats.name, NodeStats())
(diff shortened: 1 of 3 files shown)


esrally/driver/runner.py
```diff
@@ -1433,8 +1464,10 @@ def __call__(self, es, params):
                 # we can determine success if and only if the runner returns a dict. Otherwise, we have to assume it was fine.
                 elif isinstance(return_value, dict):
                     if return_value.get("success", True):
+                        self.logger.debug("%s has returned successfully", repr(self.delegate))
                         return return_value
                     else:
+                        self.logger.debug("%s has returned with an error: %s.", repr(self.delegate), return_value)
```
dliappis: I am wondering if it'd be handy to log this more eagerly than only on debug, for troubleshooting purposes and giving a more complete picture of what's going on without having to switch to debug.

Do you think it'd become very chatty?

danielmitterdorfer: IMHO it could indeed become a bit chatty because it would add a log entry every time the condition is not yet met.

danielmitterdorfer:
Thanks for your review! :)

PR #927, opened 2020-03-03
Add unit for all query charts

danielmitterdorfer:
With this commit we add a proper unit to all time series charts for
queries.


danielmitterdorfer:
Here's an example chart from our nightlies.

Before:

before

With this PR:

after

Note how the y-axis shows the unit now.

PR #932, opened 2020-03-06
Improve layout of generated dashboards

danielmitterdorfer:
With this commit we adapt the dashboard layout in the chart generator
that we also use in our nightlies, namely that the bulk-indexing charts
are wider than all other charts.

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -260,7 +260,7 @@ class BarCharts:
                 "title": title,
                 "visState": json.dumps(vis_state),
                 "uiStateJSON": BarCharts.UI_STATE_JSON,
-                "description": "",
+                "description": "gc",
                 "version": 1,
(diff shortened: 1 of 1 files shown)


esrally/chart_generator.py
```diff
@@ -1453,25 +1453,39 @@ def generate_segment_memory(chart_type, race_configs, environment):
...
+            "panelIndex": panelIndex,
+            "gridData": {
+                "x": (col * chart_width),
+                "y": (row * height),
+                "w": chart_width,
+                "h": height,
+                "i": "{}".format(panelIndex)
+            },
+            "type": "visualization",
+            "version": "6.3.2"
```
danielmitterdorfer: Adding a version was necessary to avoid that Kibana assumes a legacy format and attempts to load it in an outdated format. For newer Kibana versions we need to change the generator anyway as a couple of things have changed in 7.x.

PR #916, opened 2020-02-21
Add an asyncio-based load generator

danielmitterdorfer:
With this commit we add a new experimental subcommand `race-aync` to Rally. It
allows to specify significantly more clients than the current `race` subcommand.
The reason for this is that under the hood, `race-async` uses `asyncio` and runs
all clients in a single event loop. Contrary to that, `race` uses an actor
system under the hood and maps each client to one process.

As the new subcommand is very experimental and not yet meant to be used broadly,
there is no accompanying user documentation in this PR. Instead, we plan to
build on top of this PR and expand the load generator to take advantage of
multiple cores before we consider this usable in production (it will likely keep
its experimental status though).

In this PR we also implement a compatibility layer into the current load
generator so both work internally now with `asyncio`. Consequently, we have
already adapted all Rally tracks with a backwards-compatibility layer (see
elastic/rally-tracks#97 and rally-eventdata-track#80).

Closes #852

PR #934, opened 2020-03-20
Bump minimum Python version to 3.8

danielmitterdorfer:
With this commit we require at least Python 3.8 for Rally. This is done
as a preparation for Rally's new load generator which requires certain
Python 3.8 features (contextvars, madvise) and uses asyncio internally 
which has been improved in more recent versions of Python.

PR #935, opened 2020-03-20
Add an asyncio-based load generator

### LINKED ISSUE #852: An actorless load generator
At the moment Rally uses a process per client model (see https://github.com/elastic/rally/issues/9#issuecomment-179312337, https://github.com/elastic/rally/issues/58 and https://github.com/elastic/rally/commit/fe5dba3205c7c09a1491f349c6d4db19d89d2f9d for more details) and implements this with an actor system that allows clients (actors) to communicate with each other. The basic structure is that there is a controller actor (called `Driver`) for all clients and each client (actor) is responsible to execute scheduled tasks. However, this has several drawbacks that we've seen in practice:

* Complexity added by the actor framework: we need to spin up a separate process to communicate across nodes. Having the ability to communicate across nodes was mostly needed to spin up a cluster from the coordinator machine (where the benchmark has been started) but is less important for the load generator (although we also use it to distribute it across several machines). This was convenient in the early days but nowadays we let other tools handle remote connections during setup (e.g. Ansible) and [use Rally only to setup the individual nodes](https://github.com/elastic/rally/issues/697). Thus we don't need an actor system as much as we did before.
* Terminal output issues (the actor system performs a double-fork and thus we cannot write to the terminal)
* Issues to even startup the system on Windows that are hard to debug.
* Limit on the upper bound of supported clients (the whole system is just hanging if there are too many actors).
* Logging issues: logs are sent to a central component which then writes to the log file. If one actor issues too many logs, the actor system just becomes unresponsive (which is quite hard to debug because there are no logs to begin with).
* Debugging actor issues is quite tricky and we don't have much influence over what is logged.

Note that these limitations are due to the specific actor system implementation, and not necessarily conceptual issues. There are, however, also shortcomings that can be related to the process per client model:

* There is no built-in mechanism for clients to share state (this is not technically impossible, it's just a bit harder than with a single process model). Sharing state can be sometimes convenient e.g. if in an initial step, data is retrieved from Elasticsearch and then we can use this information in later steps. This is done for example by the eventdata track that retrieves the minimum and maximum date of log entries in Elasticsearch and uses it to choose appropriate date range parameters for simulating dashboard queries in later steps.
* Starting a process for each client is quite heavy-weight and thus it's not feasible to simulate thousands of clients with Rally.

We should explore a few options based on these shortcomings:

* Using the [Python Elasticsearch async client](https://github.com/elastic/elasticsearch-py-async).
* Using [asyncio](https://docs.python.org/3.8/library/asyncio.html) directly
* Keeping the process per client model but instead use [multiprocessing](https://docs.python.org/3.8/library/multiprocessing.html)

There are maybe other options as well but these are the choices that first come to mind. We should also refine this ticket as we proceed with our analysis.

IMHO it is ok to support only one load generator machine in the first step but we should aim to support the Rally tracks that we currently maintain.

danielmitterdorfer:
With this commit we add a new experimental subcommand `race-aync` to Rally. It
allows to specify significantly more clients than the current `race` subcommand.
The reason for this is that under the hood, `race-async` uses `asyncio` and runs
all clients in a single event loop. Contrary to that, `race` uses an actor
system under the hood and maps each client to one process.

As the new subcommand is very experimental and not yet meant to be used broadly,
there is no accompanying user documentation in this PR. Instead, we plan to
build on top of this PR and expand the load generator to take advantage of
multiple cores before we consider this usable in production (it will likely keep
its experimental status though).

In this PR we also implement a compatibility layer into the current load
generator so both work internally now with `asyncio`. Consequently, we have
already adapted all Rally tracks with a backwards-compatibility layer (see
elastic/rally-tracks#97 and elastic/rally-eventdata-track#80).

Closes #852
Relates #916


danielmitterdorfer:
Notes to reviewers: This is #916 and some more additions to address some slowdowns that we have seen. The main new commits are:

* We tighten the measurement loop now as much as possible and parse JSON responses lazily in https://github.com/elastic/rally/commit/8d64f58635b44288b7f43e0207a277aa208bfcf0
* We use `mmap` now to read bulk files in https://github.com/elastic/rally/commit/7c358e4d0e857e7998f5214a2eb72076f69fd165

All other details from https://github.com/elastic/rally/pull/916#issuecomment-590725759 still apply.

danielmitterdorfer:
> One question, is self.mm.madvise(mmap.MADV_SEQUENTIAL) something that we should strive to always have by enforcing the requirement for Python 3.8 (I am thinking of performance gains due to readhead)

We could probably add a recommendation for the latest Python version but I wouldn't go so far as to drop support for all versions before 3.8?

danielmitterdorfer:
I pushed https://github.com/elastic/rally/pull/934/commits/e49b85af80adbe25d6971d6c7075568c1a9e6eb3 in https://github.com/elastic/rally/pull/934 to bump the version to Python 3.8.

danielmitterdorfer:
@dliappis In my benchmarks I have noticed that the new load generator added a small overhead (1-2 ms) which was noticeable in queries with small service times in the single-digit millisecond range. It turns out that this has been caused by parsing the JSON response (although it's already event-based and lazy). Switching the parser backend from Python to C was (complex and) ineffective; thus I have added a parameter to skip response parsing entirely (it's on by default). Can you please take another look?

docs/track.rst
```diff
@@ -402,6 +402,7 @@ With the operation type ``search`` you can execute `request body searches <http:
         2. Rally will not attempt to serialize the parameters and pass them as is. Always use "true" / "false" strings for boolean parameters (see example below).
 
 * ``body`` (mandatory): The query body.
+* ``detailed-results`` (optional, defaults to ``true``): Records more detailed meta-data about queries. As it analyzes the corresponding response in more detail, this might incur additional overhead which can skew measurement results. Set this value to ``false`` for queries that return within single digit milliseconds to increase measurement accuracy. This flag is ineffective for scroll queries.
```
dliappis: Perhaps we should mention some important fields like `took` and `hits` fields from Elasticsearch that won't be captured when ``detailed-results:false`; might save some head scratching for people who disable this and wonder what is it that they'll be losing?

I see it is included in the [migration docs](https://github.com/elastic/rally/pull/935/commits/4c2e78a125a6b98fff25010c16066275e61f53d2)

docs/migrate.rst
```diff
@@ -9,6 +9,18 @@ Minimum Python version is 3.8.0
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Rally 1.5.0 does not determine query meta-data anymore by default to reduce the risk of client-side bottlenecks. The following meta-data fields are affected:
+
+* ``hits``
+* ``hits_relation``
+* ``timed_out``
+* ``took``
+
+If you still want to retrieve them (risking skewed results due to additional overhead), set the new property ``detailed-results`` to ``true`` for any operation of type ``search``.
```
dliappis: This contradicts what we are saying in https://github.com/elastic/rally/pull/935/commits/4fcaf9d0a17ffce744d424777c3cbc20d19caa8a#diff-a4e117fd9659940c6f1756d76f5dfe5cR405, i.e. that at least for the `search` operation the default is `true`?

danielmitterdorfer: I changed this fo `false` already in https://github.com/elastic/rally/pull/935/commits/4c2e78a125a6b98fff25010c16066275e61f53d2?

docs/track.rst
danielmitterdorfer: I pushed https://github.com/elastic/rally/pull/935/commits/5ab95230d29f9a8c256551c7e60e900be8ead364.

docs/migrate.rst
dliappis: Ah! Ok then.

PR #937, opened 2020-03-24
Fix port number in recipes docs

hub-cap:


hub-cap:
@dliappis had asked me to do this because of some errors he had with a discuss issue. I think he would be best to comment on this.

dliappis:
There are a number of motivations for switching to `39200`:

1. [In the docs](https://esrally.readthedocs.io/en/stable/configuration.html#preparation) and in our default out of the box configuration when `--target-host` is not defined, we default to port 39200.

2. It's not uncommon that there's already an instance (e.g. metricstore or left over) occupying 9200 already

I've noticed that both can lead to confusion for users, especially when using esrallyd.

Alternatively, I think we could document in [--target-hosts](https://esrally.readthedocs.io/en/stable/command_line_reference.html#target-hosts) that Rally will automatically derive and use the port defined there as the Elasticsearch http_port (and based on that transport_port+100)

danielmitterdorfer:
> Alternatively, I think we could document in --target-hosts that Rally will automatically derive and use the port defined there as the Elasticsearch http_port (and based on that transport_port+100)

I don't think we should do that given that we are about to move to the new subcommands where the port needs to be specified explicitly.

> It's not uncommon that there's already an instance (e.g. metricstore or left over) occupying 9200 already

I'd argue that when users use `--pipeline=benchmark-only` they set up the cluster themselves and use the default port (9200)? If they use Rally instead to setup the cluster then this is different but otherwise I think we actually want to point them to 9200 by default?

dliappis:
> I don't think we should do that given that we are about to move to the new subcommands where the port needs to be specified explicitly.
> 

Yeah we are in this in-between state currently; people find and still use the daemon mode and hit issues, so this was the main motivation here esp. since this magic handling of the port isn't currently documented.

> I'd argue that when users use `--pipeline=benchmark-only` they set up the cluster themselves and use the default port (9200)? If they use Rally instead to setup the cluster then this is different but otherwise I think we actually want to point them to 9200 by default?

Fair point; I'd counter argue that e.g. in Elastic Cloud this isn't the out of the box default either, but for self managed clusters one would think in most cases the port is 9200.

dliappis:
had an offline chat with @danielmitterdorfer re ^^

@hub-cap let's amend this PR to only change `9200` to `39200` for the [Benchmarking a remote cluster](https://esrally.readthedocs.io/en/stable/recipes.html#benchmarking-a-remote-cluster) example which uses Rally to spin up the cluster.

The [Distributing the load test driver](https://esrally.readthedocs.io/en/stable/recipes.html#distributing-the-load-test-driver) example assumes an externally provisioned cluster (so that's ok, no changes needed there) and the first example, [Benchmarking an existing cluster](https://esrally.readthedocs.io/en/stable/recipes.html#benchmarking-an-existing-cluster) is also about externally provisioned clusters, so no changes needed there either.

PR #938, opened 2020-03-26
Store units in results

danielmitterdorfer:
With this commit we store units for all operation-related samples and
system metrics also for summary metrics.

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -732,16 +732,17 @@ class MetricsStore:
         """
         return self._get(name, task, operation_type, sample_type, node_name, mapper)
 
-    def get_unit(self, name, task=None):
+    def get_unit(self, name, task=None, node_name=None):
         """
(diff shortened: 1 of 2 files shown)


esrally/metrics.py
```diff
@@ -732,16 +732,17 @@ def get_raw(self, name, task=None, operation_type=None, sample_type=None, node_n
         """
         return self._get(name, task, operation_type, sample_type, node_name, mapper)
 
-    def get_unit(self, name, task=None):
+    def get_unit(self, name, task=None, node_name=None):
```
hub-cap: I dont see this used in the PR, was it just added in case we need to use it?

danielmitterdorfer: Good catch! I added it in https://github.com/elastic/rally/pull/938/commits/11364ad4bc5a355a082e75e58f47bfcfd4c52b07.

PR #941, opened 2020-03-30
Improve response processing

danielmitterdorfer:
With this commit we improve response processing speed by taking the following
measures:

1. We use the raw bytes for JSON parsing instead of first converting them into a
string.
2. We expose an additional option for scroll queries to disable HTTP response
compression as we have seen that this can cause significant overhead for very
large responses (large documents). The default is (still) to have response
compression enabled.

Our experiments have shown the following changes to the 50th percentile service
time for the PMC track:

* Baseline: 3203 ms
* With measure 1: 3014 ms
* With measure 1 and measure 2: 774 ms

Relates #935


danielmitterdorfer:
Thanks for the review!

PR #943, opened 2020-04-01
Fix request timeout handling

danielmitterdorfer:
With this commit we address several issues in request timeout handling:

1. We ensure that the end of the request is properly measured
2. We set an explicit request timeout on the connection to avoid that
the default connection timeout is picked.
3. We use the proper method parameter `timeout` when falling back to the
raw transport API instead of setting the ineffective parameter
`request_timeout`.


danielmitterdorfer:
Thanks for the review :)

PR #944, opened 2020-04-01
Allow to use significantly more clients

danielmitterdorfer:
With this commit we change how clients are assigned to worker processes in the
load generator. While historically Rally has assigned one worker (process) to
one client, with the changes done in #935 we can assign multiple clients to a
single worker process now and run the clients in an asyncio event loop. This
allows us to simulate many more clients than what is possible the process-based
approach which is very heavy-weight.

By default Rally will only create as many workers as there are CPUs available on
the system. This choice can be overridden in Rally's config file with the
configuration setting `available.cores` in the section `system` to allocate more
or less workers. If multiple load generator machines note that we assume that
all of them have the same hardware specifications and only take the coordinating
machine's CPU count into account.


hub-cap:
@elasticmachine run tests

hub-cap:
@elasticmachine run tests

danielmitterdorfer:

danielmitterdorfer:
> Unrelated to this specific PR, but since #935 where we introduced the `MmapSource` it appears that with ~4000 clients (on a single machine) there _reported_ memory usage (from `top` and `pmap`) is ~`1.1TB`, so it's likely we need to look at this in more detail in a follow up.

https://github.com/elastic/rally/pull/963 to addresses this by sharing the memory-mapped file within each process. This means we memory-map a data file only once per worker (process) instead of once per client.

PR #945, opened 2020-04-01
Add an override for the rally homedir

hub-cap:


tox.ini
```diff
@@ -21,6 +21,7 @@ passenv =
     JAVA12_HOME
     JAVA13_HOME
     SSH_AUTH_SOCK
+    OVERRIDE_RALLY_HOME
```
dliappis, line 24: maybe in alphanum order here?

hub-cap:
@elasticmachine run tests

PR #947, opened 2020-04-02
Use pbzip2/pigz to decompress corpora if available

dliappis:
Decompressing large corpora using the standard bzip2/gzip libraries
can be a slow process as they only utilize one cpu core.  Take
advantage of pbzip2/pigz, if available, to speed up the process by
taking advantage of all cores.

### DIFF
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -315,14 +315,13 @@ def _do_decompress_manually(target_directory, filename, decompressor_args, decom
 
 
 def _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
-    logger = logging.getLogger(__name__)
-
     with open(os.path.join(target_directory, base_path_without_extension), "wb") as new_file:
         try:
             subprocess.run(decompressor_args + [filename], stdout=new_file, stderr=subprocess.PIPE, check=True)
         except subprocess.CalledProcessError as err:
-            console.warn(f"Failed to decompress [{filename}] with [{err.cmd}]. Falling back to default library.")
-            logger.warning("Error [%s]", err.stderr)
+            logger = logging.getLogger(__name__)
+            console.warn(f"Failed to decompress [{filename}] with [{err.cmd}]. Error [{err.stderr}]. Falling back to standard library.",
+                         logger=logger.warning)
             return False
     return True
 
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -309,7 +309,8 @@ def _do_decompress_manually(target_directory, filename, decompressor_args, decom
         if _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
             return
     else:
-        console.warn(f"{decompressor_bin} not found in PATH. Using standard library, decompression will take longer.")
+        logging.getLogger(__name__).warning("%s not found in PATH. Using standard library, decompression will take longer.",
+                                            decompressor_bin)
 
     _do_decompress_manually_with_lib(target_directory, filename, decompressor_lib(filename))
 
@@ -319,9 +320,8 @@ def _do_decompress_manually_external(target_directory, filename, base_path_witho
         try:
             subprocess.run(decompressor_args + [filename], stdout=new_file, stderr=subprocess.PIPE, check=True)
         except subprocess.CalledProcessError as err:
-            logger = logging.getLogger(__name__)
-            console.warn(f"Failed to decompress [{filename}] with [{err.cmd}]. Error [{err.stderr}]. Falling back to standard library.",
-                         logger=logger.warning)
+            logging.getLogger(__name__).warning("Failed to decompress [%s] with [%s]. Error [%s]. Falling back to standard library.",
+                                                filename, err.cmd, err.stderr)
             return False
     return True
 
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -331,7 +331,7 @@ def _do_decompress_manually_with_lib(target_directory, filename, compressed_file
 
     ensure_dir(target_directory)
     try:
-        with open("%s/%s" % (target_directory, path_without_extension), "wb") as new_file:
+        with open(os.path.join(target_directory, path_without_extension), "wb") as new_file:
             for data in iter(lambda: compressed_file.read(100 * 1024), b""):
                 new_file.write(data)
     finally:
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -17,8 +17,10 @@
 
 import bz2
 import gzip
+import logging
 import os
 import re
+import shutil
 import subprocess
 import tarfile
 import zipfile
@@ -244,6 +246,15 @@ def is_archive(name):
     return ext in [".zip", ".bz2", ".gz", ".tar", ".tar.gz", ".tgz", ".tar.bz2"]
 
 
+def is_executable(name):
+    """
+    :param name: File name to check.
+    :return: True iff given file name is executable and in PATH, all other cases False.
+    """
+
+    return shutil.which(name) is not None
+
+
 def compress(source_directory, archive_name):
     """
     Compress a directory tree.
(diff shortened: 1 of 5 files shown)


esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
     this function.
     """
     path_without_extension, extension = splitext(zip_name)
-    filename = basename(path_without_extension)
     if extension == ".zip":
         _do_decompress(target_directory, zipfile.ZipFile(zip_name))
     elif extension == ".bz2":
-        _do_decompress_manually(target_directory, filename, bz2.open(zip_name))
+        decompressor_args = ["pbzip2", "-d", "-k", "-m10000", "-c"]
+        decompressor_lib = bz2.open(zip_name)
```
danielmitterdorfer, line 292: We could avoid opening the file prematurely by just passing the function reference and calling it when needed? We can also do this with a context manager to ensure the compressed file is properly closed again afterwards. 

We're also not closing the open file in case the file can be decompressed via the binary instead of the library.

esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
     this function.
     """
     path_without_extension, extension = splitext(zip_name)
-    filename = basename(path_without_extension)
     if extension == ".zip":
         _do_decompress(target_directory, zipfile.ZipFile(zip_name))
     elif extension == ".bz2":
-        _do_decompress_manually(target_directory, filename, bz2.open(zip_name))
+        decompressor_args = ["pbzip2", "-d", "-k", "-m10000", "-c"]
```
danielmitterdorfer, line 291: Does this mean we unconditionally allow `pbzip2` to use up to 10.000MB of memory? (`-m10000`). I wonder whether this might be a bit too aggressive for constrained environments (Docker containers?). On the other hand, having a lot of RAM available for the page cache later would be beneficial in any case so I'm a bit on the fence whether we should lower it. Did you by chance measure maximum memory usage of `pbzip2` while decompressing a corpus file?

esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
...
+        decompressor_lib = bz2.open(zip_name)
+        _do_decompress_manually(target_directory, zip_name, decompressor_args, decompressor_lib)
     elif extension == ".gz":
-        _do_decompress_manually(target_directory, filename, gzip.open(zip_name))
+        decompressor_args = ["pigz", "-d", "-k", "-c"]
+        decompressor_lib = gzip.open(zip_name)
```
danielmitterdorfer, line 296: See above for my comment about prematurely opening the file.

esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
...
-def _do_decompress_manually(target_directory, filename, compressed_file):
+def _do_decompress_manually(target_directory, filename, decompressor_args, decompressor_lib):
+    decompressor_bin = decompressor_args[0]
+    base_path_without_extension = basename(splitext(filename)[0])
+
+    if is_executable(decompressor_bin):
+        if _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
+            return
+    else:
+        console.warn(f"{decompressor_bin} not found in PATH. Using default library, decompression will take longer.")
```
danielmitterdorfer, line 312: How about `{decompressor_bin} not found in PATH. Falling back to standard library. Decompression will only use one CPU and thus take longer.`?

esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
...
+
+
+def _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
+    logger = logging.getLogger(__name__)
+
+    with open(os.path.join(target_directory, base_path_without_extension), "wb") as new_file:
+        try:
+            subprocess.run(decompressor_args + [filename], stdout=new_file, stderr=subprocess.PIPE, check=True)
+        except subprocess.CalledProcessError as err:
+            console.warn(f"Failed to decompress [{filename}] with [{err.cmd}]. Falling back to default library.")
```
danielmitterdorfer, line 324: `Falling back to standard library`?. Also, `console.warn` allows to pass a `logger` instance so maybe that would be sufficient. I also think the log statement below cannot be understood properly in isolation (i.e. without also reading the console message).

esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
...
+        if _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
+            return
+    else:
+        console.warn(f"{decompressor_bin} not found in PATH. Using default library, decompression will take longer.")
+
+    _do_decompress_manually_with_lib(target_directory, filename, decompressor_lib)
+
+
+def _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
+    logger = logging.getLogger(__name__)
```
danielmitterdorfer, line 318: Maybe instantiate in the `except` block as we don't log anything in the success case?

esrally/utils/io.py
```diff
@@ -274,24 +285,55 @@ def decompress(zip_name, target_directory):
...
+
+
+def _do_decompress_manually_with_lib(target_directory, filename, compressed_file):
+    path_without_extension = basename(splitext(filename)[0])
+
     ensure_dir(target_directory)
     try:
-        with open("%s/%s" % (target_directory, filename), 'wb') as new_file:
-            for data in iter(lambda: compressed_file.read(100 * 1024), b''):
+        with open("%s/%s" % (target_directory, path_without_extension), "wb") as new_file:
```
danielmitterdorfer, line 335: Use `os.path.join` instead of `"%s/%s"`?

esrally/utils/io.py
dliappis: `-m` specifies the **max** memory usage; from the man page:

> Where # is max memory usage in 1MB steps (default 100 = 100MB)

So if an environment has less it will just whatever is available.

During the _compression+ with the same parameter (`-m10000`) on a machine with lots of ram (256GB) it reports:

```
$ pbzip2 -v -k -m10000 documents.json
Parallel BZIP2 v1.1.9     - by: Jeff Gilchrist [http://compression.ca]
[Apr. 13, 2014]               (uses libbzip2 by Julian Seward)
Major contributions: Yavor Nikolov <nikolov.javor+pbzip2@gmail.com>

         # CPUs: 12
 BWT Block Size: 900 KB
File Block Size: 900 KB
 Maximum Memory: 1410 MB
-------------------------------------------
         File #: 1 of 1
     Input Name: documents.json
    Output Name: documents.json.bz2

     Input Size: 79802445255 bytes
Compressing data...
```

whereas pmap shows:

```
                 Kbytes     RSS   Dirty 
---------------- ------- ------- -------
total kB          949904  114832  112048
```

i.e. seems to be using <1GB.

I haven't checked the usage during decompression yet.

dliappis: On the same machine during extraction `pmap -x` for the pbzip2 pid (`pbzip2 -d -k -m10000 -c /home/dliappis/.rally/benchmarks/data/nyc_taxis/documents.json.bz2`) reports:

`total kB          929564   77196   74656`

danielmitterdorfer: Ok, then we should be ok. Thanks for double-checking.

dliappis:
I am also sharing the benefits from this approach, on a machine with 12 `Intel(R) Xeon(R) CPU E5-1650 v3 @ 3.50GHz` cores and a `Micron_5100_MTFDDAK1T9TBY` nvme disk.

1. Using the `nyc_taxis` corpus as it is currently compressed with standard bzip2, decompressing with current Rally master takes: ~`29m`. Only 1 core is 100% used throughout the process.

2. Using a recompressed corpus with `pbzip2 -v -k -m10000 documents.json` and extracting using this PR: ~`7m 30s`, i.e. ~75% reduction in execution time. All cores are 100% used throughout extraction.

esrally/utils/io.py
dliappis: Addressed in c51403c60110882237bc420eca02a9739b148072

esrally/utils/io.py
dliappis: Addressed in c51403c60110882237bc420eca02a9739b148072

esrally/utils/io.py
dliappis: Addressed in fee8c00c2b8bcc1fd9ff6556c4dce6d2584138d8; I think stating that decompression will only use one cpu is a tech detail on the verge of too much info plus there are additionally (small) time savings by using using a native binary.

esrally/utils/io.py
dliappis: Addressed in 746fb50c5d0c8f82a95223410bf993782dcc0044

esrally/utils/io.py
dliappis: Addressed in 746fb50c5d0c8f82a95223410bf993782dcc0044

esrally/utils/io.py
dliappis: Addressed in c6575b4ae65417f5e685d3ca3f718851bf3354e4

esrally/utils/io.py
```diff
@@ -274,24 +285,54 @@ def decompress(zip_name, target_directory):
...
-def _do_decompress_manually(target_directory, filename, compressed_file):
+def _do_decompress_manually(target_directory, filename, decompressor_args, decompressor_lib):
+    decompressor_bin = decompressor_args[0]
+    base_path_without_extension = basename(splitext(filename)[0])
+
+    if is_executable(decompressor_bin):
+        if _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
+            return
+    else:
+        console.warn(f"{decompressor_bin} not found in PATH. Using standard library, decompression will take longer.")
```
danielmitterdorfer, line 312: On the command line this looks like:

```
[INFO] Decompressing track data from [/Users/daniel/.rally/benchmarks/data/geonames/documents-2-1k.json.bz2] to [/Users/daniel/.rally/benchmarks/data/geonames/documents-2-1k.json] ... [WARNING] pbzip2 not found in PATH. Using standard library, decompression will take longer.
[OK]
```

This looks a bit odd and I wonder whether we should just write the output to the logs? Alternatively we could maybe offer a separate function that allows to check this upfront because with `http_logs` (multiple indices) this gets a bit noisy:

```
[INFO] Decompressing track data from [/Users/daniel/.rally/benchmarks/data/http_logs/documents-181998-1k.json.bz2] to [/Users/daniel/.rally/benchmarks/data/http_logs/documents-181998-1k.json] ... [WARNING] pbzip2 not found in PATH. Using standard library, decompression will take longer.
[OK]
[INFO] Preparing file offset table for [/Users/daniel/.rally/benchmarks/data/http_logs/documents-181998-1k.json] ... [OK]
[INFO] Decompressing track data from [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json.bz2] to [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json] ... [WARNING] pbzip2 not found in PATH. Using standard library, decompression will take longer.
[OK]
[INFO] Preparing file offset table for [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json] ... [OK]
[INFO] Decompressing track data from [/Users/daniel/.rally/benchmarks/data/http_logs/documents-201998-1k.json.bz2] to [/Users/daniel/.rally/benchmarks/data/http_logs/documents-201998-1k.json] ... [WARNING] pbzip2 not found in PATH. Using standard library, decompression will take longer.
[OK]
```

esrally/utils/io.py
```diff
@@ -274,24 +285,54 @@ def decompress(zip_name, target_directory):
...
+
+
+def _do_decompress_manually_external(target_directory, filename, base_path_without_extension, decompressor_args):
+    with open(os.path.join(target_directory, base_path_without_extension), "wb") as new_file:
+        try:
+            subprocess.run(decompressor_args + [filename], stdout=new_file, stderr=subprocess.PIPE, check=True)
+        except subprocess.CalledProcessError as err:
+            logger = logging.getLogger(__name__)
+            console.warn(f"Failed to decompress [{filename}] with [{err.cmd}]. Error [{err.stderr}]. Falling back to standard library.",
+                         logger=logger.warning)
```
danielmitterdorfer, line 324: This produces:

```
[ERROR] Cannot race. Error in track preparator ('function' object has no attribute 'warning')
```

you just need to provide the `logger` instead of `logger.warning`.

Alternatively, I also wonder whether we should either fail hard here because something is clearly wrong with the `pbzip` binary or instead just write it to the logs and continue with the stdlib implementation. Here is the output from one of my tests (after I changed `logger=logger.warning` to `logger=logger`):

```
[INFO] Decompressing track data from [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json.bz2] to [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json] ... [WARNING] Failed to decompress [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json.bz2] with [['pbzip2', '/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json.bz2']]. Error [b'pbzip2: *ERROR: Input file [/Users/daniel/.rally/benchmarks/data/http_logs/documents-191998-1k.json.bz2] already has a .bz2 extension!  Skipping...\n-------------------------------------------\n']. Falling back to standard library.
```

esrally/utils/io.py
dliappis: Re: noise, however, this noise will only happen once; after the all corpora have been extracted it'll not be seen again.

Re: printed location of the warn, I initially thought about adding a newline. Then the final [OK] would land on the next line, if a warning gets emitted. I think refactoring the checks out of here is not worth the benefits. Maybe we can just stick to logging.

danielmitterdorfer: I'm fine with just logging. :)

dliappis: Switched to logging only in https://github.com/elastic/rally/pull/947/commits/476e3fd9051fa3db43cf6e85ec9a29b0f704f3de

esrally/utils/io.py
dliappis: Switched to logging only in https://github.com/elastic/rally/pull/947/commits/476e3fd9051fa3db43cf6e85ec9a29b0f704f3de

PR #812, opened 2019-11-12
Add fallbacks for ML APIs

drawlerr:
In older ES, "xpack" was still required in the URL, so the newer client in Rally would have issues.  Add fallback functionality to handle this case.

Closes #787

PR #956, opened 2020-04-07
Fix fallback implementation of ML-related runners

danielmitterdorfer:
With this commit we fix several issues of the fallback implementation of
ML-related runners. This implementation is only used for older versions
of Elasticsearch which still include the `_xpack` path component and use
the raw transport API of the Elasticsearch client. However, especially
the handling of request parameters did not conform to the API
requirements.

Relates #812

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -269,6 +269,21 @@ def mandatory(params, key, op):
                                    " parameter source." % (str(op), key))
 
 
+def escape(v):
+    """
+    Escapes values so they can be used as query parameters
+
+    :param v: The raw value. May be None.
+    :return: The escaped value.
+    """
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Note to reviewers (@hub-cap): This should be tested together with https://github.com/elastic/rally-tracks/pull/111. One way to test this (which has failed so far on `master`) is:

```
esrally --distribution-version=5.6.16 --car="4gheap" --elasticsearch-plugins="x-pack:ml" --track=nyc_taxis --challenge="append-ml" --test-mode --on-error=abort
```

hub-cap:
I did not get around to testing this today, I will tomorrow though unless @dliappis beats me to it :)

PR #957, opened 2020-04-08
Move CI script to the rally repository

hub-cap:
This script was previously in a repository that did not allow the rally
team to quickly iterate on, and instead relied on a team that is not
responsible for rally related functionality to approve changes
therein. This commit moves the script into the rally repository so that
it can be iterated on by the rally team.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -43,12 +43,12 @@ function archive {
   if [[ -d ${RALLY_DIR} ]]; then
     find ${RALLY_DIR} -name "*.log" -printf "%P\\0" | tar -cvjf ${RALLY_DIR}/${BUILD_NUMBER}.tar.bz2 -C ${RALLY_DIR} --transform "s,^,ci-${BUILD_NUMBER}/," --null -T -
   else
-    echo "Rally directory not present, this should not happen"
+    echo "Rally directory [${RALLY_DIR}] not present. Ensure the RALLY_DIR environment variable is correct"
     exit 1
   fi
 }
 
-if declare -f "$1" > /dev/null; then
+if declare -F "$1" > /dev/null; then
     $1
     exit
 else
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -1,0 +1,32 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+set -eo pipefail
+export PATH="$HOME/.pyenv/bin:$PATH"
+export TERM=dumb
+export LC_ALL=en_US.UTF-8
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+make prereq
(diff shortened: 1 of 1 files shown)


.ci/build.sh
```diff
@@ -0,0 +1,32 @@
...
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+make prereq
+make install
+make precommit
+make it
+
+# this will only be done if the build number variable is present
+if [ ! -z "$BUILD_NUMBER" ] && [ -d ".rally" ]; then
+  tar -cvjf .rally/$BUILD_NUMBER.tar.bzip $( find .rally/ -name "*.log" )
```
dliappis, line 31: You are assuming here (and in the line above) that the CWD is also where [RALLY_HOME](https://github.com/elastic/rally/commit/b4a1211bb374ac8d0a25e75d45adbfb7d12d4353) is right?

Shouldn't `RALLY_HOME` get honored, if set instead of making this assumption?

.ci/build.sh
```diff
@@ -0,0 +1,32 @@
...
+export LC_ALL=en_US.UTF-8
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+make prereq
+make install
+make precommit
+make it
+
+# this will only be done if the build number variable is present
+if [ ! -z "$BUILD_NUMBER" ] && [ -d ".rally" ]; then
```
dliappis, line 30: It think it'd be valueable to be stricter about which shell we use in the shebang and specify `#!/usr/bin/env bash` in line 1 and avoid surprises with `dash` or plain `sh`.

If that's done we could perhaps use the [bash conditionals](https://www.gnu.org/software/bash/manual/bash.html#Conditional-Constructs) `[[` / `]]` for these tests.

.ci/build.sh
hub-cap: much better idea here

.ci/build.sh
hub-cap: wow, what a rookie mistake to forget the shebang.

hub-cap:
ok I have updated the script. The general idea here is that an external caller can execute the build and archive steps independent of each other, so if build fails the caller can still execute archive. this is why it is now 2 functions and the caller must specify the function they wish to run. I have also addressed reviewer comments.

.ci/build.sh
```diff
@@ -0,0 +1,57 @@
...
+function archive {
+  # Treat unset env variables as an error, but only in this function as there are other functions that allow unset variables
+  set -u
+
+  # this will only be done if the build number variable is present
+  RALLY_DIR=${RALLY_HOME}/.rally
+  if [[ -d ${RALLY_DIR} ]]; then
+    find ${RALLY_DIR} -name "*.log" -printf "%P\\0" | tar -cvjf ${RALLY_DIR}/${BUILD_NUMBER}.tar.bz2 -C ${RALLY_DIR} --transform "s,^,ci-${BUILD_NUMBER}/," --null -T -
+  else
+    echo "Rally directory not present, this should not happen"
```
dliappis, line 46: Maybe print the actual value of `${RALLY_DIR}` for troubleshooting too?

.ci/build.sh
```diff
@@ -0,0 +1,57 @@
...
+    exit 1
+  fi
+}
+
+if declare -f "$1" > /dev/null; then
```
dliappis, line 51: We only need to check whether the function name exists right?

According to the man page, `-f` shows also the content which is not needed. Instead we can use `-F` which only shows the function name:

> The -f option will restrict the display to shell functions.  The -F option inhibits the display of function definitions; only the function name and attributes are printed.

Consider the following example:

```
$ function myf {
> echo "sdfsdf"
> }
```

```
$ declare -f myf
myf ()
{
    echo "sdfsdf"
}
```

```
## function exists
$ declare -F myf
myf
$ echo $?
0
```

```
## function not found
$ declare -F myf2
$ echo $?
1
```

hub-cap: nice. I took this from https://github.com/elastic/rally/blob/master/integration-test.sh#L634 , shall I change it there as well?

dliappis: can be done, but in a separate PR/commit plz

PR #958, opened 2020-04-08
Avoid duplicate URL parsing for async connections

danielmitterdorfer:
With this commit we provide a URL object when issuing a request instead
of a string representation to avoid expensive string parsing in aiohttp.
In our tests this has reduced the client side overhead by about one
millisecond which is important when benchmarking queries which finish
within single-digit milliseconds.

Relates #944


danielmitterdorfer:
Thanks for the review!

PR #962, opened 2020-04-14
Increase bulk size to metrics store

danielmitterdorfer:
With this commit we specify an explicit bulk size (called `chunk_size`
by the Elasticsearch Python client) to improve performance when sending
large amounts of metrics documents. After experimenting with different
bulk sizes we have settled for a bulk size of 5000 documents per bulk
(the default is 500 docs per bulk) which improved throughput by a factor
of 2 in our experiments.


danielmitterdorfer:

danielmitterdorfer:
Thanks for the review! :)

Issue #568, opened 2018-09-06
Raise an error when attempting to bulk-index without any corpora


danielmitterdorfer:
source: https://discuss.elastic.co/t/no-throughput-metrics-available-for-bulk-likely-cause-the-benchmark-ended-already-during-warmup/147368/8

The root cause of the problem in the discussion above was that the user has used outdated track syntax and wondered why no documents have been bulk-indexed. We could make this more explicit by raising an error in the bulk parameter source if the list of corpora is empty.

bartier:
Hi! Can I try work on this?

I made a test trying to run Rally like below but I removed the corpora definition for track 'percolator' in my local default repository to force this error:

`./rally --track=percolator --challenge=append-no-conflicts --kill-running-processes --distribution-version 7.6.0`

*~/.rally/benchmarks/tracks/default/percolator/track.json*
```
{% import "rally.helpers" as rally with context %}

{
  "version": 2,
  "description": "Percolator benchmark based on AOL queries",
  "indices": [
    {
      "name": "queries",
      "body": "index.json"
    }
  ],
  "operations": [
    {{ rally.collect(parts="operations/*.json") }}
  ],
  "challenges": [
    {{ rally.collect(parts="challenges/*.json") }}
  ]
}


```

I got the following error:
![image](https://user-images.githubusercontent.com/18057391/79086699-5a3cb500-7d13-11ea-8dd4-7c4752a692dc.png)

Is this a valid way to reproduce this error? If yes, I would propose your validation suggestion in `TrackSpecificationReader#_create_corpora`  if the list of corpora is empty:

```
    def _create_corpora(self, corpora_specs, indices):
        if len(corpora_specs) == 0:
            raise exceptions.TrackConfigError(f"There is no document corpora definition for track {self.name}.")
        document_corpora = []
        known_corpora_names = set()
        ...
        ...

```

With the implementation above I got the TrackConfigError when trying to use a track without any corpora: 
![image](https://user-images.githubusercontent.com/18057391/79086832-e18a2880-7d13-11ea-87b0-7d9010b4e58f.png)

By the way, there is a `None` in the error message and I'm not sure if this is something that could be avoided to the user.

danielmitterdorfer:
It is perfectly fine to define a track without a corpus (for example if you only want to run a query benchmark). I'd instead modify the constructor of `BulkIndexParamSource`. Here it determines which corpora should be used.

[...]

PR #967, opened 2020-04-17
Raise an error when attempting to bulk-index without any corpora

bartier:
This PR attempts to solve the issue #568.

As discussed in the issue, now the `BulkIndexParamSource` constructor checks whether that `self.corpora` is empty and if it is raise `exceptions.InvalidSyntax`.

**Summary of changes:**

- Explicit error is raised when list of corpora is empty and you try to bulk requests to Elasticsearch.
example
Output example when running a Rally race without corpora definition.

- Fix tests that started to fail after the new error check.
- Add new test for this new error check.

```
    def test_create_without_corpora_definition(self):
        with self.assertRaises(exceptions.InvalidSyntax) as ctx:
            params.BulkIndexParamSource(track=track.Track(name="unit-test"), params={})

        self.assertEqual("There is no document corpora definition for track unit-test. "
                         "You must add at least one before making bulk requests to Elasticsearch.", ctx.exception.args[0])
```

I hope this helps :)

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -458,6 +458,10 @@ class BulkIndexParamSource(ParamSource):
 
         self.corpora = self.used_corpora(track, params)
 
+        if len(self.corpora) == 0:
+            raise exceptions.InvalidSyntax(f"There is no document corpora definition for track {track}. You must add at "
+                                           f"least one before making bulk requests to Elasticsearch.")
+
         for corpus in self.corpora:
             for document_set in corpus.documents:
                 if document_set.includes_action_and_meta_data and self.id_conflicts != IndexIdConflict.NoConflicts:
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:

esrally/track/params.py
```diff
@@ -458,6 +458,10 @@ def __init__(self, track, params, **kwargs):
 
         self.corpora = self.used_corpora(track, params)
 
+        if len(self.corpora) == 0:
+            raise exceptions.InvalidSyntax(f"There is no document corpora definition for track {track}. You must add at "
```
danielmitterdorfer, line 462: Nit: "Corpora" is plural but we should use singular here, i.e: 

```
There is no document corpus definition for track {track}".
```

bartier:
@danielmitterdorfer I made the requested changes. Could you check again? BTW, I found a small typo in `runner.py`, I think it would be okay to add to this PR :)

danielmitterdorfer:

PR #968, opened 2020-04-20
Add integration tests with eventdata

danielmitterdorfer:
With this commit we add integration tests with the eventdata track. As
it is not included in the regular Rally tracks, it uses many features of
Rally and we often use it for benchmarks, it makes sense to integrate
testing more tightly.

Note that we do not add tests for all challenges of the eventdata track
as elastic/rally-eventdata-track#82 will remove some of them.

### DIFF
--- a/it/__init__.py
+++ b/it/__init__.py
@@ -27,7 +27,7 @@ from esrally import client
 from esrally.utils import process, io
 
 CONFIG_NAMES = ["in-memory-it", "es-it"]
-DISTRIBUTIONS = ["2.4.6", "5.6.16", "6.8.0", "7.1.1"]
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:
> This looks good. One thing Ive noticed is that we add the on-error flag to all the rally invocations, rather than passing it as a defaulted param that will put it in the it.esrally() invocation. I get that we might want to not fail on some errors, but it seems like its something that could be forgotten by a new test creator.

Thanks for the review and good point with the `on-error` flag. As discussed on another channel let's do this in a separate PR by providing a dedicated method for the invocation of the `race` subcommand as `--on-error=abort` is not applicable to every subcommand (like `info` or `list`).

PR #971, opened 2020-04-21
Add thespian logs to ci

### LINKED ISSUE #129: Improve internal module structure
Before tackling the next feature release, we should improve the module structure. I expect that we will benefit from this restructuring when we add support for distributed benchmarks in #71.

hub-cap:
Closes #129

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -21,24 +21,24 @@
 set -e
 # fail on pipeline errors, e.g. when grepping
 set -o pipefail
+# fail on any unset environment variables
+set -u
 
 function build {
-  if [[ ! -d ${RALLY_HOME} ]] ; then
-    echo "Rally home [${RALLY_HOME}] is not set or does not exist."
-    exit 1
-  fi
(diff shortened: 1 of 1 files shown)


.ci/build.sh
```diff
@@ -23,11 +23,23 @@ set -e
 set -o pipefail
 
 function build {
+  if [[ ! -d ${RALLY_HOME} ]] ; then
+    echo "Rally home [${RALLY_HOME}] is not set or does not exist."
```
dliappis, line 27: What about adding a conditional to check if the var is set, first, to help with troubleshooting?

e.g.

```
if [[ -z ${RALLY_HOME+x} ]]; then
    echo "Env var [RALLY_HOME] is unset."
    exit 1
```

Then the error message on this line can be clearer like:
`echo "\$RALLY_HOME [${RALLY_HOME}] does not exist.`

or something along those lines?

hub-cap: You might want to eyeball this again. I went ahead and just removed the check, as rally will just create the directory if it does not exist, and I ensured its set with a `set -u`, and then changed it to `set +u` so the rest of the commands would work (IIRC it was pyenv that failed if `set -u` is set.

PR #972, opened 2020-04-22
Pass actor system env vars in integration test

danielmitterdorfer:
With this commit we ensure that tox passes environment variables
prefixed with `THESPLOG_` that are related to the actor system. This
allows us to configure custom settings for integration tests.


danielmitterdorfer:
Thanks for the review!

PR #973, opened 2020-04-27
Remove warning about bogus results

danielmitterdorfer:
With the introduction of dedicated subcommands for cluster management it
has become more usual to use the benchmark-only pipeline. Thus, we
remove the warning about bogus results now.


danielmitterdorfer:
Thanks for the review.

> im curious why we decided to remove this

I've written in the PR description:

> With the introduction of dedicated subcommands for cluster management it
has become more usual to use the benchmark-only pipeline

i.e. the assumption is that users will manage clusters via these subcommands and usage of the `benchmark-only` pipeline will become the norm rather than the exception. Also, I expect the integrated command that will do everything will be removed in the near future.

PR #974, opened 2020-04-27
Remove duplicate integration test

danielmitterdorfer:
With this commit we remove an integration test for the `benchmark-only`
pipeline that exists in the old integration test script but is also
already tested in `distribution_test` in any of `test_eventdata*`.

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -302,24 +302,6 @@ function test_distribution_fails_with_wrong_track_params {
     restore_rally_log
 }
 
-function test_benchmark_only {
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
Note for reviewers: These are the tests that I am referring to the in the description that already test the same functionality: https://github.com/elastic/rally/blob/ff29cde84e71bc930d222811fa08806995e4be46/it/distribution_test.py#L59-L88.

danielmitterdorfer:
Thanks for the quick review!

Issue #975, opened 2020-04-27
Migrate all integration tests to new infrastructure


danielmitterdorfer:
In https://github.com/elastic/rally/pull/919 we added a new integration test infrastructure based on `py.test` to replace the existing shell script `integration-test.sh`. While some of the integration tests have already been migrated, a couple of them are still left:

* [x] `test_proxy_connection` https://github.com/elastic/rally/pull/1029
* [x] `test_configure` #1023
* [x] `test_distribution_fails_with_wrong_track_params`: https://github.com/elastic/rally/pull/977
* [x] `test_docker_dev_image` #1026 
* [x] Remove `integration-test.sh` and its invocation in `tox.ini` https://github.com/elastic/rally/pull/1029

bartier:
@danielmitterdorfer Can I try to migrate the test_docker_dev_image tests? I did some tests with the [first one](https://github.com/bartier/rally/blob/test-docker-dev-image/it/docker_dev_image_test.py) and it apparently worked as it should.

danielmitterdorfer:
@bartier sure, thanks for helping out. I think it would be good to get a preliminary review on the migration of the first test and then expand to further tests (in the same PR).

PR #977, opened 2020-04-29
Migrate track param check to new IT infrastructure

danielmitterdorfer:
With this commit we migrate the integration test that checks for
misspelled track parameters to the new infrastructure.

Relates #975

### DIFF
--- a/integration-test.sh
+++ b/integration-test.sh
@@ -256,52 +256,6 @@ function test_configure {
     esrally configure --assume-defaults --configuration-name="config-integration-test"
 }
 
-function test_distribution_fails_with_wrong_track_params {
-    local cfg
-    local distribution
-    # TODO check if randomization of track is possible
-    local track="geonames" # fixed value for now, as the available track params vary between tracks
-    local track_params
-    local defined_track_params
-    local undefined_track_params
-
-    random_configuration cfg
-    random_distribution dist
-
-    undefined_track_params="number_of-replicas:0" # - simulates a typo
-
-    if [[ ${track} == "geonames" ]]; then
-        defined_track_params="conflict_probability:45,"
-    fi
-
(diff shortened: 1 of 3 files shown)


it/info_test.py
```diff
@@ -16,18 +16,31 @@
 # under the License.
 
 import it
+from esrally.utils import process
 
 
-@it.random_rally_config
+@it.rally_in_mem
```
danielmitterdorfer, line 22: I opted to switch all of these tests to the in-memory configuration because it does not matter at all which metrics store implementation we use here.

integration-test.sh
```diff
@@ -256,52 +256,6 @@ function test_configure {
...
-    elif exit_if_docker_not_running && [[ ${ret_code} -ne 0 ]]; then
-        # need to use grep -P which is unavailable with macOS grep
-        if ! docker run --rm -v ${RALLY_LOG}:/rally.log:ro ubuntu:xenial grep -Pzoq '.*CRITICAL Some of your track parameter\(s\) "number_of-replicas" are not used by this track; perhaps you intend to use "number_of_replicas" instead\.\n\nAll track parameters you provided are:\n- conflict_probability\n- number_of-replicas\n\nAll parameters exposed by this track:\n*' /rally.log; then
-            error ${err_msg}
-            exit ${ret_code}
-        fi
-    fi
-    restore_rally_log
-}
-
```
ebadyano, line 304: Should we also remove the call `test_distribution_fails_with_wrong_track_params` at line 457?

danielmitterdorfer: Good catch! We should indeed. I've pushed a fix in https://github.com/elastic/rally/pull/977/commits/76b1ebf2230d65e920bd5e47f9c11a5c3e71d6e6.

danielmitterdorfer:
Thanks for the review!

PR #978, opened 2020-04-29
Remove os environment variables from env prep

### LINKED ISSUE #976: Don't preserve OS's JAVA*_HOME vars before launching Elasticsearch
## Description

When user specifies `--runtime-jdk=bundled` if `JAVA_HOME` is already set then it remains present and honored by Elasticsearch thus preferring `JAVA_HOME` than the bundled jdk (for versions of Elasticsearch coming with a bundled jdk of course)

Instead Rally should unset all JAVA*_HOME vars, specifically `JAVA_HOME`, but also `JAVA{7,8,9,10,11,...}_HOME` as a precaution for the future too.

While at it, let's also fix the weird indentation in https://github.com/elastic/rally/blob/635757e4ff70227c976febbdfeca582c6cb6a944/esrally/mechanic/launcher.py#L192 and https://github.com/elastic/rally/blob/635757e4ff70227c976febbdfeca582c6cb6a944/esrally/mechanic/launcher.py#L194

hub-cap:
The launcher pulls in local environment variables from the OS prior to
launching Elasticsearch. This can be problematic if the $JAVA_HOME
variable is locally set, but explicitly not set by the Rally
launcher. For the case of the bundled JDK, a variable java_home is
explicitly unset, which should be reconciled with the local env by also
unsetting $JAVA_HOME if it exists before launching. This commit ensures
that if java_home is unset in the code, that the local env's $JAVA_HOME
variable is also unset.

Closes #976

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -13,7 +13,7 @@ JAVA_HOME and the bundled runtime JDK
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Rally can optionally use the bundled runtime JDK by setting ``--runtime-jdk="bundled"``. This setting will use the JDK that is bundled with
-elasticsearch and not honor any ``JAVA_HOME`` settings you may have set.
+Elasticsearch and not honor any ``JAVA_HOME`` settings you may have set.
 
 Meta-Data for queries are omitted
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -1,6 +1,16 @@
 Migration Guide
 ===============
 
+
+Migrating to Rally 2.0.0
+------------------------
+
+JAVA_HOME and the bundled runtime JDK
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Rally can optionally use the bundled runtime JDK by setting ``--runtime-jdk="bundled"``. This setting will use the JDK that is bundled with
+elasticsearch and not honor any ``JAVA_HOME`` settings you may have set.
+
 Migrating to Rally 1.5.0
 ------------------------
 
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -169,7 +169,7 @@ class ProcessLauncher:
         return node
 
     def _prepare_env(self, car_env, node_name, java_home, t):
-        env = {k: v for k, v in os.environ.items() if k in self.pass_env_vars or k == "PATH"}
+        env = {k: v for k, v in os.environ.items() if k in self.pass_env_vars}
         env.update(car_env)
         if java_home:
             self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep, prepend=True)
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -175,6 +175,9 @@ class ProcessLauncher:
             self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep, prepend=True)
             # Don't merge here!
             env["JAVA_HOME"] = java_home
+        elif env.get("JAVA_HOME"):
+            # Assume that since java_home was not passed in, it should be unset in the environment if it exists
+            del env["JAVA_HOME"]
         env["ES_JAVA_OPTS"] = "-XX:+ExitOnOutOfMemoryError"
 
         # we just blindly trust telemetry here...
@@ -189,9 +192,9 @@ class ProcessLauncher:
             if k not in env:
                 env[k] = v
             elif prepend:
-                    env[k] = v + separator + env[k]
+                env[k] = v + separator + env[k]
             else:
-                    env[k] = env[k] + separator + v
+                env[k] = env[k] + separator + v
 
     @staticmethod
     def _run_subprocess(command_line, env):
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -23,7 +23,7 @@ import psutil
 
 from esrally import time, exceptions, telemetry
 from esrally.mechanic import cluster, java_resolver
-from esrally.utils import io, process
+from esrally.utils import io, opts, process
 
 
 class DockerLauncher:
@@ -126,6 +126,7 @@ class ProcessLauncher:
         self._clock = clock
         self.keep_running = self.cfg.opts("mechanic", "keep.running")
         self.logger = logging.getLogger(__name__)
+        self.pass_env_vars = opts.csv_to_list(self.cfg.opts("system", "passenv", mandatory=False, default_value="PATH"))
 
     def start(self, node_configurations):
         node_count_on_host = len(node_configurations)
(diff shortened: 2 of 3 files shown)


hub-cap:
@dliappis this commit does not remove any of the JAVAXX_HOME's, but I did not see the usefulness of that inside how we operate w/ rally today. I can do it, but I think that this PR fixes the bug that was present when we moved to the bundled jdk.

esrally/mechanic/launcher.py
```diff
@@ -175,6 +175,9 @@ def _prepare_env(self, car_env, node_name, java_home, t):
             self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep, prepend=True)
             # Don't merge here!
             env["JAVA_HOME"] = java_home
+        elif env.get("JAVA_HOME"):
```
dliappis, line 178: Should we take a cleaner (and bolder) approach and not inherit anything from the env apart from PATH?

Practically speaking I thought about replacing lines 172 and 172 with something like:
`env={k:v for k, v in os.environ.items() if k == "PATH"}`

TBH I can't think of what we need from the shell PATH to launch ES, we might just as well be ok with removing line 172 altogether, but typically all launched Elasticsearches (even in Docker) will have something in PATH so this sounds ok to me.

esrally/mechanic/launcher.py
```diff
@@ -175,6 +175,9 @@ def _prepare_env(self, car_env, node_name, java_home, t):
...
+            # Assume that since java_home was not passed in, it should be unset in the environment if it exists
+            del env["JAVA_HOME"]
```
dliappis, line 180: Could we adjust the test in launcher_test#test_bundled_jdk_not_in_path() to fail with our current code, i.e. actually check that the JAVA_HOME env var remains unset even if set in os.environ?

esrally/mechanic/launcher.py
danielmitterdorfer: I like that approach as it is much cleaner. I can think of situations when we want to set other environment variables like `LD_LIBRARY_PATH` in order set a custom allocator like [jemalloc](https://github.com/jemalloc/jemalloc) to use it for [analyzing native memory leaks](https://github.com/jemalloc/jemalloc/wiki/Use-Case%3A-Leak-Checking). 

I suggest that we leverage our config object here to only pass `PATH` by default but allow to override it with a config property. We should avoid using the config object everywhere and instead derive a property in the constructor of `ProcessLauncher` as follows:

```python
from esrally.utils import opts
#...
self.pass_env_vars = opts.csv_to_list(self.cfg.opts("system", "passenv", mandatory=False, default_value="PATH"))
```

hub-cap: Do we want to overwriting `ES_JAVA_OPTS`? Or append the ExitOnOOM to it? Ill leave it currently as overwriting, unless we want to change the behaviour here.

hub-cap: After chatting w @dliappis, we will honor the user's `ES_JAVA_OPTS` if provided in the rally.ini, otherwise we will add the flag.

hub-cap: Do we want to always glean the `PATH` even if it is not specified in the config? As is if the user adds variables, they might forget the path and then :boom:

hub-cap:
I want to ensure yall are ok w/ the functionality here. I added a check to keep `PATH` on the environment even if it was not explicitly stated in the rally config. I assume we need `PATH` and it would not make sense for a user to have to manually include that every time. If this is an invalid assumption, then please correct me!

esrally/mechanic/launcher.py
danielmitterdorfer: I tend not to apply any special logic for `PATH` and would favor if the user would clearly see what environment variables are passed. I also expect that the default will only be overridden in very special circumstances (like the example I've provided above).

hub-cap: ok I will remove the PATH special case!

tests/mechanic/launcher_test.py
```diff
@@ -247,6 +248,44 @@ def test_bundled_jdk_not_in_path(self):
...
+        cfg.add(config.Scope.application, "mechanic", "keep.running", False)
+        cfg.add(config.Scope.application, "system", "env.name", "test")
+        cfg.add(config.Scope.application, "system", "passenv", "JAVA_HOME,FOO1")
+        os.environ["JAVA_HOME"] = "/path/to/java"
+        os.environ["FOO1"] = "BAR1"
+
+        proc_launcher = launcher.ProcessLauncher(cfg)
+
+        t = telemetry.Telemetry()
+        # no JAVA_HOME -> use the bundled JDK
```
dliappis, line 264: Note: this is with ES >= 7

tests/mechanic/launcher_test.py
```diff
@@ -247,6 +248,44 @@ def test_bundled_jdk_not_in_path(self):
...
+        cfg = config.Config()
+        cfg.add(config.Scope.application, "mechanic", "keep.running", False)
+        cfg.add(config.Scope.application, "system", "env.name", "test")
+        cfg.add(config.Scope.application, "system", "passenv", "ES_JAVA_OPTS")
+        os.environ["ES_JAVA_OPTS"] = "-XX:-someJunk"
+
+        proc_launcher = launcher.ProcessLauncher(cfg)
+
+        t = telemetry.Telemetry()
+        # no JAVA_HOME -> use the bundled JDK
```
dliappis, line 282: Same comment for this comment as above.

hub-cap:
Good call on the JAVA_HOME comments. Would you like this done as part of this PR or a followup?

tests/mechanic/launcher_test.py
hub-cap: This comment is here because we are passing `java_home=None` in to the `_prepare_env` method and is not dependent on a version in the unit test

tests/mechanic/launcher_test.py
hub-cap: https://github.com/elastic/rally/pull/978#discussion_r419581290

hub-cap:
Regarding `JAVA_HOME`, it is still needed to run rally unless you specifically invoke the jvm w/ the bundled param... I unset all my JAVA*_HOME's and got

```
(.venv) ➜  rally git:(del_java_home) esrally --distribution-version=6.5.3

    ____        ____
   / __ \____ _/ / /_  __
  / /_/ / __ `/ / / / / /
 / _, _/ /_/ / / / /_/ /
/_/ |_|\__,_/_/_/\__, /
                /____/

[INFO] Preparing for race ...
[ERROR] Cannot race. ("Install a JDK with one of the versions [11, 10, 9, 8] and point to it with one of ['JAVA11_HOME', 'JAVA10_HOME', 'JAVA9_HOME', 'JAVA8_HOME', 'JAVA_HOME'].", None)
```

hub-cap:
Running with `--runtime-jdk=bundled` using `7.6.0` I got 
```
/home/baz/.rally/benchmarks/races/873a53a5-f7f2-4780-910f-cbb4f471eca5/rally-node-0/install/elasticsearch-7.6.0/jdk/bin/java -Xms1g -Xmx1g -XX:+UseConcMarkSweepGC -
```

And what is interesting, running with `bundled` and a version that had no bundled jdk, rally used `/bin/java` on my machine, which i suspect is a bug, and im going to eyeball it.

hub-cap:
As per @rjernst , the fallback from finding java on the path (which is why `/bin/java` is running) changed at 7.0. Prior to 7.0, ES would look for java on the path, whereas 7.0 and up it falls back to using the bundled JDK. 

Our code, however, allows for a `bundled` jdk to run the path jdk prior to 7.0. And not specifying a jdk version, as well as not having JAVA_HOME specified fails on 7.0 and above. 

Should we change this so the code in 7.0 and above just uses the bundled JDK if the JAVA_HOME variables are not set?

hub-cap:
I have added a note about the `bundled` jdk in the migration docs.

docs/migrate.rst
```diff
@@ -1,6 +1,16 @@
...
 
+
+Migrating to Rally 2.0.0
+------------------------
+
+JAVA_HOME and the bundled runtime JDK
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Rally can optionally use the bundled runtime JDK by setting ``--runtime-jdk="bundled"``. This setting will use the JDK that is bundled with
+elasticsearch and not honor any ``JAVA_HOME`` settings you may have set.
```
dliappis, line 12: nit: capitalization

tests/mechanic/launcher_test.py
dliappis: Ok thanks for clarifying

docs/migrate.rst
hub-cap: this will curse me till the day I die I think.

PR #984, opened 2020-05-04
Add developer info about intellij

hub-cap:


CONTRIBUTING.md
```diff
@@ -38,6 +38,13 @@ If you want to get started in the project, a good idea is to check issues labele
 You will need to fork the Rally repository and clone it to your local machine. See 
 the [Github help page](https://help.github.com/articles/fork-a-repo) for help.
 
+### Importing the project into IntelliJ IDEA
+
+Rally builds using virtualenv. When importing into IntelliJ you will need to define an appropriate Python SDK, which is provided by virtualenv.
+For more details on defining a Python SDK in IntelliJ please refer to their documentation. We recommend using the Python SDK that `make prereq` creates.
```
dliappis, line 44: Maybe we can even me more specific and just say add an "Existing environment" in the `Virtualenv Environment` dialog using `$RALLY_PATH/.venv/bin/python3` after `make prereq && make install` has been run. This should result in something like:

image

CONTRIBUTING.md
```diff
@@ -38,6 +38,13 @@ If you want to get started in the project, a good idea is to check issues labele
...
+
+In order to run tests within the IDE, ensure the `Python Integrated Tools` `Default Test Runner` is set to `pytest`.
```
dliappis, line 46: `Python Integrated Tools` / `Testing` / `Default test runner`

CONTRIBUTING.md
hub-cap: I noticed the existing ES teams info did not include pictures, and I presume its because 1) we need a new folder to store the single pic and 2) that pic is guaranteed to change when intellij shakes up the UI every release. This is why I went with no pics, but I can definitely update the wording a bit.

dliappis: Sounds reasonable to me then.

hub-cap:
@dliappis I added a bit more contextual information if you care to review this again, without adding any pictures. No need to review if you feel its ok though. I do see you already approved it. I wont request re-review, but if you see this comment and wish to review, feel free. Otherwise Ill likely merge tomorrow.

PR #986, opened 2020-05-05
Allow to override GC log parameters

danielmitterdorfer:
With this commit we expose a new telemetry parameter `gc-log-config`
that allows users to override the configuration of the GC telemetry
device. We only allow to override tags and levels but will still output
to a file.

### DIFF
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -67,6 +67,10 @@ gc
 
 The ``gc`` telemetry device enables GC logs for the benchmark candidate. You can use tools like `GCViewer <https://github.com/chewiebug/GCViewer>`_ to analyze the GC logs.
 
+If the runtime JDK is Java 9 or better, the following telemetry parameters can be specified:
+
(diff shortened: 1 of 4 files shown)


docs/telemetry.rst
```diff
@@ -67,6 +67,15 @@ gc
 
 The ``gc`` telemetry device enables GC logs for the benchmark candidate. You can use tools like `GCViewer <https://github.com/chewiebug/GCViewer>`_ to analyze the GC logs.
 
+If the runtime JDK is Java 9 or better, the following telemetry parameters can be specified:
```
hub-cap, line 70: I think `or higher` or `or greater` makes more sense than `or better` here.

danielmitterdorfer: Thanks! I'll change this before I merge.

danielmitterdorfer: Addressed in https://github.com/elastic/rally/pull/986/commits/e4deee50ecb87bf502c6716596e5a3471a8dc2d8.

PR #987, opened 2020-05-06
Validate if car allows for using the bundled JDK

### LINKED ISSUE #985: using runtime-jdk=bundled prior to elasticsearch 7 results in elasticserach using java from the path
Running esrally with `--runtime-jdk=bundled` on a version prior to 7.0.0, does not error out. Instead it just uses java it finds on the PATH, and also does not honor environment variables or the car JDK definitions

```
➜  env|grep JAVA
JAVA10_HOME=/opt/java/jdk-10.0.1
JAVA11_HOME=/opt/java/jdk-11
JAVA12_HOME=/opt/java/jdk-12
JAVA13_HOME=/opt/java/jdk-13.0.1
JAVA7_HOME=/opt/java/jdk1.7.0_80
JAVA8_HOME=/opt/java/jdk1.8.0_172
JAVA9_HOME=/opt/java/jdk-9.0.4
JAVA_HOME=/opt/java/jdk-14.0.1
JAVA14_HOME=/opt/java/jdk-14.0.1
➜  . .venv/bin/activate                                                                                            
(.venv) ➜  esrally --distribution-version=6.8.0 --runtime-jdk=bundled  
```

and in another tab

```
➜  ~ ps aux|grep java

baz      26174  264  1.9 4969716 1281528 ?     Sl   16:12   0:18 /bin/java -Xms1g -Xmx1g -XX:+UseConcMarkSweepGC -XX:CMSInitiatingOccupancyFraction=75 -XX:+UseCMSInitiatingOccupancyOnly -XX:+AlwaysPreTouch -Xss1m -Djava.awt.headless=true -Dfile.encoding=UTF-8 -Djna.nosys=true -XX:-OmitStackTraceInFastThrow -Dio.netty.noUnsafe=true -Dio.netty.noKeySetOptimization=true -Dio.netty.recycler.maxCapacityPerThread=0 -Dlog4j.shutdownHookEnabled=false -Dlog4j2.disable.jmx=true -Djava.io.tmpdir=/tmp/elasticsearch-9416369777544052768 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/home/baz/.rally/benchmarks/races/c8c3e160-9af4-40be-bfdc-80e9d89d85e8/rally-node-0/heapdump -XX:ErrorFile=/home/baz/.rally/benchmarks/races/c8c3e160-9af4-40be-bfdc-80e9d89d85e8/rally-node-0/logs/server/hs_err_pid%p.log -Djava.locale.providers=COMPAT -XX:UseAVX=2 -XX:+ExitOnOutOfMemoryError -Des.path.home=/home/baz/.rally/benchmarks/races/c8c3e160-9af4-40be-bfdc-80e9d89d85e8/rally-node-0/install/elasticsearch-6.8.0 -Des.path.conf=/home/baz/.rally/benchmarks/races/c8c3e160-9af4-40be-bfdc-80e9d89d85e8/rally-node-0/install/elasticsearch-6.8.0/config -Des.distribution.flavor=oss -Des.distribution.type=tar -cp /home/baz/.rally/benchmarks/races/c8c3e160-9af4-40be-bfdc-80e9d89d85e8/rally-node-0/install/elasticsearch-6.8.0/lib/* org.elasticsearch.bootstrap.Elasticsearch -d -p pid          
```

hub-cap:
When the bundled JDK was introduced, it was added to the car config
file. This check uses the value from the car config to ensure that the
bundled JDK is allowed to be used as specified by the config. If not, it
throws an exception.

Closes #985


hub-cap:
@elasticmachine run tests

hub-cap:
@elasticmachine update branch

hub-cap:
@danielmitterdorfer hey friend, this has passed CI checks and I would love a look. Im not 100% happy with where/how the check is, but I dont know if there is much we can do besides rely on a check everywhere this method is used, outside of the actual starting of the nodes. Is there possibly some other validation that is done before rally gets this deep into running?

esrally/mechanic/java_resolver.py
```diff
@@ -39,6 +39,9 @@ def determine_runtime_jdks():
 
     runtime_jdk_versions = determine_runtime_jdks()
     if runtime_jdk_versions[0] == "bundled":
+        if not allow_bundled_jvm:
+            raise exceptions.SystemSetupError("The bundled JDK is not allowed with the selected car(s): {}"
```
danielmitterdorfer, line 43: This decision does not depend on the chosen car but rather on the Elasticsearch version. How about we change the error message to:

```
This Elasticsearch version does not contain a bundled JDK. Please specify a different runtime JDK.
```

esrally/mechanic/java_resolver.py
```diff
@@ -21,7 +21,7 @@
 from esrally.utils import jvm
 
 
-def java_home(car_runtime_jdks, cfg):
+def java_home(car_runtime_jdks, allow_bundled_jvm, cfg):
```
danielmitterdorfer, line 24: How about we refactor the signature a bit? Ideally we could also get rid of the `cfg` object here and pass the chosen runtime JDK directly. I propose the following signature:

```python
def java_home(car_runtime_jdks, specified_runtime_jdk, provides_bundled_jdk)
```

where 

* `specified_runtime_jdk` is derived from `cfg.opts("mechanic", "runtime.jdk")`
* `provides_bundled_jdk` determines whether this Elasticsearch version ships with a bundled JDK

esrally/mechanic/provisioner.py
```diff
@@ -57,10 +54,12 @@ def docker(cfg, car, cluster_settings, ip, http_port, target_root, node_name):
 
 
 class NodeConfiguration:
-    def __init__(self, build_type, car_env, car_runtime_jdks, ip, node_name, node_root_path, binary_path, data_paths):
+    def __init__(self, build_type, car_env, car_runtime_jdks, car_allow_bundled_jvm, ip, node_name, node_root_path,
+                 binary_path, data_paths):
         self.build_type = build_type
         self.car_env = car_env
         self.car_runtime_jdks = car_runtime_jdks
+        self.car_allow_bundled_jvm = convert.to_bool(car_allow_bundled_jvm)
```
danielmitterdorfer, line 62: How about we call this variable `car_provides_bundled_jdk` (and adapt all dependent variables and strings accordingly)?

tests/mechanic/java_resolver_test.py
```diff
@@ -30,7 +30,7 @@ def test_resolves_java_home_for_default_runtime_jdk(self, resolve_jvm_path):
         cfg = config.Config()
         cfg.add(config.Scope.application, "mechanic", "runtime.jdk", None)
 
-        major, java_home = java_resolver.java_home("12,11,10,9,8", cfg)
+        major, java_home = java_resolver.java_home("12,11,10,9,8", True, cfg)
```
danielmitterdorfer, line 33: Can we switch to named parameters here?

esrally/mechanic/provisioner.py
```diff
@@ -57,10 +54,12 @@ def docker(cfg, car, cluster_settings, ip, http_port, target_root, node_name):
 
 
 class NodeConfiguration:
-    def __init__(self, build_type, car_env, car_runtime_jdks, ip, node_name, node_root_path, binary_path, data_paths):
+    def __init__(self, build_type, car_env, car_runtime_jdks, car_provides_bundled_jdk, ip, node_name, node_root_path,
+                 binary_path, data_paths):
         self.build_type = build_type
         self.car_env = car_env
         self.car_runtime_jdks = car_runtime_jdks
+        self.car_provides_bundled_jdk = convert.to_bool(car_provides_bundled_jdk)
```
danielmitterdorfer, line 62: Can we move the conversion (i.e. `convert.to_bool` to the call site(s)?) IMHO we should use the correct type `bool` already in the constructor.

tests/mechanic/java_resolver_test.py
```diff
@@ -30,7 +30,9 @@ def test_resolves_java_home_for_default_runtime_jdk(self, resolve_jvm_path):
         cfg = config.Config()
         cfg.add(config.Scope.application, "mechanic", "runtime.jdk", None)
 
-        major, java_home = java_resolver.java_home("12,11,10,9,8", cfg)
+        major, java_home = java_resolver.java_home("12,11,10,9,8",
+                                                   specified_runtime_jdk=cfg.opts("mechanic", "runtime.jdk"),
```
danielmitterdorfer, line 34: There is no need for the `Config` object anymore and therefore we should simplify the test and only pass the actual value here.

tests/mechanic/java_resolver_test.py
```diff
@@ -42,7 +44,9 @@ def test_resolves_java_home_for_specific_runtime_jdk(self, resolve_jvm_path):
         cfg = config.Config()
         cfg.add(config.Scope.application, "mechanic", "runtime.jdk", 8)
 
-        major, java_home = java_resolver.java_home("12,11,10,9,8", cfg)
+        major, java_home = java_resolver.java_home("12,11,10,9,8",
+                                                   specified_runtime_jdk=cfg.opts("mechanic", "runtime.jdk"),
```
danielmitterdorfer, line 48: There is no need for the `Config` object anymore and therefore we should simplify the test and only pass the actual value here.

tests/mechanic/java_resolver_test.py
```diff
@@ -53,9 +57,22 @@ def test_resolves_java_home_for_bundled_jdk(self):
         cfg = config.Config()
         cfg.add(config.Scope.application, "mechanic", "runtime.jdk", "bundled")
 
-        major, java_home = java_resolver.java_home("12,11,10,9,8", cfg)
+        major, java_home = java_resolver.java_home("12,11,10,9,8",
+                                                   specified_runtime_jdk=cfg.opts("mechanic", "runtime.jdk"),
```
danielmitterdorfer, line 61: There is no need for the `Config` object anymore and therefore we should simplify the test and only pass the actual value here.

tests/mechanic/java_resolver_test.py
```diff
@@ -53,9 +57,22 @@ def test_resolves_java_home_for_bundled_jdk(self):
...
         self.assertEqual(java_home, None)
+
+    def test_disallowed_bundled_jdk(self):
+
+        cfg = config.Config()
+        cfg.add(config.Scope.application, "mechanic", "runtime.jdk", "bundled")
+        cfg.add(config.Scope.application, "mechanic", "car.names", ["default"])
+
+        with self.assertRaises(exceptions.SystemSetupError) as ctx:
+            java_resolver.java_home("12,11,10,9,8", specified_runtime_jdk=cfg.opts("mechanic", "runtime.jdk"))
```
danielmitterdorfer, line 76: There is no need for the `Config` object anymore and therefore we should simplify the test and only pass the actual value here.

tests/mechanic/java_resolver_test.py
```diff
@@ -18,47 +18,35 @@
...
-        cfg.add(config.Scope.application, "mechanic", "runtime.jdk", None)
-
         major, java_home = java_resolver.java_home("12,11,10,9,8",
-                                                   specified_runtime_jdk=cfg.opts("mechanic", "runtime.jdk"),
+                                                   specified_runtime_jdk=None,
                                                    provides_bundled_jdk=True)
 
-        self.assertEqual(major, 12)
-        self.assertEqual(java_home, "/opt/jdk12")
+        self.assertEqual(major, resolve_jvm_path.return_value[0])
```
danielmitterdorfer, line 33: I think the previous implementation was more readable, even if we duplicate code. However, the scope of duplication is within the test method so this is ok IMHO.

tests/mechanic/java_resolver_test.py
```diff
@@ -18,47 +18,35 @@
...
-        cfg.add(config.Scope.application, "mechanic", "runtime.jdk", 8)
-
         major, java_home = java_resolver.java_home("12,11,10,9,8",
-                                                   specified_runtime_jdk=cfg.opts("mechanic", "runtime.jdk"),
+                                                   specified_runtime_jdk=8,
                                                    provides_bundled_jdk=True)
 
-        self.assertEqual(major, 8)
-        self.assertEqual(java_home, "/opt/jdk8")
+        self.assertEqual(major, resolve_jvm_path.return_value[0])
```
danielmitterdorfer, line 43: I think the previous implementation was more readable, even if we duplicate code. However, the scope of duplication is within the test method so this is ok IMHO.

tests/mechanic/java_resolver_test.py
hub-cap: im fine w/ either. I just hate reusing strings in a test method in general, cuz things can get out of sync. I can revert.

hub-cap:
Ive requested review again as I wasnt sure if u were meaning to approve it but just did comment, or not.

PR #988, opened 2020-05-06
Add experimental segment stats telemetry device

danielmitterdorfer:
With this commit we add a new telemetry device that writes segments
stats at the end of a benchmark to a log file. This telemetry device is
currently considered experimental (i.e. it may be removed at any time)
and is thus also intentionally undocumented.


danielmitterdorfer:
Thanks for the quick review!

PR #992, opened 2020-05-11
Cache source artifacts

danielmitterdorfer:
With this commit we add a cache for source artifacts. It can be enabled
by setting `cache = true` in the `source` section of `rally.ini`.
Caching works based on the specified revision. We don't implement an
eviction policy for now but assume that an external process cleans up
cached artifacts.

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -276,12 +276,11 @@ class CachedElasticsearchSourceSupplier:
             # this can be None if the Elasticsearch does not reside in a git repo and the user has only
             # copied all source files. In that case, we cannot resolve a revision hash and thus we cannot cache.
             if self.cached_path:
-                # noinspection PyBroadException
                 try:
                     shutil.copy(original_path, self.cached_path)
                     self.logger.info("Caching artifact in [%s]", self.cached_path)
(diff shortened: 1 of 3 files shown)


esrally/mechanic/supplier.py
```diff
@@ -223,6 +231,58 @@ def __call__(self, *args, **kwargs):
...
+        if self.cached:
+            self.logger.info("Using cached artifact in [%s]", self.cached_path)
+            binaries["elasticsearch"] = self.cached_path
+        else:
+            self.source_supplier.add(binaries)
+            original_path = binaries["elasticsearch"]
+            # this can be None if the Elasticsearch does not reside in a git repo and the user has only
+            # copied all source files. In that case, we cannot resolve a revision hash and thus we cannot cache.
+            if self.cached_path:
+                shutil.copy(original_path, self.cached_path)
```
dliappis, line 279: What is there is some issue, out of disk space, file in use etc? Should we catch `IOError`, log it, and fallback to no caching?

danielmitterdorfer: Thanks! I've pushed https://github.com/elastic/rally/pull/992/commits/937595eaa4661fe2fc2153f8b578a7e48cafb038.

PR #994, opened 2020-05-12
Report number of GCs

danielmitterdorfer:
With this commit we add two new metrics to the command line report to
show the number of young and old GCs caused by a benchmark.


danielmitterdorfer:
> Is the idea to eventually report it in our nightly benchmarks?

If we think it's a useful metric for us we can definitely add it. For now my intention was to make use of this for adhoc investigations.

PR #995, opened 2020-05-12
Don't use range for http and transport port

### LINKED ISSUE #940: Don't use a network range for [http|transport].port when installing Elasticsearch
The new [install subcommand](https://esrally.readthedocs.io/en/stable/command_line_reference.html#install) uses a range for `http.port` and `transport.tcp.port` (e.g. `http.port: 39200-39300`) even if the port is specified with `--http-port`.

We should be less lenient and configure these two properties in `elasticsearch.yml` without ranges.

dliappis:
Currently we use ranges for http and tranport port. This is error prone
if there a left over cluster forcing the Rally launched ES node to
automatically pick the next free port.

Use a specify port for http and transport port.

Closes #940


esrally/mechanic/provisioner.py
```diff
@@ -295,8 +295,8 @@ def variables(self):
             # this is the IP address that the node will be bound to. Rally will bind to the node's IP address (but not to 0.0.0.0). The
             # reason is that we use the node's IP address as subject alternative name in x-pack.
             "network_host": network_host,
-            "http_port": "%d-%d" % (self.http_port, self.http_port + 100),
-            "transport_port": "%d-%d" % (self.http_port + 100, self.http_port + 200),
+            "http_port": str(self.http_port),
```
danielmitterdorfer, line 298: Shouldn't we do the same also in `DockerProvisioner`?

dliappis: I guess this makes sense for consistency; I skipped it because inside a container the starting port will be always available (and we publish only one port externally).

dliappis: Fixed in 02c77bea34e7dbf47e5a0df6eaff5092d80bf60c

dliappis:
Updated docs in 0d8d3edd87d3e30bc08cc4814438f158e6e2a8d7

PR #996, opened 2020-05-13
Ensure artifact cache directory exists

danielmitterdorfer:
With this commit we create the artifact cache directory if it does not
exist yet. Otherwise, the artifact cannot be cached making the cache
ineffective.

Relates #992


danielmitterdorfer:
Thanks!

PR #997, opened 2020-05-19
Allow to abort on fatal errors

danielmitterdorfer:
With this commit we add a third option to the command line parameter
`on-error`. So far it has been possible to either continue or abort on
error. Now we also allow to continue except for errors that indicate
that the cluster is unreachable (e.g. because it died with an OOME).

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -670,7 +670,11 @@ Client certificates can be presented regardless of the ``verify_certs`` setting,
 ``on-error``
 ~~~~~~~~~~~~
 
-This option controls whether Rally will ``continue`` or ``abort`` when a request error occurs. By default, Rally will just record errors and report the error rate at the end of a race. With ``--on-error=abort``, Rally will immediately abort the race on the first error and print a detailed error message.
+This option controls how Rally behaves when a response error occurs. The following values are possible:
+
+* ``continue``: only records that an error has happened and will continue with the benchmark. At the end of a race, errors show up in the "error rate" metric.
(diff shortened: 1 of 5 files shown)


docs/command_line_reference.rst
```diff
@@ -670,7 +670,11 @@ Client certificates can be presented regardless of the ``verify_certs`` setting,
 ``on-error``
 ~~~~~~~~~~~~
 
-This option controls whether Rally will ``continue`` or ``abort`` when a request error occurs. By default, Rally will just record errors and report the error rate at the end of a race. With ``--on-error=abort``, Rally will immediately abort the race on the first error and print a detailed error message.
+This option controls how Rally behaves when a response error occurs. The following values are possible:
+
+* ``continue``: only records that an error has happened and will continue with the benchmark. At the end of a race, errors show up in the "error rate" metric.
+* ``continue-on-non-fatal`` (default): Behaves as ``continue`` but aborts the benchmark immediately on all fatal errors. At the moment a refused connection is considered fatal. All other errors are considered non-fatal.
```
dliappis, line 676: I had to reread a few times the second sentence (`At the moment a refused connection is considered fatal.`). I don't think we need to explain that this may (or not) change in the future.

I suggest we just state that the only fatal error is receiving "Connection Refused"

(`ECONNREFUSED`)  / (http://man7.org/linux/man-pages/man2/connect.2.html)

danielmitterdorfer: Thanks! Addressed in https://github.com/elastic/rally/pull/997/commits/9925e35c0ba680f47183d2482872638facf31431.

PR #998, opened 2020-05-20
Cache source artifacts by default

danielmitterdorfer:
With this commit we add the artifact cache introduced by #992 also for
plugins and implement a cache cleanup mechanism that cleans up artifacts
after seven days by default (can be configured with `cache.days`). As
artifact caching can now replace the pipeline `from-sources-skip-build`
we deprecate it along with `from-sources-complete` and introduce a new
pipeline `from-sources` instead.


danielmitterdorfer:
Thanks for the reviews! :)

PR #999, opened 2020-05-27
Invoke pytest consistently

danielmitterdorfer:
With this commit we always invoke the `pytest` binary and eliminate the
other variations. We also remove the obsolete `pytest-runner` and fix a
small issue with the docs environment regarding a missing whitelist.


danielmitterdorfer:
Thanks!

Issue #991, opened 2020-05-11
Add config reference documentation


danielmitterdorfer:
We currently document how users can customize Rally with the `config` subcommand in https://esrally.readthedocs.io/en/stable/configuration.html. However, we are missing a reference documentation where we can also document settings that can be specified manually, e.g. `probing.url` (introduced in https://github.com/elastic/rally/pull/909).

Note: Use `git grep "opts.*mandatory=False"` to find these settings (will return a few false positives though).

PR #1001, opened 2020-05-28
Allow to downsample request metrics

danielmitterdorfer:
With this commit we introduce a new setting
`metrics.request.downsample.factor` that allows to control that only
every N-th request metric sample is recorded in the metrics store (this
does not affect throughput calculation which will still consider every
sample). This allows to reduce the number of samples which can be useful
when benchmarks are run with many hundreds or thousands of clients which
would produce a very high number of samples.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -362,8 +362,8 @@ class Driver:
         self.quiet = False
         self.allocations = None
         self.raw_samples = []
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Note that there are no docs for this; this will be documented once we have a config reference docs page (https://github.com/elastic/rally/issues/991).

PR #1002, opened 2020-05-28
Remove experimental race-async subcommand

danielmitterdorfer:
With this commit we remove the experimental subcommand `race-async` as
all the required functionality has been integrated into Rally's regular
`race` subcommand (including support for multiple CPU core and multiple
machines).


ebadyano:
Should we also remove `race-async` reference in esbench?

PR #1003, opened 2020-05-29
Add network troubleshooting tips

danielmitterdorfer:
With this commit we improve the documentation on network issues with the
Rally daemon and also add a warning to avoid benchmarking via a VPN.


danielmitterdorfer:
Thanks for the feedback! I've mentioned now that the IP address needs to be routable.

PR #1006, opened 2020-05-29
Add info about querying challenges in the docs

### LINKED ISSUE #990: Clarify that indexing related sections are not required in a track
Background: https://discuss.elastic.co/t/rally-custom-benchmark-only-pipeline-track-without-corpora/231878?u=dliappis

We should clarify that the [track sections](https://esrally.readthedocs.io/en/stable/track.html#anatomy-of-a-track) `indices`, `corpora`, `templates` are not required, thus a query-only track can be used as shown in [this example](https://esrally.readthedocs.io/en/stable/track.html#a-track-with-a-single-task).

hub-cap:
Closes  #990

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -123,7 +123,7 @@ In the ``operations`` section you describe which operations are available for th
 
 In the ``schedule`` section you describe the workload for the benchmark, for example index with two clients at maximum throughput while searching with another two clients with ten operations per second. The schedule either uses the operations defined in the ``operations`` block or defines the operations to execute inline.
 
-In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test more than one set of scenarios. This section is optional, and more information on can be found in the :ref:`challenges section<track_challenge>`.
+In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test more than one set of scenarios. This section is optional, and more information can be found in the :ref:`challenges section<track_challenge>`.
 
 Creating a track does not require all of the above sections to be used. Tracks that are used against existing data may only rely on querying ``operations`` and can omit the ``indices``, ``templates``, and ``corpora`` sections. An example of this can be found in the :ref:`task with a single track example<track_single_task>`.
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -123,9 +123,9 @@ In the ``operations`` section you describe which operations are available for th
 
 In the ``schedule`` section you describe the workload for the benchmark, for example index with two clients at maximum throughput while searching with another two clients with ten operations per second. The schedule either uses the operations defined in the ``operations`` block or defines the operations to execute inline.
 
-In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test than one set of scenarios. If only one set of operations is needed, they can be specified in ``schedule`` instead.
+In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test more than one set of scenarios. This section is optional, and more information on ``challenges`` :ref:`can be found here <track_challenge>`.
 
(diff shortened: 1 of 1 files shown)


docs/track.rst
```diff
@@ -122,6 +123,10 @@ In the ``operations`` section you describe which operations are available for th
 
 In the ``schedule`` section you describe the workload for the benchmark, for example index with two clients at maximum throughput while searching with another two clients with ten operations per second. The schedule either uses the operations defined in the ``operations`` block or defines the operations to execute inline.
 
+In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test than one set of scenarios. If only one set of operations is needed, they can be specified in ``schedule`` instead.
```
dliappis, line 126: Two comments:

1) as you figured out, [challenges are optional](https://esrally.readthedocs.io/en/stable/adding_tracks.html?highlight=challenges#challenges); that's why they aren't in this list. I thin here we should just mention them as optional and link to the [docs](https://esrally.readthedocs.io/en/stable/adding_tracks.html?highlight=challenges#challenges) as it also explains when they could be useful.

2) `needs to test than one set of scenarios` -> `needs to test` **more** `than one set of scenarios`

docs/track.rst
```diff
@@ -122,6 +123,10 @@ In the ``operations`` section you describe which operations are available for th
...
+
+Creating a track does not require all of the above sections to be used. Tracks that are used against existing data may only rely on querying ``challenges`` and can omit the ``indices``, ``templates``, and ``corpora`` sections.
```
dliappis, line 128: As `challenges` are optional, I think we shouldn't reference it in the query scenario; we could just say `may only rely on querying operations`?

docs/track.rst
hub-cap: So are u saying put a note here but dont include them in the bullet list above?

dliappis: As it's optional, I'd tend to yes, just a note.

docs/track.rst
```diff
@@ -122,6 +123,10 @@ In the ``operations`` section you describe which operations are available for th
 
 In the ``schedule`` section you describe the workload for the benchmark, for example index with two clients at maximum throughput while searching with another two clients with ten operations per second. The schedule either uses the operations defined in the ``operations`` block or defines the operations to execute inline.
 
+In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test more than one set of scenarios. This section is optional, and more information on ``challenges`` :ref:`can be found here <track_challenge>`.
+
+Creating a track does not require all of the above sections to be used. Tracks that are used against existing data may only rely on querying ``operations`` and can omit the ``indices``, ``templates``, and ``corpora`` sections.
```
dliappis, line 128: And ideally link to https://esrally.readthedocs.io/en/stable/track.html#a-track-with-a-single-task which shows a nice query example.

docs/track.rst
```diff
@@ -122,6 +123,10 @@ In the ``operations`` section you describe which operations are available for th
 
 In the ``schedule`` section you describe the workload for the benchmark, for example index with two clients at maximum throughput while searching with another two clients with ten operations per second. The schedule either uses the operations defined in the ``operations`` block or defines the operations to execute inline.
 
+In the ``challenges`` section you describe more than one set of operations, in the event your track needs to test more than one set of scenarios. This section is optional, and more information on ``challenges`` :ref:`can be found here <track_challenge>`.
```
dliappis, line 126: This is a matter of personal preference, I am not a huge fan of the `can be found here` expression in official docs; I'd have said:

```
and more information can be found in the :ref:`challenges section<track_challenge>`. 
```

PR #1008, opened 2020-06-04
Allow to specify a runtime-jdk during installation

danielmitterdorfer:
The `install` subcommand sometimes requires to run so called install
hooks, for example to install plugins. These are installed with
Elasticsearch's `plugin` script which requires Java. Yet, if we rely on
the bundled JDK and no other JDK is installed on the system, the `java`
binary cannot be resolved because no runtime JDK is specified at all.
With this commit we expose the command line option `--runtime-jdk` so
users can provide `bundled` in the installation phase.

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -283,6 +283,11 @@ def create_arg_parser():
         "--team-path",
         help="Define the path to the car and plugin configurations to use.")
     install_parser.add_argument(
+        "--runtime-jdk",
+        type=runtime_jdk,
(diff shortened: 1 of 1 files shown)


hub-cap:
@danielmitterdorfer should this be set to `bundled` by default instead of None?

danielmitterdorfer:
@hub-cap not in my opinion because we can't guarantee that as default (there are versions without a bundled JDK + there are pre 7.0.0 versions) and this is consistent with how we handle it with other subcommands.

PR #1009, opened 2020-06-08
Remove outdated config migrations

danielmitterdorfer:
Rally can automatically migrate old versions of its configuration file.
However, the last of these migrations was added two years ago and Rally
versions that are in use should have been updated long ago. Hence, we
remove the outdated migrations while keeping the infrastructure still in
place so we can plug in new migrations when needed.


danielmitterdorfer:
🙇

PR #1010, opened 2020-06-08
Avoid redefining arguments

danielmitterdorfer:
With this commit we reenable the pylint check `R1704` which checks
whether arguments have been redefined as local variables.

Relates #838


danielmitterdorfer:
Thanks for the quick review. :)

PR #1011, opened 2020-06-08
add support for transform API's

hendrikmuhs:
add support for [transform](https://www.elastic.co/guide/en/elasticsearch/reference/master/transform-apis.html), consisting of the following parts:

 - create, start, stop, delete API support
 - transform-stats telemetry device to collect transform stats
 - basic command-line reporting
 - some basic documentation

Note: I will add definitions for the recorded metrics later after the 1st round of reviews.

### DIFF
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -134,3 +134,13 @@ Supported telemetry parameters:
 
 * ``ccr-stats-indices`` (default: all indices): An index pattern for which ccr stats should be checked.
 * ``ccr-stats-sample-interval`` (default 1): A positive number greater than zero denoting the sampling interval in seconds.
+
+transform-stats
+---------------
+
+The transform-stats telemetry device regularly calls the `transform stats API <https://www.elastic.co/guide/en/elasticsearch/reference/current/get-transform-stats.html>` and records one metrics document per transform.
+
+Supported telemetry parameters:
+
+* ``transform-stats-transforms`` (default: all transforms): A list of transforms per cluster for which transform stats should be checked.
+* ``transform-stats-sample-interval`` (default 1): A positive number greater than zero denoting the sampling interval in seconds.
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -25,7 +25,6 @@ You probably want to gain additional insights from a race. Therefore, we have ad
    node-stats       Node Stats             Regularly samples node stats
    recovery-stats   Recovery Stats         Regularly samples shard recovery stats
    ccr-stats        CCR Stats              Regularly samples Cross Cluster Replication (CCR) related stats
-   segment-stats    Segment Stats          Determines segment stats at the end of the benchmark.
    transform-stats  Transform Stats        Regularly samples transform stats
 
    Keep in mind that each telemetry device may incur a runtime overhead which can skew results.
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -18,13 +18,15 @@ You probably want to gain additional insights from a race. Therefore, we have ad
 
    Command         Name                   Description
    --------------  ---------------------  --------------------------------------------------------------------
-   jit             JIT Compiler Profiler  Enables JIT compiler logs.
-   gc              GC log                 Enables GC logs.
-   jfr             Flight Recorder        Enables Java Flight Recorder (requires an Oracle JDK or OpenJDK 11+)
-   heapdump        Heap Dump              Captures a heap dump.
-   node-stats      Node Stats             Regularly samples node stats
-   recovery-stats  Recovery Stats         Regularly samples shard recovery stats
-   ccr-stats       CCR Stats              Regularly samples Cross Cluster Replication (CCR) related stats
+   jit              JIT Compiler Profiler  Enables JIT compiler logs.
+   gc               GC log                 Enables GC logs.
+   jfr              Flight Recorder        Enables Java Flight Recorder (requires an Oracle JDK or OpenJDK 11+)
+   heapdump         Heap Dump              Captures a heap dump.
+   node-stats       Node Stats             Regularly samples node stats
+   recovery-stats   Recovery Stats         Regularly samples shard recovery stats
+   ccr-stats        CCR Stats              Regularly samples Cross Cluster Replication (CCR) related stats
+   segment-stats    Segment Stats          Determines segment stats at the end of the benchmark.
+   transform-stats  Transform Stats        Regularly samples transform stats
 
    Keep in mind that each telemetry device may incur a runtime overhead which can skew results.
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1209,7 +1209,7 @@ execute-transform
 With the operation ``execute-transform`` you can execute a full run of data transformation, it starts the transform and stops it once all data has been transformed. The transform must be created upfront. The ``execute-transform`` operation supports the following parameters:
 
 * ``transform-id`` (mandatory): The id of the transform to execute.
-* ``poll-interval`` (optional, defaults to `0.5`) How often transform stats are polled, used to set progress and check the state.
+* ``poll-interval`` (optional, defaults to `0.5`) How often transform stats are polled, used to set progress and check the state. You should not set this too low, because polling can skew the result.
 * ``transform-timeout`` (optional, defaults to `1800` (`1h`)) Overall runtime timeout of the batch transform in seconds.
 
 This operation requires at least Elasticsearch 7.5.0 (non-OSS).
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1151,6 +1151,69 @@ With the operation ``wait-for-recovery`` you can wait until an ongoing shard rec
 
 This operation is :ref:`retryable <track_operations>`.
 
+create-transform
+~~~~~~~~~~~~~~~~
+
+With the operation ``create-transform`` you can execute the `create transform API <https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html>`_. It supports the following parameters:
+
+* ``transform-id`` (mandatory): The id of the transform to create.
+* ``body`` (mandatory): Request body containing the configuration of the transform. Please see the `create transform API <https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html>`__ documentation for more details.
+* ``defer-validation`` (optional, defaults to false): When true, deferrable validations are not run. This behavior may be desired if the source index does not exist until after the transform is created.
+
+This operation requires at least Elasticsearch 7.5.0 (non-OSS). This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+This operation is :ref:`retryable <track_operations>`.
+
+start-transform
+~~~~~~~~~~~~~~~
+
+With the operation ``start-transform`` you can execute the `start transform API <https://www.elastic.co/guide/en/elasticsearch/reference/current/start-transform.html>`_. It supports the following parameters:
+
+* ``transform-id`` (mandatory): The id of the transform to start.
+* ``timeout`` (optional, defaults to empty): Amount of time to wait until a transform starts.
+
+This operation requires at least Elasticsearch 7.5.0 (non-OSS). This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+This operation is :ref:`retryable <track_operations>`.
+
+stop-transform
+~~~~~~~~~~~~~~
+
+With the operation ``stop-transform`` you can execute the `stop transform API <https://www.elastic.co/guide/en/elasticsearch/reference/current/stop-transform.html>`_. It supports the following parameters:
+
+* ``transform-id`` (mandatory): The id of the transform to stop.
+* ``force`` (optional, defaults to false): Whether to forcefully stop the transform.
+* ``timeout`` (optional, defaults to empty): Amount of time to wait until a transform stops.
+* ``wait-for-completion`` (optional, defaults to false) If set to true, causes the API to block until the indexer state completely stops.
+* ``wait-for-checkpoint`` (optional, defaults to false) If set to true, the transform will not completely stop until the current checkpoint is completed.
+
+This operation requires at least Elasticsearch 7.5.0 (non-OSS). This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+This operation is :ref:`retryable <track_operations>`.
+
+delete-transform
+~~~~~~~~~~~~~~~~
+
+With the operation ``delete-transform`` you can execute the `delete transform API <https://www.elastic.co/guide/en/elasticsearch/reference/current/delete-transform.html>`_. It supports the following parameters:
+
+* ``transform-id`` (mandatory): The id of the transform to delete.
+* ``force`` (optional, defaults to false): Whether to delete the transform regardless of its current state.
+
+This operation requires at least Elasticsearch 7.5.0 (non-OSS). This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+This operation is :ref:`retryable <track_operations>`.
+
+execute-transform
+~~~~~~~~~~~~~~~~~
+
+With the operation ``execute-transform`` you can execute a full run of data transformation, it starts the transform and stops it once all data has been transformed. The transform must be created upfront. The ``execute-transform`` operation supports the following parameters:
+
+* ``transform-id`` (mandatory): The id of the transform to execute.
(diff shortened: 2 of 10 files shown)


esrally/driver/runner.py
```diff
@@ -1535,6 +1539,76 @@ def __repr__(self, *args, **kwargs):
...
+class CreateTransform(Runner):
+    """
+    Execute the `create transform API https://www.elastic.co/guide/en/elasticsearch/reference/current/put-transform.html`_.
+    """
+
+    async def __call__(self, es, params):
+        transform_id = mandatory(params, "transform-id", self)
+        body = mandatory(params, "body", self)
+        defer_validation = params.get("defer_validation", False)
+        await es.transform.put_transform(transform_id=transform_id, body=body, defer_validation=defer_validation)
```
hendrikmuhs, line 1551: 2 questions (applies to all added runners):

1. according to Runner this should return e.g. 

```
return 1, "ops"
```

However there are also examples of returning complex structures (dicts). What's the right way to do it?

2. if the track contains an error in the config, rally still succeeds and I get no useful information in the logs (only the response code). 

How should error handling be implemented? Shall I `catch/log/raise`?

danielmitterdorfer: To question 1: It depends whether you want to add additional metadata. See the [docs](https://esrally.readthedocs.io/en/stable/adding_tracks.html#custom-runners) on possible return values. The additional return values are only useful if you configure an Elasticsearch metrics store because the metadata are attached to each metrics sample. If you only use the in-memory metrics store (the implicit default) it's not relevant.

> 2. if the track contains an error in the config, rally still succeeds and I get no useful information in the logs (only the response code).

The reason for this is that Rally continues on any errors by default and only records that an error has happened. If you configure an Elasticsearch metrics store you can see the full error details but they are not recorded in the logs (maybe they should). If you want to be more strict, you can specify `--on-error=abort` when you start the benchmark and Rally will immediately stop the benchmark and print the detailed error message it got from Elasticsearch.

> How should error handling be implemented? Shall I catch/log/raise?

Unless you need vey special handling, like calling the API differently in case the first call fails, there is no need for this. Just let the error bubble up and Rally will handle it as I've described above.

hendrikmuhs:
I realized that for the purpose of benchmarking a transform, it makes more sense to have a dedicated `execute-transform` operation that starts, stops the transform and collects the stats.

It seems still useful to have the single start and stop operations, so I do not see it as replacement but as addition. However, I am not sure about the telemetry device, anymore.

docs/telemetry.rst
```diff
@@ -134,3 +134,13 @@ Supported telemetry parameters:
 
 * ``ccr-stats-indices`` (default: all indices): An index pattern for which ccr stats should be checked.
 * ``ccr-stats-sample-interval`` (default 1): A positive number greater than zero denoting the sampling interval in seconds.
+
+transform-stats
```
ebadyano, line 138: we may also want to add `transform-stats` under `Available telemetry devices:` section above, line 27

esrally/driver/runner.py
```diff
@@ -65,6 +65,11 @@ def register_default_runners():
     register_runner(track.OperationType.CreateSnapshotRepository.name, Retry(CreateSnapshotRepository()), async_runner=True)
     register_runner(track.OperationType.WaitForRecovery.name, Retry(IndicesRecovery()), async_runner=True)
     register_runner(track.OperationType.PutSettings.name, Retry(PutSettings()), async_runner=True)
+    register_runner(track.OperationType.CreateTransform.name, Retry(CreateTransform()), async_runner=True)
```
ebadyano, line 68: Could you please add documentation for thenew operations in [track.rst](https://github.com/elastic/rally/blob/master/docs/track.rst#operations)?

esrally/driver/runner.py
```diff
@@ -1563,6 +1568,169 @@ def __repr__(self, *args, **kwargs):
...
+                                          force=force,
+                                          timeout=timeout,
+                                          wait_for_completion=wait_for_completion,
+                                          wait_for_checkpoint=wait_for_checkpoint)
+
+    def __repr__(self, *args, **kwargs):
+        return "stop-transform"
+
+
+class ExecuteTransform(Runner):
```
ebadyano, line 1625: Could you please add tests for new runner classes in https://github.com/elastic/rally/blob/master/tests/driver/runner_test.py?

tests/driver/runner_test.py
```diff
@@ -3134,6 +3134,130 @@ async def test_put_settings(self, es):
                 "indices.recovery.max_bytes_per_sec": "20mb"
             }
         })
+class CreateTransformTests(TestCase):
```
ebadyano, line 3137: nit: Could you please add two new lines before the class declaration?

docs/telemetry.rst
```diff
@@ -18,13 +18,15 @@ You probably want to gain additional insights from a race. Therefore, we have ad
...
-   recovery-stats  Recovery Stats         Regularly samples shard recovery stats
-   ccr-stats       CCR Stats              Regularly samples Cross Cluster Replication (CCR) related stats
+   jit              JIT Compiler Profiler  Enables JIT compiler logs.
+   gc               GC log                 Enables GC logs.
+   jfr              Flight Recorder        Enables Java Flight Recorder (requires an Oracle JDK or OpenJDK 11+)
+   heapdump         Heap Dump              Captures a heap dump.
+   node-stats       Node Stats             Regularly samples node stats
+   recovery-stats   Recovery Stats         Regularly samples shard recovery stats
+   ccr-stats        CCR Stats              Regularly samples Cross Cluster Replication (CCR) related stats
+   segment-stats    Segment Stats          Determines segment stats at the end of the benchmark.
```
ebadyano, line 28: Is `segment-stats` is from a different pr?

docs/track.rst
```diff
@@ -1151,6 +1151,69 @@ With the operation ``wait-for-recovery`` you can wait until an ongoing shard rec
...
+
+This operation is :ref:`retryable <track_operations>`.
+
+execute-transform
+~~~~~~~~~~~~~~~~~
+
+With the operation ``execute-transform`` you can execute a full run of data transformation, it starts the transform and stops it once all data has been transformed. The transform must be created upfront. The ``execute-transform`` operation supports the following parameters:
+
+* ``transform-id`` (mandatory): The id of the transform to execute.
+* ``poll-interval`` (optional, defaults to `0.5`) How often transform stats are polled, used to set progress and check the state.
```
ebadyano, line 1212: Does it make sense to specify the time unit for `poll-interval`? i.e. `0.5` (seconds)?

ebadyano: Also, would the poll-interval also affect the performance? Not sure if we want to mention that.

ebadyano:
What is the difference between what transform-stats telemetry device and execute-transform do?

docs/telemetry.rst
hendrikmuhs: yes, seems to have been forgotten in #988

docs/track.rst
hendrikmuhs: Yes, it might affect performance, but I can't say how much. It shouldn't be expensive.

FWIW: I debugged the progress reporting and it seems progress isn't working in rally. Is there an existing issue? Is there a plan to fix this? If not, it might be better to remove the progress report.

hendrikmuhs:
> What is the difference between what transform-stats telemetry device and execute-transform do?

see my previous comment: https://github.com/elastic/rally/pull/1011#issuecomment-641390551

`execute-transform` and the `transform telemetry device` measure the same metrics. After some thought, I still think we need both:

The `execute-transform` is operation is useful for measuring the performance of single transform executions, e.g. to test optimizations in either the transform core or in the underlying aggregation framework. These performance tests would be one-off executions.

The telemetry device is useful for testing continuous mode in transform. This is where something like the [event generator](https://github.com/elastic/rally-eventdata-track) comes into play. This is useful to test whether a cluster setup can handle a certain load of incoming data.

So, despite my previous comment, I now think we should have both.

docs/telemetry.rst
ebadyano: I think since segment-stats is experimental (based on the title in #988) we may want to document it yet.

hendrikmuhs: :thinking: I can remove the `segment-stats` line, however this feels strange to me, because this is what you get as output on the console.

hendrikmuhs: removed `segment-stats`

esrally/telemetry.py
```diff
@@ -743,6 +743,146 @@ def sample(self):
...
+        """
+        Collect Transform stats for transforms (optionally) specified in telemetry parameters and push to metrics store.
+        """
+
+        # ES returns all stats values in bytes or ms via "human: false"
+
+        import elasticsearch
+
+        try:
+            stats = self.client.transform.get_transform_stats('_all')
```
ebadyano, line 842: nit: as a convention we stick to double quotes and not single quotes.

esrally/telemetry.py
```diff
@@ -743,6 +743,146 @@ def sample(self):
...
+        except elasticsearch.TransportError as e:
+            msg = "A transport error occurred while collecting transform stats on " \
+                  "cluster [{}]".format(self.cluster_name)
+            self.logger.exception(msg)
+            raise exceptions.RallyError(msg)
+
+        for transform in stats['transforms']:
+            try:
+                if self.transforms and transform['id'] not in self.transforms:
+                    # Skip metrics for transform not part of user supplied whitelist (transform-stats-transforms)
```
ebadyano, line 853: nit: please use double quotes and not single quotes.

esrally/telemetry.py
```diff
@@ -743,6 +743,146 @@ def sample(self):
...
+        try:
+            stats = self.client.transform.get_transform_stats('_all')
+
+        except elasticsearch.TransportError as e:
+            msg = "A transport error occurred while collecting transform stats on " \
+                  "cluster [{}]".format(self.cluster_name)
+            self.logger.exception(msg)
+            raise exceptions.RallyError(msg)
+
+        for transform in stats['transforms']:
```
ebadyano, line 850: nit: please use double quotes and not single quotes.

esrally/telemetry.py
```diff
@@ -743,6 +743,146 @@ def sample(self):
...
+
+        # ES returns all stats values in bytes or ms via "human: false"
+
+        import elasticsearch
+
+        try:
+            stats = self.client.transform.get_transform_stats('_all')
+
+        except elasticsearch.TransportError as e:
+            msg = "A transport error occurred while collecting transform stats on " \
```
ebadyano, line 845: Could we please use f-strings here?

esrally/telemetry.py
```diff
@@ -743,6 +743,146 @@ def sample(self):
...
+            raise exceptions.SystemSetupError(
+                "The telemetry parameter 'transform-stats-sample-interval' must be greater than zero but was {}.".format(
+                    self.sample_interval))
+        self.specified_cluster_names = self.clients.keys()
+        self.transforms_per_cluster = self.telemetry_params.get("transform-stats-transforms", False)
+        if self.transforms_per_cluster:
+            for cluster_name in self.transforms_per_cluster.keys():
+                if cluster_name not in clients:
+                    raise exceptions.SystemSetupError(
+                        "The telemetry parameter 'transform-stats-transforms' must be a JSON Object with keys matching "
```
ebadyano, line 777: Could you please use f-strings here?

tests/telemetry_test.py
```diff
@@ -1755,6 +1760,111 @@ def test_exception_when_include_indices_metrics_not_valid(self):
...
+            mock.call("transform_documents_indexed", 3),
+            mock.call("transform_index_total", 6),
+            mock.call("transform_index_failures", 7),
+            mock.call("transform_search_total", 9),
+            mock.call("transform_search_failures", 10),
+            mock.call("transform_processing_total", 12)
+        ])
+
+        metrics_store_put_value.assert_has_calls([
+            mock.call("transform_search_time_in_ms", 8, 'ms'),
```
ebadyano, line 1862: nit: could you please use double quotes here?

esrally/driver/runner.py
```diff
@@ -1563,6 +1568,169 @@ def __repr__(self, *args, **kwargs):
...
+            if (time.time() - start_time) > transform_timeout:
+                raise exceptions.RallyAssertionError(
+                    "Transform [{}] timed out after [{}] seconds. "
+                    "Please consider increasing the timeout in the track.".format(
+                        transform_id, transform_timeout))
+            await asyncio.sleep(poll_interval)
+            stats_response = await es.transform.get_transform_stats(transform_id=transform_id)
+            state = stats_response['transforms'][0].get("state")
+
+            self._percent_completed = stats_response['transforms'][0].get("checkpointing", {}).get("next", {}).get(
```
ebadyano, line 1682: nit: as a convention we stick to double quotes and not single quotes.

esrally/driver/runner.py
```diff
@@ -1563,6 +1568,169 @@ def __repr__(self, *args, **kwargs):
...
+                "checkpoint_progress", {}).get("percent_complete", 0.0)
+
+        if state == "failed":
+            raise exceptions.RallyAssertionError(
+                "Transform [{}] failed with [{}].".format(transform_id,
```
ebadyano, line 1687: nit: double quotes around transforms

esrally/telemetry.py
```diff
@@ -743,6 +743,146 @@ def sample(self):
...
+        :param metrics_store: The configured metrics store we write to.
+        """
+        super().__init__()
+
+        self.telemetry_params = telemetry_params
+        self.clients = clients
+        self.sample_interval = telemetry_params.get("transform-stats-sample-interval", 1)
+        if self.sample_interval <= 0:
+            raise exceptions.SystemSetupError(
+                "The telemetry parameter 'transform-stats-sample-interval' must be greater than zero but was {}.".format(
```
ebadyano, line 769: Could you please use f-strings here?

esrally/driver/runner.py
```diff
@@ -1563,6 +1568,169 @@ def __repr__(self, *args, **kwargs):
...
+        start_time = time.time()
+        await es.transform.start_transform(transform_id=transform_id)
+
+        stats_response = await es.transform.get_transform_stats(transform_id=transform_id)
+        state = stats_response['transforms'][0].get("state")
+
+        while state in ('started', 'indexing'):
+            if (time.time() - start_time) > transform_timeout:
+                raise exceptions.RallyAssertionError(
+                    "Transform [{}] timed out after [{}] seconds. "
```
ebadyano, line 1675: Could you please use f-strings here?

esrally/driver/runner.py
```diff
@@ -1563,6 +1568,169 @@ def __repr__(self, *args, **kwargs):
...
+        """
+        import time
+        transform_id = mandatory(params, "transform-id", self)
+        poll_interval = params.get("poll-interval", 0.5)
+        transform_timeout = params.get("transform-timeout", 60.0 * 60.0)
+        start_time = time.time()
+        await es.transform.start_transform(transform_id=transform_id)
+
+        stats_response = await es.transform.get_transform_stats(transform_id=transform_id)
+        state = stats_response['transforms'][0].get("state")
```
ebadyano, line 1670: nit: use double quotes around transforms

esrally/driver/runner.py
ebadyano, line 1687: Could you please use f-strings here?

docs/track.rst
ebadyano: Progress reporting might be a bit of a trade of 
Functions that implement `completed` and `percent_completed` are called until `_completed` isn't set to True. The issue is that `execute-transform` just has a while loop and blocks there until completed is set to True, so `percent_completed` is never updated. For it to work we would need to have `start_transform` outside of the `execute_transform` and `execute_transform` should just check `get_transform_stats` and update `_percent_completed` outside of the loop and after the sleep and and once it's done set `_completed` to True. In this case it probably should be named something like `wait-for-transformation`. And in the track we would need to have both operations `start_transformation` followed by `wait-for-transformation`. We used to have an example of that for snapshot recovery operation, but it got changed to just waiting for the restore to complete in a loop without updating `percent_completion` to get more accurate throughout and how long it took to complete information as otherwise it could be off by the sleep time. Here is the change https://github.com/elastic/rally/commit/9f73d5d8c1f58d2e247634cfbc3eff5ee91a3750 where you can see both how it was before and after. Although i think in your case it should be okay to use percentage completed, since you are relaying on information from get_transform_stats to get the performance..

hendrikmuhs:
@ebadyano and I discussed this PR:

- `execute-transform` should be rewritten to `wait-for-transform` or merged into `stop-transform`, instead of blocking inside, it will use the completion framework and let rally run the outer polling loop (Rally will call `stop/wait-for-transform` until `_completed=true`)
  - in followup PR's the API can be extended to wait until a certain checkpoint has been reached. This will cover the continuous transform usecase (in the continuous case a transform should run several checkpoints and e.g. transform a certain amount of data/time)
- having the above, the telemetry device becomes superfluous: I will remove it from the PR(I will keep the code in case we find a usecase later)

hendrikmuhs:
closing this one in favor of #1022

PR #1014, opened 2020-06-11
Update elasticsearch library to 7.7.0, in order to pick the EQL wrapper

aleksmaus:
Picking up the newer elasticsearch client that implements EQL wrapper:
https://github.com/elastic/elasticsearch-py/blob/7.7.0/elasticsearch/client/eql.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ supported_python_versions = [(3, 8)]
 install_requires = [
     # License: Apache 2.0
     # transitive dependency urllib3: MIT
-    "elasticsearch==7.6.0",
+    "elasticsearch==7.7.0",
     # License: Apache 2.0
     # transitive dependencies:
     #   aiohttp: Apache 2.0


dliappis:
Hey @aleksmaus thanks for the PR!

Switching to a new client version is more complicated that it seems because we also rely heavily on [elasticsearch-async](https://github.com/elastic/elasticsearch-py-async); since we added the dependency, the standard elasticsearch-py library recently received [async support](https://github.com/elastic/elasticsearch-py/pull/1232/files)..

For elasticsearch-async we had to implement several abstractions and workarounds due to missing features or bugs (e.g. [here](https://github.com/elastic/rally/blob/6961f87a20e9927cd0c442c13e99b52a4179ba8b/esrally/client.py#L170-L214) and [here](https://github.com/elastic/rally/blob/6961f87a20e9927cd0c442c13e99b52a4179ba8b/esrally/async_connection.py#L28-L155)).

For these reasons, upgrading the version will take some effort specifically looking at the interaction with async and possibly switching entirely to elasticsearch-py for everything. There's always the possibility of issues and regressions and there's a brittle ecosystem that depends on accuracy (e.g. https://elasticsearch-benchmarks.elastic.co), so I am afraid this upgrade will need to wait a bit.

aleksmaus:
Sounds good. Thanks
Could port EQL to 7.6.1 instead, otherwise will just call `es.transport.perform_request` in my custom runner.

dliappis:
> otherwise will just call `es.transport.perform_request` in my custom runner.

Note that you can also use the [raw-request](https://esrally.readthedocs.io/en/stable/track.html?highlight=raw-request#raw-request) operation for one-off needs.

I'll discuss with @sethmlarson to clarify the scope of async-support in `7.7.0`; I'll open a ticket where we can track the discussion and, if you don't mind, we could close this PR?

aleksmaus:
Closing this PR, since the scope of the changes needed is larger than initially assumed. It will be tracked as a separate issue.

dliappis:
Hey @aleksmaus note that we merged today https://github.com/elastic/rally/pull/1031 and Rally now uses the Elasticsearch Python client v 7.8; you can now use the EQL wrapper.

PR #1016, opened 2020-06-12
Allow to forcefully close SSL connections

danielmitterdorfer:
With this commit we expose a new client option `enable_cleanup_closed`
which allows to forcefully close an SSL connection that might not have
been properly closed on the server-side. This avoids potentially leaking
connections.


jaymode:
Do we know if there is an issue with how Elasticsearch closes SSL connections that needs to be fixed as well?

pquentin:
For the record, there is an issue with Elasticsearch and TLS 1.3: https://github.com/elastic/elasticsearch/issues/76642

PR #1022, opened 2020-06-29
add transform support

hendrikmuhs:
add support for transform, consisting of the following parts:

 - create, start, stop, delete API support
 - basic command-line reporting

follow up of #1011, differences:

 - removed telemetry device
 - stop-transform now reports progress and per default "blocks"(from the track perspective, not within rally) until progress == `1.0` and returns stats

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -1646,29 +1646,6 @@ class WaitForTransform(Runner):
         * ``wait-for-completion``: whether to block until the transform has stopped, default true
         * ``transform-timeout``: overall runtime timeout of the transform in seconds, default 1800 (1h)
         * ``poll-interval``: how often transform stats are polled, used to set progress and check the state, default 0.5.
-
-        Returned meta data
-        `
-        The following meta data are always returned:
-
-        * ``search-time-ms``: The amount of time spent searching, in milliseconds.
-        * ``processing-time-ms``: The amount of time spent processing results, in milliseconds.
-        * ``index_time-ms``: The amount of time spent indexing, in milliseconds.
-        * ``weight``: operation-agnostic representation of the processed documents (used internally by Rally for throughput calculation).
-        * ``unit``: The unit in which to interpret, always "docs".
-        * ``ops``: A dictionary of counters, see below.
-
-        ``ops`` contains the following meta data:
-
-        * ``pages-processed``: The number of search or bulk index operations processed.
-        * ``documents-processed``: The number of documents that have been processed from the source index of the transform.
-        * ``documents-indexed``: The number of documents that have been indexed into the destination index for the transform.
-        * ``index-total``: The number of index operations on the dest index for the transform.
-        * ``index-failures``: The number of indexing failures.
-        * ``search-total``: The number of search operations on the source index for the transform.
-        * ``search-failures``: The number of search failures.
-        * ``processing-total``: The number of processing operations.
-
         """
         import time
 
@@ -1689,10 +1666,6 @@ class WaitForTransform(Runner):
                                               wait_for_checkpoint=wait_for_checkpoint)
 
         while True:
-            # sleep for a while, so stats is not called to often, note this is very basic,
-            # but because we return performance data from `_stats` this is ok
-            # await asyncio.sleep(poll_interval)
-
             stats_response = await es.transform.get_transform_stats(transform_id=transform_id)
             state = stats_response["transforms"][0].get("state")
             transform_stats = stats_response["transforms"][0].get("stats", {})
@@ -1713,17 +1686,6 @@ class WaitForTransform(Runner):
                 self._percent_completed = stats_response["transforms"][0].get("checkpointing", {}).get("next", {}).get(
                     "checkpoint_progress", {}).get("percent_complete", 0.0) / 100.0
 
-            ops = {
-                "pages-processed": transform_stats.get("pages_processed", 0),
-                "documents-processed": transform_stats.get("documents_processed", 0),
-                "documents-indexed": transform_stats.get("documents_indexed", 0),
-                "index-total": transform_stats.get("index_total", 0),
-                "index-failures": transform_stats.get("index_failures", 0),
-                "search-total": transform_stats.get("search_total", 0),
-                "search-failures": transform_stats.get("search_failures", 0),
-                "processing-total": transform_stats.get("processing_total", 0)
-            }
-
             documents_processed = transform_stats.get("documents_processed", 0)
             processing_time = transform_stats.get("search_time_in_ms", 0)
             processing_time += transform_stats.get("processing_time_in_ms", 0)
@@ -1735,12 +1697,8 @@ class WaitForTransform(Runner):
             if self._completed or (documents_processed_delta > 5000 and processing_time_delta > 500):
                 stats = {
                     "transform-id": transform_id,
-                    "search-time-ms": transform_stats.get("search_time_in_ms", 0),
-                    "processing-time-ms": transform_stats.get("processing_time_in_ms", 0),
-                    "index-time-ms": transform_stats.get("index_time_in_ms", 0),
                     "weight": transform_stats.get("documents_processed", 0),
                     "unit": "docs",
-                    "ops": ops
                 }
 
                 throughput = 0
@@ -1757,6 +1715,7 @@ class WaitForTransform(Runner):
                 self._last_processing_time = processing_time
                 return stats
             else:
+                # sleep for a while, so stats is not called to often
                 await asyncio.sleep(poll_interval)
 
     def __repr__(self, *args, **kwargs):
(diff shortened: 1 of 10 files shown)


esrally/driver/runner.py
```diff
@@ -1569,6 +1573,180 @@ def __repr__(self, *args, **kwargs):
...
+
+        if not self._start_time:
+            self._start_time = time.time()
+            await es.transform.stop_transform(transform_id=transform_id,
+                                              force=force,
+                                              timeout=timeout,
+                                              wait_for_completion=False,
+                                              wait_for_checkpoint=wait_for_checkpoint)
+
+        stats_response = await es.transform.get_transform_stats(transform_id=transform_id)
```
ebadyano, line 1689: Does it make sense to add  await asyncio.sleep(wait_period)? with wait_period parameter and some default, otherwise we keep calling get_transform_stats constantly, which could affect the performance? Kind of like you had a `poll_interval` in your previous iteration.

esrally/driver/runner.py
```diff
@@ -1569,6 +1573,180 @@ def __repr__(self, *args, **kwargs):
...
+        transform_id = mandatory(params, "transform-id", self)
+        timeout = params.get("timeout")
+
+        await es.transform.start_transform(transform_id=transform_id, timeout=timeout)
+
+    def __repr__(self, *args, **kwargs):
+        return "start-transform"
+
+
+class StopTransform(Runner):
```
ebadyano, line 1607: Does it make sense to name the operation something like `wait-for-transform-completion`? otherwise in Rally report it looks like `stop-transform` took a while to complete..although it is clear from documentation.
```
|                                                 Min Throughput | stop-transform | 1.86915e+09 | docs/s |
|                                              Median Throughput | stop-transform | 4.36236e+09 | docs/s |
|                                                 Max Throughput | stop-transform | 6.24201e+09 | docs/s |
|                                        50th percentile latency | stop-transform |     1.95756 |     ms |
|                                        90th percentile latency | stop-transform |     3.07196 |     ms |
|                                        99th percentile latency | stop-transform |      5.7514 |     ms |
|                                      99.9th percentile latency | stop-transform |     15.3971 |     ms |
|                                       100th percentile latency | stop-transform |     96.2499 |     ms |
|                                   50th percentile service time | stop-transform |     1.95976 |     ms |
|                                   90th percentile service time | stop-transform |       3.073 |     ms |
|                                   99th percentile service time | stop-transform |      5.7514 |     ms |
|                                 99.9th percentile service time | stop-transform |     15.3971 |     ms |
|                                  100th percentile service time | stop-transform |     96.2499 |     ms |
|                                                     error rate | stop-transform |           0 |      % |
```

ebadyano:
@hendrikmuhs Thank you for the changes, I left some comments above. Also, I tested the change with the track https://github.com/elastic/rally-tracks/pull/128 with metrics getting upload to elasticsearch and I see all the metrics:
```

meta.index-time-ms | 304
-- | --
  | ? meta.ops.documents-indexed | 4000
  | ? meta.ops.documents-processed | 10747184
  | ? meta.ops.index-failures | 0
  | ? meta.ops.index-total | 8
  | ? meta.ops.pages-processed | 8
  | ? meta.ops.processing-total | 8
  | ? meta.ops.search-failures | 0
  | ? meta.ops.search-total | 8
  | ? meta.processing-time-ms | 122
  | ? meta.search-time-ms | 5576


```
However with in-memory store, I don't see any metrics stored.

esrally/driver/runner.py
hendrikmuhs: I was expecting rally is doing some rate limiting. Are you sure, rally isn't?

ebadyano:
Depends on https://github.com/elastic/rally/pull/1024

esrally/driver/runner.py
hendrikmuhs: :+1: I reintroduced `poll_interval`.

hendrikmuhs:
@ebadyano @danielmitterdorfer 

Thanks for all the feedback and help.

I added support for `throughput`, it turned out that this calculation was very noisy in the 1st place reporting spikes. I avoid them now by only reporting stats if at least a certain amount of docs have been processed or certain processing time has been spend. Otherwise the values would be too small to produce reasonable results.

I followed the advise to re-introduce `WaitForTransform`, but kept `StopTransform` as admin task.

As discussed `latency` and `service time` are not making a lot of sense for benchmarking transform. 

Still, I need your help once again: I am missing a benchmark values when running on the commandline. All required metrics are int he store and using the es backend, I can chart them, but this doesn't help for the commandline. I basically need the "last" data point which contains the final throughput and processing times. In `GlobalStatsCalculator` I only figured out to get _all_ docs (and it seems like they might be truncated when using ES as metric backend).  

Alternatively I could add another runner `GetTransformStats` and use the output from that operation for reporting.

esrally/metrics.py
```diff
@@ -1849,6 +1896,26 @@ def op_metrics(op_item, key, single_value=False):
...
                     })
+            elif metric == "transform_processing_times":
+                for item in value:
+                    all_results.append({
+                        "id": item["id"],
+                        "name": "transform_processing_times",
+                        "value": {
+                            "search": item["search"],
+                            "index": item["index"],
+                            "processing": item["processing"]
```
hendrikmuhs, line 1907: this doesn't look right to me, should I instead append 3 entries to `all_results` with:

```
   {
   "id": item["id"],
   "name": "transform_search_time",
   "value": {
     "single": item["search"]
   }
   {
   "id": item["id"],
   "name": "transform_index_time",
   "value": {
     "single": item["index"]
   },
   {
   "id": item["id"],
   "name": "transform_processing_time",
   "value": {
     "single": item["processing"]
   }
```

esrally/metrics.py
```diff
@@ -1849,6 +1896,26 @@ def op_metrics(op_item, key, single_value=False):
...
+                        }
+                    })
+            elif metric == "transform_throughput":
+                for item in value:
+                    all_results.append({
+                        "id": item["id"],
+                        "name": "transform_throughput",
+                        "value": {
+                            "throughput": item["throughput"]
```
hendrikmuhs, line 1916: similar to above, use `single` instead of `throughput`?

esrally/telemetry.py
```diff
@@ -743,6 +743,181 @@ def sample(self):
...
+        """
+
+        self._record(metrics.SampleType.Normal)
+
+    def record_final(self):
+        """
+        Collect final Transform stats for transforms (optionally) specified in telemetry parameters and push to metrics store.
+        """
+
+        self._record(metrics.SampleType.Final)
```
hendrikmuhs, line 844: ^ I added `SampleType.Final` in order to query only for the last, final record in the end result.

Intermediate results are indexed with `SampleType.Normal`

I am not sure if that's an abusive usage of `SampleType`

hendrikmuhs:
Example output with ` --telemetry=transform-stats`:

```
|                               Transform search processing time | transform-browser-and-os |      13261 |     ms |
|                             Transform indexing processing time | transform-browser-and-os |        992 |     ms |
|                                 Transform task processing time | transform-browser-and-os |        265 |     ms |
|                                           Transform throughput | transform-browser-and-os | 1.3776e+06 | docs/s |
```

Example output in comparison mode:

```
|                              Transform search processing time | transform-browser-and-os |       12679 |       12260 |     -419 |     ms |
|                            Transform indexing processing time | transform-browser-and-os |         850 |         889 |       39 |     ms |
|                                Transform task processing time | transform-browser-and-os |         274 |         248 |      -26 |     ms |
|                                          Transform throughput | transform-browser-and-os | 1.44896e+06 | 1.49287e+06 |  43911.2 | docs/s |
```

esrally/telemetry.py
danielmitterdorfer: I think we can avoid introducing a new sample type for this. Given that the final sample is a total summary at the end we could prefix each metric name with `total_` in the telemetry device and then only query for that in the stats calculator?

esrally/driver/runner.py
```diff
@@ -1568,6 +1572,172 @@ def __repr__(self, *args, **kwargs):
...
+                    "unit": "docs",
+                }
+
+                throughput = 0
+                if self._completed:
+                    # take the overall throughput
+                    if processing_time > 0:
+                        throughput = documents_processed / processing_time * 1000
+                else:
+                    throughput = documents_processed_delta / processing_time_delta * 1000
```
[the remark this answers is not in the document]
hendrikmuhs: this is checked in the outer `if`, but I think its cleaner if I check it here again

esrally/metrics.py
```diff
@@ -1741,6 +1748,20 @@ def ml_processing_time_stats(self):
...
+    def total_transform_metric(self, metric_name):
+        values = self.store.get_raw(metric_name)
+        result = []
+        if values:
+            for v in values:
+                transform_id = v.get("meta", {}).get("transform_id")
+                if transform_id is not None:
+                    result.append({
+                        "id": transform_id,
+                        "value": v["value"],
```
hendrikmuhs, line 1760: should this be `"mean": v["value"]` instead, to conform with other parts?

danielmitterdorfer: I don't have a strong preference but I agree that `mean` conveys clearly how the summary metric has been calculated.

PR #1024, opened 2020-07-08
Allow runners to determine throughput

danielmitterdorfer:
With this commit Rally recognizes a new response metadata property
`throughput` from runners. When it is present, it will be used as is for
creating throughput samples. The following restrictions apply:

1. The runner must only be executed by one client. This is needed
because throughput is calculated across all clients but in this case
throughput samples are created without any further aggregation.
2. The runner should be rate-limited to avoid that too many throughput
samples are created. When throughput is calculated, all responses are
bucketed into one second buckets but when the calculation is overridden,
each sample also results in one throughput sample.

We also change the existing `IndicesRecovery` runner to use the new
property in order to calculate throughput directly.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -971,12 +971,12 @@ class Sampler:
         self.q = queue.Queue(maxsize=buffer_size)
         self.logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


ebadyano:
Thank you for the change!

PR #1026, opened 2020-07-15
[WIP] Migrate first test_docker_dev_image 

bartier:
This PR attempts to start test_docker_dev_image migration. 

At the moment, this only contains information about the first test:
https://github.com/elastic/rally/blob/b59bc099465b86b4d643bc999b85e982fae14d12/integration-test.sh#L334-L337

I was not able to run the integration tests with all them passing, I guess this may be related with my setup and some tests I guess needs a external ES to send metrics (I'm not sure about that). 

Anyway, if the project maintainers like it (and the CI of this PR occurs successfully or it is possible to obtain the root cause) I can expand the migration to the other tests.

### DIFF
--- a/esrally/utils/process.py
+++ b/esrally/utils/process.py
@@ -28,15 +28,11 @@ def run_subprocess(command_line):
     return os.system(command_line)
 
 
-def run_subprocess_in_path(path, command_line):
-    return run_subprocess(f"cd {path} && {command_line}")
-
-
-def run_subprocess_with_output(command_line, path=None):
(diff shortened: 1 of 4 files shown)


bartier:
@danielmitterdorfer Thanks! I made some changes to the code. Could you please check again?

bartier:
@danielmitterdorfer I made some changes, could you please check it again? Indeed the code looks much better now. Thanks for your feedback.

danielmitterdorfer:

PR #1027, opened 2020-07-16
Resolve JAVA_HOME for source builds lazily

danielmitterdorfer:
With this commit we resolve JAVA_HOME lazily when attempting to build
Elasticsearch. When a source build is externally provided to Rally's
source artifact cache, there is no need to actually build it. However,
if the target system does not have any JDK installed and we attempt to
install Elasticsearch, the build subsystem is eagerly initialized
(despite the artifact being already cached) and is complaining that it
cannot find a JDK to build the artifact. By lazily resolving JAVA_HOME
we avoid that error.

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -49,9 +49,14 @@ def create(cfg, sources, distribution, build, car, plugins=None):
     template_renderer = TemplateRenderer(version=es_version, os_name=target_os, arch=target_arch)
 
     if build_needed:
-        java_home = _java_home(car)
+        raw_build_jdk = car.mandatory_var("build.jdk")
+        try:
+            build_jdk = int(raw_build_jdk)
+        except ValueError:
+            raise exceptions.SystemSetupError(f"Car config key \"build.jdk\" is invalid: \"{raw_build_jdk}\" (must be int)")
+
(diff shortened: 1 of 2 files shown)


esrally/mechanic/supplier.py
```diff
@@ -49,9 +49,14 @@ def create(cfg, sources, distribution, build, car, plugins=None):
     template_renderer = TemplateRenderer(version=es_version, os_name=target_os, arch=target_arch)
 
     if build_needed:
-        java_home = _java_home(car)
+        raw_build_jdk = car.mandatory_var("build.jdk")
+        try:
+            build_jdk = int(raw_build_jdk)
+        except ValueError:
+            raise exceptions.SystemSetupError(f"Car config key \"build.jdk\" is invalid: \"{raw_build_jdk}\" (must be int)")
```
hub-cap, line 56: FWIW im perfectly fine if we use single quotes inside f strings for things like this... `f"blah 'this thing' blah"`, I hate seeing `\` when we can get around it :)

If you dont agree, thats fine by me!

danielmitterdorfer: Good point. Usually we use square brackets anyway instead of double quotes. I've pushed a change in https://github.com/elastic/rally/pull/1027/commits/690d568ccd4f148d3789712d143b9276bf433053.

PR #1028, opened 2020-07-20
Fetch artifact revision lazily on build

danielmitterdorfer:
With this commit we avoid touching the source repository when a specific
git revision is provided for an artifact (with `--revision`) and the
artifact is already present in the source artifact cache. This helps in
situations where an external process has put artifacts into Rally's
source artifact cache from a fork (in which case the artifact revision
will not be present in the main repo and the resolution would fail).

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -123,7 +123,7 @@ def create(cfg, sources, distribution, build, car, plugins=None):
                                             (plugin.name, supply_requirements[plugin.name]))
 
             if caching_enabled:
-                plugin_file_resolver = PluginFileNameResolver(plugin.name)
+                plugin_file_resolver = PluginFileNameResolver(plugin.name, plugin_version)
                 plugin_supplier = CachedSourceSupplier(source_distributions_root,
                                                        plugin_supplier,
                                                        plugin_file_resolver)
@@ -258,7 +258,7 @@ class TemplateRenderer:
         }
         r = template
         for key, replacement in substitutions.items():
-            r = r.replace(key, replacement)
+            r = r.replace(key, str(replacement))
         return r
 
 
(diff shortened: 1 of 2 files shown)


tests/mechanic/supplier_test.py
```diff
@@ -215,6 +221,39 @@ def add_es_artifact(binaries):
...
+        binaries = {}
+
+        cached_supplier.add(binaries)
+
+        self.assertEqual(0, es.fetch.call_count)
+        self.assertEqual(0, es.prepare.call_count)
+        self.assertEqual(0, es.add.call_count)
+        self.assertTrue(cached_supplier.cached)
+        self.assertIn("elasticsearch", binaries)
+        self.assertEqual("/tmp/elasticsearch-abc123-linux-x86_64.tar.gz", binaries["elasticsearch"])
```
hub-cap, line 255: Do you think its worth testing this assert in the `test_does_not_cache_when_no_revision` test above? 

```
        self.assertEqual("/path/to/artifact.tar.gz", binaries["elasticsearch"])
```

esrally/mechanic/supplier.py
```diff
@@ -328,11 +328,19 @@ def cached(self):
...
+        # fork. In that case the provided commit hash would not be present in any case in the main ES repo.
+        maybe_an_artifact = os.path.join(self.distributions_root, self.file_name)
+        if os.path.exists(maybe_an_artifact):
+            self.cached_path = maybe_an_artifact
+        else:
+            resolved_revision = self.source_supplier.fetch()
+            if resolved_revision:
+                # ensure we use the resolved revision for rendering the artifact
+                self.file_resolver.revision = resolved_revision
+                self.cached_path = os.path.join(self.distributions_root, self.file_name)
```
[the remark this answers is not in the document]
danielmitterdorfer: Ha, good catch but that's tricky. The point is that `self.file_name` is not a constant value but a (computed) property that relies on the `file_resolver`. In the first call (where we assign it to `maybe_an_artifact`) we just use the user-provided revision (if any) that is provided with `--revision`. This could be a valid git revision hash (like `d31808de80f`) but it could also be a meta-revision (like `current`). So in the first case it might resolve the artifact file name to `elasticsearch-d31808de80f-linux-x86_64.tar.gz`) and then we could indeed assign `maybe_an_artifact` here. But in the second case it would resolve it to `elasticsearch-current-linux-x86_64.tar.gz`) and that's not what we want. Instead, we need to find out what `current` means and resolve it to a real git revision hash that we can use for the file name of the cached artifact. We do this here:

```python
resolved_revision = self.source_supplier.fetch()
if resolved_revision:
    # ensure we use the resolved revision for rendering the artifact
    self.file_resolver.revision = resolved_revision
```

The `source_supplier` fetches the source tree (if needed) based on the provided revision (may also be `current`, `latest` or a timestamp) and returns the *real*, resolved git revision. We then provide that git revision to our file resolver which in turn will use it to return the corresponding file name.

tests/mechanic/supplier_test.py
danielmitterdorfer: Thanks for that; makes sense to me. I've pushed the additional assert in https://github.com/elastic/rally/pull/1028/commits/504d5c30256c3d1f8a247cf1abc147daeb81ffb0.

PR #1029, opened 2020-07-20
Add integration test for proxy

danielmitterdorfer:
With this commit we migrate the integration test for HTTP proxy connections to
the new integration test infrastructure and cleanup any remaining leftovers in
the old script (which is still used for Docker release image tests).

Relates #975


danielmitterdorfer:

danielmitterdorfer:

it/proxy_test.py
```diff
@@ -0,0 +1,95 @@
...
+    lines = process.run_subprocess_with_output(f"docker run --rm --name squid -d "
+                                               f"-v {config_dir}/squidpasswords:/etc/squid/squidpasswords "
+                                               f"-v {config_dir}/squid.conf:/etc/squid/squid.conf "
+                                               f"-p 3128:3128 datadog/squid")
+    proxy_container_id = lines[0].strip()
+    proxy = HttpProxy(authenticated_url=f"http://testuser:testuser@127.0.0.1:3128",
+                      anonymous_url=f"http://127.0.0.1:3128")
+    yield proxy
+    process.run_subprocess(f"docker stop {proxy_container_id}")
+    process.run_subprocess(f"docker rm {proxy_container_id}")
```
hub-cap, line 44: the `--rm` flag supplied to the `docker run` will take care of the docker remove when you stop it, and this ends up printing `"Error: No Such container:...."`

it/proxy_test.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+# ensures that a fresh log file is available
+@pytest.fixture(scope="function")
+def fresh_log_file():
+    cfg = it.ConfigFile(config_name=None)
+    log_file = os.path.join(cfg.rally_home, "logs", "rally.log")
+
+    if os.path.exists(log_file):
```
hub-cap, line 53: just curious, why would we not just zero out the log contents if we want a new logfile every time, instead of moving its contents. I assume the contents of the "last run" are useless if its a passing test by the time you get to the next test.

release-docker-test.sh
```diff
@@ -0,0 +1,163 @@
...
+
+function stop_and_clean_docker_container {
+    docker stop ${1} > /dev/null || true
+    docker rm ${1} > /dev/null || true
+}
+
+function kill_related_es_processes {
+    # kill all lingering Rally instances that might still be hanging
+    set +e
+    # kill all lingering Elasticsearch Docker containers launched by Rally
```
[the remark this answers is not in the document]
danielmitterdorfer: Correct. There are only Docker processes so no need to kill anything else.

it/proxy_test.py
danielmitterdorfer: I see. I think I just did a direct port of what we had in the integration test script and that's probably why I did not notice. I'll remove the `docker rm` invocation then.

it/proxy_test.py
danielmitterdorfer: The idea behind this is that we should keep the full history of the log file. I am not so concerned with CI but rather when you run the tests locally. It might be puzzling (and maybe even a problem) if we start truncating the log file.

release-docker-test.sh
```diff
@@ -0,0 +1,163 @@
...
+}
+
+function exit_if_docker_not_running {
+    if ! docker ps >/dev/null 2>&1; then
+        error "Docker is required to run integration tests. Install and run Docker and try again."
+        exit 1
+    fi
+}
+
+
```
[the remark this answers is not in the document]
danielmitterdorfer: addressed in https://github.com/elastic/rally/pull/1029/commits/daade765f8b865d58d80ec51b79c9c5633329887

it/proxy_test.py
danielmitterdorfer: Addressed in https://github.com/elastic/rally/pull/1029/commits/daade765f8b865d58d80ec51b79c9c5633329887

danielmitterdorfer:
@hub-cap can you please have another look at this?

it/proxy_test.py
hub-cap: oh ok this makes sense

danielmitterdorfer:
Thanks for the review!

PR #1031, opened 2020-07-23
Upgrade Elasticsearch client library to 7.8

### LINKED ISSUE #1017: Evaluate effort to upgrade to elasticsearch-py 7.7.0
elasticsearch-py 7.7.0 brings EQL support and https://github.com/elastic/rally/pull/1014 attempted to bump the version of the library used by Rally.

However, as mentioned in https://github.com/elastic/rally/pull/1014#issuecomment-643245590, upgrading might not be straightforward due a number of reasons:

1. Rally relies heavily on [elasticsearch-async](https://github.com/elastic/elasticsearch-py-async); since this dependency got introduced, the standard elasticsearch-py library received [async support](https://github.com/elastic/elasticsearch-py/pull/1232/files) and async development/maintenance mored there.

For elasticsearch-async we had to implement several abstractions and workarounds due to missing features or bugs (e.g. [here](https://github.com/elastic/rally/blob/6961f87a20e9927cd0c442c13e99b52a4179ba8b/esrally/client.py#L170-L214) and [here](https://github.com/elastic/rally/blob/6961f87a20e9927cd0c442c13e99b52a4179ba8b/esrally/async_connection.py#L28-L155)).

This ticket will evaluate the safety of upgrading the elasticsearch-py dependency from: https://github.com/elastic/rally/blob/6961f87a20e9927cd0c442c13e99b52a4179ba8b/setup.py#L52 to `7.7.0`.

danielmitterdorfer:
With this commit we upgrade the Elasticsearch client library to version
7.8. This version also includes a completely rewritten implementation of
the async client that is now bundled with the main client (instead of a
separate library). This allowed us to remove several workaround that we
had to implement before.

Closes #1017


danielmitterdorfer:
I did a stress test with 50.000 clients pushing data to Elasticsearch with the new implementation which worked flawlessly. I also benchmarked PMC with `master` against this implementation and did not spot any differences that cannot be attributed to run-to-run variation. Hence I'll mark it as ready for review.

sethmlarson:
:tada:

Issue #734, opened 2019-07-18
Move unit-test JSON text into resources


drawlerr:
Currently, JSON responses expected by unit tests are inline literals, which is making our python tests rather large.  We could reduce some unnecessary bulk by moving these responses into separate files.

bartier:
This looks like a old issue, is this open to help? If yes, would it be possible to have an migration example? I could replicate to the others test cases it applies to.

danielmitterdorfer:
> If yes, would it be possible to have an migration example?

I've created a draft PR in https://github.com/elastic/rally/pull/1033 with a possible approach. Once that discussion is finished and the PR is merged I think we can use that example as the basis for the rest of the tests.

danielmitterdorfer:
We've discussed this in our sync and we decided to keep the inline literal dicts so code and test data are closely linked. Hence I'm closing this issue.

PR #1033, opened 2020-07-27
Allow to load test resources from files

danielmitterdorfer:
With this commit we add a helper function `load_test_resource` that
evaluates caller context so we can automagically load JSON files with
data relevant for the current test.

Relates #734

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -16,6 +16,12 @@
 # under the License.
 
 import asyncio
+import json
+import os
+import re
+import sys
+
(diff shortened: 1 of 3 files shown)


danielmitterdorfer:
@dliappis, @ebadyano, @hub-cap I've implemented a prototype that allows us to move dict-like test data (i.e. JSON) to dedicated files instead of defining them in the code. I am putting this here as a draft PR for early discussion about the approach and I'd like to hear your thoughts about it.

danielmitterdorfer:
We've discussed this in our sync and we decided to keep the inline literal dicts so code and test data are closely linked. Hence I'm closing this PR.

PR #1039, opened 2020-07-28
Check for pointless statements

danielmitterdorfer:
With this commit we enable the linter error for pointless statements and
fix all existing issues.

Relates #838


esrally/utils/opts.py
```diff
@@ -146,13 +146,13 @@ def __init__(self, argvalue):
...
     def parse_options(self):
         def normalize_to_dict(arg):
-            from elasticsearch.client import _normalize_hosts
             """
             Return parsed comma separated host string as dict with "default" key.
             This is needed to support backwards compatible --target-hosts for single clusters that are not
             defined as a json string or file.
             """
 
+            from elasticsearch.client import _normalize_hosts
```
hub-cap, line 155: now im curious, why is this here and not at the top?

danielmitterdorfer: It is related to initialization of loggers. We noticed that especially the `elasticsearch` dependency is problematic in that respect. A long time ago we had all imports at the top but the elasticsearch loggers were initialized when a module was loaded that imported `elasticsearch` which happened before Rally had a chance to configure logging properly. Therefore we never import any elasticsearch module at the top in Rally.

Issue #1032, opened 2020-07-24
offline installation fails on linux


immon:
**Rally version** (get with `esrally --version`): 2.0.0

**Invoked command**: ./install.sh

**JVM version**: 1.8.0_171

**OS version**: Linux Ubuntu 18.04

**Description of the problem including expected versus actual behavior**:

Offline installation does not work on linux. Following steps [from the docs](https://esrally.readthedocs.io/en/stable/install.html#offline-install) gives the following error.

```
(venv) imo@lake:~/tmp/rally/esrally-dist-2.0.0$ ./install.sh 
Installing Rally 2.0.0...
Looking in links: file:///home/imo/tmp/rally/esrally-dist-2.0.0/bin
Collecting esrally==2.0.0
Collecting certifi (from esrally==2.0.0)
Collecting thespian==3.10.0 (from esrally==2.0.0)
Collecting elasticsearch==7.6.0 (from esrally==2.0.0)
Collecting elasticsearch-async==6.2.0 (from esrally==2.0.0)
Collecting Jinja2==2.10.3 (from esrally==2.0.0)
Collecting py-cpuinfo==3.2.0 (from esrally==2.0.0)
Collecting boto3==1.10.32 (from esrally==2.0.0)
Collecting psutil==5.7.0 (from esrally==2.0.0)
Collecting ijson==2.6.1 (from esrally==2.0.0)
Collecting jsonschema==3.1.1 (from esrally==2.0.0)
Collecting tabulate==0.8.5 (from esrally==2.0.0)
Collecting yappi==1.2.3 (from esrally==2.0.0)
Collecting urllib3>=1.21.1 (from elasticsearch==7.6.0->esrally==2.0.0)
Collecting aiohttp (from elasticsearch-async==6.2.0->esrally==2.0.0)
Collecting async_timeout (from elasticsearch-async==6.2.0->esrally==2.0.0)
Collecting MarkupSafe>=0.23 (from Jinja2==2.10.3->esrally==2.0.0)
Collecting s3transfer<0.3.0,>=0.2.0 (from boto3==1.10.32->esrally==2.0.0)
Collecting jmespath<1.0.0,>=0.7.1 (from boto3==1.10.32->esrally==2.0.0)
Collecting botocore<1.14.0,>=1.13.32 (from boto3==1.10.32->esrally==2.0.0)
[...]

PR #1043, opened 2020-07-30
Installation and offline dist improvements

### LINKED ISSUE #1034: [Offline Install] If Python not installed on host, install script not clear if the install aborted
**Rally version** (get with `esrally --version`): `2.0.0`

**Invoked command**: `sudo ./esrally-dist-2.0.0/install.sh`

**Configuration file (located in `~/.rally/rally.ini`))**:

**JVM version**: not installed 

**OS version**: CentOS 7

**Steps to reproduce**:
 1.Follow steps 1 - 4 in the Offline Install [guide](https://esrally.readthedocs.io/en/stable/install.html#offline-install)
 2. Step 4 results in the following events
 
```
$ sudo ./esrally-dist-2.0.0/install.sh
[sudo] password for eddie:
Installing Rally 2.0.0...
Python3 is required but not installed.
```

I suspect the install aborted because Python3 is not installed but it might make sense to be more explicit in what the outcome was. If the install did abort, maybe say "Install aborted due to Python3 being required but no installation located".

dliappis:
Improve error message during offline installation, if Python3 or pip3 
are missing.

Also due to native dependencies, make the offline package Linux specific
 adding the necessary checks.

Finally, update the installation instructions to use Python3 best 
practices.

Relates #1032
Closes #1034

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -33,9 +33,8 @@ Once ``pyenv`` is installed, install a compatible Python version::
     # see https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-global for details
     pyenv global 3.8.0
 
-    # Install pip3
-    curl -s https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-    python3 get-pip.py --user
+    # Upgrade pip
+    python3 -m pip install --user --upgrade pip
 
 git
 ~~~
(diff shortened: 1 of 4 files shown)


docs/install.rst
```diff
@@ -113,32 +112,19 @@ To find the JDK, Rally expects the environment variable ``JAVA_HOME`` to be set
 Installing Rally
 ----------------
 
-Simply install Rally with pip: ``pip3 install esrally``
-
-.. note::
-
-   Depending on your system setup you may need to prepend this command with ``sudo``.
+1. Ensure ``~/.local/bin`` is in your ``$PATH``.
+2. Install Rally: ``python3 -m pip install --user esrally``.
```
dliappis, line 116: Alternatively `--user` can be skipped and then `esrally` lives (in the form of a pyenv shim) in `$ which esrally
~/.pyenv/shims/esrally`. In this case the user would still need to either logout/login or `exec $SHELL` to make the shim active.

Since we are anyway upgrading pip (earlier) in a `--user` context, I think we should stick to that convention, which IMHO is a better practice, cleaner approach and ultimately a more discoverable location (`~/.local/bin`) for binaries to be stored. The requirement to have `~/.local/bin` is also IMHO a best practice that users should follow.

All these instructions have been tested -- at least -- in CentOS 8 and Ubuntu 18.04. More testing very welcome.

hub-cap:
Im testing this now, but ive noticed a bit about needing a `JVM` in the readme for running ES. Is this still true? Im trying it w/o one.

docs/install.rst
```diff
@@ -161,12 +147,12 @@ Offline Install
...
 
-If you are in a corporate environment where your servers do not have any access to the Internet, you can use Rally's offline installation package. Follow these steps to install Rally:
+If you are in a corporate environment using Linux servers that do not have any access to the Internet, you can use Rally's offline installation package. Follow these steps to install Rally:
 
 1. Install all prerequisites as documented above.
 2. Download the offline installation package for the `latest release <https://github.com/elastic/rally/releases/latest>`_ and copy it to the target machine(s).
-3. Decompress the installation package with ``tar -xzf esrally-dist-*.tar.gz``.
-4. Run the install script with ``sudo ./esrally-dist-*/install.sh``.
+3. Decompress the installation package with ``tar -xzf esrally-dist-linux-*.tar.gz``.
+4. Run the install script with ``sudo ./esrally-dist-linux-*/install.sh``.
```
hub-cap, line 155: The `linux-` portion of the file is not present on the release link above.

hub-cap: Also, should this be `offline-install.sh` ? Ahh it looks like we sed in teh values in the offline-install for the release version, and then its just `install.sh`.

hub-cap:
I also confirmed that the `java` stuff is still required :) Rally did not like me installing 7.8.1 without a `JAVA_HOME` !!

docs/install.rst
dliappis: > The `linux-` portion of the file is not present on the release link above.

(If I understand the question correctly): Yeah it's not present now, but it will be after we release in the future due the changes here

docs/install.rst
ebadyano: for some reason `python3 -m pip install --user esrally` keeps installing 1.4.1 rally for me

ebadyano: oh i think that's because python version is 3.6 not 3.8

dliappis:
Thanks everyone for the review.

PR #1045, opened 2020-08-05
Show processing time also in comparison reports

danielmitterdorfer:
In #935 we added a new metric `processing time` to the command line
report that can be used to determine Rally's internal overhead. While
the regular command line report shows this metric if configured in
`rally.ini`, the comparison report did not. With this commit we also
show `processing time` in the comparison report as well.

Relates #935


danielmitterdorfer:
Thanks for the review!

PR #1047, opened 2020-08-13
Add runner to wait until snapshot has been created

dliappis:
For snapshot creation (and metrics) so far we've relied only the
`create-snapshot` runner with a blocking call.

In this commit we are introducing a new runner
`wait-for-snapshot-create`to complement `create-snapshot`. This is
similar to `restore-snapshot` and `wait-for-recovery` and helps in cases
 where network connections maybe terminated making a blocking call
 unsuitable.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1119,13 +1119,13 @@ With the operation ``create-snapshot`` you can `create a snapshot <https://www.e
 * ``request-params`` (optional): A structure containing HTTP request parameters.
 
 .. note::
-    It's not recommend to rely on ``wait-for-completion=true``. Instead you should keep the default value (``False``) and use an additional ``wait-for-snapshot-create`` operation in the next step.
-    This is mandatory on the `Elastic Cloud <https://www.elastic.co/cloud>`_ or environments where Elasticsearch is sitting behind a network element that may terminate the blocking connection after a timeout.
+    It's not recommended to rely on ``wait-for-completion=true``. Instead you should keep the default value (``False``) and use an additional ``wait-for-snapshot-create`` operation in the next step.
+    This is mandatory on `Elastic Cloud <https://www.elastic.co/cloud>`_ or environments where Elasticsearch is connected via intermediate network components, such as proxies, that may terminate the blocking connection after a timeout.
 
 wait-for-snapshot-create
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-With the operation ``wait-for-snapshot-create`` you can wait until a snapshot has finished successfully.
+With the operation ``wait-for-snapshot-create`` you can wait until a `snapshot has finished successfully <https://www.elastic.co/guide/en/elasticsearch/reference/current/get-snapshot-status-api.html>`_.
 Typically you'll use this operation directly after a ``create-snapshot`` operation.
 
 It supports the following parameters:
@@ -1134,7 +1134,7 @@ It supports the following parameters:
 * ``snapshot`` (mandatory): The name of the snapshot that this operation will wait until it succeeds.
 * ``completion-recheck-wait-period`` (optional, defaults to 1 second): Time in seconds to wait in between consecutive attempts.
 
-Rally will report the achieved throughput in byte/s, the duration in seconds, the start and stop time in milliseconds and the total amount of files snapshotted as returned by the the `Elasticsearch snapshot status API call <https://www.elastic.co/guide/en/elasticsearch/reference/current/get-snapshot-status-api.html>`_.
+Rally will report the achieved throughput in byte/s.
 
 This operation is :ref:`retryable <track_operations>`.
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1119,7 +1119,24 @@ With the operation ``create-snapshot`` you can `create a snapshot <https://www.e
 * ``request-params`` (optional): A structure containing HTTP request parameters.
 
 .. note::
-    When ``wait-for-completion`` is set to ``true`` Rally will report the achieved throughput in byte/s.
+    It's not recommend to rely on ``wait-for-completion=true``. Instead you should keep the default value (``False``) and use an additional ``wait-for-snapshot-create`` operation in the next step.
+    This is mandatory on the `Elastic Cloud <https://www.elastic.co/cloud>`_ or environments where Elasticsearch is sitting behind a network element that may terminate the blocking connection after a timeout.
+
+wait-for-snapshot-create
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+With the operation ``wait-for-snapshot-create`` you can wait until a snapshot has finished successfully.
+Typically you'll use this operation directly after a ``create-snapshot`` operation.
+
+It supports the following parameters:
+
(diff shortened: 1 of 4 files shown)


docs/track.rst
```diff
@@ -1119,7 +1119,24 @@ With the operation ``create-snapshot`` you can `create a snapshot <https://www.e
 * ``request-params`` (optional): A structure containing HTTP request parameters.
 
 .. note::
-    When ``wait-for-completion`` is set to ``true`` Rally will report the achieved throughput in byte/s.
+    It's not recommend to rely on ``wait-for-completion=true``. Instead you should keep the default value (``False``) and use an additional ``wait-for-snapshot-create`` operation in the next step.
```
danielmitterdorfer, line 1122: nit: "not recommend" -> "not recommended"

docs/track.rst
```diff
@@ -1119,7 +1119,24 @@ With the operation ``create-snapshot`` you can `create a snapshot <https://www.e
...
+    This is mandatory on the `Elastic Cloud <https://www.elastic.co/cloud>`_ or environments where Elasticsearch is sitting behind a network element that may terminate the blocking connection after a timeout.
```
danielmitterdorfer, line 1123: * on the Elastic Cloud -> on Elastic Cloud?
* "sitting behind a network element" -> "connected via intermediate network components, such as proxies, that may terminate ..."?

esrally/driver/runner.py
```diff
@@ -1464,33 +1465,66 @@ async def __call__(self, es, params):
...
+            response = await es.snapshot.status(repository=repository,
+                                                snapshot=snapshot,
+                                                ignore_unavailable=True)
+
+            if "snapshots" in response:
+                response_state = response["snapshots"][0]["state"]
+                # Possible states:
+                # https://www.elastic.co/guide/en/elasticsearch/reference/current/get-snapshot-status-api.html#get-snapshot-status-api-response-body
+                if response_state == "FAILED":
+                    self.logger.error("Snapshot [%s] failed. Response status:\n%s", snapshot, json.dumps(response))
```
danielmitterdorfer, line 1497: * "Response status" -> "Response"? (as it is the actual full, response). I also think we should pretty-print it by specifying e.g. `indent=2` in `json.dumps`.

esrally/driver/runner.py
```diff
@@ -1464,33 +1465,66 @@ async def __call__(self, es, params):
...
+                    raise exceptions.RallyAssertionError(
```
danielmitterdorfer, line 1498: Nit: Any reason why it is formatted this way? I tried putting everything on the same line and ended up with a line width of 110 which should still be fine?

esrally/track/track.py
```diff
@@ -420,7 +420,7 @@ class OperationType(Enum):
     RawRequest = 5
     WaitForRecovery = 6
     CreateSnapshot = 7
```
danielmitterdorfer, line 422: I think we could move `CreateSnapshot` again to administrative actions (i.e. anything > 1000) because Rally would not report request metrics by default for administrative operations (it's usually not interesting to know).

tests/driver/runner_test.py
```diff
@@ -2756,6 +2757,140 @@ async def test_create_snapshot_wait_for_completion(self, es):
             }
         })
 
+        params = {
```
danielmitterdorfer, line 2760: Unfortunately I cannot comment at the specific line but in `test_create_snapshot_no_wait` we still mock `es.snapshot.status` and assert later on that it is not called. IMHO we should remove this now because there is no chance we'd ever call it anymore in the new runner implementation.

tests/driver/runner_test.py
```diff
@@ -2779,36 +2914,56 @@ async def test_create_snapshot_wait_for_completion(self, es):
...
+        es.snapshot.status.return_value = as_future(snapshot_status)
+
+        params = {
+            "repository": "backups",
+            "snapshot": "snapshot-001",
+        }
+
+        r = runner.WaitForSnapshotCreate()
+
+        logger = logging.getLogger("esrally.driver.runner")
```
danielmitterdorfer, line 2959: how about we just use the runner's logger with `logger = r.logger` (you can then probably just inline it in the `with` statement below? This would also make it more refactoring safe.

docs/track.rst
```diff
@@ -1119,7 +1119,24 @@ With the operation ``create-snapshot`` you can `create a snapshot <https://www.e
...
+With the operation ``wait-for-snapshot-create`` you can wait until a snapshot has finished successfully.
+Typically you'll use this operation directly after a ``create-snapshot`` operation.
+
+It supports the following parameters:
+
+* ``repository`` (mandatory): The name of the snapshot repository to use.
+* ``snapshot`` (mandatory): The name of the snapshot that this operation will wait until it succeeds.
+* ``completion-recheck-wait-period`` (optional, defaults to 1 second): Time in seconds to wait in between consecutive attempts.
+
+Rally will report the achieved throughput in byte/s, the duration in seconds, the start and stop time in milliseconds and the total amount of files snapshotted as returned by the the `Elasticsearch snapshot status API call <https://www.elastic.co/guide/en/elasticsearch/reference/current/get-snapshot-status-api.html>`_.
```
danielmitterdorfer, line 1137: I am not sure I'd document all the metadata that we return (except for the throughput)? Otherwise we should probably also document the respective metric keys and make clear that these metadata are only available with an Elasticsearch metrics store (contrary to the throughput).

docs/track.rst
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

docs/track.rst
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

esrally/driver/runner.py
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

esrally/driver/runner.py
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

esrally/track/track.py
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

tests/driver/runner_test.py
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5 + 2b242b2f6596c784674b908853e8ffbbdc83240e + 6523ec90ae29a2a70272865fdaa13d19c66c0bff

tests/driver/runner_test.py
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

docs/track.rst
dliappis: Addressed in 32ee4abf6a15d4158992c9843958b51e9a766cb5

dliappis:
Thanks for your comments! Could you PTAL?

Issue #1050, opened 2020-08-20
Add polling mode to force-merge runner


dliappis:
## Background

Currently the [force-merge](https://esrally.readthedocs.io/en/stable/track.html#force-merge) runner relies on the blocking nature of the corresponding [ES API Call](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html#forcemerge-blocks).

When the connection times out, e.g. due to a proxy, we document the use of a follow-up [index-stats](https://esrally.readthedocs.io/en/stable/track.html#index-stats) operation:

```
    "operation-type": "index-stats",
    "index": "_all",
    "condition": {
        "path": "_all.total.merges.current",
        "expected-value": 0
    },
    "retry-until-success": true
```

However, this is not bulletproof. Merges might momentarily show `0` and then continue again.

## Improving the force-merge runner

We propose adding a `mode` property to the existing force-merge runner, with two possible values:

- `blocking` (default): This will keep the operation as it is now and rely on the APIs blocking nature. **NOTE**: I am think of `default` due to backwards compatibility, but we could consider not having it the default.
- `polling`: In this mode, the force-merge API call should be executed with a deliberately small [request-timeout](https://github.com/elastic/rally/blob/be8622aa3fd45c5c186bec82fe86bbefe6060be4/esrally/driver/runner.py#L617-L620). Then we should wait until no `forcemerge` tasks are present in the `tasks` API.

## Implementation details:

This is an example output from the tasks API while a forcemerge is running:

<details><summary>forcemerge in tasks</summary>

```
{
  "nodes" : {
    "ajOeYAKfR0uuloeQnT9rsQ" : {
      "name" : "instance-XXXX",
      "transport_address" : "...",
      "host" : "...",
      "ip" : "...",
      "roles" : [
        "data",
        "ingest",
        "master",
        "remote_cluster_client",
        "transform"
      ],
      "attributes" : {
        ...
      },
      "tasks" : {
        "ajOeYAKfR0uuloeQnT9rsQ:110681" : {
          "node" : "ajOeYAKfR0uuloeQnT9rsQ",
          "id" : 110681,
          "type" : "direct",
          "action" : "cluster:monitor/tasks/lists[n]",
          "description" : "",
          "start_time_in_millis" : 1597338489380,
          "running_time_in_nanos" : 98452,
          "cancellable" : false,
          "parent_task_id" : "ajOeYAKfR0uuloeQnT9rsQ:110680",
          "headers" : { }
[...]

PR #1051, opened 2020-08-21
Force Merge Runner Improvements -  Polling

gingerwizard:
This adds a polling mode to the force-merge runner and in the process closes https://github.com/elastic/rally/issues/1050

Some considerations:

1. The change adds a `mode` parameter. This defaults to `blocking` which causes the current behaviour. 
2. If set to `polling` we ignore the `request-timeout` parameter and timeout the force_merge API call after 1s. We then poll the tasks API at an interval of `poll-period` (default 10s), until no force-merge tasks are listed.
3. Currently there is no timeout on the tasks API calls. It is possible this is susceptible to cluster pressure and may itself timeout under heavy load. However, i believe this represents a valid error condition.
4. The 1s timeout is a workaround to the force_merge API not supporting an async call and returning a task id. If multiple force_merge calls are issued in different runners in parallel, all force merges must complete to unblock. This seems low risk as consecutive calls are effectively a noop if `max-num-segments` has been reached.

Will add docs once parameter naming and usage is agreed.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -172,3 +172,114 @@ Rally will push metrics to the metric store configured in 1. and they can be vis
 To tear down everything issue ``./stop.sh``.
 
 It is possible to specify a different version of Elasticsearch for step 3. by setting ``export ES_VERSION=<the_desired_version>``.
+
+Identifying when errors have been encountered
+--------------------------------------------------------------
+
+Custom track development can be error prone especially if you are testing a new query. A number of reasons can lead to queries returning errors.
+
+Consider a simple example Rally operation::
+
+    {
+      "name": "geo_distance",
+      "operation-type": "search",
+      "index": "logs-*",
+      "body": {
+        "query": {
+           "geo_distance": {
+              "distance": "12km",
+              "source.geo.location": "40,-70"
+           }
+        }
+      }
+    }
+
+This query requires the field ``source.geo.location`` to be mapped as a ``geo_point`` type. If incorrectly mapped, Elasticsearch will respond with an error. 
+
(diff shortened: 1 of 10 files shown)


gingerwizard:
@dliappis This is ready for an initial review now all CI tests pass.

dliappis (changes requested):
Thank you for this super useful addition!

I left a number of style comments (unfortunately we skip many errors that pylint would normally report, see #838)

I haven't tested this yet, will do this asap.

gingerwizard:
@dliappis i believe i have addressed the formatting issues. If you're happy with the approach i'll add docs.

docs/recipes.rst
```diff
@@ -172,3 +172,114 @@ Rally will push metrics to the metric store configured in 1. and they can be vis
...
+	    "hits_relation" : "eq",
+	    "timed_out" : false,
+	    "took" : 110,
+	    "success" : true
+	  },
+	  "task" : "scroll",
+	  "operation" : "scroll",
+	  "operation-type" : "Search"
+	}
+
```
dliappis, line 285: FYI These doc fragmants seem to have come from https://github.com/elastic/rally/pull/1051/commits/89ed934ad78088260620f742b9e2e170b5a9b9b6 (they've already been merged in https://github.com/elastic/rally/pull/1049).

dliappis (commented):
@gingerwizard I tested this yesterday in a real long benchmark where force merging down to 1 segment takes a very long time the polling mode worked marvelously. Let's proceed with docs.

docs/recipes.rst
dliappis: @gingerwizard I still see this unrelated recipe in docs (coming from the older PR).

dliappis (commented):
Have you added docs already for the polling mode of the force merge runner? I didn't see anything in this review cycle.

dliappis:
@elasticmachine test this

dliappis:
@elasticmachine run tests

PR #1057, opened 2020-09-14
Polling force merge

gingerwizard:
Cleaner version of Force Merge Support 
https://github.com/elastic/rally/pull/1051

1. The change adds a `mode` parameter. This defaults to `blocking` which causes the current behaviour. 
2. If set to `polling` we ignore the `request-timeout` parameter and timeout the force_merge API call after 1s. We then poll the tasks API at an interval of `poll-period` (default 10s), until no force-merge tasks are listed.
3. Currently there is no timeout on the tasks API calls. It is possible this is susceptible to cluster pressure and may itself timeout under heavy load. However, i believe this represents a valid error condition.
4. The 1s timeout is a workaround to the force_merge API not supporting an async call and returning a task id. If multiple force_merge calls are issued in different runners in parallel, all force merges must complete to unblock. This seems low risk as consecutive calls are effectively a noop if `max-num-segments` has been reached.


docs/track.rst
```diff
@@ -524,6 +524,8 @@ force-merge
 With the operation type ``force-merge`` you can call the `force merge API <http://www.elastic.co/guide/en/elasticsearch/reference/current/indices-forcemerge.html>`_. On older versions of Elasticsearch (prior to 2.1), Rally will use the ``optimize API`` instead. It supports the following parameters:
 
 * ``index`` (optional, defaults to the indices defined in the ``indices`` section or ``_all`` if no indices are defined there): The name of the index that should be force-merged.
+* ``mode`` (optional, default to ``blocking``): In the default ``blocking`` mode the Elasticsearch client blocks until the operation returns or times out as dictated by the :ref:`client-options <_clr_client_options>`. In mode `polling` the client timeout is ignored. Instead, the api call is given 1s to complete. If the operation has not finished, the operator will poll every ``poll-period`` until all force merges are complete.
```
dliappis, line 527: I was wondering if need to mention any changes to existing behavior in https://github.com/elastic/rally/blob/master/docs/migrate.rst; AFAIU the default behavior remains the same, so I think we are good here.

esrally/track/params.py
```diff
@@ -621,12 +621,16 @@ def __init__(self, track, params, **kwargs):
         self._index_name = params.get("index", default_index)
         self._max_num_segments = params.get("max-num-segments")
         self._request_timeout = params.get("request-timeout")
+        self._poll_period = params.get("poll-period", 1)
```
dliappis, line 624: Isn't the default `poll-period` supposed to be `10s`?

dliappis (commented):
This is basically LGTM from me. Left one question about the default poll-period, which I think is currently `1` in the code?

docs/track.rst
gingerwizard: Yes default behaviour is identical.

esrally/track/params.py
gingerwizard: Well spotted @dliappis. Bumped to 10. This is maybe a bit conservative but the users can always configure.

dliappis: Maybe we could have a tiny unittest to ensure this gets checked properly?

dliappis (approved):
LGTM (provided CI becomes green after the latest commit).

I left a suggestion that perhaps we could benefit from a unit test checking the default `poll-period` but there's no need for another review cycle.

Issue #335, opened 2017-10-23
Ability to hook into and customize data download and distribution


cdahlqvist:
Rally currently handles download, decompression, and partitioning of data files automatically in the background. It would be very useful to have hooks into this process so that it can be overridden or customized for specific tracks. I do not necessarily see any need to customize the download and decompression, but it would be useful if it was possible to perform some level of preprocessing on the files before they are being used and also control how the data is distributed across worker threads.

cdahlqvist:
I realized it may be useful to add a bit more context around this feature.

The built-in functionality to download and extract data files is very useful but does not offer a lot of flexibility around how to use and distribute the data. If we were able to hook into this process at various stages it would be possible to e.g. pre-process the downloaded and extracted file and break up records into parts that allows real-time timestamps to efficiently be inserted while using this data as a template. This would allow us to avoid wasting CPU cycles on parsing, altering and serializing each event repeatedly.

danielmitterdorfer:
I'm closing this one in favor of the slightly more generic issue https://github.com/elastic/rally/issues/1066. Its scope is a bit wider, i.e. it will include everything that is requested here.

PR #1058, opened 2020-09-14
Noop for Async Elasticsearch client

gingerwizard:
Currently any custom async runners must make an Elasticsearch call. If they don't the code responsible for invoking the runner fails as it expects a "request_end" key on the context - https://github.com/elastic/rally/blob/master/esrally/driver/driver.py#L1338-L1343

We anticipate runners in the future which might do something else rather than query ES. This represents a larger effort and needs a clear scope and discussion. For now, this introduces a simple `noop` on the ES client. This can be called by a customer runner, after which it can proceed to do any desired work. This also provides the opportunity for the user to introduce their own timing, with a simple precondition check that timing can't result in a negative value.

By default the noop records no time.

### DIFF
--- a/esrally/client.py
+++ b/esrally/client.py
@@ -30,6 +30,7 @@ class EsClientFactory:
     """
     Abstracts how the Elasticsearch client is created. Intended for testing.
     """
+
     def __init__(self, hosts, client_options):
         self.hosts = hosts
         self.client_options = dict(client_options)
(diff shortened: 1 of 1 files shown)


danielmitterdorfer (commented):
I had a look at the PR that motivated this change (https://github.com/elastic/rally-tracks/pull/131). The purpose of a runner is to encapsulate remote API calls. Request timing infrastructure is instead implemented in the client that is injected into the runner. Should Rally support other target systems than Elasticsearch in the future then we should ensure that the respective client implementation also provides timing information keeping the runner free of such details and focussing only on the intercation with the remote system.

I think the affected runner in https://github.com/elastic/rally-tracks/pull/131 should only be used as a temporary workaround and we should instead add support in Rally to explicitly pre-process / generate data as a dedicated step (see https://github.com/elastic/rally/issues/335) and then remove the runner added in https://github.com/elastic/rally-tracks/pull/131. Meanwhile I would prefer that we keep that workaround encapsulated in the runner by retrieving the request context explicitly, setting the necessary keys and closing this PR. Wdyt?

PR #1059, opened 2020-09-14
Bump ES Client, Asyncio client and add pytest-asyncio

gingerwizard:
Bumps ES client to 7.9.1.
Introduces `pytest-asyncio`. This allows use of fixtures in mock tests for async methods. `pytest` needs moving up in minors to support.

All tests pass.


dliappis (approved):
LGTM

PR #1062, opened 2020-09-17
Document how to benchmark an Elastic Cloud cluster

danielmitterdorfer:
With this commit we add an example to our docs that demonstrates how to
benchmark an Elastic Cloud cluster. It is very similar to benchmarking
any other existing cluster but the details have confused users in the
past so we add this example for clarification.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -3,6 +3,19 @@ Tips and Tricks
 
 This section covers various tips and tricks in a recipe-style fashion.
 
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM thank you!

marcelopm:
Is benchmarking a Elastic Cloud cluster no longer supported?

I can't find the doc changes from the previously merged pull request on README.md

Issue #1063, opened 2020-09-17
Fail fast is leading `/` is missing from `path` in raw-transport operation


dliappis:
While we do document specifically that a leading `/` is mandatory in [raw-request docs](https://github.com/elastic/rally/blob/master/docs/track.rst#raw-request) in case it's missing it can be extremely hard to debug as logs (this is with `DEBUG` enabled on the elasticsearch logger) can look like:

```
2020-09-17 10:49:01,935 ActorAddr-(T|:42129)/PID:16991 elasticsearch WARNING GET https://04254962XXXXXX.my.cloud.region:9243/indexname/_settings [status:400 request:0.011s]
2020-09-17 10:49:01,935 ActorAddr-(T|:42129)/PID:16991 elasticsearch DEBUG > None
2020-09-17 10:49:01,935 ActorAddr-(T|:42129)/PID:16991 elasticsearch DEBUG < b'400 Bad Request'
2020-09-17 10:49:01,936 ActorAddr-(T|:42129)/PID:16991 elasticsearch WARNING Undecodable raw error response from server: Extra data: line 1 column 5 (char 4)
2020-09-17 10:49:01,938 ActorAddr-(T|:42129)/PID:16991 esrally.driver.driver INFO Total run duration: 0.014675 seconds.
2020-09-17 10:49:01,936 ActorAddr-(T|:42129)/PID:16991 esrally.driver.driver ERROR Could not execute schedule
Traceback (most recent call last):
  File "/home/esbench/rally/esrally/driver/driver.py", line 1340, in __call__
    total_ops, total_ops_unit, request_meta_data = await execute_single(runner, self.es, params, self.on_error)
  File "/home/esbench/rally/esrally/driver/driver.py", line 1451, in execute_single
    raise exceptions.RallyAssertionError(msg)
esrally.exceptions.RallyAssertionError: ("Request returned an error. Error type: transport, Description: b'400 Bad Request'", None)
```

We use [yarl](https://github.com/aio-libs/yarl) which relies on `urllib` for some things like `urlunsplit` (ref: https://github.com/aio-libs/yarl/blob/959e8630f512d61838d20b392e3e7a48adb05849/yarl/_url.py#L258-L262) and it doesn't verify the path conformity.

We should instead do the check in the [runner](https://github.com/elastic/rally/blob/c578034e283dbfe893f9dbc921f84936dd2bcf47/esrally/driver/runner.py#L1421-L1435) and fail accordingly.

PR #1064, opened 2020-09-18
Fail fast if leading `/` is missing from `path` in raw-request

ebadyano:
Check for mandatory leading `/` in path parameter of raw-request. Raise RallyException if it's missing.

Closes https://github.com/elastic/rally/issues/1063


dliappis (approved):
LGTM (left one nit)

ebadyano:
@dliappis Thank you for the review!

PR #1065, opened 2020-09-29
Allow to specify corpus meta-data

danielmitterdorfer:
With this commit we add a new property `meta` to the `corpus` property
and its nested documents. This allows users to specify arbitrary
additional meta-data that can e.g. be used by custom runners.


gingerwizard (approved):
LGTM. Thanks for this it will be allow us to centralize alot of custom config duplicated amongst operators. Maybe the docs could illustrate this is its primary purpose?

Issue #1000, opened 2020-05-28
Allow custom headers


vincentarnaud90:
**Describe the feature**:
Hello, 
I have a custom elasticsearch plugin that requires HTTP headers and following [this post](https://discuss.elastic.co/t/custom-headers/234669) on dicuss I'd like to know if it's possible to add the ability to pass custom headers either through `--client-options` and/or on a per operation basis by allowing an `headers` key in operations on track definition, thus allowing not to systematically use the `raw-request` operation.

danielmitterdorfer:
Closed by https://github.com/elastic/rally/pull/1070

PR #1070, opened 2020-10-05
Add support for custom headers and request-timeout

DJRickyB:
Addresses #567 and #1000 
Well it became trivial to add `headers` and `opaque-id` support for everything in the course of this refactor so I added that too.

Done:
* support request-timeout parameter for most operation types
* support opaque-id parameter for most operation types
* support headers dict parameter for most operation types
* document expanded parameter support


DJRickyB:
P.S. I have signed the contributor agreement, will troubleshoot the mismatch

esrally/driver/runner.py
```diff
@@ -643,11 +645,7 @@ async def __call__(self, es, params):
         except elasticsearch.TransportError as e:
             # this is caused by older versions of Elasticsearch (< 2.1), fall back to optimize
             if e.status_code == 400:
-                if max_num_segments:
-                    await es.transport.perform_request("POST", f"/_optimize?max_num_segments={max_num_segments}",
-                                                       timeout=request_timeout)
-                else:
-                    await es.transport.perform_request("POST", "/_optimize", timeout=request_timeout)
+                await es.transport.perform_request(method="POST", url="/_optimize", params=merge_params)
```
DJRickyB, line 648: `transport.perform_request()` expects the `timeout` parameter in the `params` arg.  Throws
```
TypeError: perform_request() got an unexpected keyword argument 'timeout'
```

esrally/driver/runner.py
```diff
@@ -310,21 +310,24 @@ async def __call__(self, es, params):
...
+        * ``detailed-results``: If ``True``, the runner will analyze the response and add detailed meta-data. Defaults
+                                to ``False``. Note that this has a very significant impact on performance and will very
+                                likely cause a bottleneck in the benchmark driver so please be very cautious enabling
+                                this feature. Our own measurements have shown a median overhead of several thousand
+                                times (execution time is in the single digit microsecond range when this feature is
+                                disabled and in the single digit millisecond range when this feature is enabled; numbers
+                                based on a bulk size of 500 elements and no errors). For details please refer to the
+                                respective benchmarks in ``benchmarks/driver``.
+        * ``request-timeout``: a non-negative float indicating the client-side timeout for the operation.  If not
+                               present, defaults to ``None`` and potentially falls back to the global timeout setting.
```
hub-cap, line 330: Please apply the same line length formatting as the other multi line length comments

esrally/driver/runner.py
```diff
@@ -614,17 +619,14 @@ class ForceMerge(Runner):
...
-        # a special handling for the force-merge API.
         request_timeout = params.get("request-timeout")
         mode = params.get("mode")
         merge_params = {"request_timeout": request_timeout}
         if max_num_segments:
             merge_params["max_num_segments"] = max_num_segments
         try:
             if mode == "polling":
-                # we ignore the request_timeout if we are in polling mode and deliberately timeout early
+                # no reason to wait as long as a whole {polling-period} (minimum 1 second)
```
hub-cap, line 629: What was the reasoning for changing the comment here?

esrally/driver/runner.py
hub-cap: it looks like you removed the logic here about `max_num_segments`. Good catch about the timeout though, lets definitely fix that here once you revert the code change that deleted the `if` stmt

hub-cap:
Hi, @DJRickyB , thank you very much for your contribution! This is good work so far. I have one overall suggestion that might make the code cleaner. It also requires a bit of context.

We hope to one day utilize more than just the timeout in the future, for instance, [the opaque-id](https://github.com/elastic/elasticsearch/issues/31521), so it might make sense to consolidate all of the different `Runner` implementations to use a single method, such as [this method](https://github.com/elastic/rally/blob/master/esrally/driver/runner.py#L915), and use it to populate all of the Runner classes. As is, many of them do some very similar operations with their `params` to extract the `request-params`, and maybe we can use a common method to both extract that, as well as ensure we populate the timeout. What do you think of this idea/suggestion? 

I have also added just a few minor nits to the pull request that I saw right off the bat, before realizing we could extract the common functionality into the Runners.

hub-cap (changes requested):
Oops! I meant to post that comment into this review, not as just a comment. here are the nits I mentioned in the review itself!

esrally/driver/runner.py
```diff
@@ -310,21 +310,24 @@ async def __call__(self, es, params):
 
         * ``body``: containing all documents for the current bulk request.
         * ``bulk-size``: the number of documents in this bulk.
-        * ``action_metadata_present``: if ``True``, assume that an action and metadata line is present (meaning only half of the lines
-        contain actual documents to index)
+        * ``action_metadata_present``: if ``True``, assume that an action and metadata line is present (meaning only
```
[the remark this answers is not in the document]
DJRickyB: This is a big reason I stopped at a draft PR for feedback before doing too much.  I'd like to document  my parameter additions but the docs in this file are not uniform, and don't all comply with pep8, so I wasn't sure I would compound tech debt.  

If it's ok with the team I'll go through and format it like e.g. elasticsearch-py, and push all those changes in a single commit so its review can be separate.  If not, I'll just make each param addition fit in with its local comment block

esrally/driver/runner.py
DJRickyB: it took me a while to understand why 1 was acceptable.  it's only obvious to me now in the diff that i replaced quite a bit of context also.  I will change this to include both comments if that is ok

esrally/driver/runner.py
DJRickyB: max_num_segments exists in the merge_params and gets applied to the query string by the perform_request method

esrally/driver/runner.py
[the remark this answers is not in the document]
hub-cap: I like this approach.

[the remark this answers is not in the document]
DJRickyB: I was going to ask to clarify which approach you like but I just decided to put formatting changes in a separate PR.  will revert this and other noted lines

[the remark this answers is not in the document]
hub-cap: feel free to @mention me to get my attention on that PR

DJRickyB:
Thank you Michael!  Your timing is perfect, I was planning on finishing this PR this weekend.  Please pardon the force-push, it's not usually my style but I wanted a way to get prior git history to apply to the comment lines I had touched

esrally/driver/runner.py
```diff
@@ -142,6 +142,28 @@ async def __call__(self, *args):
     async def __aexit__(self, exc_type, exc_val, exc_tb):
         return False
 
+    def _default_kw_params(self, params):
+        # map of API kwargs to Rally config parameters
+        kw_dict = {
+            "body": "body",
+            "index": "index",
+            "opaque_id": "opaque-id",
```
DJRickyB, line 150: because this exists in the queryparams decorator applied to all API calls that we use, it was trivial to add it here for handling.  the other thing I want to add is `headers`, which technically collides with the opaque_id but precedence is handled here: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L140

esrally/driver/runner.py
```diff
@@ -477,15 +502,16 @@ async def __call__(self, es, params):
...
+            response = await es.bulk(params=bulk_params, **api_kwargs)
         else:
-            response = await es.bulk(body=params["body"], index=index, doc_type=params.get("type"), params=bulk_params)
+            response = await es.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
 
         stats = self.detailed_stats(params, bulk_size, response) if detailed_results else self.simple_stats(bulk_size, response)
 
         meta_data = {
-            "index": str(index) if index else None,
+            "index": params.get('index'),
```
hub-cap, line 514: pls use double quotes here instead of single

tests/driver/runner_test.py
```diff
@@ -2010,7 +2154,40 @@ async def test_accepts_better_cluster_status(self, es):
...
+        es.cluster.health.assert_called_once_with(params={"wait_for_status": "yellow"})
+
+    @mock.patch("elasticsearch.Elasticsearch")
+    @run_async
+    async def test_cluster_health_with_timeout_and_headers(self, es):
+        es.cluster.health.return_value = as_future({
+            "status": "green",
+            "relocating_shards": 0
+        })
+        r = runner.ClusterHealth()
```
hub-cap, line 2166: we try not to have single value names for variables, lets make this a bit more descriptive

tests/driver/runner_test.py
```diff
@@ -2099,6 +2276,66 @@ async def test_creates_multiple_indices(self, es):
             mock.call(index="indexB", body={"settings": {}}, params=request_params)
         ])
 
+    @mock.patch("elasticsearch.Elasticsearch")
+    @run_async
+    async def test_create_with_timeout_and_headers(self, es):
+        es.indices.create.return_value = as_future()
+
+        r = runner.CreateIndex()
```
hub-cap, line 2284: same w single character variable names

hub-cap (changes requested):
wow this looks great now! I love how its not touching all the tests. Also, Thank you SO VERY much for making the PR name more descriptive. Just some minor nits now and this will be good to go.

One other thing, go ahead and merge master in `git merge upstream/master` and run `make precommit`, and it will show you one line length failure you have. This will fail before the tests will run. Once you are done w this ill make sure our CI infra can run the tests.

hub-cap:
@elasticmachine ok to test

hub-cap:
grr, my magic incantation is incorrect.

@elasticmachine add to allowlist please

hub-cap:
ok at this point w/ the proper incantation, your PR will now run tests for each commit. If it fails for some reason, go ahead and fix it up (or ask for help if need be), and push, and it should run the tests again.

DJRickyB:
is there somewhere i can push new test tracks to exercise the functionality? good to see it looks like no regressions :)

hub-cap:
If you push some test tracks, it would be rad to just push them to a branch, and then link the commands you use (assuming the repos were checked out locally). Ill test them myself to double check and then we can call this done and merge it. Im off tomorrow, but I should be able to get to them by monday if you push some branches up to your fork of rally tracks.

hub-cap:
Also, I forgot to add, if you are using and existing corpora, you can use the built in rally `--test-mode` flag which will only ingest a small sample set of data, so your feedback loop is much tighter when testing track changes.

DJRickyB:
Thanks, sorry for the delay.  Running with local test tracks and watching with wireshark revealed a couple of misses in params.py based on how parameter parsing for a given operation was set up.  The changes I made add 3 `None` values to the params dictionary for a few types, so now the size check tests have changed in a few places (see failures from your CI).

I'm inclined to delete the size checks as there may be more of these in the future, and replace them with a variety of checks on the new parameter keys.  Thoughts?

FYI I'm just using this locally: https://github.com/elastic/rally-tracks/commit/d7b1ac4d92ca6e49b326f72d2f24ff4b612457f4

hub-cap (approved):
yay, its merge time!

PR #1071, opened 2020-10-05
Fix parallel completed-by task identification

### LINKED ISSUE #1068: completed-by property in parallel element wrongly expects operation-type and not task name
Documentation for the [completed-by](https://github.com/elastic/rally/blame/master/docs/track.rst#L452) property in the parallel element read:

> Allows to define the name of one task in the tasks list.

It seems, however, that there's a bug and right now it expects the operation-type instead.

This is easily reproduced with the following track:

<details><summary>track.json</summary>

```
{
  "name": "test-completed-by-bug",
  "description": "Track to reproduce bug about completed-by property of parallel element.",
  "schedule": [
    {
      "parallel": {
        "completed-by": "run-node-stats",
        "tasks": [
          {
            "name": "check-cluster-health",
            "operation": {
              "operation-type": "cluster-health"
            }
          },
          {
            "name": "run-node-stats",
            "operation": {
              "operation-type": "node-stats"
            },
            "iterations": 30
          }
        ]
      }
    }
  ]
}
```
</details>

If I store this `track.json` in a dedicated directory and run Rally against an ES cluster running locally on port 19200, I get:

```
# run elasticsearch, listening on port 19200
$ docker run --rm -e "discovery.type=single-node" -p 19200:9200 docker.elastic.co/elasticsearch/elasticsearch:7.9.1

# run the test track
$ ~/source/elastic/rally-parallel-bug $ esrally --target-hosts=127.0.0.1:19200 --track-path=$PWD


[ERROR] Cannot race. Error in race control (("Track 'rally-parallel-bug' is invalid. 'parallel' element for challenge 'default' is marked with 'completed-by' with task name 'run-node-stats' but no task with this name exists.", None))
```

this problem goes away when I change `completed-by` to `node-stats`.

hub-cap:
The code that parses the track previously used the operation name to
identify which task completes a parallel loop. This was fine if tasks
were not explicitly because the task name defaults to the operation
name. If a user explicitly creates a task name, this would fail. The fix
is to use the task name instead, which still defaults to the operation
name if not provided, which preserves compatibility.

Closes #1068


danielmitterdorfer:

esrally/track/loader.py
```diff
@@ -1263,7 +1264,7 @@ def parse_task(self, task_spec, ops, challenge_name, default_warmup_iterations=N
                                               default_value=default_time_period),
                           clients=self._r(task_spec, "clients", error_ctx=op.name, mandatory=False, default_value=1),
                           # this will work because op_name must always be set, i.e. it is never `None`.
-                          completes_parent=(op.name == completed_by_name),
```
danielmitterdorfer, line 1266: Nit: I think we can get rid of the comment in the line above as well as it is not applicable anymore IMHO.

danielmitterdorfer (approved):
Thanks for the fix. I left one nit but apart from that LGTM.

PR #1075, opened 2020-10-07
Ensure that request error meta-data is string

danielmitterdorfer:
With this commit we ensure that the `error-description` property that is
added to request metadata on errors always contains a string. This
avoids serialization issues later on when these objects are written to
the metrics store.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -1432,7 +1432,11 @@ async def execute_single(runner, es, params, on_error):
         elif e.info:
             request_meta_data["error-description"] = "%s (%s)" % (e.error, e.info)
         else:
-            request_meta_data["error-description"] = e.error
+            if isinstance(e.error, bytes):
+                error_description = e.error.decode("utf-8")
+            else:
+                error_description = str(e.error)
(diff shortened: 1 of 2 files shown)


esrally/driver/driver.py
```diff
@@ -1432,7 +1432,11 @@ async def execute_single(runner, es, params, on_error):
         elif e.info:
             request_meta_data["error-description"] = "%s (%s)" % (e.error, e.info)
         else:
-            request_meta_data["error-description"] = e.error
+            if isinstance(e.error, bytes):
+                error_description = e.error.decode("utf-8")
+            else:
+                error_description = str(e.error)
```
gingerwizard, line 1438: Would it be soo bad to leave as None given thats the only other viable type? For safety this probably makes sense.

gingerwizard (approved):

esrally/driver/driver.py
danielmitterdorfer: As discussed elsewhere I'd rather be a bit more cautious here and always enforce conversion to a string.

PR #1076, opened 2020-10-07
Avoid defining and not using variables

hub-cap:
With this commit we reenable the pylint check `W0612` which checks
whether a variable is defined but unused.

Relates #838


esrally/config.py
```diff
@@ -527,7 +527,7 @@ def migrate(config_file, current_version, target_version, out=print, i=input):
         raise ConfigError("The config file in {} is too old. Please delete it and reconfigure Rally from scratch with {} configure."
                           .format(config_file.location, PROGRAM_NAME))
 
-    prompter = Prompter(i=i, o=out, assume_defaults=False)
+    Prompter(i=i, o=out, assume_defaults=False)
```
danielmitterdorfer, line 530: How about we remove the entire statement as it does not have any effect?

danielmitterdorfer (commented):
Thanks! I left one comment. Can you please also provide a description for PRs to describe the purpose of this change? Specificically: I had to google what this warning is about to find out what I'm supposed to review.

esrally/config.py
hub-cap: I was afraid of unintended side affects here, so I did not remove. But Ill assume you know that it doesnt do something magic and remove :)

danielmitterdorfer (approved):
LGTM

PR #1077, opened 2020-10-07
Avoid using exit() or quit()

hub-cap:
With this commit we reenable the pylint check `R1722` which checks
whether sys.exit() is properly used.

Relates #838


dliappis (approved):
LGTM thanks!

PR #1080, opened 2020-10-07
Avoid reimporting modules

hub-cap:
With this commit we reenable the pylint check `W0404` which checks
whether an import has already been imported.

Relates #838


gingerwizard (approved):
LGTM

hub-cap:
@elasticmachine update branch

PR #1081, opened 2020-10-08
Avoid using incorrect number of parameters

hub-cap:
With this commit we reenable the pylint check `W0221` which checks
whether the correct number of parameters is used in an implemented
or overridden method.

Relates #838


danielmitterdorfer (approved):
Thanks! LGTM

hub-cap:
@elasticmachine update branch

PR #1082, opened 2020-10-08
Avoid assigning an expression to nothing

hub-cap:
With this commit we reenable the pylint check `W0106` which checks
whether an expression not a function call is assigned to nothing.

Relates #838


ebadyano:
lgtm, thank you

PR #1083, opened 2020-10-08
Avoid using protected members outside scope

hub-cap:
With this commit we reenable the pylint check `W0212` which checks
whether protected members are used outside of their classes scope.

Relates #838


gingerwizard (commented):
@hub-cap you have a conflict here to resolve. 

Whilst disabling warnings for tests makes sense, have we checked we can't mitigate some of these by just exposing accessors for the variables?

hub-cap:
The way I look at it, since these are almost all tests,. it doesn't make a ton of sense to expose members for them. 

Looks like there are a few new things to clean up, im working on them now.

gingerwizard:
@hub-cap see https://github.com/elastic/rally/pull/1106#pullrequestreview-523194430 why this maybe failed. @danielmitterdorfer corrected the reference

gingerwizard:
@hub-cap i dont think we need to make the changes to the static merge method anymore. See https://github.com/elastic/rally/pull/1106

esrally/track/params.py
```diff
@@ -430,31 +430,28 @@ def __init__(self, track, params, templates, **kwargs):
...
+                        body = self.create_or_merge(template.content, ["template", "settings"], settings)
                         self.template_definitions.append((template.name, body))
         else:
             raise exceptions.InvalidSyntax("Please set the properties 'template' and 'body' for the "
                                            f"{params.get('operation-type')} operation or declare composable and/or component "
                                            "templates in the track")
 
-    @staticmethod
-    def _create_or_merge(content, path, new_content):
+    def create_or_merge(self, content, path, new_content):
```
gingerwizard, line 440: I don't think we need to do this anymore - See #1106

esrally/track/params.py
```diff
@@ -430,31 +430,28 @@ def __init__(self, track, params, templates, **kwargs):
...
                 if sub_path not in content:
                     content[sub_path] = {}
                 content = content[sub_path]
-            CreateComposableTemplateParamSource.__merge(content, new_content)
+            self.merge(content, new_content)
         return original_content
 
-    @staticmethod
-    def __merge(dct, merge_dct):
+    def merge(self, dct, merge_dct):
```
gingerwizard, line 450: see above

gingerwizard (commented):
See changes due to #1106

hub-cap:
I reverted the code for that, ty @gingerwizard for the pointer.

gingerwizard (approved):
LGTM now we have the other change. You have a conflict but this looks easily fixed.

PR #1084, opened 2020-10-08
Avoid passing too few parameters into function calls

hub-cap:
With this commit we reenable the pylint check `E1120` which checks
whether function calls contain the correct number of arguments.

Relates #838


danielmitterdorfer (approved):
Thanks! LGTM

hub-cap:
@elasticmachine update branch

PR #1088, opened 2020-10-13
Use the new pip dependency resolver

hub-cap:
There is currently an issue with incompatible versions in our dependency
of aiohttp, which is causing installations to fail. Using the stricter
pip 2020 resolver, which is off by default in pip 20, fixes this
issue. The new resolver, which will be the new default in future pip
versions, is not ready for all use cases, but works fine for our use
case, so we are choosing to use it.

Relates https://discuss.python.org/t/announcement-pip-20-2-release/4863
Relates https://github.com/aio-libs/aiohttp/commit/2d36ef6911a1d49001381da40d658a70d64168fa
Relates https://github.com/aio-libs/aiohttp/issues/4987

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,9 +60,9 @@ check-venv:
 
 install: venv-create
 	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install --upgrade pip setuptools wheel
-	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install -e .
+	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install -e . --use-feature=2020-resolver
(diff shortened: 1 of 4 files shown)


sethmlarson (approved):
LGTM!

hub-cap:
Closing in favor of https://github.com/elastic/elasticsearch-py/pull/1401

dliappis:
Based on an offline discussion, we think there's value having this rather than relying on the next released version based on https://github.com/elastic/elasticsearch-py/pull/1401, so I've reopened this.

sethmlarson (approved):

dliappis (approved):
LGTM

danielmitterdorfer (approved):
LGTM

PR #1091, opened 2020-10-16
Check registered parameter sources more strictly

danielmitterdorfer:
With this commit we add a stricter check when parameter sources are
registered with Rally. This ensures only functions or classes are
registered. If a user mistakenly registers e.g. an instance,
registration will fail immediately.

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -15,13 +15,13 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import inspect
 import logging
 import math
 import numbers
 import operator
 import random
 import time
-import types
 from enum import Enum
 
 from esrally import exceptions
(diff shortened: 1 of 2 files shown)


esrally/track/params.py
```diff
@@ -43,18 +43,24 @@ def param_source_for_operation(op_type, track, params, task_name):
...
+    if inspect.isfunction(param_source):
         return DelegatingParamSource(track, params, param_source)
     else:
         return param_source(track, params)
 
 
+def ensure_valid_param_source(param_source):
+    if not inspect.isfunction(param_source) and not inspect.isclass(param_source):
+        raise exceptions.RallyAssertionError(f"Parameter source [{param_source}] must be either a function or a class.")
+
```
gingerwizard, line 55: nit: 2 spaces or one between methods? We seem to have an inconsistency

gingerwizard (approved):
LGTM. I'll refer to @danielmitterdorfer on spacing preferences.

esrally/track/params.py
danielmitterdorfer: Thanks for pointing this out.

Unfortunately, the linter that we have configured for this project (`pylint`) cannot check this but I ran [pycodestyle](https://pypi.org/project/pycodestyle/) against the code base locally and only found one spot where we only use one line instead of two lines:

```
(.venv) daniel@io:Projects/rally ‹param-source-check›$ pycodestyle esrally | grep E302
esrally/track/loader.py:726:1: E302 expected 2 blank lines, found 1
```

I'll fix that separately.

PR #1094, opened 2020-10-16
Add support for Google Cloud Storage buckets

dliappis:
This commit adds support for GCS hosted buckets storing source data.

While at it, switch `utils/net.py` tests to pytest.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -225,10 +225,10 @@ The ``corpora`` section contains all document corpora that are used by this trac
 
 Each entry in the ``documents`` list consists of the following properties:
 
-* ``base-url`` (optional): A http(s), S3 or GS URL that points to the root path where Rally can obtain the corresponding source file. Rally can also download data from private S3 or GS buckets if access is properly configured:
+* ``base-url`` (optional): A http(s), S3 or Google Storage URL that points to the root path where Rally can obtain the corresponding source file. Rally can also download data from private S3 or Google Storage buckets if access is properly configured:
 
   * S3 according to `docs <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration>`_.
-  * GS: Either using `client library authentication <https://cloud.google.com/storage/docs/reference/libraries#setting_up_authentication>`_ or by presenting an `oath2 token <https://cloud.google.com/storage/docs/authentication>`_ via the ``GOOGLE_AUTH_TOKEN`` environment variable, typically done using: ``export GOOGLE_AUTH_TOKEN=$(gcloud auth print-access-token)``.
+  * Google Storage: Either using `client library authentication <https://cloud.google.com/storage/docs/reference/libraries#setting_up_authentication>`_ or by presenting an `oauth2 token <https://cloud.google.com/storage/docs/authentication>`_ via the ``GOOGLE_AUTH_TOKEN`` environment variable, typically done using: ``export GOOGLE_AUTH_TOKEN=$(gcloud auth print-access-token)``.
 * ``source-format`` (optional, default: ``bulk``): Defines in which format Rally should interpret the data file specified by ``source-file``. Currently, only ``bulk`` is supported.
 * ``source-file`` (mandatory): File name of the corresponding documents. For local use, this file can be a ``.json`` file. If you provide a ``base-url`` we recommend that you provide a compressed file here. The following extensions are supported: ``.zip``, ``.bz2``, ``.gz``, ``.tar``, ``.tar.gz``, ``.tgz`` or ``.tar.bz2``. It must contain exactly one JSON file with the same name. The preferred file extension for our official tracks is ``.bz2``.
 * ``includes-action-and-meta-data`` (optional, defaults to ``false``): Defines whether the documents file contains already an action and meta-data line (``true``) or only documents (``false``).
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -225,7 +225,10 @@ The ``corpora`` section contains all document corpora that are used by this trac
 
 Each entry in the ``documents`` list consists of the following properties:
 
-* ``base-url`` (optional): A http(s) or S3 URL that points to the root path where Rally can obtain the corresponding source file. Rally can also download data from private S3 buckets if access is properly `configured <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration>`_.
+* ``base-url`` (optional): A http(s), S3 or GS URL that points to the root path where Rally can obtain the corresponding source file. Rally can also download data from private S3 or GS buckets if access is properly configured:
+
+  * S3 according to `docs <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration>`_.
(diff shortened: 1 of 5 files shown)


esrally/utils/net.py
```diff
@@ -94,17 +95,64 @@ def __call__(self, bytes_amount):
...
+
+    with open(local_path, "wb") as local_fp:
+        media_url = _build_gcs_object_url(bucket_name, bucket_path)
+        download = ChunkedDownload(media_url, chunk_size, local_fp)
+        # allow us to calculate the total bytes
+        download.consume_next_chunk(transport)
+        if not expected_size_in_bytes:
+            expected_size_in_bytes = download.total_bytes
+        while not download.finished:
+            if progress_indicator and download.bytes_downloaded and download.total_bytes:
```
hub-cap, line 139: is there ever a case where we wont have a `progress_indicator`?

esrally/utils/net.py
```diff
@@ -94,17 +95,64 @@ def __call__(self, bytes_amount):
...
+    bucket_and_path = url[5:]  # s3:// or gs:// prefix for now
     bucket_end_index = bucket_and_path.find("/")
     bucket = bucket_and_path[:bucket_end_index]
     # we need to remove the leading "/"
     bucket_path = bucket_and_path[bucket_end_index + 1:]
 
-    logger.info("Downloading from S3 bucket [%s] and path [%s] to [%s].", bucket, bucket_path, local_path)
-    _download_from_s3_bucket(bucket, bucket_path, local_path, expected_size_in_bytes, progress_indicator)
+    logger.info("Downloading from [%s] bucket [%s] and path [%s] to [%s].", blobstore, bucket, bucket_path, local_path)
+    blob_downloader[blobstore](bucket, bucket_path, local_path, expected_size_in_bytes, progress_indicator)
```
hub-cap, line 155: entirely up to you, but i find `if` statements easier to reason about than looking up some string->function call magic

hub-cap (approved):
This is great! Just 2 minor nits/Q's and thats it from me.

docs/track.rst
```diff
@@ -225,7 +225,10 @@ The ``corpora`` section contains all document corpora that are used by this trac
 
 Each entry in the ``documents`` list consists of the following properties:
 
-* ``base-url`` (optional): A http(s) or S3 URL that points to the root path where Rally can obtain the corresponding source file. Rally can also download data from private S3 buckets if access is properly `configured <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration>`_.
+* ``base-url`` (optional): A http(s), S3 or GS URL that points to the root path where Rally can obtain the corresponding source file. Rally can also download data from private S3 or GS buckets if access is properly configured:
```
danielmitterdorfer, line 228: typo: `GS` -> `GCS`

docs/track.rst
```diff
@@ -225,7 +225,10 @@ The ``corpora`` section contains all document corpora that are used by this trac
...
+
+  * S3 according to `docs <https://boto3.amazonaws.com/v1/documentation/api/latest/guide/quickstart.html#configuration>`_.
+  * GS: Either using `client library authentication <https://cloud.google.com/storage/docs/reference/libraries#setting_up_authentication>`_ or by presenting an `oath2 token <https://cloud.google.com/storage/docs/authentication>`_ via the ``GOOGLE_AUTH_TOKEN`` environment variable, typically done using: ``export GOOGLE_AUTH_TOKEN=$(gcloud auth print-access-token)``.
```
danielmitterdorfer, line 231: typo: `GS` -> `GCS`

danielmitterdorfer, line 231: typo: `oath2` -> `oauth2`.

setup.py
```diff
@@ -79,7 +79,13 @@ def str_from_file(name):
     # License: Apache 2.0
     "yappi==1.2.3",
     # License: BSD
-    "ijson==2.6.1"
+    "ijson==2.6.1",
+    # License: Apache 2.0
+    # transitive dependencies:
+    #   google-crc32c: Apache 2.0
+    "google-resumable-media==1.1.0",
```
danielmitterdorfer, line 86: You also need to update the script `create-notice.sh` accordingly. I also wonder whether we should investigate to make S3 and GCS optional dependencies and avoid pulling them in core Rally. I'd be ok if we tackle this separately though.

esrally/utils/net.py
```diff
@@ -146,8 +194,9 @@ def download(url, local_path, expected_size_in_bytes=None, progress_indicator=No
     """
     tmp_data_set_path = local_path + ".tmp"
     try:
-        if url.startswith("s3"):
-            expected_size_in_bytes = download_s3(url, tmp_data_set_path, expected_size_in_bytes, progress_indicator)
+        scheme = urllib3.util.parse_url(url).scheme
+        if scheme in ["s3", "gs"]:
```
danielmitterdorfer, line 198: I wonder whether we should start encapsulating this now that we add a third scheme. We could implement this as a `Downloader` that supports certain schemes, register them upon startup and implement `def supports(self, url: str) -> bool` to check to which `Downloader` to pass a URL instead of spreading this logic in multiple places. Wdyt?

danielmitterdorfer (changes requested):
Thanks! I left some comments around the design and we need to add the new dependencies to our notice file.

docs/track.rst
dliappis: The scheme is gs not gcs (i.e. `gs://bucket`). Also the service itself is called Google Storage (https://cloud.google.com/storage).

danielmitterdorfer: While Amazon calls the product "S3", the official name of the equivalent Google service seems to be "Cloud Storage" according to the docs you point to. Should we then refer to it by that name instead of a different acronym? Or is "GS" the official acronym (I did not find it in Google's documentation)?

setup.py
dliappis: Great catch. Done in 27419b61228b130511f2ced2e08f6bc32ac37a42 (where I also fixed the URL for `yarl` which was actually downloading html).
As discussed offline, will create an issue for making the S3 dependencies optional.

esrally/utils/net.py
dliappis: As discussed offline, we'll defer this for now, but certainly look at it again, if we are to extend to another scheme.

docs/track.rst
dliappis: Fixed in 27419b61228b130511f2ced2e08f6bc32ac37a42

docs/track.rst
dliappis: Standardizing on `Google Storage` and fixed in 27419b61228b130511f2ced2e08f6bc32ac37a42

docs/track.rst
dliappis: Fixed in c3fa7b9b61a9b5dffa98d9af57ad07468f0ca6cd

esrally/utils/net.py
dliappis: We use it in https://github.com/elastic/rally/blob/1bdec4ff774e9c66640124437189bda1dd9cc58d/esrally/mechanic/supplier.py#L519 and https://github.com/elastic/rally/blob/1bdec4ff774e9c66640124437189bda1dd9cc58d/esrally/track/loader.py#L410 but want to be able to support a case (in the future?) were it shouldn't be used.

esrally/utils/net.py
dliappis: Thanks, used it because it is a pattern we use elsewhere too.

danielmitterdorfer (approved):
Thanks for iterating. LGTM

PR #1097, opened 2020-10-22
Upgrade google-auth to latest

danielmitterdorfer:
With this commit we upgrade google-auth from 1.21.1 to the currently
latest version 1.22.1.


dliappis (approved):
LGTM (assuming that CI will also be green :) )

PR #1092, opened 2020-10-16
Add Support for Datastreams

gingerwizard:
Addresses https://github.com/elastic/rally/issues/1053

This PR adds support for data streams. I've tried to address this in a backwards compatible way i.e. it should in no way impact current use of indices. The PR itself is of limited use until we have https://github.com/elastic/rally/issues/1054 as we would be entirely reliant on dynamic mappings otherwise. Assuming this PR is good, i will address this functionality next and then document together ie. i propose we don't document until https://github.com/elastic/rally/issues/1054 is addressed.

This PR supports creating and deleting data streams through new explicit operations. Furthermore, we attempt to provide the same support for data streams as indices. More specifically, `data-streams` is a top level key in the track and `target-data-stream` can be specified on a corpus documents. These definitions will be re-usable in:

1. Force Merge
2. Search
3. Bulk Indexing. This is probably the most invasive as we support a create action.

In all cases we prefer the specification of an index over a data stream if both are specified e.g. we don't error on the specification of both.

Finally, there are cases where internally a data stream value is used to set an `index` construct. This can appear confusing but the ES api doesn't extinguish also e.g. a bulk operation would contain `{"create": {"_index": "logs-k8-application.log-default"}}` when indexing into a data stream

I've held off implementing statistics. I suspect we will want to implement the stats endpoint for datastreams as a telemetry device.

Finally, PR doesn't include integration tests. We can either wait for https://github.com/elastic/rally/issues/1054 or add limited now e.g. creation and deletion but not indexing.

PR #1099, opened 2020-10-22
Composable + Component template support

gingerwizard:
This adds support for composable + component templates thus addressing https://github.com/elastic/rally/issues/1099.

We add two keys to the track definition "composable-templates" and "component-templates". Some important notes:

1. Relies on the user ensuring all component templates required by the composable templates are provided by the user i.e. there is no check that the component templates referenced by the composable templates are provided.
2. The user must ensure all component templates are inserted before attempting to insert composable template i.e. they must use `create-component-templates` before `create-composable-templates` in their track. We could combine these into a single runner+parameter source if we don't see any benefit from seperating them - thus ensuring the order is handled for the user. @dliappis 
3. This adds to the PR https://github.com/elastic/rally/pull/1092 effectively making it useful. I will follow up with IT tests verifying these work nicely together. Finally, docs when we are happy to announce.

This is additive only and does not impact the now deprecated legacy templates. We need to consider how we will remove these and probably eventually promote "composable-templates"  to "templates".

Closes #1099


esrally/track/params.py
```diff
@@ -1122,6 +1203,10 @@ def read_bulk(self):
 register_param_source_for_operation(track.OperationType.DeleteDataStream, DeleteDataStreamParamSource)
 register_param_source_for_operation(track.OperationType.CreateIndexTemplate, CreateIndexTemplateParamSource)
 register_param_source_for_operation(track.OperationType.DeleteIndexTemplate, DeleteIndexTemplateParamSource)
+register_param_source_for_operation(track.OperationType.CreateComponentTemplate, CreateComponentTemplateParamSource)
+register_param_source_for_operation(track.OperationType.DeleteComponentTemplate, DeleteComponentTemplateParamSource)
+register_param_source_for_operation(track.OperationType.CreateComposableTemplate, CreateComposableTemplateParamSource)
+register_param_source_for_operation(track.OperationType.DeleteComposableTemplate, DeleteIndexTemplateParamSource)
```
gingerwizard, line 1209: Small note - we use existing `DeleteIndexTemplateParamSource` for `DeleteComposableTemplate`- didn't seem worth duplicating.

esrally/track/track.py
```diff
@@ -403,14 +419,15 @@ def __repr__(self):
 
     def __hash__(self):
         return hash(self.name) ^ hash(self.meta_data) ^ hash(self.description) ^ hash(self.challenges) ^ \
-               hash(self.indices) ^ hash(self.data_streams) ^ hash(self.templates) ^ hash(self.corpora)
+               hash(self.indices) ^ hash(self.templates) ^ hash(self.composable_templates) ^ hash(self.component_templates) \
```
gingerwizard, line 422: I have little confidence any of these __hash__ functions actually work but not because of this PR - you can't hash an iterable. Separate PR needed

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
         return "delete-data-stream"
 
 
+class CreateComponentTemplate(Runner):
+    """
+    Execute the `PUT index template API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-component-template.html>`_.
```
danielmitterdorfer, line 1112: Nit: In the docs we call it the "Put component template API", not the "PUT index template API".

esrally/driver/runner.py
```diff
@@ -23,7 +23,7 @@
 import types
 from collections import Counter, OrderedDict
 from copy import deepcopy
-
+from elasticsearch.client import _make_path
```
danielmitterdorfer, line 26: Can we please avoid top-level imports of Elasticsearch (logging issues)? Also, I am not certain why we need this? Can you please clarify?

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+                                                    params=request_params)
+        return len(templates), "ops"
+
+    def __repr__(self, *args, **kwargs):
+        return "create-component-template"
+
+
+class DeleteComponentTemplate(Runner):
+    """
+    Execute the `PUT index template API
```
danielmitterdorfer, line 1129: It's the "Delete component template API"

esrally/track/track.py
```diff
@@ -144,6 +144,20 @@ def __eq__(self, other):
         return self.name == other.name
 
 
+class ComponentTemplate(IndexTemplate):
```
danielmitterdorfer, line 147: Do we rely on polymorphism anywhere? I don't think we should rely on inheritance just to safe some code. Also [index templates are deprecated in Elasticsearch](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates-v1.html) and we should therefore decouple the two concepts as well in Rally.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
 
 
+class CreateComponentTemplate(Runner):
+    """
+    Execute the `PUT index template API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-component-template.html>`_.
+    """
+
+    async def __call__(self, es, params):
+        templates = mandatory(params, "templates", self)
+        request_params = params.get("request-params", {})
```
danielmitterdorfer, line 1117: This parameter should be mandatory as the parameter source needs to set it.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+
+class DeleteComponentTemplate(Runner):
+    """
+    Execute the `PUT index template API
+    <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-component-template.html>`_.
+    """
+
+    async def __call__(self, es, params):
+        template_names = mandatory(params, "templates", self)
+        only_if_exists = params.get("only-if-exists", False)
```
danielmitterdorfer, line 1135: This parameter should be mandatory as the parameter source needs to set it.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+        request_params = params.get("request-params", {})
```
danielmitterdorfer, line 1136: This parameter should be mandatory as the parameter source needs to set it.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+
+        async def _exists(name):
+            return await es.transport.perform_request(
```
danielmitterdorfer, line 1139: What's the reason for this low-level API call? Is this (yet?) unsupported by the Python client? Can you please add a clarifying comment?

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+
+
+class CreateComposableTemplate(Runner):
+    """
+    Execute the `PUT index template API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-template.html>`_.
+    """
+
+    async def __call__(self, es, params):
+        templates = mandatory(params, "templates", self)
+        request_params = params.get("request-params", {})
```
danielmitterdorfer, line 1165: This parameter should be mandatory as the parameter source needs to set it.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+
+
+class DeleteComposableTemplate(Runner):
+    """
+    Execute the `PUT index template API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-template.html>`_.
+    """
+
+    async def __call__(self, es, params):
+        template_names = mandatory(params, "templates", self)
+        only_if_exists = params.get("only-if-exists", False)
```
danielmitterdorfer, line 1182: This parameter should be mandatory as the parameter source needs to set it.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+        request_params = params.get("request-params", {})
```
danielmitterdorfer, line 1183: This parameter should be mandatory as the parameter source needs to set it.

esrally/resources/track-schema.json
```diff
@@ -207,6 +207,61 @@
       "type": "object",
       "description": "Meta-information which will be added to each metrics-record of this track."
     },
+    "component_templates": {
```
danielmitterdorfer, line 210: By convention we use hyphens for user-facing names if possible (sometimes it's not). Can we rename this to `component-templates`? Also, the loader code actually uses the hyphenated name already.

esrally/resources/track-schema.json
```diff
@@ -207,6 +207,61 @@
...
+            "description": "Component template file name."
+          }
+        },
+        "required": [
+          "name",
+          "template"
+        ]
+      }
+    },
+    "composable_templates": {
```
danielmitterdorfer, line 233: By convention we use hyphens for user-facing names if possible (sometimes it's not). Can we rename this to `composable-templates`? Also, the loader code actually uses the hyphenated name already.

esrally/driver/runner.py
```diff
@@ -1103,6 +1107,101 @@ def __repr__(self, *args, **kwargs):
...
+        ops_count = 0
+
+        for template_name, delete_matching_indices, index_pattern in template_names:
```
danielmitterdorfer, line 1186: Given that this expands to a tuple I think `templates` would be a more appropriate name than `template_names`?

esrally/track/loader.py
```diff
@@ -1041,6 +1046,18 @@ def _create_index(self, index_spec, mapping_dir):
...
+    def _create_component_template(self, tpl_spec, mapping_dir):
+        name = self._r(tpl_spec, "name")
+        template_file = self._r(tpl_spec, "template")
+        template_file = os.path.join(mapping_dir, template_file)
+        idx_tmpl_src = TemplateSource(mapping_dir, template_file, self.source)
+        with self.source(template_file, "rt") as f:
+            idx_tmpl_src.load_template_from_string(f.read())
+            template_content = self._load_template(
+                idx_tmpl_src.assembled_source,
+                f"definition for index template {name} in {template_file}")
```
danielmitterdorfer, line 1058: "definition for **component** template ..."?

esrally/track/params.py
```diff
@@ -368,6 +369,86 @@ def params(self):
...
+                dct[k] = merge_dct[k]
+
+    def params(self):
+        return {
+            "templates": self.template_definitions,
+            "request-params": self.request_params
+        }
+
+
+class CreateComponentTemplateParamSource(CreateComposableTemplateParamSource):
```
danielmitterdorfer, line 448: How about we use an abstract base class and derive two sub classes from it? I think such a class hierarchy would be more consistent.

esrally/track/params.py
danielmitterdorfer: Ok for me.

danielmitterdorfer (commented):
I did an initial pass.

esrally/driver/runner.py
gingerwizard: Agreed lets be consistent.

esrally/driver/runner.py
gingerwizard: Exactly its not currently supported. I'll import locally as well and add a comment.

esrally/driver/runner.py
gingerwizard: see below. Moving to local import.

esrally/track/track.py
gingerwizard: So whilst i agree we don't seem to, `IndexTemplate` is being used for Composable templates as well - so when we remove support for the deprecated version the class `IndexTemplate` will remain. Consider also how we use the old param source https://github.com/elastic/rally/pull/1099/files/42de57f76c6f103b2b3d20c298684ba76f521bd3#r512749427.

We can still decouple if we're generally against use of polymorphism in rally.

esrally/track/params.py
gingerwizard: I agree this is more consistent with the actual concept

gingerwizard:
@danielmitterdorfer resolved all issues raised except use of polymorphism and `IndexTemplate` - see comment.

esrally/track/track.py
gingerwizard: Maybe we create a new class `ComposableTemplate`. Both this and `ComponentTemplate` can inherit from `IndexTemplate` for now and later we move `IndexTemplate` to be a ABC? Alternatively we create these two new classes and define in full, leaving a note for them to exploit an ABC once the name `IndexTemplate` is available?

gingerwizard:
discussed with @danielmitterdorfer we will add integration tests to this PR and raise a sibling PR with docs for data streams + composable/component templates.

esrally/track/track.py
danielmitterdorfer: I did not imply to be against polymorphism, quite the contrary: We should rely on inheritance if we have polymorphic code. However, I was arguing against using inheritance "just" to save code. The current implementation of `ComponentTemplate` calls the `IndexTemplate` constructor as follows:

```python
super().__init__(name, None, content, delete_matching_indices=False)
```

By specifying less constructor parameters than the base class (the second parameter `pattern` is set to `None`) we indicate that this property is not needed for the subclass, which violates the contract of the base class: every index template requires a matching index pattern. But now that we derive component templates from index templates, which don't require an index pattern, we violate this invariant. Thus, `ComponentTemplate` cannot be a subclass of `IndexTemplate`.

danielmitterdorfer (commented):
Looks quite good already! I left an additional [comment on the topic of inheritance from `IndexTemplate`](https://github.com/elastic/rally/pull/1099#discussion_r514227439)

esrally/track/track.py
gingerwizard: Ok i now see your point. I'll pull ComponentTemplate into a separate cluster. When legacy Index template support is removed we should move `IndexTemplate` to an abstract base class with an appropriate contract for both `ComponentTemplate` and `ComposableTemplate`. I'll add a note accordingly.

gingerwizard:
Docs have been pushed for both composable templates and data streams. Ready for review.

Discussed IT with @danielmitterdorfer. We agreed that we will move the observability track to use these new features and test there. Once the track is moved to the rally tracks repo, IT tests will be moved here.

Some other notes on the docs:

1. When defining legacy index templates in the track, the current docs mention 

> delete-matching-indices (optional, defaults to true): Delete all indices that match the provided index pattern before start of the benchmark.

I see no logic which does this. The user is required to specify a `delete-index-templates` operation. The docs for the new composable templates reflect this.

2. Currently, we don't allow the user to specify `indices`  and `data-streams`  in the track. However, we don't error if a user doesn't specify `indices` or `data-streams` but refers to the them in the corpora and uses both i.e. with `target-data-stream` or `target-index`. They could therefore use a mixture. We also don't require these values exist in the `indices`/`data-streams`. Should we warn or validate more strongly here @danielmitterdorfer @danielmitterdorfer ?

3. Currently we require users to use the `component-templates` and `composable-templates` in the right order i.e. a user must create the former first or they will have issues. I've documented this but we could validate in future.

docs/track.rst
```diff
@@ -194,8 +201,8 @@ Example::
         }
     ]
 
-templates
-.........
+templates (deprecated)
```
danielmitterdorfer, line 204: In which context are they deprecated? Do you mean in Elasticsearch or in Rally? If it's the former, I don't think we should document so explicitly what is deprecated in a certain Elasticsearch version given the wide range of major versions.

docs/track.rst
```diff
@@ -215,6 +222,65 @@ Example::
         }
     ]
 
+data-streams
+............
+
+The ``data-streams`` section contains a list of all data streams that are used by this track.   Cannot be used if the ``indices`` section is specified.
```
danielmitterdorfer, line 228: Nit: Two spaces after the first sentence instead of one.

docs/track.rst
```diff
@@ -215,6 +222,65 @@ Example::
...
+
+* ``name`` (mandatory): Component template name
+* ``template`` (mandatory): Component template file name
+
+Example::
+
+    "component-templates": [
+        {
+            "name": "my-default-component-template",
+            "index-pattern": "my-index-*"
```
danielmitterdorfer, line 280: The `index-pattern` property is not defined above. Did you mean to specify the `template` property here?

docs/track.rst
```diff
@@ -236,9 +302,15 @@ Each entry in the ``documents`` list consists of the following properties:
 * ``document-count`` (mandatory): Number of documents in the source file. This number is used by Rally to determine which client indexes which part of the document corpus (each of the N clients gets one N-th of the document corpus). If you are using parent-child, specify the number of parent documents.
 * ``compressed-bytes`` (optional but recommended): The size in bytes of the compressed source file. This number is used to show users how much data will be downloaded by Rally and also to check whether the download is complete.
 * ``uncompressed-bytes`` (optional but recommended): The size in bytes of the source file after decompression. This number is used by Rally to show users how much disk space the decompressed file will need and to check that the whole file could be decompressed successfully.
+<<<<<<< HEAD
```
danielmitterdorfer: Merge issue?

docs/track.rst
```diff
@@ -236,9 +302,15 @@ Each entry in the ``documents`` list consists of the following properties:
...
+* ``target-index``: Defines the name of the index which should be targeted for bulk operations.  Rally will automatically derive this value if you have defined exactly one index in the ``indices`` section. Ignored if ``includes-action-and-meta-data`` is ``true``.
+* ``target-type`` (optional): Defines the name of the document type which should be targeted for bulk operations. Rally will automatically derive this value if you have defined exactly one index in the ``indices`` section and this index has exactly one type. Ignored if ``includes-action-and-meta-data`` is ``true`` or if a ``target-data-stream`` is specified. Types have been removed in Elasticsearch 7.0.0 so you must not specify this property if you want to benchmark Elasticsearch 7.0.0 or later.
+* ``target-data-stream``: Defines the name of the data stream which should be targeted for bulk operations.  Rally will automatically derive this value if you have defined exactly one index in the ``data-streams`` section. Ignored if ``includes-action-and-meta-data`` is ``true``.
+=======
```
danielmitterdorfer: Merge issue?

docs/track.rst
```diff
@@ -236,9 +302,15 @@ Each entry in the ``documents`` list consists of the following properties:
...
 * ``target-index``: Defines the name of the index which should be targeted for bulk operations. Rally will automatically derive this value if you have defined exactly one index in the ``indices`` section. Ignored if ``includes-action-and-meta-data`` is ``true``.
 * ``target-type`` (optional): Defines the name of the document type which should be targeted for bulk operations. Rally will automatically derive this value if you have defined exactly one index in the ``indices`` section and this index has exactly one type. Ignored if ``includes-action-and-meta-data`` is ``true``. Types have been removed in Elasticsearch 7.0.0 so you must not specify this property if you want to benchmark Elasticsearch 7.0.0 or later.
 * ``meta`` (optional): A mapping of arbitrary key-value pairs with additional meta-data for a source file.
+>>>>>>> ed7aaad49bec980612a7a1ec7f9c9c50e447b674
```
danielmitterdorfer: Merge issue?

docs/track.rst
```diff
@@ -269,6 +342,22 @@ Here we define a single document corpus with one set of documents::
         }
       ]
 
+Here we define a single document corpora with one set of documents using data streams instead of indices::
```
danielmitterdorfer, line 342: "single document corpora" -> "corpus"?

docs/track.rst
```diff
@@ -835,22 +924,210 @@ This is an administrative operation. Metrics are not reported by default. Report
...
+
+
+.. note::
+    If your component template is used in composable templates ensure these are created after using the ``create-composable-template`` operation.
+
+This is an administrative operation. Metrics are not reported by default. Reporting can be forced by setting ``include-in-reporting`` to ``true``.
+
+This operation is :ref:`retryable <track_operations>`.
+
+create-index-template (deprecated)
```
danielmitterdorfer, line 1112: Remove "deprecated" here (see my reasoning above)?

docs/track.rst
```diff
@@ -835,22 +924,210 @@ This is an administrative operation. Metrics are not reported by default. Report
...
+              "enabled": false
+            }
+          }
+        }
+      }
+    }
+
+
+.. note::
+    If your component template is used in composable templates ensure these are created after using the ``create-composable-template`` operation.
```
danielmitterdorfer, line 1106: Did you mean: "these are created after using the ``create-comonent-template`` operation."?

esrally/track/params.py
```diff
@@ -16,6 +16,7 @@
 # under the License.
 
 import collections
+import inspect
```
danielmitterdorfer, line 19: Is this in here due to a merge?

danielmitterdorfer (commented):
I did a pass over the docs and found a few minor things but I think we're close.

danielmitterdorfer:
Regarding your notes:


>>    delete-matching-indices (optional, defaults to true): Delete all indices that match the provided index pattern before start of the benchmark.
>
> I see no logic which does this. The user is required to specify a delete-index-templates operation. The docs for the new composable templates reflect this.

I think this might be a leftover from the times when Rally managed indices more implicitly (without the user defining tasks specifically).

>   Currently, we don't allow the user to specify indices and data-streams in the track. However, we don't error if a user doesn't specify indices or data-streams but refers to the them in the corpora and uses both i.e. with target-data-stream or target-index. They could therefore use a mixture. We also don't require these values exist in the indices/data-streams. Should we warn or validate more strongly here @danielmitterdorfer @danielmitterdorfer ?

We could consider this as an improvement in the future but IMHO that's not a blocker.

> Currently we require users to use the component-templates and composable-templates in the right order i.e. a user must create the former first or they will have issues. I've documented this but we could validate in future.

Also this one might be nice to have in the future but not a blocker here.

docs/track.rst
gingerwizard: yes i was thinking we might remove support but given our discussions re ES support for 2 majors +1.

docs/track.rst
gingerwizard: 3 even!

docs/track.rst
gingerwizard: good catch, thanks

docs/track.rst
gingerwizard: no, but clearer would be `If your component template is used in composable templates, ensure these are created after specifying this operation using the ``create-composable-template`` operation.`

esrally/track/params.py
gingerwizard: I dont think so, i think its because of collections being added above. Its pushed `inspect` down.

danielmitterdorfer (approved):
Thanks for adding this! LGTM

PR #1103, opened 2020-11-03
Accept some differing units for throttling

danielmitterdorfer:
With #1100 we introduced more flexible throttling which requires that
the unit in which requests are throttled and the unit in which they are
reported, are aligned. This causes issues at the moment with scroll
requests which are throttled in ops/s but reported in pages/s. Similarly we have
also throttled bulk requests based on ops/s sometimes but their throughput is
reported in docs/s.

With this commit we correct the unit mismatch for this special case in
order to stay backwards-compatible with older versions of Rally. We also
ensure that throttling is preserved in test mode so our integration
tests can spot such issues in the future.

Relates #1100

### DIFF
--- a/esrally/driver/scheduler.py
+++ b/esrally/driver/scheduler.py
@@ -205,6 +205,9 @@ class Unthrottled(Scheduler):
     def next(self, current):
         return 0
 
+    def __str__(self):
+        return "unthrottled"
+
 
 class DeterministicScheduler(SimpleScheduler):
     """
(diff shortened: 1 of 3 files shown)


dliappis (approved):
LGTM.

I think we should change (in master) the corresponding scroll queries in geonames and pmc e.g. https://github.com/elastic/rally-tracks/blob/ea52c9f601f1b5fbe485941b081d2bba9e445c20/geonames/challenges/default.json#L104-L105 to have throughput specified as "20 pages/s".

danielmitterdorfer:
> I think we should change (in master) the corresponding scroll queries in geonames and pmc e.g. https://github.com/elastic/rally-tracks/blob/ea52c9f601f1b5fbe485941b081d2bba9e445c20/geonames/challenges/default.json#L104-L105 to have throughput specified as "20 pages/s".

I agree; what we have here is only temporary for backwards-compatibility. After we've merged this, I'll raise a corresponding issue in rally-tracks. I think we could do the change in rally-tracks after we've released 2.1.0 so there is a long enough grace period for people to upgrade.

danielmitterdorfer:
I've created cleanup issues for the affected tracks:

* https://github.com/elastic/rally-tracks/issues/139
* https://github.com/elastic/rally-eventdata-track/issues/90

danielmitterdorfer:

PR #1105, opened 2020-11-03
Drop support for Elasticsearch 2.x

### LINKED ISSUE #822: Drop support for Elasticsearch 2.x
Elasticsearch [2.x went EOL on 2018-02-28](https://www.elastic.co/support/eol). According our own [version policy in Rally](https://esrally.readthedocs.io/en/stable/versions.html) we keep support in Rally for two more years.

This ticket is meant as a place to collect what needs to be changed to drop support for 2.x.

* [x] Adapt our [version policy page](https://esrally.readthedocs.io/en/stable/versions.html) to indicate we've dropped support for Elasticsearch 2.x.
* [x] Remove support for environment variables in car configurations.
 
Similar to #715, we need to have a grace period in rally-teams and rally-tracks before we drop support there.

Marking as blocked because we should not work on this before 2020-02-28.

danielmitterdorfer:
With this commit we drop support in Rally to benchmark Elasticsearch
2.x. This version is end-of-life since February 28, 2018.

Closes #822

### DIFF
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -266,7 +266,7 @@ You can also show details about your track with ``esrally info --track-path=~/ra
     5. force-merge
     6. query-match-all (8 clients)
 
-Congratulations, you have created your first track! You can test it with ``esrally --distribution-version=6.0.0 --track-path=~/rally-tracks/tutorial``.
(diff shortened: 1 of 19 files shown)


dliappis (approved):
LGTM thanks for this cleanup

PR #1106, opened 2020-11-04
Make template merging Python 3.9 compatible

danielmitterdorfer:
With this commit we change a deprecated import of `collections.Mapping`
to `collections.abc.Mapping` which would not work with Python 3.9
anymore. We also correct the class name in the referenced internal merge
method.

Relates #1099

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -446,15 +446,15 @@ class CreateTemplateParamSource(ABC, ParamSource):
                 if sub_path not in content:
                     content[sub_path] = {}
                 content = content[sub_path]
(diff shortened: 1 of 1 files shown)


gingerwizard (approved):
LGTM. I'm surprised the original tests past i.e. the static method is inherited. @hub-cap this might be why your new lint rule failed in https://github.com/elastic/rally/pull/1083

danielmitterdorfer:
Thanks for the review. I think even with the wrong reference it is still able to resolve the method but it's at least confusing and that's why I cleaned it up here.

PR #1108, opened 2020-11-04
Add support for FOSSA

danielmitterdorfer:
With this commit we add support in Rally to run license checks on third-party
dependencies with a separate tool called `fossa`. Note that this work is only a
preparation on the Rally side. This is intended to be run in CI and needs to be
supplemented by a corresponding CI job.


hub-cap:
I am going to spin up a test machine to ensure this works properly

@elasticmachine run tests

hub-cap (approved):
This looks good. It complained about an environment variable but I dont think that needs to be solved in this PR.

My only question is ive seen some projects make the `project` field the entire github URL, whereas some do not. I dont know enought about fossa to know what is correct, and I assume you do :)

hub-cap:
Not sure how I could have altered the tests by being on the machine but Im going to rerun them just in case on a pristine machine that I do not get on.

@elasticmachine run tests

danielmitterdorfer:
Thanks for the review!

> This looks good. It complained about an environment variable but I dont think that needs to be solved in this PR.

I checked the logs from the CI run as well and I think this is because I've created this PR before #1105 has been merged to master. I've merged master now and retriggered a CI run.

> My only question is ive seen some projects make the project field the entire github URL, whereas some do not. I dont know enought about fossa to know what is correct, and I assume you do :)

The Github URL is autopopulated when you create a fossa configuration on the command line (with `fossa init`) and this property is used in the fossa UI to identify the project (see also their [docs](https://github.com/fossas/fossa-cli/blob/master/docs/config-file.md/#project-optional)). I've explicitly changed this so it shows the nicer name "Rally" instead of the somewhat clunky URL. I hope that answers your question.

PR #1109, opened 2020-11-04
Throttle tasks based on presence of a scheduler

danielmitterdorfer:
Previously Rally has considered a task only throttled if it specified a
target throughput. This assumes that there is a meaningful target
throughput to specify though. Consider a more complex case where we want
to simulate a traffic pattern that changes over the course of a
(simulated) day. In that case there is no single "right" target
throughput that can be specified.

With this commit we consider tasks also throttled if no target
throughput is specified but the user has supplied a scheduler
explicitly. We assume that the scheduler will use different parameters
specified on the task to determine its behavior.

### DIFF
--- a/esrally/driver/scheduler.py
+++ b/esrally/driver/scheduler.py
@@ -114,7 +114,7 @@ def is_legacy_scheduler(scheduler_class):
     target throughput.
     """
     constructor_params = inspect.signature(scheduler_class.__init__).parameters
-    return len(constructor_params) == 2 and "params" in constructor_params
+    return len(constructor_params) >= 2 and "params" in constructor_params
 
 
 def is_simple_scheduler(scheduler_class):
--- a/esrally/driver/scheduler.py
+++ b/esrally/driver/scheduler.py
@@ -196,7 +196,7 @@ class LegacyWrappingScheduler(Scheduler):
         self.legacy_scheduler = legacy_scheduler_class(task.params)
 
     def next(self, current):
-        return self.legacy_scheduler(current)
+        return self.legacy_scheduler.next(current)
 
 
 class Unthrottled(Scheduler):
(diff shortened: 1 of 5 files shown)


gingerwizard:
I believe we're going to have other issues here. In the `UnitAwareScheduler` we delegate to the underlying scheduler but don't consider which API is being used - specifically line 288

`self.scheduler = self.scheduler_class(self.task, target_throughput)`

danielmitterdorfer:
> In the UnitAwareScheduler we delegate to the underlying scheduler but don't consider which API is being used

It is an precondition of the `UnitAwareScheduler` that only "simple" scheduler classes are passed in its constructor and this is checked in:

https://github.com/elastic/rally/blob/8961770558e70d7b6aa61327d37d81078da48dce/esrally/driver/scheduler.py#L102-L104

gingerwizard (commented):
@danielmitterdorfer some test failures here. Change in principle makes sense once the int tests are fixed.

danielmitterdorfer:
> @danielmitterdorfer some test failures here. Change in principle makes sense once the int tests are fixed.

The test failure has been caused by a misclassification of the `UtilizationScheduler` class. Previously a scheduler was classified as "legacy scheduler" iff  it had exactly two constructor parameters (`self` and `params`) and one of them was called `params`. However, the [constructor signature of `UtilizationScheduler`](https://github.com/elastic/rally-eventdata-track/blob/3cdffcafc1b4ebdae5b82e2a8bd30de015a014c8/eventdata/schedulers/utilization_scheduler.py#L33) is:

```python
def __init__(self, params, perf_counter=time.perf_counter):
```

The additional parameter is needed for unit tests but prevented classification as a legacy scheduler. I have loosened the check now in https://github.com/elastic/rally/pull/1109/commits/9b72a3c17e443c3f4b05e0d3017884dc33cd7d9f and fixed a follow-up issue uncovered by the tests in https://github.com/elastic/rally/pull/1109/commits/c8c70b25e032999a74b4a6988b582bf8499c4f78. IMHO it's ok to be a bit lenient here because we should remove support for legacy schedulers (along with this check) with Rally 2.1.0.

gingerwizard (approved):
LGTM

PR #1113, opened 2020-11-18
Avoid assignment from None

danielmitterdorfer:
With this commit we reenable the pylint warning `assignment-from-none`
which checks whether a `None` return value from a function is assigned
to a variable. We also audited all such issues in the code and disabled
the warning accordingly.

Relates #838


hub-cap (approved):
LGTM

PR #1115, opened 2020-11-18
Error wrapper script on unclean local state

### LINKED ISSUE #911: Make run.sh update emit an error when git update fails
When invoking rally/rallyd with `./rally` or `./rallyd` by default it will attempt to update from source.

However, it's very meticulous due to the use of `git status --porcelain` and if there is a rogue file in the repo it will just skip the update in https://github.com/elastic/rally/blob/5b40bd5529ddfd4980b8efe8ca37b74897fce748/run.sh#L104-L105.

Let's actually emit an error [in the else section](https://github.com/elastic/rally/blob/5b40bd5529ddfd4980b8efe8ca37b74897fce748/run.sh#L104-L105) so that the situation is more user friendly.

Note: this might break the existing workflow for some users working with the dev version of Rally but I don't think it needs to be flagged as a breaking change.

danielmitterdorfer:
With this commit we exit the Rally wrapper script with an error if the
local working copy contains uncommitted files. This avoids skipping
automatic updates silently if uncommitted files are unintentionally
present. Users that still want to continue running can do so by
specifying `--skip-update`.

Closes #911

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -103,8 +103,9 @@ then
       #else
       # offline - skipping update
       fi
-    # else
-      # Uncommitted changes - don't upgrade, just run
+    else
+      >&2 echo "There are uncommitted changes. Please cleanup your working copy or specify --skip-update."
+      exit 1
     fi
 #else -> No self update
 fi


hub-cap (approved):
excellent!

PR #1112, opened 2020-11-17
Execute complex request hierarchies

danielmitterdorfer:
With this commit we add the ability to Rally to execute arbitrarily complex
request hierarchies in benchmarks. This is implemented with a new `composite`
operation type. It allows to nest multiple so-called streams which are executed
concurrently. At the moment, the request hierarchy needs to be specified inline
in the track. It is recommended to store more complex request hierarchies in a
separate file and include it with Jinja's `include` directive.

Apart from this change, we also align the internal mapping of operation-type
names to `OperationType` enumeration constants. This is mainly done to simplify
usage and implementation of the `Composite` runner as it can directly use the
operation-types that are specified in the track rather than converting them
internally.

PR #1116, opened 2020-11-19
Ensure admin tasks are always executed

danielmitterdorfer:
When a user specified `--exclude-tasks`, Rally did not only exclude the
specified tasks but also all administrative tasks. This happened because
of a broken logic that always added an admin task filter internally. Due
to another bug, administrative tasks where never correctly detected but
we fixed that broken logic in #1112, which finally uncovered this issue.

With this commit we ensure now that administrative tasks are always
executed regardless whether the user excludes them. We also increase
test coverage in this part of the code.

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -693,27 +693,31 @@ def filter_tasks(t, filters, exclude=False):
 
     logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


dliappis (approved):
LGTM

PR #1117, opened 2020-11-19
Don't change scheduler on failed requests

danielmitterdorfer:
With this commit we check whether a request has a non-zero weight before
attempting to calculate a target throughput. This ensures that
throughput is only calculated / adjusted if the request has returned
successfully and we're able to derive a throughput.


dliappis (approved):
LGTM thanks!

PR #1118, opened 2020-11-19
Allow index pattern for source-index in shrink-index operation

dliappis:
This commit adds support for specifying several source indices using the
standard Elasticsearch Multi-target syntax for the shrink-index
runner.

This helps automate shrink operations for multiple indices.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1303,7 +1303,7 @@ shrink-index
 With the operation ``shrink-index`` you can execute the `shrink index API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-shrink-index.html>`_. Note that this does not correspond directly to the shrink index API call in Elasticsearch but it is a high-level operation that executes all the necessary low-level operations under the hood to shrink an index. It supports the following parameters:
 
 * ``source-index`` (mandatory): The name of the index that should be shrinked. Multiple indices can be defined using the `Multi-target syntax <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_.
-* ``target-index`` (mandatory): The name of the index that contains the shrinked shards. If multiple indices match ``source-index``, the common prefix will appended to ``target-index``.
+* ``target-index`` (mandatory): The name of the index that contains the shrinked shards. If multiple indices match ``source-index``, one shrink operation will execute for every matching index. Each shrink operation will use a modified ``target-index``: the unique suffix of the source index (derived by removing the common prefix of all matching source indices) will be appended to ``target-index``. See also the example below.
 * ``target-body`` (mandatory): The body containing settings and aliases for ``target-index``.
 * ``shrink-node`` (optional, defaults to a random data node): As a first step, the source index needs to be fully relocated to a single node. Rally will automatically choose a random data node in the cluster but you can choose one explicitly if needed.
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1302,8 +1302,8 @@ shrink-index
 
 With the operation ``shrink-index`` you can execute the `shrink index API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-shrink-index.html>`_. Note that this does not correspond directly to the shrink index API call in Elasticsearch but it is a high-level operation that executes all the necessary low-level operations under the hood to shrink an index. It supports the following parameters:
 
-* ``source-index`` (mandatory): The name of the index that should be shrinked.
-* ``target-index`` (mandatory): The name of the index that contains the shrinked shards.
+* ``source-index`` (mandatory): The name of the index that should be shrinked. Multiple indices can be defined using the `Multi-target syntax <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_.
+* ``target-index`` (mandatory): The name of the index that contains the shrinked shards. If multiple indices match ``source-index``, the common prefix will appended to ``target-index``.
 * ``target-body`` (mandatory): The body containing settings and aliases for ``target-index``.
 * ``shrink-node`` (optional, defaults to a random data node): As a first step, the source index needs to be fully relocated to a single node. Rally will automatically choose a random data node in the cluster but you can choose one explicitly if needed.
 
@@ -1324,6 +1324,24 @@ Example::
     }
 
 This will shrink the index ``src`` to ``target``. The target index will consist of one shard and have one replica. With ``shrink-node`` we also explicitly specify the name of the node where we want the source index to be relocated to.
+
+The following example ``src*`` matches a list of indices ``src-a,src-b``::
+
+    {
+      "operation-type": "shrink-index",
+      "shrink-node": "rally-node-0",
+      "source-index": "src*",
+      "target-index": "target",
+      "target-body": {
+        "settings": {
+          "index.number_of_replicas": 1,
+          "index.number_of_shards": 1,
+          "index.codec": "best_compression"
+        }
+      }
+    }
+
+and will reindex ``src-a`` as ``target-a`` and ``src-b`` as ``target-b``.
 
 This operation is :ref:`retryable <track_operations>`.
 
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -23,6 +23,8 @@ import sys
 import types
 from collections import Counter, OrderedDict
 from copy import deepcopy
+from os.path import commonprefix
+
 import ijson
 
 from esrally import exceptions, track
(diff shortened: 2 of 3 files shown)


esrally/driver/runner.py
```diff
@@ -1320,46 +1329,56 @@ async def _wait_for(self, es, idx, description):
...
+                target_body["settings"] = {}
+            target_body["settings"]["index.routing.allocation.require._name"] = None
+            target_body["settings"]["index.blocks.write"] = None
+            # kick off the shrink operation
+            index_suffix = remove_prefix(source_index, source_indices_stem)
+            final_target_index = target_index if len(index_suffix) == 0 else target_index+index_suffix
+            await es.indices.shrink(index=source_index, target=final_target_index, body=target_body)
+
+            self.logger.info("Waiting for shrink to finish for index [%s] ...", source_index)
+            await self._wait_for(es, final_target_index, "shrink for index [{}]".format(final_target_index))
```
dliappis, line 1380: We could run things in parallel by moving this to a separate loop later, however, I am not sure whether a high level of concurrent shrink operations will actually be beneficial or cause contention and provide minimal advantage (esp. if only one shrink node has been provided).

esrally/driver/runner.py
```diff
@@ -1320,46 +1329,56 @@ async def _wait_for(self, es, idx, description):
...
+        source_indices = list(source_indices_get.keys())
+        source_indices_stem = commonprefix(source_indices)
+
         target_index = mandatory(params, "target-index", self)
+
         # we need to inject additional settings so we better copy the body
         target_body = deepcopy(mandatory(params, "target-body", self))
         shrink_node = params.get("shrink-node")
         # Choose a random data node if none is specified
+        node_names = [shrink_node] if shrink_node else []
```
danielmitterdorfer, line 1342: I think this makes the code a bit hard to read as you operate on two variables (`node_names` and `shrink_node`) in several branches and check `shrink_node` multiple times.

How about this instead?

```python
        # Choose a random data node if none is specified
        if shrink_node:
            node_names = [shrink_node]
        else:
            node_names = []
            # choose a random data node
            node_info = await es.nodes.info()
            for node in node_info["nodes"].values():
                if "data" in node["roles"]:
                    node_names.append(node["name"])
            if not node_names:
                raise exceptions.RallyAssertionError("Could not choose a suitable shrink-node automatically. Please specify it explicitly.")
```

This makes it easier to reason because there is only one top-level conditional and then we populate `node_names` in each branch separately.

esrally/driver/runner.py
```diff
@@ -23,6 +23,8 @@
 import types
 from collections import Counter, OrderedDict
 from copy import deepcopy
+from os.path import commonprefix
```
danielmitterdorfer, line 26: While I see that it's useful, I find it a bit odd to use a path-related library method for string processing. Although it's not ideal I'm ok if we keep it here.

esrally/driver/runner.py
```diff
@@ -301,8 +303,15 @@ def mandatory(params, key, op):
...
         return params[key]
     except KeyError:
-        raise exceptions.DataError("Parameter source for operation '%s' did not provide the mandatory parameter '%s'. Please add it to your"
-                                   " parameter source." % (str(op), key))
+        raise exceptions.DataError(
+            f"Parameter source for operation '{str(op)}' did not provide the mandatory parameter '{key}'. "
+            f"Add it to your parameter source and try again.")
+
+
+def remove_prefix(string, prefix):
```
danielmitterdorfer, line 311: Too bad we don't have Python 3.9 as minimum version requirement as this has been recently added to the [standard library](https://docs.python.org/3/library/stdtypes.html#str.removeprefix). I wonder whether we should put a TODO into the code?

esrally/driver/runner.py
```diff
@@ -1320,46 +1329,56 @@ async def _wait_for(self, es, idx, description):
...
+                                          body={
+                                              "settings": {
+                                                  "index.routing.allocation.require._name": shrink_node,
+                                                  "index.blocks.write": "true"
+                                              }
+                                          },
+                                          preserve_existing=True)
+
+            self.logger.info("Waiting for relocation to finish for index [%s] ...", source_index)
+            await self._wait_for(es, source_index, "shard relocation for index [{}]".format(source_index))
```
danielmitterdorfer, line 1368: We could use an f-string here?

esrally/driver/runner.py
danielmitterdorfer, line 1380: We could use an f-string here?

danielmitterdorfer: Agreed, I'd not complicate things unless there is a strong benefit.

esrally/driver/runner.py
```diff
@@ -1320,46 +1329,56 @@ async def _wait_for(self, es, idx, description):
...
+            # kick off the shrink operation
+            index_suffix = remove_prefix(source_index, source_indices_stem)
+            final_target_index = target_index if len(index_suffix) == 0 else target_index+index_suffix
+            await es.indices.shrink(index=source_index, target=final_target_index, body=target_body)
+
+            self.logger.info("Waiting for shrink to finish for index [%s] ...", source_index)
+            await self._wait_for(es, final_target_index, "shrink for index [{}]".format(final_target_index))
+            self.logger.info("Shrinking [%s] to [%s] has finished.", source_index, final_target_index)
         # ops_count is not really important for this operation...
         return 1, "ops"
```
danielmitterdorfer, line 1383: I think we could use the number of shrunk indices here for the operation count?

docs/track.rst
```diff
@@ -1302,8 +1302,8 @@ shrink-index
 
 With the operation ``shrink-index`` you can execute the `shrink index API <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-shrink-index.html>`_. Note that this does not correspond directly to the shrink index API call in Elasticsearch but it is a high-level operation that executes all the necessary low-level operations under the hood to shrink an index. It supports the following parameters:
 
-* ``source-index`` (mandatory): The name of the index that should be shrinked.
-* ``target-index`` (mandatory): The name of the index that contains the shrinked shards.
+* ``source-index`` (mandatory): The name of the index that should be shrinked. Multiple indices can be defined using the `Multi-target syntax <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_.
+* ``target-index`` (mandatory): The name of the index that contains the shrinked shards. If multiple indices match ``source-index``, the common prefix will appended to ``target-index``.
```
danielmitterdorfer, line 1306: Isn't it the other way around? The common prefix will be removed and instead the unique parts of each source index name will be appended to the string specified via `target-index`?

danielmitterdorfer (commented):
Nice convenience improvement. Looks fine for the most part but I left a few comments.

docs/track.rst
dliappis: Of course :facepalm: ! Addressed in a3cab0e5a5470d7dcdc1964d6c48c8d0a32969ab

esrally/driver/runner.py
dliappis: Yes I noticed that too while coding writing this. Added a TODO in a3cab0e5a5470d7dcdc1964d6c48c8d0a32969ab.

esrally/driver/runner.py
dliappis: Addressed in a3cab0e5a5470d7dcdc1964d6c48c8d0a32969ab (also removed please following https://developers.google.com/style/tone#politeness-and-use-of-please)

esrally/driver/runner.py
dliappis: Done in 52f28a6e3729b9a6f9006bcb56099a5e0efc9836

esrally/driver/runner.py
dliappis: Done in a3cab0e5a5470d7dcdc1964d6c48c8d0a32969ab

esrally/driver/runner.py
dliappis: Done in a3cab0e5a5470d7dcdc1964d6c48c8d0a32969ab

dliappis:
@danielmitterdorfer Thanks for the review! I addressed your comments and requested a second review.

docs/track.rst
```diff
@@ -1303,7 +1303,7 @@ shrink-index
...
 
 * ``source-index`` (mandatory): The name of the index that should be shrinked. Multiple indices can be defined using the `Multi-target syntax <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_.
-* ``target-index`` (mandatory): The name of the index that contains the shrinked shards. If multiple indices match ``source-index``, the common prefix will appended to ``target-index``.
+* ``target-index`` (mandatory): The name of the index that contains the shrinked shards. If multiple indices match ``source-index``, one shrink operation will execute for every matching index. Each shrink operation will use a modified ``target-index``: the unique suffix of the source index (derived by removing the common prefix of all matching source indices) will be appended to ``target-index``. See also the example below.
```
danielmitterdorfer, line 1306: Much clearer, thanks!

danielmitterdorfer (approved):
Thanks! LGTM

PR #1121, opened 2020-11-23
Ensure that sleep tasks are timed

danielmitterdorfer:
With this commit we ensure that also the `sleep` task is included in
time measurements. Regular runners in Rally issue requests against
Elasticsearch and time measurement is done via callbacks in the request
lifecycle but the `sleep` runner is different because it only waits for
a certain time without any calls.


dliappis (approved):
LGTM

PR #1122, opened 2020-11-23
Remove implicit request timeout from force merge operation in polling

dliappis:
Currently in polling mode, the force merge operation uses a very short
timeout (1s) which, in case of slow response times, might actually
cause the operation to not get executed.

Since #1070 we can specify per operation request-timeout and we should
honor that one instead.


danielmitterdorfer (approved):
LGTM

PR #1124, opened 2020-11-24
Treat some zero target-intervals as unthrottled

### LINKED ISSUE #1120: UnitAwareScheduler does not support target_interval
I ran into an issue using a track w/ `target-interval` instead of `target-throughput` and it was bombing out [here](https://github.com/elastic/rally/blob/master/esrally/driver/scheduler.py#L268).

danielmitterdorfer:
With this commit we adjust the predicate that considers a task as
unthrottled. Previously a task was unthrottled if and only if its target
throughput property was not defined and there was also not schedule set.
This lead to problems when a target-interval of zero was set on task
with an explicitly defined schedule (such as poisson). Now we treat
tasks as unthrottled if and only if the target throughput property is
not defined and the schedule is either unspecified or one of the
built-in schedules. This ensures that a target-interval of zero is
recognized as unthrottled for all built-in schedulers.

Closes #1120

### DIFF
--- a/tests/track/track_test.py
+++ b/tests/track/track_test.py
@@ -260,33 +260,29 @@ class TaskFilterTests(TestCase):
 
 class TaskTests(TestCase):
     def task(self, schedule=None, target_throughput=None, target_interval=None):
-        op = track.Operation("bulk-index", track.OperationType.Bulk.name)
+        op = track.Operation("bulk-index", track.OperationType.Bulk.to_hyphenated_string())
         params = {}
-        if target_throughput:
+        if target_throughput is not None:
             params["target-throughput"] = target_throughput
-        if target_interval:
+        if target_interval is not None:
             params["target-interval"] = target_interval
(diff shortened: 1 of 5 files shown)


tests/track/track_test.py
```diff
@@ -260,33 +260,29 @@ def test_task_op_type_filter(self):
...
     def test_unthrottled_task(self):
         task = self.task()
         self.assertIsNone(task.target_throughput)
-        self.assertFalse(task.throttled)
 
-    def test_task_with_scheduler_is_throttled(self):
-        task = self.task(schedule="daily-traffic-pattern")
+    def test_target_interval_zero_treated_as_unthrottled(self):
+        task = self.task(target_interval=0)
         self.assertIsNone(task.target_throughput)
```
hub-cap, line 277: yay!

hub-cap (approved):
Thank you for fixing this bug!

PR #1126, opened 2020-11-26
Respect specified order in composite task

danielmitterdorfer:
With this commit we ensure that operations that are intertwined with
streams don't get executed before streams but rather respect the order
in which they have been specified in the track.

Relates #1112


gingerwizard (approved):
Thanks a nice improvement and useful in practical use cases

PR #1127, opened 2020-11-30
Avoid too many public methods

danielmitterdorfer:
With this commit we re-enabled the pylint check `R0904` which checks for
too many public methods. We have eliminated the warnings by using
different strategies:

* We have merged too fine-grained methods into more coarse-grained ones.
* We have eliminated some methods that have been called once or never.
* We have properly marked protected methods.
* Where appropriate we have disabled the warning (in tests).

Relates #838


dliappis (approved):
LGTM

PR #1129, opened 2020-12-02
Support async search API

### LINKED ISSUE #1114: Support async search in Rally
Currently our [search](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/async-search.html#submit-async-search) operation supports the standard synchronous way of issuing queries and is subject to client timeouts.

We'd like to also support issuing [async_search](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/async-search.html#submit-async-search) queries. This is a fundamental component for long running queries, such as the ones related to slower tiers like cold/frozen and features like searchable snapshots.

danielmitterdorfer:
With this commit we introduce support for Elasticsearch's `async_search`
API. We introduce three new runners for that:

* `submit-async-search`
* `get-async-search`
* `delete-async-search`

We restrict these runners to the scope of a `composite` operation
because we need to guarantee that they are executed by a single client
and the operations are able to share context (in that case the async
search id).

Closes #1114


esrally/driver/runner.py
```diff
@@ -1999,13 +2004,121 @@ def __repr__(self, *args, **kwargs):
...
+
+    def __repr__(self, *args, **kwargs):
+        return "submit-async-search"
+
+
+def async_search_ids(op_names):
+    if isinstance(op_names, str):
+        subjects = [op_names]
+    else:
+        subjects = op_names
```
dliappis, line 2027: or just one line `subjects = [op_names] if isinstance(op_names, str) else op_names`?

docs/track.rst
```diff
@@ -1780,6 +1780,112 @@ The following diagram depicts in which order requests are executed; the specific
...
+              {
+                "stream": [
+                  {
+                    "operation-type": "submit-async-search",
+                    "name": "search-1",
+                    "body": {
+                      "query": {
+                        "match_all": {}
+                      },
+                      "request-params": {
```
dliappis, line 1839: I think `request_params` is wrongly inside the `body` here, should be one level outside (as done below).

dliappis (changes requested):
Great job! Also tested it locally with a minimal track based on the example and worked fine.
Left two comments, main thing is to fix the level of `request_params` in the example.

esrally/driver/runner.py
danielmitterdorfer: Thanks, that looks nicer indeed. Changed in https://github.com/elastic/rally/pull/1129/commits/ed81c409ad72181111a5ffe86973ca5ec53ae8cf.

docs/track.rst
danielmitterdorfer: d'oh. Good catch! Addressed in https://github.com/elastic/rally/pull/1129/commits/5bfd464cd43d8384e4c311392e4a7093922fafdc

danielmitterdorfer:
@dliappis thanks for the review. I've addressed your comments. Can you please have another look?

dliappis (approved):
LGTM thanks!

PR #1131, opened 2020-12-03
Use top-level imports

danielmitterdorfer:
With this commit we re-enable the pylint warning `C0415` which checks
whether modules are imported outside of the toplevel. While we move most
of the imports to toplevel, we keep some imports more targeted, most
notable the `elasticsearch` module so we ensure that the library picks
up Rally's logging configuration properly even in a (multiprocess) actor
system.

Relates #838


danielmitterdorfer:
The CI failure looked spurious and also did not reproduce locally.

dliappis (approved):
LGTM

PR #1133, opened 2020-12-07
Clarify latency-vs-service-time FAQ

DaveCTurner:
I found the answer to the FAQ on latency vs service time unclear. It doesn't really define what `latency` means in throughput-throttled mode, and I'm not sure I fully understand the analogy it uses. I'm opening this to propose some alternative words, based on my interpretation of the previous answer, but I wouldn't be surprised if I'm still misunderstanding the details.


docs/faq.rst
```diff
@@ -25,15 +25,37 @@ First of all: Please (please, please) do NOT run Rally against your production c
...
+request and receiving the response.
+
+The explanation of ``latency`` depends on your choice of benchmarking mode:
+
+* **Throughput benchmarking mode**: In this mode, Rally will issue requests as
+  fast as it can: as soon as it receives a response to one request it will
+  generate and send the next request. In this mode ``latency`` and
+  ``service_time`` are identical.
+
+* **Throughput-throttled mode**: You may prefer to run benchmarks with a target
```
dliappis, line 48: Internally Rally uses a concept called "Schedule". There are different ways to alter the traffic that Rally generates by issuing requests:

- target-throughput (in requests/sec)
- target-interval (inverse of target-throughput i.e. amount of secs to wait per request)
- use poisson distribution or a custom schedule of your own

As all of those actually end up influencing the schedule that Rally will use to issue requests, I wonder if we should be talking about the schedule here briefly and adjust wording below, e.g. in line 51/52 instead of "according to this target" "according to the schedule".

However ... I think this concept is tricky by itself and by trying to provide a very generic answer (as I suggest) we might be confusing the reader, whereas the description in this PR is easier to understand by using a concrete example (via target throughput). @danielmitterdorfer WDYT?

dliappis (commented):
Thank you for this clarification! This topic pops up frequently and I think a more engineering/factual explanation wrt how Rally builds the request traffic actually helps.

I left a comment, mostly for discussion; after all I think, for the sake of keeping the explanation simple, we should just stick to target-throughput as done in this PR.

docs/faq.rst
DaveCTurner: Ack, thanks for the clarification. I'll try and generify the wording and maybe link to the schedule docs if I can work out how.

DaveCTurner: I pushed some more generic words in https://github.com/elastic/rally/pull/1133/commits/18ae7f27e4f191b153c184e74fdef5b7cdaf6a45 which I don't think make it any more complicated.

I tried to work out how to cross-reference the docs on the `schedule` bit of a track but timed out after a couple of minutes of trying to understand what on earth the [Sphinx docs](https://www.sphinx-doc.org/en/1.7/markup/inline.html) were trying to tell me. A cross-ref on the first occurrence of the word `schedule` would be very helpful IMO, could someone else push an update that adds one?

danielmitterdorfer: > could someone else push an update that adds one?

@DaveCTurner I've added one; it's available if you merge the latest master into your branch. You can use it as follows:

```
:ref:`your custom link text here <track_choose_schedule>`
```

DaveCTurner: Thanks @danielmitterdorfer, done in 4f999a72de81c2f703ed41b8ce9151d93a58a6fd.

danielmitterdorfer (approved):
Thanks for doing this! The wording sounds good me.

dliappis (approved):
LGTM thank you for this clarification!

PR #1136, opened 2020-12-16
Support non default repositories in chart generator

dliappis:
Honor "track-repository" property, along side the track property,
while generating charts.

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -1371,14 +1371,19 @@ class TimeSeriesCharts:
 
 
 class RaceConfigTrack:
-    def __init__(self, cfg, name=None):
-        self.cached_track = self.load_track(cfg, name)
+    def __init__(self, cfg, repository, name=None):
+        self.repository = repository
+        self.cached_track = self.load_track(cfg, name=name)
 
     def load_track(self, cfg, name=None, params=None):
         if not params:
             params = {}
+        # required in case a previous track using a different repository has specified the revision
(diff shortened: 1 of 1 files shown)


esrally/chart_generator.py
```diff
@@ -1371,14 +1371,19 @@ def index(environment, race_configs, title):
...
+    def __init__(self, cfg, repository, name=None):
+        self.repository = repository
+        self.cached_track = self.load_track(cfg, name=name)
 
     def load_track(self, cfg, name=None, params=None):
         if not params:
             params = {}
+        # required in case a previous track using a different repository has specified the revision
+        if cfg.opts("track", "repository.name") != self.repository:
+            cfg.add(config.Scope.applicationOverride, "track", "repository.revision", None)
```
danielmitterdorfer, line 1383: (Unrelated to this PR) I wonder whether we should disallow specifying a track revision for the generator via a command line parameter.

danielmitterdorfer (approved):
LGTM. I left one thought but not meant to be addressed by this PR. I am not certain I understood your PR description as it says:

> Honor "track-revision" property, along side the track property, [...]

Did you mean to say "Honor the track-repository property"?

dliappis:
> Did you mean to say "Honor the track-repository property"?

Right, I've changed it.

esrally/chart_generator.py
dliappis: Yes I think this would make sense given that we allow mixing tracks from different repositories.

Issue #365, opened 2017-11-17
Allow to parameterize queries


danielmitterdorfer:
Parameterizing queries (e.g. varying the search term in each iteration) requires a custom parameter source but it would be great if we could allow users to just specify values directly in the track.

Implementation note:

We could specify a path which part of the JSON should be replaced, e.g. `query.term.useragent` to replace`useragent` in the following query:

```json
"body": {
  "query": {
    "term": {
      "useragent": "Opera/5.11 (Windows 98; U) [en]"
    }
  }
}
```

However, we also need to support paths with JSON arrays.

Further ideas: Allow to read parameters from files, not just from static arrays.

danielmitterdorfer:
@talevy mentioned an additional and very related use case for this in #780: Pointing Rally to a file that contains a list of queries to run. In that case we'd parameterize the whole `body`. We should also consider to record in the metrics store which parameter has been used. If a simple plain-text file is provided, all we have is the value or the line number, e.g:

```
{"query": { "match_all": {}}}
{"query": {"exists": {"field": "user"}}}
```

In that case, the raw metrics sample could contain, e.g.:

```json
"op-params": {
    "body": 0
}
```

_The `0` refers to the `match_all` query in the first line, i.e. index zero)._

Alternatively, we could allow a different format, e.g. JSON as input which would allow to specify a well-known name per parameter:

```json
[
  {
    "key": "match-all",
    "value": "{\"query\": { \"match_all\": {}}}"
  },
  {
    "key": "user-exists",
    "value": "{\"query\": {\"exists\": {\"field\": \"user\"}}}"
  }
]
```

If such an input format is used, the raw sample could contain more useful meta-data:

```json
"op-params": {
    "body": "match-all"
}
```

danielmitterdorfer:
We don't intend to work on this for the forseeable future and recommend instead to use the `composite` operation introduced in https://github.com/elastic/rally/pull/1112 with a custom parameter source instead to simulate realistic queries.

PR #1138, opened 2020-12-17
Remove experimental query param substitution

danielmitterdorfer:
For a while we had experimental support to replace query body parameters
from a list of candidates. However, this feature was always
undocumented, we never used it ourselves in any benchmarks with the new
`composite` operation type we have better ways to support realistic
queries. Therefore we remove support for this feature with this commit.

Relates #365


gingerwizard (approved):
LGTM. Agreed its unlikely this would be that useful in reality - any query manipulation often needs to be dynamic e.g. based on the current timestamp, and this will be hard to support in json syntax.

PR #1139, opened 2020-12-17
Create default configuration automatically

danielmitterdorfer:
With this commit we remove the `configure` subcommand. Previously users
had to invoke this subcommand to configure Rally. Instead we recommend
now that users edit `rally.ini` themselves if needed and create a
default configuration implicitly in case it is not present.


danielmitterdorfer:
Although we remove a subcommand it's ok not to declar this as a breaking change as new users would not notice it because the getting starting experience has been adapted and existing users also won't notice it because their configuration does not need to be changed.

dliappis (approved):
Looks and works great LGTM!

PR #1140, opened 2020-12-17
Improve responsiveness of track preparation

danielmitterdorfer:
With this commit we offload track preparation to a background thread.
This ensures that the respective Rally actor is not blocked anymore and
can respond timely. One benefit of this change is that any log messages
that are emitted during track preparation show up in the log file close
to when they are emitted instead of at the end of track preparation.


gingerwizard (approved):
LGTM. I'm familiar with the actor system sufficiently to understand how the threading works and how we block subsequent tasks from being run until this thread completes. Code standard wise it looks good.

PR #1144, opened 2021-01-05
support point-in-time API and search-after pagination

### LINKED ISSUE #1141: Support Point-in-time API
Currently Rally only supports the scroll API for paginating search results. Elasticsearch has introduced a [point-in-time API](https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html) that we should also support similar to how we do it with async search since https://github.com/elastic/rally/pull/1129 (i.e. also assume the same restrictions such as it is only allowed in the context of a `composite` task). We should specifically make sure to add support for [`search_after`](https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after) using the point-in-time API.

DJRickyB:
opening as a draft to source input on the following.
[ ] structure of functionality as 3 new operation types vs adding functionality to existing query type
[ ] naming of search-after operation
[ ] search-after operation's parameterization using query instead of body
[ ] behavior of ClosePointInTime (should it pop the pit_id out of the CompositeContext?)

also todo
[ ] updated documentation

closes #1141


esrally/driver/runner.py
```diff
@@ -2068,6 +2071,62 @@ def __repr__(self, *args, **kwargs):
...
+        body["size"] = size
+        pit_op = params.get("with-point-in-time-from")
+        request_params, headers = self._transport_request_params(params)
+
+        while True:
+            if pit_op:
+                pit_id = CompositeContext.get(pit_op)
+                body["pit"] = {"id": pit_id,
+                               "keep_alive": "1m" }
+            props = await self._raw_search(es, doc_type=None, index=None, body=body.copy(), params=request_params, headers=headers)
```
DJRickyB, line 2102: note here we leave eager parsing on and materialize the response.  this is because we need the last hit's sort field in order to know what to search_after

esrally/driver/runner.py
```diff
@@ -2068,6 +2071,62 @@ def __repr__(self, *args, **kwargs):
...
 
 
+class OpenPointInTime(Runner):
+    async def __call__(self, es, params):
+        index = mandatory(params, "index", self)
+        response = es.open_point_in_time(index=index,
+                                         params=params)
+        id = response.get("id")
+        op_name = mandatory(params, "name", self)
+        # is name uniqueness enforced somewhere? this may collide with async searches
```
danielmitterdorfer, line 2081: Uniqueness is not enforced but we could add a check to `CompositeContext#put`.

esrally/driver/runner.py
danielmitterdorfer, line 2102: Is the `body.copy()` necessary here?

danielmitterdorfer (commented):
Thanks for adding this. I'm only providing high-level feedback here as this is a draft PR.

Because of how PIT is used in this PR it would make sense to include it directly in the search after runner. However, PIT is not tied to search after so I propose that we keep them separate (as you've already implemented it) because this allows us to use PIT more flexibly in the future. I think it would make sense to remove the `pit_id` from `CompositeContext` when closing it to expose bugs early on although the composite context is dropped when it goes out of scope.

A couple of things about the search after runner:

* Currently it will exhaustively paginate over search results. That may not be realistic and we should maybe include an additional parameter to allow early termination (e.g. after a certain number of pages).
* I'd probably just name the runner `SearchAfter`.
* I think it's fine to restrict support for this runner to `composite` structures as you did.
* I'd probably have the user just provide the initial body instead of having separate keys for `query` and `sort`. Then you also don't need a separate `results-per-page` and users can add other properties as well (see the [request body docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-search.html#search-search-api-request-body) for supported properties)
* I see how the body parsing is necessary but I fear the overhead might be quite high (I recall it was very noticeable for scroll requests) and I wonder whether we can "convince" the `ijson` parser to parse from the bottom instead of the top, or even extract it ourselves. I think it would pay off to measure the overhead here with a reasonably large response (see https://github.com/elastic/rally/pull/941 for more details for regular scrolls) to make sure we don't introduce accidental client-side bottlenecks.

esrally/driver/runner.py
DJRickyB: interesting problem...

If you take it out, a test that asserts the correct order of a changing PIT key was used in this loop fails, due to (apparently) the `.._called_with` APIs themselves use objects as references rather than doing a deep copy.  The mutation in this loop causes the mock to ultimately assert that the _final_ iteration of the body is what was passed each time the ES client is exercised.

It is not functionally required, but I also desire to have the test in place that checks the thing is properly updated.  Thoughts?

esrally/driver/runner.py
DJRickyB: Ok, so have the context itself raise an error if the key already exists?  I think this is doubly supportive of the notion that close_pit should also pop the pit_id from the CompositeContext 👍

DJRickyB:
>     Currently it will exhaustively paginate over search results. That may not be realistic and we should maybe include an additional parameter to allow early termination (e.g. after a certain number of pages).

Seems easy

>     I'd probably just name the runner SearchAfter.

As it stands today, it sub-classes the Query runner.  Is the separate runner preferable to just supporting it as a configuration in the existing runner?

>     I think it's fine to restrict support for this runner to composite structures as you did.

To be clear, only the pit operations are restricted to composite structures.  If SearchAfter doesn't use a PIT, it doesn't require a live CompositeContext (this assertion is in the tests as they are now)

>     I'd probably have the user just provide the initial body instead of having separate keys for query and sort. Then you also don't need a separate results-per-page and users can add other properties as well (see the request body docs for supported properties)

This makes me wonder more about merging this functionality into the `search` operation

>     I see how the body parsing is necessary but I fear the overhead might be quite high (I recall it was very noticeable for scroll requests) and I wonder whether we can "convince" the ijson parser to parse from the bottom instead of the top, or even extract it ourselves. I think it would pay off to measure the overhead here with a reasonably large response (see #941 for more details for regular scrolls) to make sure we don't introduce accidental client-side bottlenecks.

I bet an end-anchor-oriented regexp could be fairly performant, at least compared to full dict materialization.  Will investigate

esrally/driver/runner.py
danielmitterdorfer: 🤦. Makes sense now with your explanation and I fear you have already chosen the most pragmatic approach here. Maybe a comment is warranted here.

esrally/driver/runner.py
danielmitterdorfer: Yes, sounds good to me.

danielmitterdorfer:
> As it stands today, it sub-classes the Query runner. Is the separate runner preferable to just supporting it as a configuration in the existing runner?

I'm really fine either way. The approach you're proposing (integrating into the query runner) sounds a bit more flexible so I'd lean towards that.

> I bet an end-anchor-oriented regexp could be fairly performant, at least compared to full dict materialization. Will investigate

Thanks!

esrally/driver/runner.py
gingerwizard: Reading through this implementation, you seem to always paginate pages to exhaustion? I'm not sure entirely sure this is what we will always want to do when using this functionality.  Could we possibly limit the number of pages or the total number of results to paginate?

DJRickyB:
thanks both for your input. closing this in favor of a(n eventual) clean PR

PR #1147, opened 2021-01-12
Don't require a name to register a track processor

danielmitterdorfer:
With this commit we drop the `name` argument from the registration
method for track processors. Contrary to runners, schedulers and
parameter sources there is no need to refer to track processors by name
and this parameter has been introduced by mistake.

Relates #1135


gingerwizard (approved):
LGTM

PR #1148, opened 2021-01-13
Make docs build stricter

danielmitterdorfer:
With this commit we treat warnings as errors when building docs with
`make docs`. We also remove the obsolete docs configuration from tox as
documentation should only be built via the make target. Finally, we
upgrade the RTD theme to its latest version to avoid
`RemovedInSphinx30Warning` when building docs.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -55,7 +55,7 @@ clean:
 
 .PHONY: html
 html:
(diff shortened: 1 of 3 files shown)


DJRickyB (approved):
LGTM

PR #1151, opened 2021-01-14
Use regular dependency resolver in pip

danielmitterdorfer:
In order to workaround issues with the old PIP resolver we have resorted
to the pip command line flag `--use-feature=2020-resolver`. In newer
versions of pip this doesn't have any effect but issues a warning:

```
--use-feature=2020-resolver no longer has any effect,
since it is now the default dependency resolver in pip.
This will become an error in pip 21.0.
```

In order to avoid this warning, we drop the flag from all invocations of
pip in our development workflow and our Docker images.


danielmitterdorfer:

DJRickyB (approved):
LGTM

PR #1154, opened 2021-01-18
Allow to filter tasks by tag

danielmitterdorfer:
With this commit we allow users to filter tasks by tags. Tasks define a
new property `tags` and we add a new `tag` keyword prefix for task
filters. A usage example is:

```
--exclude-tasks="tag:setup"
```

By adding this task filter, Rally executes all tasks but the ones that
contain this tag.

### DIFF
--- a/esrally/track/track.py
+++ b/esrally/track/track.py
@@ -704,7 +704,7 @@ class TaskNameFilter:
         return isinstance(other, type(self)) and self.name == other.name
 
     def __str__(self, *args, **kwargs):
-        return "filter for task name [%s]" % self.name
+        return f"filter for task name [{self.name}]"
 
 
 class TaskOpTypeFilter:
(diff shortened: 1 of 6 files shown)


esrally/track/track.py
```diff
@@ -721,7 +721,24 @@ def __eq__(self, other):
...
-        return "filter for operation type [%s]" % self.op_type
+        return f"filter for operation type [{self.op_type}]"
+
+
+class TaskTagFilter:
+    def __init__(self, tag_name):
+        self.tag_name = tag_name
+
+    def matches(self, task):
+        return self.tag_name in task.tags
```
gingerwizard, line 732: This appears to mean case sensitive matching? Is this what we want? I assume `include-tasks` is currently case sensitive?

gingerwizard (approved):
LGTM. Surprisingly simple change. Only comment on case sensitivity. I'd lean towards the behaviour we have now on other filtering and document accordingly.

esrally/track/track.py
danielmitterdorfer: Correct, this behavior is consistent with current behavior, i.e. all other task filters are also case-sensitive.

danielmitterdorfer: I pushed an additional note in the docs about case sensitivity of task filters in https://github.com/elastic/rally/pull/1154/commits/6cd13b22e6fce3dcd3acdc5a7bd6370af7957cd7.

Issue #1142, opened 2020-12-22
Always require a subcommand


danielmitterdorfer:
Currently Rally will default to running a benchmark when it is invoked without any subcommand. This is intransparent both to the user and in the code. Also, this is in conflict to the direction we want to move, namely getting away from the integrated handling of setting up Elasticsearch and running a benchmark in one command in favor of using dedicated subcommands (`install`, `start`, `stop`, `race`). Should we therefore require that a user always specifies a subcommand?

danielmitterdorfer:
We have disussed this in our sync and decided to move forward with this.

PR #1155, opened 2021-01-19
Deprecate invoking Rally without a subcommand

danielmitterdorfer:
With this commit we deprecate running Rally without a subcommand. This
makes the user's intention explicit and avoids any leniency in Rally
around subcommand handling. Once a subcommand is required, command line
help can also be improved by only showing command line flags relevant to
a specific subcommand.

Relates #1142

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -589,7 +589,8 @@ def derive_sub_command(args):
     sub_command = args.subcommand
     if sub_command is None:
         logger = logging.getLogger(__name__)
-        console.warn("Invoking Rally without a subcommand is deprecated. Specify the 'race' subcommand explicitly.", logger=logger)
+        console.warn("Invoking Rally without a subcommand is deprecated and will be required with Rally 2.1.0. Specify "
+                     "the 'race' subcommand explicitly.", logger=logger)
         return "race"
     else:
         return sub_command
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -587,8 +587,12 @@ def create_arg_parser():
 
 def derive_sub_command(args):
     sub_command = args.subcommand
-    # TODO: Can we select a subcommand by default?
-    return sub_command if sub_command is not None else "race"
+    if sub_command is None:
+        logger = logging.getLogger(__name__)
+        console.warn("Invoking Rally without a subcommand is deprecated. Specify the 'race' subcommand explicitly.", logger=logger)
+        return "race"
+    else:
+        return sub_command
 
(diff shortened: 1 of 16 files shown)


esrally/rally.py
```diff
@@ -587,8 +587,12 @@ def runtime_jdk(v):
 
 def derive_sub_command(args):
     sub_command = args.subcommand
-    # TODO: Can we select a subcommand by default?
-    return sub_command if sub_command is not None else "race"
+    if sub_command is None:
+        logger = logging.getLogger(__name__)
+        console.warn("Invoking Rally without a subcommand is deprecated. Specify the 'race' subcommand explicitly.", logger=logger)
```
dliappis, line 592: I wonder, have we ever considered https://docs.python.org/3/library/warnings.html for such warnings?

dliappis: Should we specify that the subcommand `race` will be required starting with Rally 2.1.0?

dliappis (approved):
LGTM left a question and a suggestion.

esrally/rally.py
danielmitterdorfer: I attemped to use the warnings module now but there are two issues:

1. We configure Rally to redirect warnings to the log. We do this because of urllib3 which would otherwise write to the terminal (see https://github.com/elastic/rally/blob/5bf0fed3ddf442d19265d314f8d50bce95ad4127/esrally/log.py#L129).
2. Even if we disable this behavior, the output is slightly verbose:

```
/Users/daniel/Projects/rally/esrally/rally.py:592: FutureWarning: Invoking Rally without a subcommand is deprecated. Specify the 'race' subcommand explicitly.
  warnings.warn("Invoking Rally without a subcommand is deprecated. Specify the 'race' subcommand explicitly.", category=FutureWarning)
```

I've included a note about 2.1.0 in the message in https://github.com/elastic/rally/pull/1155/commits/7513e561a0734a516801d17858b4e400b15e37bf.

dliappis: Thanks for taking a look at the warnings module! This was just something to consider, given the ^^ observations doesn't make sense to use in this project right now.

Also thanks for enhancing the message in https://github.com/elastic/rally/commit/7513e561a0734a516801d17858b4e400b15e37bf, I think it'll help the user base prepare promptly before upgrading to the next minor.

danielmitterdorfer: Well, thanks for bringing it up! I'll wait until the CI build is finished and will merge the PR then.

PR #1157, opened 2021-01-25
Add support for assertions

danielmitterdorfer:
With this commit we add a new `assertions` property for operations. When
this property is present and assertions are enabled with `--enable-assertions`,
Rally checks all assertions and raises an error if the check fails.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -579,7 +579,7 @@ Use assertions for sanity checks, e.g. to ensure a query returns results. Assert
 * ``condition``: The following conditions are supported: ``<``, ``<=``, ``==``, ``>=``, ``>=``.
 * ``value``: The expected value.
 
-While assertions are always evaluated and raise an error when they are violated, Rally's default behavior is to only record errors and continue executing the benchmark. To abort on assertion errors, run Rally with ``--on-error=abort``.
+Assertions are disabled by default and can be enabled with the command line flag ``--enable-assertions``. A failing assertion aborts the benchmark.
 
 Example::
 
@@ -607,7 +607,7 @@ Example::
 
     This requires to set ``detailed-results`` to ``true`` so the search operation gathers additional meta-data, such as the number of hits.
 
-If this assertion fails, and Rally is executed with ``--on-error=abort`` it exits with the following error message::
+If assertions are enabled with ``--enable-assertions`` and this assertion fails, it exits with the following error message::
 
     [ERROR] Cannot race. Error in load generator [0]
         Cannot run task [term]: Expected [hits] to be > [0] but was [0].
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -566,9 +566,53 @@ Each operation consists of the following properties:
 * ``name`` (mandatory): The name of this operation. You can choose this name freely. It is only needed to reference the operation when defining schedules.
 * ``operation-type`` (mandatory): Type of this operation. See below for the operation types that are supported out of the box in Rally. You can also add arbitrary operations by defining :doc:`custom runners </adding_tracks>`.
 * ``include-in-reporting`` (optional, defaults to ``true`` for normal operations and to ``false`` for administrative operations): Whether or not this operation should be included in the command line report. For example you might want Rally to create an index for you but you are not interested in detailed metrics about it. Note that Rally will still record all metrics in the metrics store.
+* ``assertions`` (optional, defaults to ``None``): A list of assertions that should be checked. See below for more details.
 * ``request-timeout`` (optional, defaults to ``None``): The client-side timeout for this operation.  Represented as a floating-point number in seconds, e.g. ``1.5``.
 * ``headers`` (optional, defaults to ``None``): A dictionary of key-value pairs to pass as headers in the operation.
 * ``opaque-id`` (optional, defaults to ``None`` [unused]): A special ID set as the value of ``x-opaque-id`` in the client headers of the operation.  Overrides existing ``x-opaque-id`` entries in ``headers`` (case-insensitive).
+
+**Assertions**
+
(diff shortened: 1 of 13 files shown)


danielmitterdorfer:
This has been caused by a transient failure:

```
[2021-01-25T12:53:07,103][ERROR][o.e.b.Bootstrap          ] [rally-node-0] Exception
org.elasticsearch.http.BindHttpException: Failed to bind to [39200]
[...]
Caused by: java.net.BindException: Address already in use
```

docs/track.rst
```diff
@@ -1619,9 +1663,18 @@ composite
...
 
-* ``requests`` (mandatory): A list that specifies the request streams to execute. Each stream consists of operations of either type ``raw-request`` to send requests to Elasticsearch or ``sleep`` to simulate client processing time. Streams execute concurrently, operations within a stream sequentially. It is possible to nest streams. See below for specific examples.
+* ``requests`` (mandatory): A list that specifies the request streams to execute. Streams execute concurrently, operations within a stream sequentially. It is possible to nest streams. See below for specific examples.
 * ``max-connections`` (optional: defaults to unbounded): The maximum number of concurrent connections per client executing this composite operation. By default, the operation itself does not restrict the number of connections but is bound to Rally's network connection limit. Therefore raise the number of available network connections appropriately (see :ref:`command line reference <clr_client_options>`).
 
+The ``composite`` operation only supports the following operation-types:
+
+* ``raw-request``
+* ``sleep``
+* ``search``
```
gingerwizard, line 1673: Adding search to composite 👍

docs/track.rst
```diff
@@ -566,10 +566,54 @@ Each operation consists of the following properties:
...
+* ``assertions`` (optional, defaults to ``None``): A list of assertions that should be checked. See below for more details.
 * ``request-timeout`` (optional, defaults to ``None``): The client-side timeout for this operation.  Represented as a floating-point number in seconds, e.g. ``1.5``.
 * ``headers`` (optional, defaults to ``None``): A dictionary of key-value pairs to pass as headers in the operation.
 * ``opaque-id`` (optional, defaults to ``None`` [unused]): A special ID set as the value of ``x-opaque-id`` in the client headers of the operation.  Overrides existing ``x-opaque-id`` entries in ``headers`` (case-insensitive).
 
+**Assertions**
+
+Use assertions for sanity checks, e.g. to ensure a query returns results. Assertions need to be defined with the following properties. All of them are mandatory:
+
+* ``property``: A dot-delimited path to the meta-data field to be checked. Only meta-data fields that are returned by an operation are supported.
```
gingerwizard, line 578: Maybe im missing something but i cant find a list of meta fields without looking at the runner. Is this in the docs somewhere?

esrally/driver/runner.py
```diff
@@ -2042,15 +2124,26 @@ async def __call__(self, es, params):
...
+        stats = {}
+        for search_id, search in async_search_ids(searches):
             response = await es.async_search.get(id=search_id,
                                                  params=request_params)
-            success = success and not response["is_running"]
+            is_running = response["is_running"]
+            success = success and not is_running
+            if not is_running:
+                stats[search] = {
+                    "hits": response["response"]["hits"]["total"]["value"],
```
gingerwizard, line 2135: Do we document these as the "meta-fields" available for assertions?

esrally/track/params.py
```diff
@@ -515,24 +517,12 @@ def __init__(self, track, params, **kwargs):
             self.query_params["pages"] = pages
         if results_per_page:
             self.query_params["results-per-page"] = results_per_page
+        if "assertions" in params:
+            self.query_params["assertions"] = params["assertions"]
```
gingerwizard, line 521: Should we error if `detailed-results` is False and `assertions` is True?

esrally/driver/runner.py
```diff
@@ -311,6 +324,69 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+        self.predicates = {
+            ">": self.greater_than,
+            ">=": self.greater_than_or_equal,
+            "<": self.smaller_than,
+            "<=": self.smaller_than_or_equal,
+            "==": self.equal,
+        }
+
+    def greater_than(self, expected, actual):
+        return actual > expected
```
gingerwizard, line 341: I notice there are no type restrictions here? i.e. we could in theory compare strings. Is this intentional and left to the user to ensure valid types are being compared?

gingerwizard (approved):
A few clarifications if possible but more for my understanding re typing and meta fields

docs/track.rst
danielmitterdorfer: You did not miss anything; this is not documented at the moment. I propose we do this in a follow-up PR.

esrally/driver/runner.py
danielmitterdorfer: As indicated above: No, it's currently undocumented.

esrally/track/params.py
danielmitterdorfer: Good idea. :)

esrally/driver/runner.py
danielmitterdorfer: Yes, it's intentional that we're not too restrictive here.

esrally/track/params.py
danielmitterdorfer: I've pushed https://github.com/elastic/rally/pull/1157/commits/8286f9fd17eb04666bb370200e81d97f4b924e24 to address your proposal. Can you please have another look @gingerwizard?

esrally/driver/runner.py
gingerwizard: 👍 maybe we could warn on type comparisons but probably not worth the cost.

PR #1160, opened 2021-01-26
Include empty mean value in summary stats

danielmitterdorfer:
With this commit we add `mean` also in summary stats when no value can
be derived. This is consistent with the handling of all other summary
metrics and ensures command line reporting can rely on stats keys always
being present.


gingerwizard (approved):
LGTM

PR #1162, opened 2021-01-27
Log trace of uncaught exception in load generator

danielmitterdorfer:
With this commit we enhance the handlers for uncaught exceptions
in the load generator to log the full exception trace before forwarding
the exception *message* to the coordinating process. Previously, the
coordinating process would then log the exception message but we lost
the actual trace which makes diagnosis harder.


danielmitterdorfer:

dliappis (approved):
LGTM

PR #1163, opened 2021-01-28
Make some document properties extensible

danielmitterdorfer:
The `Documents` domain class currently assumes static corpora and thus
populates all properties eagerly via instance fields. There are cases
however when we want to resolve these attributes lazily because the
respective set of documents is generated at a later stage in the
benchmark and we cannot derive these properties upfront. With this
commit we make the relevant properties extensible so they can be
resolved lazily in subclass implementations.

### DIFF
--- a/esrally/track/track.py
+++ b/esrally/track/track.py
@@ -213,9 +213,9 @@ class Documents:
         self.document_archive = document_archive
         self.base_url = base_url
         self.includes_action_and_meta_data = includes_action_and_meta_data
-        self.number_of_documents = number_of_documents
-        self.compressed_size_in_bytes = compressed_size_in_bytes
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
Note to reviewers: I was pondering to expose all instance fields via properties but opted against it because it would make the implementation of this class much larger for little value (except consistency).

esrally/track/track.py
```diff
@@ -227,6 +227,30 @@ def has_compressed_corpus(self):
     def has_uncompressed_corpus(self):
         return self.document_file is not None
 
+    @property
```
DJRickyB, line 230: TIL

DJRickyB (approved):
LGTM

PR #1164, opened 2021-02-01
Update .pylintrc to remove exception

ebadyano:
We currently don't have `Wrong hanging indentation` errors
 Relates to https://github.com/elastic/rally/issues/838


ebadyano:
build failed, seems like connection issues @elasticmachine test this please

danielmitterdorfer (approved):
Nice catch! Thanks for the cleanup.

Can you please update the PR title / description **and** the commit title / body when you merge this so it matches the intent (Why are we doing the change?) instead of the action (What are we changing?). See https://github.com/elastic/rally/pull/1131 for an example.

ebadyano:
@danielmitterdorfer thank you for the review!

PR #1165, opened 2021-02-02
Fallback to closest minor branch for track/team repositories

dliappis:
So far Rally's fallback logic for picking the track/team repository
branch has been basically either that of exact match, or failing that
use the matching major version.

This means that e.g. a major change introduced in 7.11 would mean that
e.g. rally-teams would require branches 7.11, 7.12 etc. throughout the
entire lifecycle of the 7 version.

This commit alters this fallback logic and instead will prefer the
nearest minor branch, if available.

Also convert all `utils/versions.py` tests to `pytest`.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -45,7 +45,7 @@ Rally implements a fallback logic in order of specificity up to the minor versio
 Assuming the track repository has several branches, the order is:
 
 1. Exact branch matches; e.g. if the repo contains branches `7`, `7.1` and `7.10.2` and Elasticsearch version is `7.10.2`, `7.10.2` will be checked out.
-2. Nearest minor matches;  e.g. if the repo contains branches `7`, `7.1` and `7.10` and Elasticsearch version is `7.10.2`, `7.10` will be checked out. Alternatively if version is `7.9`, `7.1` will be checked out.
+2. Nearest prior minor matches;  e.g. if the repo contains branches `7`, `7.1` and `7.10` and Elasticsearch version is `7.10.2`, `7.10` will be checked out. Alternatively if version is `7.9`, `7.1` will be checked out.
 3. Major branch matches; e.g. if the repo contains branches `7` and `7.10` and Elasticsearch version is `7.1`, `7` will be checked out.
 4. Failing everything, `master` will be elected, e.g. if the repo contains branches `6`, `7` and `master` and Elasticsearch version is `8.1.0`, `master` will be checked out.
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -44,10 +44,10 @@ Rally implements a fallback logic in order of specificity up to the minor versio
 
 Assuming the track repository has several branches, the order is:
 
-1. Exact branch matches; e.g. if the repo contains branches `7`, `7.1` and `7.10.2` and Elasticsearch version is `7.10.2`, `7.10.2` will be chosen as the track repo.
-2. Nearest minor matches;  e.g. if the repo contains branches `7`, `7.1` and `7.10` and Elasticsearch version is `7.10.2`, `7.10` will be chosen as the track repo. Alternatively if version is `7.9`, `7.1` will be chosen.
-3. Major branch matches; e.g. if the repo contains branches `7` and `7.10` and Elasticsearch version is `7.1`, `7` will be chosen as the track repo.
-4. Failing everything, `master` will be elected, e.g. if the repo contains branches `6`, `7` and `master` and Elasticsearch version is `8.1.0`, `master` will be chosen as the track repo.
+1. Exact branch matches; e.g. if the repo contains branches `7`, `7.1` and `7.10.2` and Elasticsearch version is `7.10.2`, `7.10.2` will be checked out.
+2. Nearest minor matches;  e.g. if the repo contains branches `7`, `7.1` and `7.10` and Elasticsearch version is `7.10.2`, `7.10` will be checked out. Alternatively if version is `7.9`, `7.1` will be checked out.
+3. Major branch matches; e.g. if the repo contains branches `7` and `7.10` and Elasticsearch version is `7.1`, `7` will be checked out.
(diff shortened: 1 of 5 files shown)


docs/track.rst
```diff
@@ -38,12 +38,18 @@ Alternatively, you can store Rally tracks also in a dedicated git repository whi
...
+.. _track-repositories-fall-back-logic:
 
-* The branch `6.0.0-alpha1` will be chosen for the version ``6.0.0-alpha1`` of Elasticsearch.
-* The branch `5` will be chosen for all versions for Elasticsearch with the major version 5, e.g. ``5.0.0``, ``5.1.3`` (provided there is no specific branch).
+Rally implements a fallback logic in order of specificity up to the minor version level, so you don't need to define a branch for each patch release of Elasticsearch.
 
-Rally tries to use the branch with the best match to the benchmarked version of Elasticsearch.
+Assuming the track repository has several branches, the order is:
+
+1. Exact branch matches; e.g. if the repo contains branches `7`, `7.1` and `7.10.2` and Elasticsearch version is `7.10.2`, `7.10.2` will be chosen as the track repo.
```
danielmitterdorfer, line 47: Maybe say "will be chosen as branch" or "checked out" instead of "will be chosen as the track repo"?

docs/track.rst
```diff
@@ -38,12 +38,18 @@ Alternatively, you can store Rally tracks also in a dedicated git repository whi
...
+2. Nearest minor matches;  e.g. if the repo contains branches `7`, `7.1` and `7.10` and Elasticsearch version is `7.10.2`, `7.10` will be chosen as the track repo. Alternatively if version is `7.9`, `7.1` will be chosen.
```
danielmitterdorfer, line 48: I think the term "nearest" is too vague. Consider the version is `7.8.0` and we have branches `7.9` and `7.1`. To me, `7.9` is "nearer than" `7.1` because it's only one minor version distance. Therefore we need to refer to this behavior more accurately, e.g. as "nearest prior minor"?

Note that the term "nearest" is used elsewhere too (e.g. in code comments).

esrally/utils/repo.py
```diff
@@ -103,7 +103,7 @@ def update(self, distribution_version):
 
     def _find_matching_tag(self, distribution_version):
         tags = git.tags(self.repo_dir)
-        for version in versions.versions(distribution_version):
+        for version, _ in versions.VersionVariants(distribution_version).all_versions:
```
danielmitterdorfer, line 106: Would it make sense to expose this as a function instead? This would maybe read a bit nicer:

```python
for version, _ in versions.variants_of(distribution_version):
    ...
```

tests/utils/versions_test.py
```diff
@@ -15,64 +15,126 @@
...
+            "Exact match"
+
+        assert versions.best_match(["7", "7.10", "master"], "7.1.0") == "7", \
+            "Best match is major version"
+
+        assert versions.best_match(["7", "7.11", "7.2", "5", "6", "master"], "7.11.0") == "7.11",\
+            "Best match for specific minor version"
+
+        assert versions.best_match(["7", "7.11", "7.2", "5", "6", "master"], "7.12.0") == "7.11",\
+            "If no exact match, best match is the closest minor"
```
danielmitterdorfer, line 128: I interpret "closest" as a synonym for "nearest" in this context. IMHO we should stick to one term consistently.

danielmitterdorfer (commented):
Thanks for the PR. I left a couple of comments, mostly around terminology.

docs/track.rst
dliappis: Addressed in 3ae2fa484ec82d3aac09aa045e9f16b1f4bcc120

docs/track.rst
dliappis: Addressed in 861a176a7466163096ac2af17b431df48d7e2a3f

esrally/utils/repo.py
dliappis: The class `VersionVariants` is used in `utils/versions.py` leveraging state. `repo.py` doesn't require state, just the list expansion of all versions.
Are you suggesting `variants_of()` could be a utility function within `util/versions.py` that does the object instantiation/calling of `.all_versions()` or?

tests/utils/versions_test.py
dliappis: Addressed in 861a176a7466163096ac2af17b431df48d7e2a3f

esrally/utils/repo.py
danielmitterdorfer: Yes, I am proposing a utility function along the lines of:

```python
def variants_of(distribution_version):
    for version, _ in versions.VersionVariants(distribution_version).all_versions
        yield version
```

This implementation would even get rid of the superfluous second return value of `#all_versions()` and give us the chance to keep the class `VersionVariants` internal to to the `versions` module.

dliappis: Ok fixed in 13ba1ad7dce96530611619a4eb8d3999a3640074

danielmitterdorfer (approved):
Thanks for iterating! LGTM

dliappis:
@danielmitterdorfer it occurs to me that the final step, before merging this is to create the following branches:

`rally-tracks`

| branch | based_on | intended for ES versions |
| --------- | ------------- | ---------------------------- |
| 7.2 | 7 | >=7.2.0 and <7.11.0 (otherwise e.g. ES 7.10.2 would use Rally tracks `7.1`) |

`rally-teams`

| branch | based_on | intended for ES versions |
| --------- | ------------- | ---------------------------- |
| 6.3 | 6 | >=6.3.0 and <7.0.0 (otherwise ES 6.8.7 would use rally-teams `6.2`) |

WDYT?

danielmitterdorfer:
As discussed offline this sounds reasonable.

dliappis:
[Earlier comment](https://github.com/elastic/rally/pull/1165#issuecomment-773254232) has now been addressed:

- New `7.2` rally-tracks branch: https://github.com/elastic/rally-tracks/tree/7.2
- New `6.3` rally-teams branch: https://github.com/elastic/rally-teams/tree/6.3

PR #1167, opened 2021-02-03
Clarify num of docs in corpora when action and metadata is used

dliappis:
In the typical cases of corpora that doesn't include an action-and-meta
-data line, defining the number of documents is simple and equivalent
to the number of lines.

This commit clarifies that this calculation differs when using the
`includes-action-and-meta-data` property and that number of documents
should not include action-and-meta-data lines.


docs/track.rst
```diff
@@ -298,7 +298,12 @@ Each entry in the ``documents`` list consists of the following properties:
...
 * ``source-file`` (mandatory): File name of the corresponding documents. For local use, this file can be a ``.json`` file. If you provide a ``base-url`` we recommend that you provide a compressed file here. The following extensions are supported: ``.zip``, ``.bz2``, ``.gz``, ``.tar``, ``.tar.gz``, ``.tgz`` or ``.tar.bz2``. It must contain exactly one JSON file with the same name. The preferred file extension for our official tracks is ``.bz2``.
-* ``includes-action-and-meta-data`` (optional, defaults to ``false``): Defines whether the documents file contains already an action and meta-data line (``true``) or only documents (``false``).
+* ``includes-action-and-meta-data`` (optional, defaults to ``false``): Defines whether the documents file contains already an `action and meta-data <https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-desc>`_ line (``true``) or only documents (``false``).
+
+    .. note::
+
+        In this case the ``documents`` property should only reflect the number of documents and not additionally include the number of action and metadata lines.
```
danielmitterdorfer, line 305: "In this case" refers to the case when `includes-action-and-meta-data` is set to `true`. Maybe we should say instead: 

```
When ``includes-action-and-meta-data`` is set to ``true``...
```

danielmitterdorfer (approved):
Looks fine. I left a suggestion but no need for another review round.

docs/track.rst
dliappis: Thanks. It's actually an interesting situation when a corpus contains action and metadata line but this property isn't true: it'll try to index them, frequently failing due to strict mappings :)

I've reworded with slightly fewer words to avoid cognitive load from repetition of terms in 20c839e0f6fe5329d47b8f60edfb1eae63738941

PR #1168, opened 2021-02-03
Require unit for bulk operations

danielmitterdorfer:
With this commit we remove the deprecated logic to automatically set a
unit for bulk operations when the parameter source does not provide one.
Note that this only affects custom parameter sources because Rally's
default bulk parameter source will provide this parameter.

Relates #1100

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -1,6 +1,16 @@
 Migration Guide
 ===============
 
+Migrating to Rally 2.1.0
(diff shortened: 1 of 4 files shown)


danielmitterdorfer:
Note that this PR is targeting the next minor release 2.1.0, not 2.0.4.

dliappis (approved):
LGTM thanks!

PR #1169, opened 2021-02-03
Document meta-data returned by operations

danielmitterdorfer:
With this commit we add a new `meta data` section to each operation in
the track reference documentation which documents which meta data are
returned. This information can be used to add assertions for custom
tracks. We also enhance a couple of existing runners so that they return
a `dict` structure instead of a tuple because only `dicts` are supported
for assertions.

Relates #1157

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -457,135 +457,6 @@ class BulkIndex(Runner):
          in ``benchmarks/driver``.
         * ``request-timeout``: a non-negative float indicating the client-side timeout for the operation.  If not present, defaults to
          ``None`` and potentially falls back to the global timeout setting.
-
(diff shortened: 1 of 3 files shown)


esrally/driver/runner.py
```diff
@@ -457,135 +457,6 @@ async def __call__(self, es, params):
          in ``benchmarks/driver``.
         * ``request-timeout``: a non-negative float indicating the client-side timeout for the operation.  If not present, defaults to
          ``None`` and potentially falls back to the global timeout setting.
-
```
danielmitterdorfer, line 460: As this would now duplicate user docs I've removed it here.

gingerwizard (approved):
LGTM.

PR #1171, opened 2021-02-03
Add request-id to search operator

gingerwizard:
This allows a user to specify a request-id to a search operator. This will be passed down to the runner from the parameter source and include in metrics if specified. @danielmitterdorfer if you're happy with this approach ill push docs.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -904,6 +904,8 @@ class Query(Runner):
     * `pages`: Number of pages to retrieve at most for this scroll. If a scroll query does yield less results than the specified number of
                pages we will terminate earlier.
     * `results-per-page`: Number of results to retrieve per page.
+    * `request-id`: a user specified string identifying the query. This will be included in rally metrics to allow query
+                execution to be tracked.
 
     Returned meta data
 
@@ -920,7 +922,13 @@ class Query(Runner):
     For scroll queries we also return:
 
     * ``pages``: Total number of pages that have been retrieved.
+
+    If specified, we also return:
+
+    * `request_id``: The user specified id identifying the query.
+
     """
+
     async def __call__(self, es, params):
         if "pages" in params and "results-per-page" in params:
             return await self.scroll_query(es, params)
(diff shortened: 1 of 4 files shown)


esrally/driver/runner.py
```diff
@@ -951,28 +959,22 @@ async def request_body_query(self, es, params):
...
-                "timed_out": timed_out,
-                "took": took
-            }
-        else:
-            return {
-                "weight": 1,
-                "unit": "ops",
-                "success": True
-            }
+                "request_id": request_id
```
danielmitterdorfer, line 966: Is this a leftover? Because in the next line, the `request_id` is conditionally added to this dict.

danielmitterdorfer (commented):
Nice addition! I think there is one leftover but apart from that it looks fine. Can you please add the new parameter `request-id` also to the operation docs in the track reference? I'll take care in https://github.com/elastic/rally/pull/1169 that the additional meta-data property is documented.

esrally/driver/runner.py
gingerwizard: Well spotted and also a lint issue

dliappis (approved):
LGTM

esrally/driver/runner.py
```diff
@@ -775,7 +775,11 @@ class Query(Runner):
     * `pages`: Number of pages to retrieve at most for this scroll. If a scroll query does yield less results than the specified number of
                pages we will terminate earlier.
     * `results-per-page`: Number of results to retrieve per page.
+    * `request-id`: a user specified string identifying the query. This will be included in rally metrics to allow query
+                execution to be tracked.
+
```
dliappis, line 780: @gingerwizard as mentioned by the lint failure in CI, this newline needs to go away.

gingerwizard:
Discussed this with @danielmitterdorfer this doesn't bring us much sadly. For normal searches, we already index the operation name in the `operation` field. The composite operation doesn't propagate meta data and provide metadata for the searches inside it. This is therefore not that useful. Closing.

PR #1172, opened 2021-02-04
Improve isolation in race control tests

danielmitterdorfer:
With this commit we rely on pytest fixtures for proper cleanup even when
a test case fails. Previously, cleanup was only working when a test
succeeded leading to unwanted side-effects between tests.

### DIFF
--- a/tests/racecontrol_test.py
+++ b/tests/racecontrol_test.py
@@ -16,79 +16,93 @@
 # under the License.
 
 import os
+import re
 import unittest.mock as mock
-from unittest import TestCase
+
+import pytest
 
 from esrally import config, exceptions, racecontrol
 
 
-class RaceControlTests(TestCase):
-    def test_finds_available_pipelines(self):
-        expected = [
-            ["from-sources", "Builds and provisions Elasticsearch, runs a benchmark and reports results."],
-            ["from-sources-complete", "Builds and provisions Elasticsearch, runs a benchmark and reports results [deprecated]."],
-            ["from-sources-skip-build", "Provisions Elasticsearch (skips the build), runs a benchmark and reports results [deprecated]."],
-            ["from-distribution", "Downloads an Elasticsearch distribution, provisions it, runs a benchmark and reports results."],
-            ["benchmark-only", "Assumes an already running Elasticsearch instance, runs a benchmark and reports results"],
-        ]
+@pytest.fixture
+def running_in_docker():
+    os.environ["RALLY_RUNNING_IN_DOCKER"] = "true"
+    # just yield anything to signal the fixture is ready
+    yield True
+    del os.environ["RALLY_RUNNING_IN_DOCKER"]
 
-        self.assertEqual(expected, racecontrol.available_pipelines())
 
-    def test_prevents_running_an_unknown_pipeline(self):
-        cfg = config.Config()
-        cfg.add(config.Scope.benchmark, "race", "pipeline", "invalid")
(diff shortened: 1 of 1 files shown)


tests/racecontrol_test.py
```diff
@@ -16,79 +16,93 @@
...
+def test_prevents_running_an_unknown_pipeline():
+    cfg = config.Config()
+    cfg.add(config.Scope.benchmark, "race", "pipeline", "invalid")
+    cfg.add(config.Scope.benchmark, "mechanic", "distribution.version", "5.0.0")
 
-        del racecontrol.pipelines[test_pipeline_name]
+    with pytest.raises(exceptions.SystemSetupError) as ex:
+        racecontrol.run(cfg)
+
+    assert re.match(r"Unknown pipeline \[invalid]. List the available pipelines with [\S]+? list pipelines.",
```
dliappis, line 75: We could also use the modern feature [match](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest.raises) feature of `pytest` and simply:

```
    with pytest.raises(
            exceptions.SystemSetupError,
            match=r"Unknown pipeline \[invalid]. List the available pipelines with [\S]+? list pipelines."):
        racecontrol.run(cfg)
```

thus removing the need to use a named resource for the exception and a separate assert (and more importantly having to know which property of the exception object holds the message :) )

tests/racecontrol_test.py
```diff
@@ -16,79 +16,93 @@
...
-        cfg.add(config.Scope.benchmark, "mechanic", "distribution.version", "")
+def test_fails_without_benchmark_only_pipeline_in_docker(running_in_docker, unittest_pipeline):
+    cfg = config.Config()
+    cfg.add(config.Scope.benchmark, "race", "pipeline", "unit-test-pipeline")
 
+    with pytest.raises(exceptions.SystemSetupError) as ex:
         racecontrol.run(cfg)
 
-        mock_pipeline.assert_called_once_with(cfg)
+    assert ex.value.message == "Only the [benchmark-only] pipeline is supported by the Rally Docker image.\n" \
```
dliappis, line 95: Same suggestion as above, for leveraging [match](https://docs.pytest.org/en/stable/reference.html?highlight=raises#pytest.raises), and using `re.escape()` since this is a literal match i.e.

```
    with pytest.raises(
            exceptions.SystemSetupError,
            match=re.escape(
                "Only the [benchmark-only] pipeline is supported by the Rally Docker image.\n"
                "Add --pipeline=benchmark-only in your Rally arguments and try again.\n"
                "For more details read the docs for the benchmark-only pipeline in "
                "https://esrally.readthedocs.io/en/latest/pipelines.html#benchmark-only\n"
            )):
        racecontrol.run(cfg)
```

dliappis (approved):
Good catch. LGTM, left two suggestions.

tests/racecontrol_test.py
danielmitterdorfer: Nice, thanks for the improvement! I've pushed this in https://github.com/elastic/rally/pull/1172/commits/dff0ee0fa9b109627dfbca1beec71435337e0d19.

tests/racecontrol_test.py
danielmitterdorfer: Nice, thanks for the improvement! I've pushed this in https://github.com/elastic/rally/pull/1172/commits/dff0ee0fa9b109627dfbca1beec71435337e0d19.

PR #1174, opened 2021-02-04
Remove deprecated source pipelines

danielmitterdorfer:
With this commit we remove the deprecated pipelines
`from-sources-skip-build` and `from-sources-complete` which are replaced
by the pipeline `from-sources`.


DJRickyB (approved):
LGTM.  i poked around in my own copy locally and didn't see any further gotchas

danielmitterdorfer:
Thanks for the quick review!

PR #1175, opened 2021-02-04
Always require a subcommand

### LINKED ISSUE #1142: Always require a subcommand

danielmitterdorfer:
With this commit we remove leniency in subcommand handling and always
require a subcommand. This makes it obvious what command Rally is
executing and as a nice side-effect makes the command line help more
useful as well.

Closes #1142

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -11,6 +11,18 @@ Previously, Rally has implicitly used the unit ``docs`` for bulk operations. Wit
 
     esrally.exceptions.DataError: Parameter source for operation 'bulk-index' did not provide the mandatory parameter 'unit'. Add it to your parameter source and try again.
 
+Rally requires a subcommand
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Previously a subcommand was optional when running a benchmark. With Rally 2.1.0 a subcommand is always required. So instead of invoking::
+
+    esrally --distribution-version=7.10.0
+
+Invoke Rally with the ``race`` subcommand instead::
+
+    esrally race --distribution-version=7.10.0
+
+
 Migrating to Rally 2.0.4
 ------------------------
 
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -28,6 +28,18 @@ Invoke Rally with the ``race`` subcommand instead::
     esrally race --distribution-version=7.10.0
 
 
+Running without a track is deprecated
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Previously Rally has implicitly chosen the geonames track as default when ``--track`` was not provided. To ensure the workload is chosen deliberately, it is now deprecated to not specify a track. So instead of invoking::
+
+    esrally race --distribution-version=7.10.0
+
+Invoke Rally with ``--track=geonames`` instead::
+
+    esrally race --distribution-version=7.10.0 --track=geonames
+
+
 Migrating to Rally 2.0.4
 ------------------------
 
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -31,7 +31,7 @@ Invoke Rally with the ``race`` subcommand instead::
 Running without a track is deprecated
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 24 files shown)


danielmitterdorfer:
This requires a bit more work for command dispatching; I've converted it to a draft for now and unassigned Rick as a reviewer.

danielmitterdorfer:
Additional work in this PR:

* We have removed some obsolete parameters during the migration (like `--keep-cluster-running` which can be achieved now with dedicated subcommands)
* We have deprecated the default choice for `--track` (see also #1176 for follow-up work in 2.2.0).
* We have removed configuration keys when the command line argument was only provided to the function immediately called after command dispatching (e.g. comparison reporting, `--kill-running-processes)

I also decided *not* do to a couple of related refactorings here:

* Remove `--skip-build` flag for the `install` subcommand. This can be done in a dedicated PR.
* Restructure argument parser construction, config object preparation and command execution into a dedicated `Command` class. While this would nicely encapsulate these related aspects, the changeset would have been even larger so I decided against it. We could do this though in a follow-up PR at a later point.

danielmitterdorfer:
Looking at the logs, the CI failure looked transient:

```
2021-02-08 09:14:28,91 -not-actor-/PID:39539 esrally.rally ERROR Cannot run subcommand [race].
  File "/var/lib/jenkins/workspace/elastic+rally+pull-request/.tox/py38/lib/python3.8/site-packages/esrally/rally.py", line 857, in dispatch_sub_command
[...]
  File "/var/lib/jenkins/workspace/elastic+rally+pull-request/.tox/py38/lib/python3.8/site-packages/esrally/racecontrol.py", line 251, in race
    raise exceptions.RallyError(result.message, result.cause)
esrally.exceptions.RallyError: Worker [6] has exited prematurely.
```

There are no further details in the logs.

docs/migrate.rst
```diff
@@ -16,6 +16,30 @@ Pipelines from-sources-complete and from-sources-skip-build are removed
...
+
+Invoke Rally with the ``race`` subcommand instead::
+
+    esrally race --distribution-version=7.10.0
+
+
+Running without a track is deprecated
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Previously Rally has implicitly chosen the geonames track as default when ``--track`` was not provided. To ensure the workload is chosen deliberately, it is now deprecated to not specify a track. So instead of invoking::
```
dliappis, line 34: > To ensure the workload is chosen deliberately, 

What if we explain a bit more e.g.

> We want users to make a conscious choice of the workload and not specifying the track explicitly is deprecated (to be removed in a future version). So, instead of invoking::

esrally/rally.py
```diff
@@ -730,30 +704,183 @@ def generate(cfg):
...
+            cfg.add(config.Scope.applicationOverride, "mechanic", "preserve.install", convert.to_bool(args.preserve_install))
+            cfg.add(config.Scope.applicationOverride, "mechanic", "skip.rest.api.check", convert.to_bool(args.skip_rest_api_check))
+
+            configure_reporting_params(args, cfg)
+
+            race(cfg, args.kill_running_processes)
         elif sub_command == "generate":
+            cfg.add(config.Scope.applicationOverride, "generator", "chart.type", args.chart_type)
+            cfg.add(config.Scope.applicationOverride, "generator", "output.path", args.output_path)
+            configure_track_params(arg_parser, args, cfg)
```
dliappis, line 861: This is blocking generation.

dliappis (changes requested):
Thanks for this massive PR. Left a few comments.

docs/migrate.rst
danielmitterdorfer: Thanks; I've pushed a slighty modified version of this in https://github.com/elastic/rally/pull/1175/commits/0e8a57081661817d1f1646469a76bf512e54b4af.

esrally/rally.py
danielmitterdorfer: Fixed in https://github.com/elastic/rally/pull/1175/commits/c89aaaef84e521a6049561da50eb032ea74d6116.

dliappis (approved):
LGTM!

PR #1177, opened 2021-02-08
Use operation name consistently in metrics samples

danielmitterdorfer:
With this commit we align the value that is used for the `operation`
property in request metrics. Previously, only the `latency` samples
contained the correct operation name but `service_time` and
`processing_time` duplicated the task name instead.


ebadyano (approved):
LGTM

PR #1178, opened 2021-02-09
Always rely on source artifact caching

danielmitterdorfer:
With this commit we remove the undocumented command line parameter
`--skip-build` which was useful before Rally has implemented source
artifact caching. As Rally is aware whether an artifact needs to built,
this flag can be removed.

This is a follow-up from #1175.

Relates #1175


ebadyano (approved):
LGTM!

PR #1181, opened 2021-02-11
Remove leftover from resolver cleanup

danielmitterdorfer:
With this commit we remove one leftover reference in our tox
configuration which still mentions the 2020-resolver.

Relates #1151


dliappis (approved):
LGTM already (provided CI is green), as I tested this offline and locally and works fine (and seems that the pip version in tox's venv the latest).

Issue #1184, opened 2021-02-15
Removal of 2020-resolver enforcing breaks non developer installations


dliappis:
As described in https://github.com/elastic/rally/pull/1183#issuecomment-779094529, the removal of explicitly using the `2020-resolver` of pip in #1151 breaks installations of the latest [2.0.4 Rally release](https://github.com/elastic/rally/releases/tag/2.0.4) for several people who haven't upgraded their pip.

Note: We are requiring pip to be upgraded in [here](https://esrally.readthedocs.io/en/stable/install.html#python) but this command won't help for venv installations.

## Reproduction

<details><summary>Reproduction with Python 3.8.0</summary>

```
$ docker run --rm -ti python:3.8.0 /bin/bash

root@6c5413504db2:/# python3 --version
Python 3.8.0

root@6c5413504db2:/# python3 -mpip --version
pip 19.3.1 from /usr/local/lib/python3.8/site-packages/pip (python 3.8)

root@6c5413504db2:/# python3 -m pip install --user --upgrade pip
Collecting pip
  Downloading https://files.pythonhosted.org/packages/fe/ef/60d7ba03b5c442309ef42e7d69959f73aacccd0d86008362a681c4698e83/pip-21.0.1-py3-none-any.whl (1.5MB)
     |████████████████████████████████| 1.5MB 850kB/s 
Installing collected packages: pip
  WARNING: The scripts pip, pip3 and pip3.8 are installed in '/root/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Successfully installed pip-21.0.1
WARNING: You are using pip version 19.3.1; however, version 21.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

root@6c5413504db2:/# python3 -mpip --version
pip 21.0.1 from /root/.local/lib/python3.8/site-packages/pip (python 3.8)

root@6c5413504db2:/# python3 -m venv .venv

root@6c5413504db2:/# source .venv/bin/activate

(.venv) root@6c5413504db2:/# python3 -m pip install esrally
Collecting esrally

... several output lines snipped ...

ERROR: botocore 1.13.50 has requirement urllib3<1.26,>=1.20; python_version >= "3.4", but you'll have urllib3 1.26.3 which is incompatible.
[...]

PR #1185, opened 2021-02-15
Fix installation docs after #1151

dliappis:
With the `2020-resolver` requirement removed in #1151, installation in
virtualenv will fail due older pip versions not using the new
dependency resolver by default[1].

This commit updates the docs such that users always need to upgrade
`pip`, either in a local or `virtualenv` environments prior to esrally
installation.

[1] https://discuss.python.org/t/announcement-pip-20-3-release/5948

Relates #1184

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -27,11 +27,11 @@ We recommend to use `pyenv <https://github.com/pyenv/pyenv>`_ to manage installa
 Once ``pyenv`` is installed, install a compatible Python version::
 
     # Install Python
-    pyenv install 3.8.0
+    pyenv install 3.8.7
 
     # select that version for the current user
     # see https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-global for details
-    pyenv global 3.8.0
+    pyenv global 3.8.7
 
     # Upgrade pip
     python3 -m pip install --user --upgrade pip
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -113,16 +113,20 @@ Installing Rally
...
 
 VirtualEnv Install
 ------------------
 
 You can also use Virtualenv to install Rally into an isolated Python environment without sudo.
 
-1. Set up a new virtualenv environment in a directory with ``python3 -m venv .``
-2. Activate the environment with ``source /path/to/virtualenv/dir/bin/activate``
-3. Install Rally with ``python3 -m pip install esrally``
+1. Set up a new virtualenv environment in a directory with ``python3 -m venv .venv``
```
dliappis, line 124: One thing to note here that is different to the `--user` case is that even with e.g. python 3.8.7 with comes with `pip 21.0.1`, creating a `venv` brings pip `20.2.3`. Starting with Python `3.9.0`, there's a new flag [--upgrade-deps](https://docs.python.org/3.9/library/venv.html) that can do this in one go.

docs/install.rst
```diff
@@ -113,16 +113,20 @@ Installing Rally
...
+2. Activate the environment with ``source /path/to/virtualenv/.venv/bin/activate``
```
danielmitterdorfer, line 125: I wonder whether we could simplify this path here assuming we are already in the current directory in the step above, i.e.

```
source .venv/bin/activate
```

danielmitterdorfer (approved):
Thanks for the fix! I left one optional suggestion. LGTM.

docs/install.rst
dliappis: We could, I think we have it here because a few lines after we reference this step at a point where the user might not be in the same directory anymore:

> Whenever you want to use Rally, run the activation script (step 2 above) first.

danielmitterdorfer: Good point. Then let's be safe here.

PR #1190, opened 2021-02-18
Add support for search_after and point-in-time

### LINKED ISSUE #1141: Support Point-in-time API

DJRickyB:
This change adds support for `search_after` usage and `point-in-time` interactions.  Specifically:
* Defines `open-point-in-time` and `close-point-in-time` operations, for use in composite contexts
* Adds support to the query/search runner for `search_after` pagination, using a new operation type `paginated-search`
* Defines `scroll-search` operation type for usability
* Moves runner parsing logic to its own module for isolation
* Adds test and documentation for the above

Bonus:
* Re-factors existing query runner to reduce duplicated code
* Adds a benchmark file for demonstrating and experimenting with parsing things in the vein of `detailed-results` and the properties required for `search_after` support
* A minor documentation fix (corrected `get-async-search` to `delete-async-search`)
* Style fixes in docs with extra whitespace

Closes #1141


.pylintrc
```diff
@@ -5,7 +5,7 @@
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
 # run arbitrary code.
-extension-pkg-whitelist=
+extension-pkg-whitelist=ujson
```
DJRickyB, line 8: only needed for tests.  not actually needed since it's a benchmark test, can remove if desired

docs/track.rst
```diff
@@ -883,8 +883,13 @@ Properties
 * ``body`` (mandatory): The query body.
 * ``response-compression-enabled`` (optional, defaults to ``true``): Allows to disable HTTP compression of responses. As these responses are sometimes large and decompression may be a bottleneck on the client, it is possible to turn off response compression.
 * ``detailed-results`` (optional, defaults to ``false``): Records more detailed meta-data about queries. As it analyzes the corresponding response in more detail, this might incur additional overhead which can skew measurement results. This flag is ineffective for scroll queries.
-* ``pages`` (optional): Number of pages to retrieve. If this parameter is present, a scroll query will be executed. If you want to retrieve all result pages, use the value "all".
-* ``results-per-page`` (optional):  Number of documents to retrieve per page for scroll queries.
+* ``results-per-page`` (optional): Number of results to retrieve per page.  This maps to the Search API's ``size`` parameter, and can be used for paginated and non-paginated searches.  Defaults to ``10``
```
DJRickyB, line 886: should this just be called `size`?  there didn't seem to be a good reason to make this only for scroll/pagination given it maps to something consequential to the single request_body search

esrally/driver/runner.py
```diff
@@ -793,111 +771,147 @@ async def request_body_query(self, es, params):
...
+                if pit_op:
+                    # per the documentation the response pit id is most up-to-date
+                    CompositeContext.put(pit_op, parsed.get("pit_id"))
+
+                if results.get("hits") / size > page:
+                    body["search_after"] = last_sort
+                else:
+                    # done. clean body for next iteration
+                    for item in ["pit", "search_after"]:
+                        body.pop(item, None)
```
DJRickyB, line 813: took this route to avoid a deep copy of the body parameter.   personally i think the `query` should be the operation parameter, and the body assembled within the runner, then we wouldn't have this issue.  open to other ideas

esrally/driver/runner.py
```diff
@@ -2051,6 +2065,34 @@ def __repr__(self, *args, **kwargs):
...
+
+
+class ClosePointInTime(Runner):
+    async def __call__(self, es, params):
+        pit_op = params.get("with-point-in-time-from")
+        pit_id = CompositeContext.get(pit_op)
+        request_params = params.get("request-params", {})
+        body = {"id": pit_id}
+        await es.close_point_in_time(body=body, params=request_params, headers=None)
+        CompositeContext.remove(pit_op)
```
DJRickyB, line 2090: wait until this step to remove in case we want to make this retryable

esrally/track/params.py
```diff
@@ -732,6 +738,48 @@ def percent_completed(self):
...
 
+class OpenPointInTimeParamSource(ParamSource):
+    def __init__(self, track, params, **kwargs):
+        super().__init__(track, params, **kwargs)
+        if len(track.indices) == 1:
+            default_target = track.indices[0].name
+        elif len(track.data_streams) == 1:
+            default_target = track.data_streams[0].name
+        else:
+            default_target = None
```
DJRickyB, line 749: all this default_target stuff is copied directly from the QueryParamSource to support a default index selection.  it doesn't need to be here, or else it can be extracted to its own function, if desired

esrally/track/params.py
```diff
@@ -732,6 +738,48 @@ def percent_completed(self):
...
+        # indices are preferred but data streams can also be queried the same way
+        target_name = params.get("index")
+        type_name = params.get("type")
+        if not target_name:
+            target_name = params.get("data-stream", default_target)
```
DJRickyB, line 754: is `data-stream` intentionally undocumented for `search`? also undocumented here

setup.py
```diff
@@ -51,7 +51,7 @@ def str_from_file(name):
     # transitive dependencies:
     #   urllib3: MIT
     #   aiohttp: Apache 2.0
-    "elasticsearch[async]==7.9.1",
+    "elasticsearch[async]==7.10.1",
```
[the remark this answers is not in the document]
DJRickyB: I was under the impression the point-in-time APIs I'm using were added in elasticsearch-py 7.10.  if that ends up not being the case, I'll revert this, otherwise I'll open the (blocking) PR and discuss timing internally with the team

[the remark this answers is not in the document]
danielmitterdorfer: Sorry for not being clear enough. Yes, we need the 7.10 client for this PR. I was proposing that we raise a separate PR just to upgrade the client which we will merge prior to this PR. In the case something goes wrong the cause is easier to isolate. Does it make more sense now?

docs/track.rst
DJRickyB: can/should i deprecate `results-per-page` here?

setup.py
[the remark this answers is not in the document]
DJRickyB: Yep!

esrally/driver/runner.py
```diff
@@ -793,111 +771,147 @@ async def request_body_query(self, es, params):
...
+                    pit_id = CompositeContext.get(pit_op)
+                    body["pit"] = {"id": pit_id,
+                                   "keep_alive": "1m" } # todo parameterize second keep_alive
+
+                response = await self._raw_search(es, doc_type=None, index=index, body=body.copy(), params=request_params, headers=headers)
+                parsed, last_sort = parsing.extract_search_after_properties(response, bool(pit_op), results.get("hits"))
+                results["pages"] = page
+                results["weight"] = page
+                if results.get("hits") is None:
+                    results["hits"] = parsed.get("hits.total.value")
```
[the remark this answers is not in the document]
DJRickyB: i was under the impression `hits.total.value` was inclusive of the entire search result, and would be the same from each page

docs/track.rst
```diff
@@ -883,8 +883,13 @@ Properties
...
+* ``with-point-in-time-from`` (optional): The `name` of an ``open-point-in-time`` operation. Issues query with a `pit_id` parsed from the referenced operation.  Note that this requires usage of a ``composite`` operation containing both the ``open-point-in-time`` task and this search.  Only used when ``use-search-after`` is `True`
+
+If the following parameters are present in addition, a `paginated query <https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html>`_ will be issued:
+
+* ``pages`` (optional): Number of pages to retrieve (at most) for this search. If a query yields fewer results than the specified number of pages we will terminate earlier. To retrieve all result pages, use the value "all".
+* ``use-search-after`` (optional): If ``True``, use the search_after mechanism in the Search API.  If ``False`` but ``pages`` is defined, use the Scroll API to paginate.
```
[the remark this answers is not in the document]
DJRickyB: My concern then becomes backward compatibility.  I've been trying to think about how to better address the addition per your point as well as preserve existing scroll behavior, if possible.  If that's desirable, I'd argue what I have here is appropriate in the intermediate term.  

Perhaps we support but deprecate scrolling in the `search` operation, and add a new `scroll-search` and `paginated-search` (with description indicating it's using the search_after mechanism, as is recommended by the docs).

> I am not sure whether we should map the implementation 1:1 to the user.

for the record I think this is a fantastic idea, and it never occurred to me

esrally/driver/parsing.py
```diff
@@ -0,0 +1,70 @@
...
+                break
+    except ijson.IncompleteJSONError:
+        # did not find all properties
+        pass
+
+    parsed.update(parsed_lists)
+    return parsed
+
+
+def extract_search_after_properties(response: BytesIO, get_point_in_time: bool, hits_total):
```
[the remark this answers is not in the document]
DJRickyB: I'm going to move them back.  This was originally motivated by a desire to eventually move more parsing to purpose-built extraction mechanisms but in practice they did not yield a huge hit to use the ijson parser generically (as evidenced by the fact that I use `parse()` inside `extract_search_after_properties()`

docs/track.rst
```diff
@@ -927,6 +924,122 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
...
+paginated-search
+~~~~~~~~~~~~~~~~
+
+With the operation type ``paginated-search`` you can execute `paginated searches <https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#search-after>`_, specifically using the ``search_after`` mechanism.
+
+Properties
+""""""""""
+
+* ``index`` (optional): An `index pattern <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_ that defines which indices or data streams should be targeted by this query. Only needed if the ``indices`` or ``data-streams`` section contains more than one index or data stream respectively. Otherwise, Rally will automatically derive the index or data stream to use. If you have defined multiple indices or data streams and want to query all of them, just specify ``"index": "_all"``.
+* ``type`` (optional): Defines the type within the specified index for this query. By default, no ``type`` will be used and the query will be performed across all types in the provided index. Also, types have been removed in Elasticsearch 7.0.0 so you must not specify this property if you want to benchmark Elasticsearch 7.0.0 or later.
```
danielmitterdorfer, line 936: Given that `search_after` is only available in 7.x and types are not available we should not expose this parameter.

docs/track.rst
```diff
@@ -927,6 +924,122 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
...
+* ``cache`` (optional): Whether to use the query request cache. By default, Rally will define no value thus the default depends on the benchmark candidate settings and Elasticsearch version.
+* ``request-params`` (optional): A structure containing arbitrary request parameters. The supported parameters names are documented in the `Search URI Request docs <https://www.elastic.co/guide/en/elasticsearch/reference/current/search-uri-request.html#_parameters_3>`_.
+
+    .. note::
+        1. Parameters that are implicitly set by Rally (e.g. `body` or `request_cache`) are not supported (i.e. you should not try to set them and if so expect unspecified behavior).
```
danielmitterdorfer, line 941: Nit: rst requires double-backticks here for `body` and `request_cache`.

docs/track.rst
```diff
@@ -927,6 +924,122 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
...
+"""""""""
+
+The following meta data are always returned:
+
+* ``weight``: "weight" of an operation. Always 1 for regular queries and the number of retrieved pages for scroll queries.
+* ``unit``: The unit in which to interpret ``weight``. Always "ops" for regular queries and "pages" for scroll queries.
+* ``success``: A boolean indicating whether the query has succeeded.
+* ``hits``: Total number of hits for this query.
+* ``hits_relation``: whether ``hits`` is accurate (``eq``) or a lower bound of the actual hit count (``gte``).
+* ``timed_out``: Whether the query has timed out. For scroll queries, this flag is ``true`` if the flag was ``true`` for any of the queries issued.
```
danielmitterdorfer, line 986: This mentions scrolls when you want actually mean the point in time API? As we have a dedicated section for this API, I don't think we need to qualify this statement anyway. How about we simply say:

> ``timed_out``:  Whether any of the any of the queries issued has timed out.

docs/track.rst
```diff
@@ -927,6 +924,122 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
...
```
danielmitterdorfer, line 1040: See my earlier comment in the section about point in time queries.

docs/track.rst
```diff
@@ -927,6 +924,122 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
...
+* ``took``` Value of the the ``took`` property in the query response. For scroll queries, this value is the sum of all ``took`` values in query responses.
```
danielmitterdorfer, line 987: Similar to above.

docs/track.rst
```diff
@@ -927,6 +924,122 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
...
```
danielmitterdorfer, line 1041: Similar to above.

esrally/driver/runner.py
```diff
@@ -912,12 +955,38 @@ async def scroll_query(es, params):
...
+
+            parsed = parse(response, properties)
+
+            # standardize these before returning...
+            parsed["hits.total.value"] = parsed.pop("hits.total.value", parsed.pop("hits.total", 0))
+            parsed["hits.total.relation"] = parsed.get("hits.total.relation", "eq")
+
+            response_str = response.getvalue().decode("UTF-8")
+            index_of_last_sort = response_str.rfind('"sort"')
+            sort_pattern = r"sort\":([^\]]*])"
```
danielmitterdorfer, line 974: Can we turn this into a class variable and compile it with `re.compile`? Can you please also add documentation what's the intent behind this regex and how to read it so it's easier to maintain this in the future? 

I think it would also be great to extract `_extract_search_after_properties` into a top-level function and add dedicated tests for it. This would also make it easier to understand the possible inputs and cases that you're covering here.

If you do that, then you need to somehow reference the compiled regex, so you could implement it as a small class. I'm thinking about something along the lines of this:

```python
class SearchAfterExtractor:
    sort_pattern = re.compile(...)

    def __call__(response: BytesIO, get_point_in_time: bool, hits_total: int) -> (dict, List):
        ...

# reference this then in the Query runner via:
class Query:
    def __init__(...):
        self.extract_search_after_properties = SearchAfterExtractor()
```

esrally/driver/runner.py
```diff
@@ -912,12 +955,38 @@ async def scroll_query(es, params):
...
+            if last_sort_str is not None:
+                last_sort = json.loads(last_sort_str.group(1))
+            else:
+                last_sort = None
+            return parsed, last_sort
+
+        search_method = params.get("operation-type")
+        if search_method == "paginated-search":
+            return await _search_after_query(es, params)
+        # also allow if pages is defined, for backward compatiblity
```
danielmitterdorfer, line 985: I think we should deprecate the old approach after a grace period and log a warning here. Can you please raise an issue for this?

esrally/track/track.py
```diff
@@ -631,7 +632,7 @@ def from_hyphenated_string(cls, v):
             return OperationType.IndexStats
         elif v == "node-stats":
             return OperationType.NodeStats
-        elif v == "search":
+        elif v in ["search", "scroll-search", "paginated-search"]:
```
danielmitterdorfer, line 635: I think this is leaking an implementation detail (one runner class for several operation types) into other parts of the code. How about we instead register the `Query` runner for all three operation types? See:

https://github.com/elastic/rally/blob/35b1fbab6cf35d9e41b6b3db50af1e3a4ce0e424/esrally/driver/runner.py#L45

danielmitterdorfer (commented):
Thanks for iterating. I did another pass through the code and left a couple of suggestions.

esrally/driver/runner.py
DJRickyB: great idea

esrally/driver/runner.py
DJRickyB: deprecating now, opened #1215 for removal

esrally/track/track.py
DJRickyB: looks like i got halfway to doing that then forgot it was happening haha

esrally/driver/runner.py
```diff
@@ -793,111 +771,147 @@ async def request_body_query(self, es, params):
...
+            index = params.get("index", "_all")
+            pit_op = params.get("with-point-in-time-from")
+            results = {
+                "unit": "pages",
+                "success": True,
+                "took": 0
+            }
+            if pit_op:
+                # these are disallowed as they are encoded in the pit_id
+                for item in ["index", "routing", "preference"]:
```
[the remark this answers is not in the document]
DJRickyB: discussed offline, we will not go this route

tests/driver/runner_test.py
```diff
@@ -4748,6 +4748,309 @@ async def test_delete_async_search(self, es):
...
+        r = runner.Query()
+
+        async with runner.CompositeContext():
+            runner.CompositeContext.put(pit_op, pit_id)
+            await r(es, params)
+            # make sure pit_id is updated afterward
+            self.assertEqual("fedcba9876543211", runner.CompositeContext.get(pit_op))
+
+        es.transport.perform_request.assert_has_calls([mock.call('GET', '/_search', params={},
+                                                                 body={'query': {'match-all': {}},
```
danielmitterdorfer, line 4874: Nit: Personally, I find the following formatting easier to read:

```python
 body={
     'query': {
         'match-all': {}
     },
       'sort': [{
           'timestamp': 'asc',
           'tie_breaker_id': 'asc'
       }],
       'size': 2,
       'pit': {
           'id': '0123456789abcdef',
           'keep_alive': '1m'
       }
 },
```

I'm fine if we keep it as is but I want to offer it as a thought.

danielmitterdorfer (commented):
Thanks for iterating! I left some minor suggestions and corrections on the docs and some formatting. `SearchAfterExtractor` looks good but I think we can use one shared instance per runner.

docs/track.rst
```diff
@@ -2464,6 +2581,90 @@ Meta-data
...
+With the ``close-point-in-time`` operation you can close a previously opened point in time from another task.
+
+Properties
+""""""""""
+
+* ``with-point-in-time-from``: Denotes the ``name`` of a given ``open-point-in-time`` operation (from the same composite stream) whose `pit_id` should be closed.
+
+**Example**
+
+In this example, a point-in-time is opened, used by a ``search_after``-based search operation, and closed::
```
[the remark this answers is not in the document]
DJRickyB: i was going for a "this is what we are doing to elasticsearch" description rather than a "this is what we are doing in rally" description, but I take this hesitation as sufficient evidence it should be changed (so it was)

danielmitterdorfer (approved):
Thanks for iterating. Looks good now! :)

PR #1192, opened 2021-02-19
Fix compare subcommand

dliappis:
The refactoring in #1175 broken the compare subcommand due to reliance
on the `show-in-report` cli argument. As prior to the PR the subcommand
parsing logic was obscure the problem surfaced now.

This commit defaults the parameter `show-in-report` for the compare
subcommand to True. This parameter could be removed in the future,
the reason for adding it in the past was that some users relied on
parsing the report file and the comparison adds fields
(like percentiles) dynamically.

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -198,6 +198,10 @@ def create_arg_parser():
         "--report-file",
         help="Write the command line report also to the provided file.",
         default="")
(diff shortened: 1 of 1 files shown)


danielmitterdorfer (approved):
LGTM

Issue #1195, opened 2021-02-23
Support Query Ramp-up


gingerwizard:
Currently, after a warmup period, all query tasks are executed together at the same time. After the first execution, they respect their configured scheduler for subsequent executions.

In cases with a high number of tasks (e.g. executing using a parallel block) this absence of a ramp up, can produce an initial extremely high load and result in excessive utilization. This is best illustrated through an example:

![image](https://user-images.githubusercontent.com/12695796/108863295-f482b000-75e8-11eb-8c89-405a1e341e53.png)

Even considering the warmup period (darker in the first visual) it's apparent we suffer from a spike before the model stabilizes. 

We should consider both short term improvements here e.g. maybe tasks call their scheduler before executing the first time?, 
and longer-term possible improvements e.g. a ramp-up period.

@danielmitterdorfer

danielmitterdorfer:
> We should consider both short term improvements here e.g. maybe tasks call their scheduler before executing the first time

I had a look at this proposal and the change for this is actually pretty small:

https://github.com/elastic/rally/blob/387596872c2e4724f37d7890eaf6d10e422f264e/esrally/driver/driver.py#L1742-L1771

We basically need to call the scheduler's `next` method before the generator yields insted of afterwards.

gingerwizard:
This maybe has the potential to cause a deviation in nightly performance @danielmitterdorfer if the number of requests isn't high - especially around our 95th and 99th percentile. If anything it might bring greater stability to the results.

danielmitterdorfer:
I don't think that it would cause a deviation in the nightlies for most tracks because we only test with a single client and the effect only shows up with multiple clients. But for more complex benchmarks this is more likely an issue.

PR #1199, opened 2021-02-24
Throttle requests from the beginning

danielmitterdorfer:
With this commit we consult the scheduler prior to issuing a request
instead of afterwards. If we don't do that, clients can coordinate
creating a large initial load spike in Elasticsearch. With this
countermeasure, it is possible that clients avoid this initial spike if
a non-determistic scheduler, such as the poisson scheduler is chosen.

Relates #1195

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -4,6 +4,11 @@ Migration Guide
 Migrating to Rally 2.1.0
 ------------------------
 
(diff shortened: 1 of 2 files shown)


gingerwizard (approved):
LGTM. Maybe worth noting in the migration guide that this may cause different results where large number of clients are used?

PR #1203, opened 2021-03-09
Log results of cluster health check

### LINKED ISSUE #1150: Log results of cluster health check
Usually the cluster health operation is run in a loop until it reaches the expected status. This is problematic for cases when cluster health `green` is expected and a the benchmark creates an index with one replica on a single node cluster because the benchmark will hang indefinitely without giving an indication in the logs. In order to help users diagnose this case, we should add a log statement every time cluster health is checked, indicating the expected and the actual status.

danielmitterdorfer:
When the cluster health check fails e.g. due to invalid setup where we
check for a green cluster on a single node but indices have been
configured with replicas, the benchmark appears to hang indefinitely.
The reason is that the cluster health check is retried by default but
there is no indication in the logs.

With this commit we log the result of the cluster health check. We also
raise the log level of retry log messages so users can analyze the
behavior by inspecting logs.

Closes #1150

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -967,13 +967,16 @@ class ClusterHealth(Runner):
         cluster_status = result["status"]
         relocating_shards = result["relocating_shards"]
 
-        return {
+        result = {
             "weight": 1,
             "unit": "ops",
             "success": status(cluster_status) >= status(expected_cluster_status) and relocating_shards <= expected_relocating_shards,
             "cluster-status": cluster_status,
             "relocating-shards": relocating_shards
         }
+        self.logger.info("%s: expected status=[%s], actual status=[%s], relocating shards=[%d], success=[%s].",
+                         repr(self), expected_cluster_status, cluster_status, relocating_shards, result["success"])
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
Here are some examples of the behavior from my local testing:

In the success case we see in the logs:

```
PID:10002 esrally.driver.runner INFO cluster-health: expected status=[green], actual status=[green], relocating shards=[0], success=[True].
```

When the cluster health status is not reached, we get:

```
PID:11154 esrally.driver.runner INFO [cluster-health] has timed out. Retrying in [0.50] seconds.
```

and additionally, the Python client issues a warning right before the log line above:

```
PID:11154 elasticsearch WARNING GET http://127.0.0.1:39200/_cluster/health/logs-*?wait_for_status=green&wait_for_no_relocating_shards=true [status:408 request:30.006s]
```

DJRickyB (approved):
This is great, thank you.  Upping the log level for retry is a good idea

danielmitterdorfer:
Thanks for the review!

PR #1207, opened 2021-03-11
Expose an API to handle file offset tables

### LINKED ISSUE #1204: Api for updating offset tables
We have instances where custom tracks need to update and maintain offset tables for files, to allow fast seeking. Rally supports a method `prepare_file_offset_table` but this isn't exposed for external use. Any support should probably allow the addition of offsets and possibly the seeking into the parent file based on iterating over the offsets.

danielmitterdorfer:
With this commit we encapsulate handling of file offset tables for data
files in a dedicated class. This class implements the necessary
low-level handling only exposing a higher-level API for consumers.

Closes #1204

### DIFF
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -406,6 +406,109 @@ def has_extension(file_name, extension):
     return ext == extension
 
 
+class FileOffsetTable:
+    """
+    The FileOffsetTable represents a persistent mapping from lines in a data file to their offset in bytes in the
+    data file. This helps bulk-indexing clients to advance quickly to a certain position in a large data file.
+    """
+    def __init__(self, data_file_path, offset_table_path, mode):
+        """
+        Creates a new FileOffsetTable instance. The constructor should not be called directly but instead the
+        respective factory methods should be used.
+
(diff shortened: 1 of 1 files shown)


esrally/utils/io.py
```diff
@@ -406,6 +406,109 @@ def has_extension(file_name, extension):
...
+    """
+    def __init__(self, data_file_path, offset_table_path, mode):
+        """
+        Creates a new FileOffsetTable instance. The constructor should not be called directly but instead the
+        respective factory methods should be used.
+
+        :param data_file_path: The absolute path to the data file. This file is assumed to exist at this point.
+        :param offset_table_path: The absolute path to the corresponding offset table file. Only required to exist
+                                  for read operations on the data file.
+        :param mode: The mode in which the file offset table should be opened.
```
gingerwizard, line 422: Can we not infer this? If an offset file is provided we can always open for read?

gingerwizard: never mind this will never be called directly.

gingerwizard (approved):
LGTM

PR #1208, opened 2021-03-11
Remove unreferenced code

danielmitterdorfer:
With this commit we remove some code in the `io` module that has only
been used in tests.


ebadyano (approved):
LGTM, thank you

PR #1209, opened 2021-03-15
Ensure stable mappings for metrics

danielmitterdorfer:
With this commit we disable dynamic detection of dates for documents in
the metrics store. By disabling this feature, any dynamic values (like
track parameters or user tags) that look like dates are mapped as
keywords instead of as dates. These values should only be treated as
keywords and are not meant to be used e.g. in range queries.

Furthermore, this ensures that we don't cause trouble when date formats
conflict in such dynamic values. Note that properties that are
explicitly mapped as dates (e.g. `@timestamp`) are not affected by this
change and will still be mapped as dates.

### DIFF
--- a/esrally/resources/metrics-template.json
+++ b/esrally/resources/metrics-template.json
@@ -8,6 +8,7 @@
   },
   "mappings": {
     "_doc": {
+      "date_detection": false,
       "dynamic_templates": [
         {
           "strings": {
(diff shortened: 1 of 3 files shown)


dliappis (commented):
Thank you, yay for less unpredictability in data types.

One question:

For existing monthly indices that have already ingested, say, a track param as a date, will this change break things due to type conflicts?

danielmitterdorfer:
> For existing monthly indices that have already ingested, say, a track param as a date, will this change break things due to type conflicts?

No, because this only changes the index template which applies only to newly created indices. So the change then applies only to the new index.

dliappis (approved):
LGTM

PR #1214, opened 2021-03-17
Collect metrics on composite subtasks

### LINKED ISSUE #1173: Generate meta documents for search operations in composite
Currently metrics are reported for composite queries at the level of the operation. This makes identifying performance of the actual contained search requests impossible. It is proposed that a metrics document be generated for every search operation. The current meta data is sufficient.  We may want to require this to be actively enabled by the user given its potential to generate a large number of documents.

danielmitterdorfer:
Previously, Rally could only measure request timings on top-level
requests. This is usually what is want but we introduced `composite`
tasks recently, which allow to issue multiple concurrent requests in the
context of one task. It was not possible to measure the service time of
these requests so far.

With this commit we introduce a notion of "dependent timings" that can
be attached to measurement samples of a task. These are stored as
additional service time samples in the metrics store and can be used to
analyze behavior in more detail.

We also introduce a slight change in how "absolute time" and "relative
time" need to be interpreted. Previously, these times represented the
absolute and relative point in time when the request sample has been
created (which happens *after* a response has been received). Now, these
represent the point in time when the request has been sent to
Elasticsearch which is more meaningful and also simplifies correlating
timestamps from other tools such as tcpdump.

Closes #1173

### DIFF
--- a/esrally/client.py
+++ b/esrally/client.py
@@ -26,6 +26,87 @@ from esrally import exceptions, doc_link
 from esrally.utils import console, convert
 
 
+class RequestContextManager:
+    """
+    Ensures that request context span the defined scope and allow nesting of request contexts with proper propagation.
+    This means that we can span a top-level request context, open sub-request contexts that can be used to measure
+    individual timings and still measure the proper total time on the top-level request context.
+    """
+    def __init__(self, client):
+        self.client = client
+        self.ctx = None
+        self.token = None
+
+    async def __aenter__(self):
+        self.ctx, self.token = self.client.init_request_context()
+        return self
+
+    @property
+    def request_start(self):
+        return self.ctx["request_start"]
+
+    @property
+    def request_end(self):
+        return self.ctx["request_end"]
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        # propagate earliest request start and most recent request end to parent
+        request_start = self.request_start
+        request_end = self.request_end
+        self.client.restore_context(self.token)
+        # don't attempt to restore these values on the top-level context as they don't exist
+        if self.token.old_value != contextvars.Token.MISSING:
+            self.client.update_request_start(request_start)
+            self.client.update_request_end(request_end)
+        self.token = None
+        return False
+
+
+class RequestContextHolder:
+    """
+    Holds request context variables. This class is only meant to be used together with RequestContextManager.
+    """
+    request_context = contextvars.ContextVar("rally_request_context")
+
+    def new_request_context(self):
+        return RequestContextManager(self)
+
+    @classmethod
+    def init_request_context(cls):
+        ctx = {}
+        token = cls.request_context.set(ctx)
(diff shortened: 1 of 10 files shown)


danielmitterdorfer:
Here's a screenshot of the breakdown of requests in Kibana's discover view that shows the individual request timings:

request-timings

Note how the timeline starts with the top-level (composite) request and then sub-request are added with their correct timestamps as the operations are executed.

I've used the following track for this:

```json
{
  "schedule": [
    {
      "name": "sleep-a-while",
      "operation": {
        "operation-type": "composite",
        "requests": [
          {
            "stream": [
              {
                "operation-type": "sleep",
                "name": "sleep-a1",
                "duration": 1
              },
              {
                "operation-type": "sleep",
                "name": "sleep-a2",
                "duration": 1
              },
              {
                "operation-type": "sleep",
                "name": "sleep-a3",
                "duration": 1
              },
              {
                "operation-type": "sleep",
                "name": "sleep-a4",
                "duration": 1
              }
            ]
          },
          {
            "stream": [
              {
                "operation-type": "sleep",
                "name": "sleep-b1",
                "duration": 0.5
              },
              {
                "operation-type": "sleep",
                "name": "sleep-b2",
                "duration": 1
              },
              {
                "operation-type": "sleep",
                "name": "sleep-b3",
                "duration": 1
              },
              {
                "operation-type": "sleep",
                "name": "sleep-b4",
                "duration": 1
              }
            ]
          }

        ]
      }
    }
  ]
}
```

which can be run with Rally using (the Elasticsearch version doesn't really matter as it's only doing sleep operations):

```
esrally race --track-path=composite_test.json --distribution-version=7.10.0 --on-error=abort
```

esrally/client.py
```diff
@@ -26,6 +26,87 @@
...
 
 
+class RequestContextManager:
+    """
+    Ensures that request context span the defined scope and allow nesting of request contexts with proper propagation.
+    This means that we can span a top-level request context, open sub-request contexts that can be used to measure
+    individual timings and still measure the proper total time on the top-level request context.
+    """
+    def __init__(self, client):
+        self.client = client
```
gingerwizard, line 36: Initially unclear to me that client here is a ref to the RequestContextHolder. A small comment would of made this easier to follow

gingerwizard (approved):
This code touches a lot of things that took me a while to fully understand - even now I'm not 100% comfortable. I don't see any glaring issues so LGTM. I would like to see higher-level tests that exercise timing collection for composite - this would be useful for following the code in execution. Right now unit tests are quite low level as far as I can tell. Maybe we could enhance one of the Composite tests?

danielmitterdorfer:
> I would like to see higher-level tests that exercise timing collection for composite 

Thanks for the feedback. I've added a test in https://github.com/elastic/rally/pull/1214/commits/710d9213bffb20a08b90ce13ea56e7566f54003c.

danielmitterdorfer:

PR #1217, opened 2021-03-23
Fix auto build of docs

dliappis:
PR #1193 introduced a bug with readthedocs.org in relying on presenting
the MIN_PY_VER env var via the Makefile. However the Makefile isn't
invoked during automated build of readthedocs.

This commit fixes automatic builds of our docs again.


danielmitterdorfer (approved):
Sweet and simple. Thanks!

Issue #1197, opened 2021-02-24
Store total time of task in rally-results metrics record


dliappis:
Currently, after executing a benchmark it's not so easy to get a breakdown of the total duration per task.
It can be calculated indirectly from the detailed `rally-metrics` index (when using a dedicated metrics store, but this ticket focuses on that case) by e.g. having a breakdown per task for for each task getting the maximum of `relative-time`.

Instead we should push an additional metrics record in `rally-results` called `duration` that captures the total time the task took to complete.

Issue #1198, opened 2021-02-24
Use milliseconds instead of microseconds for relative time


dliappis:
When using an Elasticsearch metrics store, metrics stored in `rally-metrics` contain a `relative-time` key in every record (as this starts from `0` it's useful for comparisons).
The unit of this metric is μs (microseconds) and as such unnecessarily granular and trappy in visualizations and data analysis.

We should change this to `ms` (milliseconds).

As per #1221 the schedule is:

- [x] Rally 2.1.0: `relative-time` is deprecated and Rally adds a new field `relative-time-ms` which contains the relative time in milliseconds. Done in #1221.
- [x] Rally 2.2.0: `relative-time` will be dropped. Rally only populates the field `relative-time-ms` which contains the relative time in milliseconds. Done in #1231.
- [x] Rally 2.2.1: `relative-time` will be reintroduced and contain the relative time in milliseconds. The field `relative-time-ms` will be deprecated. Done in https://github.com/elastic/rally/pull/1273.
- [x] Rally 2.3.0: `relative-time-ms` will be dropped. Done in https://github.com/elastic/rally/pull/1291.

PR #1220, opened 2021-03-26
Store duration time of task in rally-results metrics record

ebadyano:
Closes: https://github.com/elastic/rally/issues/1197


esrally/metrics.py
```diff
@@ -1673,6 +1675,11 @@ def total_transform_metric(self, metric_name):
     def error_rate(self, task_name, operation_type):
         return self.store.get_error_rate(task=task_name, operation_type=operation_type, sample_type=SampleType.Normal)
 
+    def duration(self, task_name):
+        values = self.store.get_raw("service_time", task_name, mapper=lambda doc: doc["relative-time"])
```
danielmitterdorfer, line 1679: This can potentially return millions of documents (e.g. in a benchmark with thousands of clients) and we're only interested in one value. I wonder whether this would justify a specialized query method in the metrics store as it would be much cheaper only to retrieve the maximum value directly?

I also want to point out that `relative-time` is determined *before* a request is issued:

https://github.com/elastic/rally/blob/62595bde7319b77895bc8c5b726c9cd298daef30/esrally/driver/driver.py#L1060-L1062

So strictly speaking `duration` does not account for the service time of the last request. If this is a long-running request (e.g. in frozen tier benchmarks) the duration measurement could be several minutes off. This could be specifically misleading if there is only one measurement iteration as `duration` would report something in order of several milliseconds when the actual duration would more be like several minutes.

So if we settle on a special query method, I propose that the API allows us to retrieve one sample ordered by a criterion. E.g. we could amend the current `get_one` method in the metrics store:

https://github.com/elastic/rally/blob/62595bde7319b77895bc8c5b726c9cd298daef30/esrally/metrics.py#L619

to allow sorting:

```python
def get_one(self, name, sample_type=None, node_name=None, task=None, sort_key=None, sort_reverse=False):
```

_Note: I've borrowed the names of the new parameters `sort_key` an `sort_reverse` from Python's [`sorted`](https://docs.python.org/3/library/functions.html#sorted) function for an idiomatic API._

What do you think?

esrally/metrics.py
```diff
@@ -1821,6 +1830,7 @@ def add_op_metrics(self, task, operation, throughput, latency, service_time, pro
             "service_time": service_time,
             "processing_time": processing_time,
             "error_rate": error_rate,
+            "duration_time": duration
```
danielmitterdorfer, line 1833: I'd argue that `duration` by itself conveys its meaning already and the `_time` suffix is redundant? (see also https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/time/Duration.html)

danielmitterdorfer (commented):
Thanks for the PR! That's very helpful. I left a couple of suggestions. 

I also think that a prerequisite for this is https://github.com/elastic/rally/issues/1198 so we have consistent time scale. Otherwise all measurements would be in milliseconds except for relative-time, which would be denoted in microseconds.

danielmitterdorfer:
> I also think that a prerequisite for this is #1198 so we have consistent time scale.

I've pushed https://github.com/elastic/rally/pull/1221 now and we should base the duration on the newly introduced field `relative-time-ms` in this PR here.

esrally/metrics.py
ebadyano: Sounds good, I will use that, thank you for the suggestion!

ebadyano: @danielmitterdorfer interestingly the current implementation for get_one uses get and the just returns values[0] so it would still return all the values and then pick the first one locally.
https://github.com/elastic/rally/blob/62595bde7319b77895bc8c5b726c9cd298daef30/esrally/metrics.py#L633
 Was your idea to modify the search query so that it would only return one value?

danielmitterdorfer: Yes, we should implement `get_one` in both subclasses (`InMemoryMetricsStore` and `EsMetricsStore`) and remove the generic implementation in the base class.

esrally/metrics.py
```diff
@@ -825,6 +828,29 @@ def _get(self, name, task, operation_type, sample_type, node_name, mapper):
         self.logger.debug("Metrics query produced [%s] results.", result["hits"]["total"])
         return [mapper(v["_source"]) for v in result["hits"]["hits"]]
 
+    def get_one(self, name, sample_type=None, node_name=None, task=None, mapper=lambda doc: doc["value"],
+                sort_key=None, sort_reverse=False):
+        order = "desc"
```
danielmitterdorfer, line 833: Maybe this would be simpler?

```python
order = "desc" if sort_reverse else "asc"
```

esrally/metrics.py
```diff
@@ -825,6 +828,29 @@ def _get(self, name, task, operation_type, sample_type, node_name, mapper):
...
+        if not sort_reverse:
+            order = "asc"
+        if sort_key:
```
danielmitterdorfer, line 836: Alternatively you could implement this block as follows:

```python
query = {
    "query": self._query_by_name(name, task, None, sample_type, node_name),
    "size": 1
}
if sort_key:
    query["sort"] = [{sort_key: {"order": order}}]
```

This avoids repetition of common elements.

esrally/metrics.py
```diff
@@ -825,6 +828,29 @@ def _get(self, name, task, operation_type, sample_type, node_name, mapper):
...
+            }
+        else:
+            query = {
+                "query": self._query_by_name(name, task, None, sample_type, node_name),
+                "size": 1
+            }
+        self.logger.debug("Issuing get against index=[%s], query=[%s].", self._index, query)
+        result = self._client.search(index=self._index, body=query)
+        self.logger.debug("Metrics query produced [%s] results.", result["hits"]["total"])
+        return mapper(result["hits"]["hits"][0]["_source"])
```
danielmitterdorfer, line 852: This would fail with a key error if there are no hits. Can we check this and return `None` if there are no hits?

esrally/metrics.py
```diff
@@ -1069,6 +1095,18 @@ def _get(self, name, task, operation_type, sample_type, node_name, mapper):
                 (node_name is None or doc.get("meta", {}).get("node_name") == node_name)
                 ]
 
+    def get_one(self, name, sample_type=None, node_name=None, task=None, mapper=lambda doc: doc["value"],
+                sort_key=None, sort_reverse=False):
+        if sort_key:
+            docs = sorted(self.docs, key=lambda k: k[sort_key], reverse=sort_reverse)
+        else:
+            docs = self.docs
+        for doc in docs:
```
danielmitterdorfer, line 1104: I think we need to `return None` if there are no hits (i.e. after the for loop).

tests/metrics_test.py
```diff
@@ -1583,6 +1585,7 @@ def test_calculate_global_stats(self):
         self.assertEqual(collections.OrderedDict(
             [("50_0", 200), ("100_0", 210), ("mean", 200), ("unit", "ms")]), opm["service_time"])
         self.assertAlmostEqual(0.3333333333333333, opm["error_rate"])
+        self.assertAlmostEqual(709*1000, opm["duration"])
```
danielmitterdorfer, line 1588: Shouldn't we able to use `assertEqual` here as this is an integer value?

tests/metrics_test.py
```diff
@@ -1595,6 +1598,7 @@ def test_calculate_global_stats(self):
         self.assertEqual(17.2, stats.ml_processing_time[0]["median"])
         self.assertEqual(36.0, stats.ml_processing_time[0]["max"])
         self.assertEqual("ms", stats.ml_processing_time[0]["unit"])
+        self.assertAlmostEqual(600*1000, opm2["duration"])
```
danielmitterdorfer, line 1601: Shouldn't we able to use `assertEqual` here as this is an integer value?

danielmitterdorfer (commented):
Thanks for iterating! I left a couple more comments. Can you please also implement tests for `get_one` for both metrics store implementations (including tests for handling an empty search result)?

tests/metrics_test.py
ebadyano: Opps thank you for catching

tests/metrics_test.py
```diff
@@ -1290,6 +1397,40 @@ def tearDown(self):
...
+
+        self.metrics_store.close()
+
+        self.metrics_store.open(InMemoryMetricsStoreTests.RACE_ID, InMemoryMetricsStoreTests.RACE_TIMESTAMP,
+                                "test", "append-no-conflicts", "defaults")
+
+        actual_duration = self.metrics_store.get_one("service_time", task="task1", mapper=lambda doc: doc["relative-time-ms"],
+                                                     sort_key="relative-time-ms", sort_reverse=True)
+
+        self.assertEqual(None, actual_duration)
```
danielmitterdorfer, line 1432: We can use `self.assertIsNone` here.

danielmitterdorfer (approved):
Thanks for iterating! I left one comment but no need for another review round. LGTM

ebadyano:
Thank you for the review @danielmitterdorfer

PR #1221, opened 2021-03-26
Store relative-time in milliseconds

danielmitterdorfer:
Previously `relative-time` has denoted the relative time in microseconds
since the start of a task. This is inconvenient as all other times are
denoted in milliseconds. Therefore, we will migrate this field to be
denoted in milliseconds as well. In the first step we deprecate the
current field and introduce a new field `relative-time-ms` that is used
for the transition period.

Relates #1198

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -15,6 +15,7 @@ Here is a typical metrics record::
           "race-id": "6ebc6e53-ee20-4b0c-99b4-09697987e9f4",
           "@timestamp": 1461213093093,
           "relative-time": 10507328,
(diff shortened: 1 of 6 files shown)


danielmitterdorfer:
I did not mark this PR as breaking as Rally will continue to work as before although `relative-time` is now deprecated. The newly introduced field `relative-time-ms` will just be added to new metrics documents, even if they are added to an existing index.

dliappis (approved):
LGTM.

As discussed elsewhere, we could consider adding a `unit` field in a future PR.

PR #1228, opened 2021-03-31
Natively support parallelized track preparation

### LINKED ISSUE #1166: Natively support parallelized track preparation
The new track processor API implemented https://github.com/elastic/rally/pull/1135 supports two customize two steps in the track lifecycle:

1. `on_after_load_track`
2. `on_prepare_track`

While the first phase doesn't take a lot of time, the second one might take up more resources as many GB of data may be downloaded, extracted or generated. Both steps are currently implemented as a single-threaded, blocking operation. We should support that `on_prepare_track` can leverage multiple cores.

### Requirements

* The implementation should leverage the actor system for concurrency, not multithreading (which would be ineffective in Python on a CPU-bound problem).
* The available level of parallelism should be determined based on the number of CPU cores (there is already support for this in the `Driver` actor).
* It should be transparent to track processor implementations that they are running inside an actor system.

### Hints / Implementation notes

* Track preparation is handled entirely by the `TrackPreparationActor`

DJRickyB:
In this commit we change the flow of the undocumented `on_prepare_track` API to utilize separate actors for processing, instead of a thread pool.

In doing so, the interface of `on_prepare_track` has changed to return an iterable of tuple `(Callable, dict)`, where the `dict` represents named arguments to the given Callable. Track processors which implement this method are expected to yield function/parameter combinations that are effectively partitions of work.  On each load driver machine, the `TrackPreparationActor` will spin up a number of `TaskExecutionActor`s equal to the number of available CPUs. Then, each `TaskExecutionActor` will request a unit of work (an instance of `WorkerTask`, which encapsulates the Callable+parameters) and execute it, running in a loop until instructed to be idle with an empty task message, and then to shut down when there are no further track processors to execute.

closes #1166


DJRickyB:
Two big things I want to accomplish next:
* figure out a good way to stand up unit testing for our actors
* (related) merge the functionality of the `driver.TaskExecutionActor` and the similar existing `driver.Worker` Actor so that the generic nature of delegating tasks to simple actors that background them is maintained in one place

esrally/driver/driver.py
```diff
@@ -313,42 +336,157 @@ def load_local_config(coordinator_config):
...
+                               force_update=True)
+
+        if tpr.processors():
+            # do this locally...
+            for processor in tpr.processors():
+                processor.on_after_load_track(self.track)
+            # ...and delegate the rest
+            self.children = [self._create_task_worker() for _ in range(num_cores(self.cfg))]
+            for processor in tpr.processors():
+                self.processors.put(processor)
```
danielmitterdorfer, line 448: I am not certain why we can't enqueue all tasks upfront here? IMHO that would simplify the interaction model between this actor and its children. Every time we receive a `ReadyForWork` message we'd reply with another task until the queue is drained. We then need to reply with `TrackPrepared` to the original sender once all children have finished their outstanding work (for which we'd yet need to implement state handling logic).

An alternative would be to send all tasks round-robin to children, which would queue them themselves (in their thread pool's queue) and only respond when they're done with all submitted tasks. This would simplify state handling in this actor because we could rely on the base class implementation of `send_to_children_and_transition` and only need to send `TrackPrepared` from the callback method (which is invoked once all children have finished their tasks).

esrally/driver/driver.py
```diff
@@ -26,11 +26,13 @@
 import queue
 import threading
 import time
+from enum import Enum
 
 import thespian.actors
 
 from esrally import actor, config, exceptions, metrics, track, client, paths, PROGRAM_NAME, telemetry
 from esrally.driver import runner, scheduler
+from esrally.track.loader import TrackProcessorRegistry, load_track_plugins
```
danielmitterdorfer, line 35: Can we expose these two in the `__init__.py` of `esrally.track` so that the (low-level) `loader` submodule is not directly used?

esrally/driver/driver.py
```diff
@@ -313,42 +336,157 @@ def load_local_config(coordinator_config):
...
+        new_msg = DoTask(next_task, self.cfg)
+        self.logger.debug("Track Preparator sending %s to %s", vars(new_msg), sender)
+        self.send(sender, new_msg)
+
+    def receiveMsg_WorkerIdle(self, msg, sender):
+        self.transition_when_all_children_responded(sender, msg, self.Status.PROCESSOR_RUNNING,
+                                                    self.Status.PROCESSOR_COMPLETE, self.resume)
+
+
+def num_cores(conf):
```
danielmitterdorfer, line 485: Nit: So far we've referred to the configuration either as `config` or `cfg`, maybe we should use on of these here as well.

esrally/track/loader.py
```diff
@@ -391,55 +395,40 @@ def used_corpora(t):
...
+                logging.getLogger(__name__).info("Resolved data root directory for document corpus [%s] in track [%s] "
+                                                 "to [%s].", corpus.name, track.name, data_root)
+                if len(data_root) == 1:
+                    preparator.prepare_document_set(document_set, data_root[0])
+                # attempt to prepare everything in the current directory and fallback to the corpus directory
+                elif not preparator.prepare_bundled_document_set(document_set, data_root[0]):
+                    preparator.prepare_document_set(document_set, data_root[1])
 
     def on_prepare_track(self, track, data_root_dir):
+        self.track = track
```
danielmitterdorfer, line 421: Is a field necessary here? As far as I could see `track` is only used locally?

danielmitterdorfer (commented):
Thanks for the PR; I am already looking forward to a properly parallelized API. :)

I left some ideas and suggestions after doing a first pass.

esrally/driver/driver.py
```diff
@@ -75,6 +77,26 @@ class TrackPrepared:
     pass
 
 
+class StartTaskLoop:
```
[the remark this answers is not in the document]
DJRickyB: Given how many message classes are simply just a configuration and track, maybe we just look now to collapse them to one type?  I'm worried about the specificity of applying what you're saying to the task execution workers, and for instance this (potential) line which looks like it's using a generic message anyway:
```python
            self.send_to_children_and_transition(self, Init(self.track, self.cfg), self.Status.INITIALIZING,
                                                 self.Status.PROCESSOR_RUNNING)
```

esrally/driver/driver.py
DJRickyB: calling this **The Grouping Comment** for reference elsewhere.

The logic here is to prevent overlap of different processors occurring at the same time.  Put another way, all given tasks will finish for a registered processor before starting the tasks for the next one.  I could pre-seed all the tasks here in this partitioned fashion, but that's not what you're talking about, so I'm guessing there would be no need (provided the way I implemented it to behave makes sense)

esrally/driver/driver.py
```diff
@@ -313,42 +336,157 @@ def load_local_config(coordinator_config):
...
+                self.send(self.parent, ReadyForWork())
         else:
             self.wakeupAfter(datetime.timedelta(seconds=self.wakeup_interval))
 
 
+class TrackPreparationActor(actor.RallyActor):
+    class Status(Enum):
+        INITIALIZING = "initializing"
+        PROCESSOR_RUNNING = "processor running"
+        PROCESSOR_COMPLETE = "processor complete"
```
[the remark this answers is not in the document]
DJRickyB: see **The Grouping Comment**.  This is supposed to be an intermediate signal that we can move to the next processor

esrally/driver/driver.py
```diff
@@ -313,42 +336,157 @@ def load_local_config(coordinator_config):
...
+    @actor.no_retry("track preparator")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_ActorExitRequest(self, msg, sender):
+        self.logger.info("ActorExitRequest received. Forwarding to children")
+        for child in self.children:
+            self.send(child, msg)
+
+    @actor.no_retry("track preparator")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_BenchmarkFailure(self, msg, sender):
+        # sent by our generic worker; give context and forward
+        msg.message = "Track preparator has detected a benchmark failure. Notifying master..."
```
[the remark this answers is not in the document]
DJRickyB: I think the source message is likely enough.  I'm just going to forward it

esrally/driver/driver.py
```diff
@@ -313,42 +336,157 @@ def load_local_config(coordinator_config):
...
+            self.send_to_children_and_transition(self, StartTaskLoop(self.track.name, self.cfg), self.Status.PROCESSOR_COMPLETE,
+                                                 self.Status.PROCESSOR_RUNNING)
+        else:
+            self.send(self.original_sender, TrackPrepared())
+
+    def _seed_tasks(self, processor):
+        self.tasks = list(processor.on_prepare_track(self.track, self.data_root_dir))
+
+    def _create_task_worker(self):
+        child = self.createActor(TaskExecutionActor)
```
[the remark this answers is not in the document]
DJRickyB: good catch, leftover debugging stuff

esrally/track/loader.py
```diff
@@ -77,19 +83,17 @@ def register_track_processor(self, processor):
             processor.decompressor = Decompressor()
         self.track_processors.append(processor)
 
+    def processors(self):
+        if not self.track_processors:
```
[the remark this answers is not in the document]
DJRickyB: it's a bit confusing but the `TaskFilterTrackProcessor` and `TestModeTrackProcessor` are executed separately from this workflow.  Otherwise somewhere around this code block we need to say "always do these two but sometimes do this one." The code got kind of sloppy so I still invoke those two in [line 200/204] in a separate, dedicated TPR instance

esrally/track/loader.py
```diff
@@ -77,19 +83,17 @@ def register_track_processor(self, processor):
...
+            self.register_track_processor(DefaultTrackPreparator(self.base_config))
+        return self.track_processors
+
     def on_after_load_track(self, track):
```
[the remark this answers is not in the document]
DJRickyB: I agree.  after my last comment I think I'll maintain a list of `TrackProcessorRegistry.required_processors = [TaskFilterTrackProcessor, TestModeTrackProcessor ]` and then make the TrackPreparationActor code work like
```python
for processor in [*tpr.required_processors, *tpr.processors]:
    processor.on_after_load_track
    self.processors.put(processor)
```

Or however that syntax works

esrally/track/loader.py
DJRickyB: I was debugging this line, `track` is poisoned because of `from esrally.track import track`.  I'll make this clean without setting an instance field

esrally/utils/modules.py
```diff
@@ -61,9 +61,10 @@ def _load_component(self, component_name, module_dirs):
         root_module_name = "%s.%s" % (component_name, self.component_entry_point)
 
         for p in self._modules(module_dirs, component_name):
+            if p.endswith(".__init__"):
```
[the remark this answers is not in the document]
DJRickyB: probably the addition is not required, though I'll say it works fine.  I will remove it and test.  `importlib.reload(m)` was removed due to it causing pickling errors whenever we'd send a function over the wire for the TaskExecutionWorker

solution was gleaned from reading SO and articles similar to [this one](https://stackoverflow.com/questions/1412787/picklingerror-cant-pickle-class-decimal-decimal-its-not-the-same-object).  Couldn't figure out why reload was invoked here, or how to cause removing it to break something, everything seems to work fine

esrally/driver/driver.py
```diff
@@ -313,42 +336,157 @@ def load_local_config(coordinator_config):
...
-                self.send(self.original_sender, TrackPrepared())
+                self.send(self.parent, ReadyForWork())
         else:
             self.wakeupAfter(datetime.timedelta(seconds=self.wakeup_interval))
 
 
+class TrackPreparationActor(actor.RallyActor):
+    class Status(Enum):
+        INITIALIZING = "initializing"
+        PROCESSOR_RUNNING = "processor running"
```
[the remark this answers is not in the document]
DJRickyB: (meant to say this previously) see **The Grouping Comment**. This is supposed to be an intermediate signal

esrally/utils/modules.py
```diff
@@ -63,7 +63,6 @@ def _load_component(self, component_name, module_dirs):
         for p in self._modules(module_dirs, component_name):
             self.logger.debug("Loading module [%s]", p)
             m = importlib.import_module(p)
-            importlib.reload(m)
```
DJRickyB, line 66: this change does not seem to break anything, in testing.  this change was made to support pickling functions for the actor system, as it was breaking the class references after each reload (see a relevant [SO](https://stackoverflow.com/questions/1412787/picklingerror-cant-pickle-class-decimal-decimal-its-not-the-same-object))

DJRickyB:
Changed target to 2.2.0

esrally/track/loader.py
```diff
@@ -191,17 +184,13 @@ def _load_single_track(cfg, track_repository, track_name):
     try:
         track_dir = track_repository.track_dir(track_name)
         reader = TrackFileReader(cfg)
-
         current_track = reader.read(track_name, track_repository.track_file(track_name), track_dir)
-
         track_processor = TrackProcessorRegistry(cfg)
```
danielmitterdorfer, line 188: As the purpose of this class has changed from being a composite track processor to a registry I propose that we rename this variable e.g. to `registry`.

esrally/track/loader.py
```diff
@@ -83,16 +81,11 @@ def register_track_processor(self, processor):
             processor.decompressor = Decompressor()
         self.track_processors.append(processor)
 
+    @cached_property
```
danielmitterdorfer, line 84: IMHO it's not ideal that this property has a side-effect. It could also lead to issues that are hard to track down, e.g. when we want to register additional track processors lazily, i.e. after we've already retrieved the `processors` property once, because then additionally registered processors won't be recognized without any indication (e.g. an exception).

How about we provide a method to ensure explicitly that the default behavior (`DefaultTrackPreparator`) is available?

danielmitterdorfer (commented):
Thanks for iterating. I left a few more suggestions.

esrally/track/loader.py
DJRickyB: I understand the first part of this, but not 
> How about we provide a method to ensure explicitly that the default behavior (DefaultTrackPreparator) is available?

sorry.  Does changing this to `@property` circumvent the need for this suggestion? If not, I may need more explanation here or offline

DJRickyB: discussed offline.  will handle in my next commit

DJRickyB:
@elasticmachine run tests

danielmitterdorfer (commented):
I have a couple more comments about error handling but overall it looks quite promising.

danielmitterdorfer (approved):
Thanks for iterating! LGTM

PR #1229, opened 2021-04-05
remove invalid monitoring_type from example execution

DJRickyB:
looks like `https` is not a valid value here, should either say `http` or `local`


ebadyano (approved):
LGTM! Thank you for catching this!

PR #1231, opened 2021-04-06
Drop relative-time metric

danielmitterdorfer:
With this commit we drop the metric `relative-time` which contained a
timestamp in microseconds relative to the start of a task. It is
superseeded by the metric `relative-time-ms` which has been introduced
for a transition period between Rally 2.1.0 and Rally 2.3.0. With Rally
2.4.0 we will align all time-related metrics to be denoted in
milliseconds.

Relates #1198


danielmitterdorfer:
While this change can already be reviewed I intend to target this for the 2.2.0 release (i.e. not into any patch releases in the 2.1 series that we might release meanwhile).

ebadyano (approved):
LGTM! Thank you!

PR #1234, opened 2021-04-07
Allow to specify static responses

danielmitterdorfer:
With this commit we add the ability to short-circuit Rally's networking layer
and instead of targeting a real cluster, Rally will return static user-defined
responses. This simplifies client-side analysis for bottlenecks or load driver
behavior in general which previously required to setup a stub server (e.g.
nginx)


esrally/async_connection.py
```diff
@@ -16,6 +86,65 @@ async def text(self, encoding=None, errors="strict"):
...
+                raise ValueError(f"Unknown body encoding [{body_encoding}] for path [{path}]")
+
+            self.responses.append((path, matcher, body))
+
+    @staticmethod
+    def always():
+        # pylint: disable=unused-variable
+        def f(p):
+            return True
+        return f
```
dliappis, line 121: alternatively we could replace the closures with more Pythonic (IMHO? :) ) lambdas e.g.

`return lambda f: True`

or for `startswith`:
```
def startswith(path_pattern):
    return lambda f: f.startswith(path_pattern)
```

esrally/async_connection.py
```diff
@@ -16,6 +86,65 @@ async def text(self, encoding=None, errors="strict"):
...
+
+    @staticmethod
+    def startswith(path_pattern):
+        def f(p):
+            return p.startswith(path_pattern)
+        return f
+
+    @staticmethod
+    def endswith(path_pattern):
```
dliappis, line 130: One thought, we could enhance it in the future and support simple shell like patterns like `*_snapshot/frozen*` easily with https://docs.python.org/3/library/fnmatch.html. This could be quite useful.

docs/command_line_reference.rst
```diff
@@ -682,6 +688,63 @@ Client certificates can be presented regardless of the ``verify_certs`` setting,
...
+
+.. note::
+   Paths are evaluated from top to bottom. Therefore, place more restrictive paths at the top of the file.
+
+Save the above responses as ``responses.json`` and execute a benchmark as follows::
+
+    esrally race --track=geonames --challenge=append-no-conflicts-index-only --pipeline=benchmark-only --skip-rest-api-check --distribution-version=8.0.0 --client-options="static_responses:'responses.json'"
+
+.. note::
+   Use ``--skip-rest-api-check`` to ensure that Rally skips any probing requests that are not directly related to the benchmark (e.g. checks whether the cluster is up).
```
dliappis, line 746: Could we automatically skip rest api checks when `static_responses` is defined?

dliappis (commented):
Thanks! This is a very useful PR for stubbing ES responses easily.

I left a few questions/suggestions.

esrally/async_connection.py
danielmitterdorfer: I implemented this with lambdas earlier but then got a PEP-8 violation warning. :)

dliappis: LOL whaaaaaat?

esrally/async_connection.py
danielmitterdorfer: Good idea. Before we enhance it, we should probably run microbenchmarks to get a better grasp of the overhead though because we're on the hot code path.

docs/command_line_reference.rst
danielmitterdorfer: Thanks for the idea! I've pushed a change in https://github.com/elastic/rally/pull/1234/commits/af77d1bd3d670bdee1ae37c07dd0b3af77a4a275. In that change I've also:

* Turned `--skip-rest-api-check` into an undocumented flag again. With the changes in this PR it's also likely that we won't need that flag at all in the future but I'd like to keep it for now.
* Improved logging
* Avoided issuing *any* additional requests (probing for cluster version, telemetry devices) when static responses are enabled.

docs/command_line_reference.rst
```diff
@@ -682,6 +683,60 @@ Client certificates can be presented regardless of the ``verify_certs`` setting,
...
+        "body-encoding": "json"
+      }
+    ]
+
+.. note::
+   Paths are evaluated from top to bottom. Therefore, place more restrictive paths at the top of the file.
+
+Save the above responses as ``responses.json`` and execute a benchmark as follows::
+
+    esrally race --track=geonames --challenge=append-no-conflicts-index-only --pipeline=benchmark-only --distribution-version=8.0.0 --client-options="static_responses:'responses.json'"
```
dliappis, line 738: It took me a few seconds to realize that whether that while specifying a `pipeline` it won't really target any ES. One might even use the default pipeline, ES will get launched but won't be used.

Should we clarify with a note here or above in the `client-options` section that when `static_responses:'file'` is used, the use should also specify the `benchmark-only` pipeline and the `distribution-version` as Rally doesn't have a way to derive the version automatically as it'd normally do.

dliappis (approved):
Left one suggestion (up to you to implement), but LGTM. This is very useful!

docs/command_line_reference.rst
danielmitterdorfer: Thanks! I've added a note in https://github.com/elastic/rally/pull/1234/commits/37c0900542b1a088abf1c05869fad4462c222c84.

Issue #1202, opened 2021-03-02
Line up decimal places please!


nik9000:
When I run `esrally compare` I get a table that contains this bit:
```
| 1666.85 | 1555.73 | -111.116 | ms |
| 37.6112 |  38.016 |  0.40475 | ms |
| 2875.04 | 2934.54 |   59.493 | ms |
|  1515.4 | 1151.45 | -363.957 | ms |
```

That is hard for me to read so I always manually reformat it to:
```
| 1666.85   | 1555.73  | -111.116   | ms |
|   37.6112 |   38.016 |    0.40475 | ms |
| 2875.04   | 2934.54  |   59.493   | ms |
| 1515.4    | 1151.45  | -363.957   | ms |
```

Is that something rally could do by default?

danielmitterdorfer:
We use the `tabulate` library to render tables and [column alignment can be configured](https://github.com/astanin/python-tabulate#column-alignment). Currently, Rally uses `numalign=right`:

https://github.com/elastic/rally/blob/1562fde791dc3c2343d935b5ce2dd02a15ca2e8a/esrally/reporter.py#L78

If we change this to `decimal` we get the output that you describe. IMHO that makes the layout a bit more "noisy" so I'd rather not enable this by default. Maybe we can make this configurable though.

pquentin:
I believe this was done in #1236 and can be closed now.

PR #1236, opened 2021-04-09
add support for numbers alignment in table output (#1202)

cmanning09:
previously numbers alignment was hardcoded to 'right'. This change adds
the ability to set a configuration value 'numbers.align' in the
formatting section of the config to alter the format of the table output.


danielmitterdorfer:
Thanks for your interest in Rally and your PR! We'll start the review process shortly.

DJRickyB:
Hi! Thank you for this.  For the `race` subcommand things work as I would expect, but I wanted to note that when I use Rally's `compare` subcommand I get a failure, and it prints an error if the configuration is not defined:
```
[ERROR] Cannot compare. No value for mandatory configuration: section='reporting', key='numbers.align'
```

From the log:
```2021-04-12 16:15:10,875 -not-actor-/PID:59308 esrally.rally ERROR Cannot run subcommand [compare].
  File "/Users/rick.boyd/dev/rally/esrally/config.py", line 178, in opts
    return self._opts[self._k(scope, section, key)]
KeyError: (<Scope.application: 1>, 'reporting', 'numbers.align')

During handling of the above exception, another exception occurred:

  File "/Users/rick.boyd/dev/rally/esrally/rally.py", line 774, in dispatch_sub_command
    reporter.compare(cfg, args.baseline, args.contender)
  File "/Users/rick.boyd/dev/rally/esrally/reporter.py", line 48, in compare
    ComparisonReporter(cfg).report(
  File "/Users/rick.boyd/dev/rally/esrally/reporter.py", line 315, in __init__
    self.numbers_align = config.opts("reporting", "numbers.align")
  File "/Users/rick.boyd/dev/rally/esrally/config.py", line 183, in opts
    raise exceptions.ConfigError(f"No value for mandatory configuration: section='{section}', key='{key}'")
esrally.exceptions.ConfigError: No value for mandatory configuration: section='reporting', key='numbers.align'
```

cmanning09:
Good catch. I have addressed the issue you raised. Please let me know if there is anything else.

DJRickyB:
wonderful! thank you @cmanning09. This code functions as I'd expect, and as you indicated in the initial commit it addresses #1202 nicely.  

I have one more ask/iteration of you: this functionality should be documented.  please consider whether it will be supported similarly to the `reporting.format` key, in that it's only documented as the CLI `--report-format` flag (in which case more code would be required), or if users should always specify it in the `[reporting]` section of their local .ini.  Note that the documentation for that section states `This section defines how metrics are stored.` and so would need to be updated for the new meaning.

Thank you!

cmanning09:
I addressed your concern by adding support through CLI flags for both race and compare subcommands. Is there anything else?

DJRickyB:
Beautiful! I've tested it and it works as expected.  Please add the documentation and then I think we're all set. (Should be in docs/command_line_reference.rst)

DJRickyB (approved):
Thank you!

DJRickyB:
@elasticmachine run tests

DJRickyB:
@cmanning09, just as an update: I am having trouble getting this branch to pass our internal CI.  I don't believe this is due to your code, so please bear with me.  Once tests pass cleanly I will merge this one in

DJRickyB:
@elasticmachine update branch

DJRickyB:
@elasticmachine run tests

Issue #1107, opened 2020-11-04
Drop support for Elasticsearch 5.x


danielmitterdorfer:
Elasticsearch [5.x went EOL on 2019-03-11](https://www.elastic.co/support/eol). According our own [version policy in Rally](https://esrally.readthedocs.io/en/stable/versions.html) we keep support in Rally for two more years.

This ticket is meant as a place to collect what needs to be changed to drop support for 5.x.

* [x] Adapt our [version policy page](https://esrally.readthedocs.io/en/stable/versions.html) to indicate we've dropped support for Elasticsearch 5.x. - done in https://github.com/elastic/rally/pull/1237
* [x] Remove support to specify cluster settings in the track: https://github.com/elastic/rally/issues/915 - done in https://github.com/elastic/rally/pull/1238
 
Similar to #715, we need to have a grace period in rally-teams and rally-tracks before we drop support there.

Marking as blocked because we should not work on this before 2021-03-11 .

danielmitterdorfer:
Closing as the respective PRs have been merged. I've pushed the tag `v5` in rally-teams (https://github.com/elastic/rally-teams/tree/v5) and rally-tracks (https://github.com/elastic/rally-tracks/tree/v5).

PR #1237, opened 2021-04-12
Drop support for Elasticsearch 5.x

danielmitterdorfer:
Elasticsearch 5.x is end of life since 2019-03-11. According to our
version policy in Rally we drop support two years after the end of life
date.

Relates #1107


README.rst
```diff
@@ -16,8 +16,6 @@ You want to benchmark Elasticsearch? Then Rally is for you. It can help you with
 
 We have also put considerable effort in Rally to ensure that benchmarking data are reproducible.
 
-In general, Rally works with all versions of Elasticsearch starting from 1.x but benchmarking with plugins and benchmarking source builds will only work from Elasticsearch 5.0 onwards.
```
ebadyano, line 19: Do we want to specify 6.x onwards instead of totally removing it?

ebadyano:
Do we want to somehow prevent people running with `esrally race  --distribution-version=5.6.16`?

ebadyano:
A few more spots where we reference 5.0.0:
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/esrally/mechanic/supplier.py#L141
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/esrally/resources/track-schema.json#L494
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/esrally/utils/versions.py#L74

There are also a few references to 5.x. in rally tracks README file

README.rst
danielmitterdorfer: Our README file mostly exists so people that open the Rally repo on Github have an entry point. This file closely resembles our index page at https://esrally.readthedocs.io/en/stable/ which also does not document this so I opted for consistency across the two pages.

danielmitterdorfer:
Thanks for your review @ebadyano. I've pushed https://github.com/elastic/rally/pull/1237/commits/1a9b53c08a8f845bf8eab3d8670825159285c87a to address your feedback. Can you please have another look?

To your comments:

* I did not touch `VersionVariants` now because 5.x is only referenced in a source code comment and only serves as an example how that piece of code is supposed to work.
* I've removed the 5.x reference in the other places
* We could probably fail fast when people attempt to benchmark an unsupported version but I'm honestly not sure whether we should actively prohibit it. I see this as out of scope for this PR but we could discuss this in next week's sync. I've added it to the agenda.

ebadyano:
Looks good, but i think I found a few more references:
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/docs/pipelines.rst#L52
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/docs/command_line_reference.rst#L550
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/docs/recipes.rst#L87
https://github.com/elastic/rally/blob/7eaee22d9fb7632b817507a0c7afa20fb7e8b7c2/docs/track.rst#L658


Also do we want to add anything in the migrate guide?

ebadyano (approved):
no need for another iteration after the 3 more spots i mentioned above. Thank you, LGTM!

danielmitterdorfer:
Thanks for your review! I've removed the remaining references in https://github.com/elastic/rally/pull/1237/commits/0359c6444819902c3d74cee809f42bcfae21b601.

PR #1238, opened 2021-04-12
Disallow to specify cluster settings in the track

### LINKED ISSUE #915: Remove support to specify cluster settings in the track
So far we allow to specify a `cluster-settings` element in the track. This is trappy because it only works in certain circumstances, namely when Rally sets up the benchmark candidate itself. 

However, as we split Rally's functionality into individual subcommands, this should be done differently:

* static cluster settings should be defined via `rally-teams` and the [`additional_cluster_settings`](https://github.com/elastic/rally-teams/tree/master/cars/v1/vanilla)
* dynamic cluster settings should be defined via the [put-settings](https://esrally.readthedocs.io/en/stable/track.html#put-settings) operation in Rally

danielmitterdorfer:
With this commit we don't honor the challenge property
`cluster-settings` anymore as this is mixing multiple concerns (cluster
setup and running the benchmark). Instead, we encourage users to use the
`put-settings` API for dynamic cluster settings and the existing
mechanism in rally-teams to inject static cluster settings.

Closes #915


it/__init__.py
```diff
@@ -29,7 +29,7 @@
 from esrally.utils import process
 
 CONFIG_NAMES = ["in-memory-it", "es-it"]
-DISTRIBUTIONS = ["5.6.16", "6.8.0", "7.6.0"]
+DISTRIBUTIONS = ["6.8.0", "7.6.0"]
```
danielmitterdorfer, line 32: We're doing this here (as well) to decouple this PR from https://github.com/elastic/rally/pull/1237 as rally-tracks still uses this feature for 5.x clusters.

dliappis (approved):
LGTM

PR #1241, opened 2021-04-14
fix JAVA_HOME determination for ES 8.x

### LINKED ISSUE #1191: Don't set JAVA_HOME starting with ES 8.0
Starting with Elasticsearch 8.0 (https://github.com/elastic/elasticsearch/pull/69149) `JAVA_HOME` doesn't get honored. One should rely on `ES_JAVA_HOME` instead. This will also be supported starting with `7.12.0` having higher priority over `JAVA_HOME` and using `JAVA_HOME` will be deprecated.

Practically speaking for nightly benchmarks we generally rely on the bundled jdk, but if any user or internal tool needs to specify a different java version while we do add the bin directory in [PATH](https://github.com/elastic/rally/blob/5679af3a8f071d27fe04a2e5326d05edffbdd04e/esrally/mechanic/launcher.py#L174) we also specify [JAVA_HOME](https://github.com/elastic/rally/blob/5679af3a8f071d27fe04a2e5326d05edffbdd04e/esrally/mechanic/launcher.py#L176) which we define at least in https://github.com/elastic/rally/blob/5679af3a8f071d27fe04a2e5326d05edffbdd04e/esrally/mechanic/launcher.py#L176.

As far as I understand we won't be broken right now given that `JAVA_HOME/bin` is in the PATH and Elasticsearch 8.0.0 will just ignore the environment variable `JAVA_HOME`. Therefore the priority for fixing this is not immediate.

Regarding the fix I think we need to:

1. Rely on `ES_JAVA_HOME` when the distribution-version is >=8.0.0
2. Rely on `JAVA_HOME` always when building Elasticsearch from source (when using the pipeline [from-sources](https://esrally.readthedocs.io/en/stable/pipelines.html#from-sources)
3. Adjust and enhance unit tests/IT tests as required

DJRickyB:
closes #1191

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -172,7 +172,9 @@ class ProcessLauncher:
         env = {k: v for k, v in os.environ.items() if k in self.pass_env_vars}
         if java_home:
             self._set_env(env, "PATH", os.path.join(java_home, "bin"), separator=os.pathsep, prepend=True)
-            # Don't merge here!
+            # This property is the higher priority starting in ES 7.12.0, and is the only supported java home in >=8.0
+            env["ES_JAVA_HOME"] = java_home
+            # TODO remove this when ES <8.0 becomes unsupported by Rally
             env["JAVA_HOME"] = java_home
         if not env.get("ES_JAVA_OPTS"):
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM thank you

PR #1243, opened 2021-04-15
Ensure destructive actions work regardless of wildcard input

### LINKED ISSUE #1152: Ensure destructive actions work independently of action.destructive_requires_name
There are a couple of destructive APIs in Elasticsearch that allow wildcards (i.e. `*` or `_all`) by default (configurable with the setting `action.destructive_requires_name`). The default value for this setting will change with https://github.com/elastic/elasticsearch/pull/66908. We have put a stop-gap solution in place in https://github.com/elastic/rally-teams/pull/58 to set this value to `false` for all clusters that are configured by Rally. However, we should ensure that we always use explicit names in destructive actions as not all cluster might be configured by Rally. The affected APIs are:

* Adding an index block
* Open / close index API
* Delete Index / data stream

Note: I am not aware that we add index block or use the open / close index APIs anywhere but I wanted to mention them for completeness.

For the delete index / data stream APIs which are supported out of the box by Rally should choose one the following approaches:

* Temporarily set `action.destructive_requires_name` to `false` via the cluster settings API and reset it afterwards.
* Resolve index names containing patterns via an API call and enumerate all index names explicitly.

When we have implemented support in Rally, we should also revert the stop gap solution implemented in https://github.com/elastic/rally-teams/pull/58  after a short grace period.

DJRickyB:
As of version 8.0, Elasticsearch sets `action.destructive_requires_name` to `true` by default.

For Rally's purposes we set this value to `false` while executing `delete-index` operations so that we transparently support wildcard operations.

Closes #1152


esrally/driver/runner.py
```diff
@@ -1172,16 +1189,18 @@ async def __call__(self, es, params):
...
-        for index_name in indices:
-            if not only_if_exists:
-                await es.indices.delete(index=index_name, params=request_params)
-                ops += 1
-            elif only_if_exists and await es.indices.exists(index=index_name):
-                self.logger.info("Index [%s] already exists. Deleting it.", index_name)
-                await es.indices.delete(index=index_name, params=request_params)
-                ops += 1
-
+        prior_destructive_setting = await set_destructive_requires_name(es, False, request_params)
```
DJRickyB, line 1192: it occurred to me to do this conditionally but I thought this would be less fragile

ebadyano:
LGTM, but I tried to test it and I cannot seem to get it to fail without your change for the recent elasticsearch revision and with elastic/rally-teams#58. Am I missing something?

DJRickyB:
looks like the vanilla car template sets it explicitly. 
```
# ---------------------------------- Various -----------------------------------
#
# Allow wildcard deletion of indices:
#
action.destructive_requires_name: false
```

 previously i had tested it using only wireshark tracing to ensure responses were what it looked like it was supposed to. I am now also looking into getting this to fail, but even changing this value locally (and confirming it was applied with `--preserve-install`) I cannot seem to get it to do so

DJRickyB:
I have done the low-level testing between the old/new rally and enabled/disabled `destructive_requires_name` to ensure this resolves the problem. Under no current conditions does the 400 response cause `delete-index` to fail the race, but trying to delete-index without setting `destructive_requires_name` to `false` does indeed fail, silently (albeit with a logged warning).

I tried to get better correctness by setting `ignore=[404]` on the API, but 400 still fails silently.  Submitting to you as-is

ebadyano (approved):
LGTM, Thank you

PR #1244, opened 2021-04-20
Avoid ephemeral ports in integration tests

danielmitterdorfer:
Previously, some integration tests have spun up clusters in the
ephemeral port range (on Linux this port range usually starts with port
32768) which can negatively influence test stability. With this commit
we bind test clusters in integration tests to ports below the ephemeral
port range to avoid this issue.


DJRickyB (approved):
LGTM

PR #1245, opened 2021-04-20
Strictly check for supported version

danielmitterdorfer:
With this commit we add a strict check whether the supplied
Elasticsearch version is supported. We check the
`--distribution-version` command line argument for validity but also
the automatically derived distribution version because the distribution
version doesn't have to be specified in all cases.

Relates #1107


danielmitterdorfer:

esrally/rally.py
```diff
@@ -56,6 +56,14 @@ def runtime_jdk(v):
             except argparse.ArgumentTypeError:
                 raise argparse.ArgumentTypeError(f"must be a positive number or 'bundled' but was {v}")
 
+    def supported_es_version(v):
```
dliappis, line 59: Reading the definition of [type](https://docs.python.org/3/library/argparse.html#type) it seems to be that using type to do more advanced content based validation like here is a bit of a stretch. (hence the weird `if v:` and `return v`).

I think for now this is ok, but in a future refactoring that we've discussed about offline anyway, we could leverage [action](https://docs.python.org/3/library/argparse.html#action)?

dliappis (changes requested):
This seems to work fine!

Let's remove the unused dot file (AFAICS it's unused) and then it's ready for merging.

.es-version.txt
```diff
@@ -0,0 +1 @@
+6.0.0
```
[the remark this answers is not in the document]
danielmitterdorfer: That was a leftover. I'll remove.

esrally/es-version.txt
```diff
@@ -0,0 +1 @@
+6.0.0
```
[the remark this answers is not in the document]
danielmitterdorfer: Addressed in https://github.com/elastic/rally/pull/1245/commits/e1d7e1978e6a99f3d685a17862c478ff601c20e6.

esrally/rally.py
danielmitterdorfer: I was applying the same logic as for other validations but you have a valid point. I can raise a follow-up issue.

danielmitterdorfer:
@dliappis thanks for your review. I've addressed your comments. Can you please have another look?

dliappis (approved):
Thanks for iterating. LGTM.

esrally/rally.py
danielmitterdorfer: I've raised https://github.com/elastic/rally/issues/1249 so we won't forget.

PR #1250, opened 2021-04-22
Add note about ES version support

danielmitterdorfer:
With this commit we add a note about the fact that we've dropped support
for ES 5.x and Rally will refuse to start with unsupported versions.


gingerwizard (approved):
LGTM

PR #1252, opened 2021-04-25
Make S3 support optional

### LINKED ISSUE #1095: Make S3 dependency in Rally optional
To reduce the amount of dependencies, installation time and possibility of conflicts with the existing Python environment the `boto3` library could be made optional (similar to [async](https://github.com/elastic/elasticsearch-py/blob/252f39e0b5648ba13960ec84504b05d58512383b/setup.py#L105) in [elasticsearch-py](https://github.com/elastic/elasticsearch-py)).

The mechanics of this are illustrated [here](https://setuptools.readthedocs.io/en/stable/userguide/dependency_management.html?highlight=extras_require#optional-dependencies).

To install Rally including S3 support one would need to issue `python3 -mpip install esrally[s3]` and this change would need to be documented properly.

pquentin:
Closes #1095.

With this pull request, boto3 is no longer installed by default but should be installed using `python -m pip install esrally[s3]` instead.

When a S3 `base-url` is provided, here's what users see:

```
[INFO] Preparing for race ...
[ERROR] S3 support is optional. Install it with `python -m pip install esrally[s3]`

[ERROR] Cannot race. Error in track preparator
        No module named 'boto3'

...
```

So, we let the `ImportError` propagate, but also explain how to get S3 support.

To test this manually in an existing virtual environment, uninstalling boto3 manually is not enough, as `pkg_resource.require("esrally")` will complain that it's missing. Recreating the virtual environment from scratch and removing any .egg-info files should fix this.

For what it's worth, I haven't managed to get the integration tests to work locally, but they don't seem to use S3, so they probably still pass.

(This is only about S3 as suggested by #1095, but I'll be happy to do the same thing for GCS if this PR gets merged.)

### DIFF
--- a/esrally/utils/net.py
+++ b/esrally/utils/net.py
@@ -70,10 +70,22 @@ class Progress:
         self.p.finish()
 
 
+def _fake_import_boto3():
+    # This function only exists to be mocked in tests to raise an ImportError, in
+    # order to simulate the absence of boto3
+    pass
+
+
 def _download_from_s3_bucket(bucket_name, bucket_path, local_path, expected_size_in_bytes=None, progress_indicator=None):
     # pylint: disable=import-outside-toplevel
-    # lazily initialize S3 support - we might not need it
-    import boto3.s3.transfer
+    # lazily initialize S3 support - it might not be available
+    try:
+        _fake_import_boto3()
+        import boto3.s3.transfer
+    except ImportError:
+        console.error("S3 support is optional. Install it with `python -m pip install esrally[s3]`")
+        raise
+
 
     class S3ProgressAdapter:
         def __init__(self, size, progress):
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -112,7 +112,7 @@ To find the JDK, Rally expects the environment variable ``JAVA_HOME`` to be set 
 Optional dependencies
 ---------------------
 
-S3 support is optional and can be installed using the ``s3`` extra. If you need S3 support, install ``esrally[s3]`` instead of just ``esrally``, but otter than that follow the instructions below.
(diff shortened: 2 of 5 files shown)


esrally/utils/net.py
```diff
@@ -70,10 +70,22 @@ def finish(self):
         self.p.finish()
 
 
+def _fake_import_boto3():
+    # This function only exists to be mocked in tests to raise an ImportError, in
+    # order to simulate the absence of boto3
```
pquentin, line 75: This is quite ugly, but the alternative is to use `del sys.modules["boto3"]`, which is an ugly and global hack, and likely to break things. While we use this hack succesfully in urllib3 to test support for Python versions without ssl, this:

 * has caused [huge headaches in a project that adds async support to urllib3](https://github.com/python-trio/hip/pull/42#issuecomment-386304273) and
 * would prevent running tests in parallel.

So I would suggest avoiding touching `sys.modules`.

DJRickyB:

pquentin:
Thanks for running the tests! I'll fix the lint

pquentin:
@DJRickyB Can you please run the tests again? I guess @elasticmachine won't listen if I try to ask myself

DJRickyB:
@elasticmachine run tests

DJRickyB (changes requested):
@pquentin thanks a ton! I have one remaining requirement before approval: please preserve the behavior where `make install` ends up installing boto3+dependencies. for our purposes I think it's ok if `[s3]` resolution occurs in either `make install-user` or a new `make` target altogether

pquentin:
@DJRickyB Thank you for the review! Done

DJRickyB (approved):
Looks good to me.  Thank you once again for your contribution

pquentin:
Thank _you_ for the quick reviews! I'll work on making GCS support optional next.

DJRickyB:
@pquentin thank you, but we are looking at keeping GCS support first-class for the time being

PR #1254, opened 2021-04-28
assert developers have block storage dependencies installed

DJRickyB:
While reviewing #1252, we found boto3 got de-referenced for the developer's environment (in other words, when invoking `make install` in a clean environment).  With this commit we assert with tests that the proper dependencies are in place


pquentin (approved):
Thanks!

dliappis (approved):
LGTM

Issue #1242, opened 2021-04-15
Record shard and segment allocation


gingerwizard:
Shard allocation can be critical to identifying potential performance issues e.g. hot spots.

We need a means to record which shards have been allocated to which nodes and their segment count. Ideally we also record the index and data stream to which these shards belong.

A runner might be sufficient initially, but given shard allocation changes during the lifetime of a test e.g. as indices rollover, a telemetry device similar to node stats would be preferable.

dliappis:
Let's implement this as a telemetry device.

PR #1258, opened 2021-04-29
Record shard allocation

ebadyano:
Relates to https://github.com/elastic/rally/issues/1242


ebadyano:
@gingerwizard I started with _cat/shards API for now, it gives you  storage and docs count. do you think it's sufficient or would you prefer segments? I could use node stats api with ?level-shards, which will have segement info as well, but then it gives node id instead of of node name.. not sure if that's important.

gingerwizard:
@ebadyano could you provide some sample docs here. We defn need the node name for each doc. Shard level analysis is sufficient although a segment count will be better.

ebadyano:
For _cat/shards reposne `"geonames         4     p      STARTED    211237     63mb 127.0.0.1 rally-node-0"`

The **sample doc** would like like this:
```
{'name': 'shard-stats', 
'shard-id': '4', 
'index': 'geonames', 
'prirep': 'p', 
'docs': '211237', 
'store': '63mb', 
'node': 'rally-node-0'} 
```
**metadata**:
```
{'cluster': 'default'}
```

@gingerwizard I wonder if we want to have store always in bytes? And for primary/replica instead of `p` vs `r` could be `True/False`

gingerwizard:
So i think we need:
1. A count of segments if possible
2. Yes re `primary: true` vs p/r. Easier to filter.
3. And yes we need to convert all units to bytes so we can perform math at agg time + chart.

esrally/telemetry.py
```diff
@@ -568,6 +568,134 @@ def record(self):
                 self.metrics_store.put_doc(doc, level=MetaInfoScope.cluster, meta_data=shard_metadata)
 
 
+class ShardStats(TelemetryDevice):
+    """
+    Gathers different shard per index stats every minute.
```
danielmitterdorfer, line 573: The documentation here refers to the (current) default value of `shard-stats-sample-interval`. Maybe remove this information?

esrally/telemetry.py
```diff
@@ -568,6 +568,134 @@ def record(self):
...
+    """
+
+    internal = False
+    command = "shard-stats"
+    human_name = "Shard Stats"
+    help = "Regularly samples _cat/shards"
```
danielmitterdorfer, line 579: The API we use should be an implementation detail and we should probably not mention it here (at best in the user docs).

esrally/telemetry.py
```diff
@@ -568,6 +568,134 @@ def record(self):
...
+        return "shard stats"
+
+    def record(self):
+        """
+        Collect _cat/shards for indexes (optionally) specified in telemetry parameters and push to metrics store.
+        """
+        # pylint: disable=import-outside-toplevel
+        import elasticsearch
+        try:
+            stats = self.client.cat.shards(index=self.indices)
```
danielmitterdorfer, line 672: Can we avoid cat APIs here? Quoting from the [Elasticsearch docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/cat.html):

> cat APIs are only intended for human consumption [...]. For application consumption, we recommend using a corresponding JSON API.

esrally/telemetry.py
```diff
@@ -568,6 +568,134 @@ def record(self):
...
+        except elasticsearch.TransportError:
+            msg = "A transport error occurred while collecting _cat/shards on cluster [{}]".format(self.cluster_name)
```
danielmitterdorfer, line 674: We should use f-strings for all new code.

esrally/telemetry.py
```diff
@@ -568,6 +568,134 @@ def record(self):
...
+        if isinstance(indices_per_cluster, str):
+            self.indices_per_cluster = {opts.TargetHosts.DEFAULT: indices_per_cluster}
+        else:
+            self.indices_per_cluster = indices_per_cluster
+
+        if self.indices_per_cluster:
+            for cluster_name in self.indices_per_cluster.keys():
+                if cluster_name not in clients:
+                    raise exceptions.SystemSetupError(
+                        "The telemetry parameter 'shard-stats-indices' must be a JSON Object with keys matching "
```
danielmitterdorfer, line 616: We should use f-strings for all new code.

danielmitterdorfer (commented):
I left some comments: we should avoid the cat API, also user documentation is missing.

esrally/telemetry.py
pquentin: Have you considered using https://github.com/asottile/pyupgrade as part of `make lint` to enforce f-strings everywhere? There are ways to [avoid ruining git blame](https://github.com/psf/black/#migrating-your-code-style-without-ruining-git-blame) if that's a concern.

danielmitterdorfer: Ah, thanks for the idea; I wasn't aware of that. [pyupgrade](https://github.com/asottile/pyupgrade) could really be a nice approach to migrate to f-strings across the code base. We already discussed a while back using black for automatic formatting in another tool that we maintain but did not get around to introducing it yet so mid-term this could indeed be an option.

ebadyano:
@gingerwizard Example of a doc pushed to the metric store:

```
            {
                "name": "shard-stats",
                "shard-id": "0",
                "index": "geonames",
                "primary": True,
                "docs": 1000,
                "store": 212027,
                "node": "rally0"
            }

**Comment by gingerwizard** [comment]
The structure LGTM @ebadyano and fullfils the requirements thanks. I will now review the code.

File: esrally/telemetry.py [hunk 6]
```diff
@@ -568,6 +568,138 @@ def record(self):
...
+
+        self.telemetry_params = telemetry_params
+        self.clients = clients
+        self.sample_interval = telemetry_params.get("shard-stats-sample-interval", 60)
+        if self.sample_interval <= 0:
+            raise exceptions.SystemSetupError(
+                f"The telemetry parameter 'shard-stats-sample-interval' must be greater than zero but was {self.sample_interval}.")
+
+        self.specified_cluster_names = self.clients.keys()
+        indices_per_cluster = self.telemetry_params.get("shard-stats-indices", False)
```
gingerwizard, line 604: is `False` a valid value here? Shouldn't this be a * by default i.e. we monitor all indices? 
Its not clear what happens if nothing is specified for `shard-stats-indices` given `specified_cluster_names` will not be set.

esrally/telemetry.py
```diff
@@ -664,36 +662,42 @@ def __str__(self):
...
 
+        for node_stats in sample["nodes"].values():
+            node_name = node_stats["name"]
+            collected_node_stats = collections.OrderedDict()
+            collected_node_stats["name"] = "shard-stats"
+            shard_stats = node_stats["indices"].get("shards")
+
+            for index_name, stats in shard_stats.items():
+                for curr_shard in stats:
+                    for shard_id, curr_stats in curr_shard.items():
```
danielmitterdorfer, line 688: Can we make this more robust in case of missing properties? A benchmark should not fail because a property wasn't contained in the response.

esrally/telemetry.py
```diff
@@ -568,6 +568,138 @@ def record(self):
...
+        if isinstance(indices_per_cluster, str):
+            self.indices_per_cluster = {opts.TargetHosts.DEFAULT: indices_per_cluster}
+        else:
+            self.indices_per_cluster = indices_per_cluster
+
+        if self.indices_per_cluster:
+            for cluster_name in self.indices_per_cluster.keys():
+                if cluster_name not in clients:
+                    raise exceptions.SystemSetupError(
+                        f"The telemetry parameter 'shard-stats-transforms' must be a JSON Object with keys "
```
danielmitterdorfer, line 615: The parameter is called `shard-stats-indices`, can you please align the names? Also, we should add user docs.

danielmitterdorfer (commented):
Thanks for iterating. I left a couple more comments. Can you please ensure we have user docs?

esrally/telemetry.py
```diff
@@ -568,6 +568,138 @@ def record(self):
...
+                        doc = {
+                            "name": "shard-stats",
+                            "shard-id": shard_id,
+                            "index": index_name,
+                            "primary": curr_stats["routing"].get("primary"),
+                            "docs": curr_stats["docs"].get("count"),
+                            "store": curr_stats["store"].get("size_in_bytes"),
+                            "segments_count":  curr_stats["segments"].get("count"),
+                            "node": node_name
+                        }
```
gingerwizard, line 698: I'm assuming other meta is automatically added e.g. the env id etc, that we have on other docs.

gingerwizard (changes requested):
thanks @ebadyano this largely looks good. A few comments re exceptions and I'm not clear how we support no `shard-stats-indices`. I think we need a better default value

esrally/telemetry.py
gingerwizard: maybe `{opts.TargetHosts.DEFAULT: "*"}` as the default.

esrally/telemetry.py
```diff
@@ -568,6 +568,138 @@ def record(self):
...
+        # allow the user to specify either an index pattern as string or as a JSON object
+        if isinstance(indices_per_cluster, str):
+            self.indices_per_cluster = {opts.TargetHosts.DEFAULT: indices_per_cluster}
+        else:
+            self.indices_per_cluster = indices_per_cluster
+
+        if self.indices_per_cluster:
+            for cluster_name in self.indices_per_cluster.keys():
+                if cluster_name not in clients:
+                    raise exceptions.SystemSetupError(
```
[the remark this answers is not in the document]
danielmitterdorfer: f-strings in exceptions are fine but for logging we should use `%` formatting so the logging framework can do the formatting lazily if needed.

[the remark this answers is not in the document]
ebadyano: since this is an exception going too leave it as f-string

esrally/telemetry.py
ebadyano: I assume so, since I don't see any special things for that in rally

esrally/telemetry.py
```diff
@@ -568,6 +568,138 @@ def record(self):
...
+        :param clients: A dict of clients to all clusters.
+        :param metrics_store: The configured metrics store we write to.
+        """
+        super().__init__()
+
+        self.telemetry_params = telemetry_params
+        self.clients = clients
+        self.sample_interval = telemetry_params.get("shard-stats-sample-interval", 60)
+        if self.sample_interval <= 0:
+            raise exceptions.SystemSetupError(
```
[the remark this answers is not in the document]
ebadyano: since this is an exception going too leave it as f-string

esrally/telemetry.py
ebadyano: So, this was leftover from when i used _cat api and there was a way to pass index parameter to the query to only get stats for specific indices. Looks like nodes api in elasticsearch py client doesn't allow that. I can still keep and process on rally side if we think it could be useful. thoughts?

danielmitterdorfer: We don't expect that there are many indices in the cluster that are not created because of the benchmark so I'd opt for simplicity and remove the parameter.

gingerwizard: agreed any non-benchmark indices will be negligible. I'd like to filter out internal and system indices but I think we can do that in our dashboarding.

docs/telemetry.rst
```diff
@@ -165,3 +166,25 @@ Supported telemetry parameters:
...
+
+The shard-stats telemetry device regularly calls the `cluster node-stats API with level=shard parameter <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html>`_ and records one metrics document per shard.
+
+Example of the recorded document::
+
+   {
+     "name": "shard-stats",
+     "shard-id": "0",
+     "index": "geonames",
+     "primary": True,
```
danielmitterdorfer, line 181: We should show example documents in JSON, rather than Python dict format. Can you please lower-case this?

docs/telemetry.rst
```diff
@@ -165,3 +166,25 @@ Supported telemetry parameters:
...
+     "docs": 1000,
+     "store": 212027,
+     "segments_count": 8,
```
danielmitterdorfer, line 184: We mix hyphens (see `shard-id`) and underscores here but it would be nicer to read if this is consistent. Can you please use hyphens everywhere?

docs/telemetry.rst
```diff
@@ -165,3 +166,25 @@ Supported telemetry parameters:
 
 * ``searchable-snapshots-stats-indices`` (default: None): A string with the index/index pattern, or list of indices/index patterns that searchable snapshots stats should additionally be collected from. If unset, only cluster level stats will be collected.
 * ``searchable-snapshots-stats-sample-interval`` (default 1): A positive number greater than zero denoting the sampling interval in seconds.
+
+shard-stats
+--------------
+
+The shard-stats telemetry device regularly calls the `cluster node-stats API with level=shard parameter <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-stats.html>`_ and records one metrics document per shard.
```
danielmitterdorfer, line 173: Nit: "node-stats" -> "node**s**-stats"

docs/telemetry.rst
```diff
@@ -165,3 +166,25 @@ Supported telemetry parameters:
...
+
+Example of the recorded document::
```
danielmitterdorfer, line 175: Nit: "the" -> "a"

esrally/telemetry.py
```diff
@@ -568,6 +568,112 @@ def record(self):
...
+                            "shard-id": shard_id,
+                            "index": index_name,
+                            "primary": curr_stats.get("routing", {}).get("primary"),
+                            "docs": curr_stats.get("docs", {}).get("count", -1),
+                            "store": curr_stats.get("store", {}).get("size_in_bytes", -1),
+                            "segments_count":  curr_stats.get("segments", {}).get("count", -1),
+                            "node": node_name
+                        }
+                        self.metrics_store.put_doc(doc, level=MetaInfoScope.cluster, meta_data=shard_metadata)
+                        print(f"shards {doc}")
```
danielmitterdorfer, line 674: Leftover?

danielmitterdorfer (commented):
Thanks for the docs. I left a couple of small comments but I think we're almost there now.

esrally/telemetry.py
ebadyano: oops, thank you for catching

danielmitterdorfer: Can you please remove the print statement?

danielmitterdorfer (commented):
Once we get rid of a leftover I think we're good. :)

ebadyano:
that's one stubborn print!

danielmitterdorfer (approved):
With the `print` now gone, LGTM :)

PR #1259, opened 2021-04-30
improvements to chart generator

DJRickyB:
With this commit we:
* Add a `-yyyymmdd` suffix to generated dashboard names for easier version maintenance
* Set memory-based timeseries charts on a logarithmic scale
* Fail if `--chart-spec-path` is invalid

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -19,7 +19,6 @@ import glob
 import json
 import logging
 import uuid
-from datetime import datetime
 
 from esrally import track, config, exceptions
 from esrally.utils import io, console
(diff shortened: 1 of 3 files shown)


ebadyano:
@DJRickyB Thank you for the changes. I'm not totally sure about the `-yyyymmdd` suffix, the chart title is what displayed in the dashboards so it would be displayed like this:
image

DJRickyB:
Here's an example generated dashboard for group 3 (geopointshape): https://ae582947d1ed4df0adc39c2d047e051a.eu-central-1.aws.cloud.es.io:9243/goto/78cecc8b5a2433858b42e57e40b151a6

The date suffix is only visible if you click through to the dashboard, and only in the dashboard title.  Maybe another format could be more natural looking, like `-rev20210429`, or if you don't think it's useful or worth doing I'll just remove it

DJRickyB:
after discussing offline, @ebadyano pointed out we don't get the intended benefit so I've changed the naming behavior back to what it was before.  here's a newly generated geopointshape dashboard for demonstration: https://elasticsearch-benchmark-analytics.elastic.co/app/dashboards#/view/7361f6ef-1595-4c03-99ba-29cd05f155d0

ebadyano (approved):
LGTM, thank you

PR #1264, opened 2021-05-12
Pass JAVA16_HOME in build

ebadyano:
With this commit we pass the environment variable JAVA16_HOME in the build if it is defined.


danielmitterdorfer (approved):
LGTM

PR #1266, opened 2021-05-19
Support gradual ramp-up

### LINKED ISSUE #1195: Support Query Ramp-up

danielmitterdorfer:
With this commit we allow users to ramp-up load gradually by specifying
the task property `ramp-up-time-period`. If a non-zero value is
specified, Rally will gradually add clients during that time period
until the target client count as specified by `clients` is reached. This
reduces the potential for initial load spikes when running with many
concurrent clients.

Closes #1195


danielmitterdorfer:
This can be tested with the following configuration:

1. Create a track and store it as `ramp-up.json`:

```json
{
  "version": 2,
  "schedule": [
    {
      "operation": {
        "name": "query-match-all",
        "operation-type": "search",
        "index": "_all",
        "body": {
          "query": {
            "match_all": {}
          }
        }
      },
      "clients": 100,
      "ramp-up-time-period": 120,
      "warmup-time-period": 120,
      "time-period": 120,
      "target-throughput": 1000
    }
  ]
}
```

2. We'll use static responses, so also create `responses.json`:

```json
[
  {
    "path": "*/_search",
    "body": {
        "took": 1,
        "timed_out": false,
        "hits":{
            "total" : {
                "value": 1,
                "relation": "eq"
            }
        }
    },
    "body-encoding": "raw"
  },
  {
    "path": "*",
    "body": {},
    "body-encoding": "json"
  }
]
```

3. Run a race:

```
esrally race --track-path=ramp-up.json --pipeline=benchmark-only --distribution-version=8.0.0 --client-options="static_responses:'responses.json'"
```

If we specify an Elasticsearch metrics store, we can visualize the ramp-up. The following chart shows how query throughput is linearly increasing during ramp-up and is then kept steadily at the specified target througput:

throughput-ramp-up

docs/track.rst
```diff
@@ -447,6 +447,7 @@ The ``schedule`` element contains a list of tasks that are executed by Rally, i.
 * ``clients`` (optional, defaults to 1): The number of clients that should execute a task concurrently.
 * ``warmup-iterations`` (optional, defaults to 0): Number of iterations that each client should execute to warmup the benchmark candidate. Warmup iterations will not show up in the measurement results.
 * ``iterations`` (optional, defaults to 1): Number of measurement iterations that each client executes. The command line report will automatically adjust the percentile numbers based on this number (i.e. if you just run 5 iterations you will not get a 99.9th percentile because we need at least 1000 iterations to determine this value precisely).
+* ``ramp-up-time-period`` (optional, defaults to 0): Rally will start clients gradually. It reaches the number specified by ``clients`` at the end of the specified time period. See the section on :ref:`ramp-up <track_ramp_up>` for more details.
```
DJRickyB, line 450: presumably this value is in seconds?

esrally/track/loader.py
```diff
@@ -1407,29 +1407,41 @@ def parse_parallel(self, ops_spec, ops, challenge_name):
...
         clients = self._r(ops_spec, "clients", error_ctx="parallel", mandatory=False)
         completed_by = self._r(ops_spec, "completed-by", error_ctx="parallel", mandatory=False)
 
         # now descent to each operation
         tasks = []
         for task in self._r(ops_spec, "tasks", error_ctx="parallel"):
             tasks.append(self.parse_task(task, ops, challenge_name, default_warmup_iterations, default_iterations,
-                                         default_warmup_time_period, default_time_period, completed_by))
+                                         default_warmup_time_period, default_time_period, default_ramp_up_time_period,
+                                         completed_by))
```
DJRickyB, line 1419: the number of positional arguments here is a smell, consider encapsulating in a config dataclass?

danielmitterdorfer: I agree that this method declares quite some parameters by now. However, given that this is de facto a private method that has only one call site that passes these parameters, I'd not go so far as to introduce a class for that?

docs/track.rst
danielmitterdorfer: Good point. Addressed in https://github.com/elastic/rally/pull/1266/commits/5232776893b4f2ced328f06c3f2235c2c20ccf9b.

docs/track.rst
```diff
@@ -568,6 +569,21 @@ If you want as much reproducibility as possible you can choose the `deterministi
...
+~~~~~~~~~~~~
+
+For benchmarks involving many clients it can be useful to increase load gradually. This avoids load spikes at the beginning of a benchmark when Elasticsearch is not yet warmed up. Rally will gradually add more clients over time but each client will already attempt to reach its specified target throughput. The diagram below shows how clients are added over time:
+
+.. image:: track-ramp-up.png
+   :alt: How ramp-up works
+
+.. note::
+
+    As the number of clients and thus the throughput varies during ramp-up, you should set ``warmup-time-period`` at least to the same value as ``ramp-up-time-period`` to get stable measurement results.
```
gingerwizard, line 584: My first question was "do ramp up and warm up overlap and are they related?" This answers this but not directly. I think we could be more explicit.  Maybe the image tells a thousand words?

gingerwizard: "warmup-time-period greater than or equal to" I think is clearer? Also, can we see a viable use case where the warm-up is less than the ramp-up? I don't think we want to be still ramping after warmup? If not, can we actually prevent this?

esrally/track/loader.py
```diff
@@ -1452,17 +1465,26 @@ def parse_task(self, task_spec, ops, challenge_name, default_warmup_iterations=N
...
-            self._error("Operation '%s' in challenge '%s' defines a warmup time period of '%d' seconds and '%d' iterations. Please do not "
-                        "mix time periods and iterations." % (op.name, challenge_name, task.warmup_time_period, task.iterations))
+            self._error(f"Operation '{op.name}' in challenge '{challenge_name}' defines a warmup time period of "
+                        f"{task.warmup_time_period} seconds and {task.iterations} iterations but mixing time periods "
+                        f"and iterations is not allowed.")
+
+        if (task.warmup_iterations is not None or task.iterations is not None) and task.ramp_up_time_period is not None:
+            self._error(f"Operation '{op.name}' in challenge '{challenge_name}' defines a ramp-up time period of "
+                        f"{task.ramp_up_time_period} seconds as well as {task.warmup_iterations} warmup iterations and "
+                        f"{task.iterations} iterations but mixing time periods and iterations is not allowed.")
```
gingerwizard, line 1487: Re above and the fact I don't think we want to be still ramping after warmup, can we prevent and error here?

gingerwizard (commented):
Some feedback re relationship between params

docs/track.rst
danielmitterdorfer: Yes, let's prevent this. I've pushed 518d66c0bea9cb7b8c77ddd92ba1b25fb678a46d to enforce this behavior.

esrally/track/loader.py
danielmitterdorfer: Yes, let's prevent this. I've pushed 518d66c0bea9cb7b8c77ddd92ba1b25fb678a46d to enforce this behavior.

danielmitterdorfer:
@DJRickyB can you please have another look?

esrally/track/loader.py
DJRickyB: i was thinking it would be applicable to both this and Task.__init__() but the call sites are mainly dozens of tests, so yes, retracted for now

DJRickyB (approved):
LGTM. generated/inspected docs locally

PR #1267, opened 2021-05-19
Consider new API calls in static response example

danielmitterdorfer:
With this commit we amend the example response file in our documentation
to also include a static response for the cluster settings API. This is
needed because Rally issues this call to determine the value of
`action.destructive_requires_name` prior to issuing a delete index
request.

Relates #1243

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -716,6 +716,13 @@ Here we define the necessary responses for a track that bulk-indexes data::
         "body-encoding": "json"
       },
       {
+        "path": "/_cluster/settings",
+        "body": {
+          "transient": {}
+        },
+        "body-encoding": "json"
+      },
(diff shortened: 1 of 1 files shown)


docs/command_line_reference.rst
```diff
@@ -715,6 +715,13 @@ Here we define the necessary responses for a track that bulk-indexes data::
         },
         "body-encoding": "json"
       },
+      {
+        "path": "/_cluster/settings",
+        "body": {
+          "transient": {}
```
DJRickyB, line 721:
```suggestion
          "persistent": {},
          "transient": {}
```

DJRickyB (approved):
LGTM.  suggested an optional change for completeness sake (this is what is returned in totality when no custom settings are applied)

PR #1270, opened 2021-05-20
Include requests when installing google-resumable-media

### LINKED ISSUE #1269: Missing transitive dependency `requests` from the `google-resumable-media` library
## Description of the problem

If a user [installs Rally from pip](https://esrally.readthedocs.io/en/stable/install.html#installing-rally) or from a local checkout without installing development dependencies, and attempts to use a track that references a corpus stored in a Google Storage bucket, `esrally race` will fail with:

`[ERROR] Cannot race. Error in track preparator (The requests library is not installed, please install the requests package to use the requests transport.)`

## Root cause

When we added Google Storage support for bucket in #1094 we leveraged the [google resumable media]() package to allow us support the progress indicator.

Specifically we are using [download.py#consume_next_chunk](https://github.com/elastic/rally/blob/d9dcb2ccc7e4cadee5d1d8e26571360683e26ef9/esrally/utils/net.py#L148) which in turn [relies](https://github.com/googleapis/google-resumable-media-python/blob/e254f3e8e704a031773a0ca1c11679cfef7e94e4/google/resumable_media/requests/download.py#L347-L348) on the [requests]() package.

The requests package is specified as an extra dependency though in google resumable media [dependencies](https://github.com/googleapis/google-resumable-media-python/blob/e254f3e8e704a031773a0ca1c11679cfef7e94e4/setup.py#L33-L34) and we missed that in Rally's [setup.py](https://github.com/elastic/rally/blob/d9dcb2ccc7e4cadee5d1d8e26571360683e26ef9/setup.py#L80)

Note that when installing Rally [for development](https://esrally.readthedocs.io/en/stable/developing.html#installation-instructions-for-development), certain third party libraries specify `requests` as a transitive dependency and everything works ok with GS buckets, masking this bug.

dliappis:
This commit adds the requests package (which is an extra dependency,
i.e. listed as optional) when installing the
google-resumable-media pip package.

Currently this is missing when installing Rally without development
dependencies and will cause tracks relying on GS hosted private buckets
to fail.

Closes #1269

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,7 +77,7 @@ install_requires = [
     # License: Apache 2.0
     # transitive dependencies:
     #   google-crc32c: Apache 2.0
-    "google-resumable-media==1.1.0",
+    "google-resumable-media[requests]==1.1.0",
     # License: Apache 2.0
     "google-auth==1.22.1"
 ]


danielmitterdorfer (commented):
Thanks for addressing! LGTM

danielmitterdorfer (approved):

PR #1273, opened 2021-05-28
Re-introduce relative-time property

danielmitterdorfer:
With this commit we introduce the property `relative-time` for metrics
documents again. Contrary to Rally versions prior to 2.1.0, this
property contains the relative time since Rally started execution of a
task denoted in milliseconds, instead of microseconds. We also
deprecate the ancillary property `relative-time-ms` which will be
dropped in the next minor release in favor of `relative-time`.

Relates #1198


ebadyano (approved):
LGTM

PR #1274, opened 2021-05-31
Add example for raw-request

danielmitterdorfer:
With this commit we add an example Rally track that demonstrates how to
use the `raw-request` operation. As the description in the reference
docs is quite abstract this example should clarify usage.


ebadyano (approved):
LGTM

PR #1275, opened 2021-05-31
Remove segment memory chart

danielmitterdorfer:
As segment memory statistics are dropped in
https://issues.apache.org/jira/browse/LUCENE-9387, we drop the related
chart from the chart generator.

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -442,11 +442,6 @@ class BarCharts:
         }
 
     @staticmethod
-    def segment_memory(title, environment, race_config):
-        # don't generate segment memory charts for releases
-        return None
-
-    @staticmethod
     def query(environment, race_config, q):
         metric = "service_time"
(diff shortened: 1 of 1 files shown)


ebadyano:
Do we also need a follow up pr to cleanup telemetry/reporter or will it just work and not produce the metrics when the statistics are dropped form lucene?

ebadyano (approved):
LGTM, thank you

danielmitterdorfer:
> Do we also need a follow up pr to cleanup telemetry/reporter or will it just work and not produce the metrics when the statistics are dropped form lucene?

@ebadyano I've checked this before raising this PR and I think we're good here. Rally will issue a warning that it could not read these metrics but it will continue to run; the reporter will not report any metrics without a value. As we also support versions prior to ES master we need to keep the code in the telemetry device. The only aspect that we could consider is retrieving the metrics only for versions < 8.0 but I'd defer that consideration until Lucene 9 is available on master.

PR #1277, opened 2021-06-04
Add chart support for observability queries

ebadyano:
Observability queries do't have target throughput and instead have `workflow-interval`

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -1689,8 +1689,7 @@ class RaceConfig:
                 #
                 # We should refactor the chart generator to make this classification logic more flexible so the user can specify
                 # which tasks / or types of operations should be used for which chart types.
-                if sub_task.operation.type == "search" or sub_task.operation.type == "composite" or sub_task.operation.type == "eql"\
-                    or sub_task.operation.type == "paginated-search" or sub_task.operation.type == "scroll-search"\
+                if sub_task.operation.type in ["search", "composite", "eql", "paginated-search", "scroll-search"]\
                     or "target-throughput" in sub_task.params or "target-interval" in sub_task.params:
                     task_names.append(sub_task)
         return task_names
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -1658,7 +1658,6 @@ class RaceConfig:
         task_names = []
         for task in self.track.find_challenge_or_default(self.challenge).schedule:
             for sub_task in task:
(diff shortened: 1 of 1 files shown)


esrally/chart_generator.py
```diff
@@ -1666,7 +1666,8 @@ def throttled_tasks(self):
                 #
                 # We should refactor the chart generator to make this classification logic more flexible so the user can specify
                 # which tasks / or types of operations should be used for which chart types.
-                if "target-throughput" in sub_task.params or "target-interval" in sub_task.params or sub_task.operation.type == "eql":
+                if "target-throughput" in sub_task.params or "target-interval" in sub_task.params\
```
danielmitterdorfer, line 1669: How about we instead use the operation type as an indicator? I propose the following operation types (in addition to `eql`):

* `search`
* `composite`
* `paginated-search`
* `scroll-search`

danielmitterdorfer (commented):
Thanks! I left a suggestion how to categorize throttled tasks.

esrally/chart_generator.py
ebadyano: hmm, just realize that if we only check for type of operation we will miss raw-requests like in https://github.com/elastic/rally-tracks/pull/179/files, I wonder if we still want to keep ` "target-throughput" in sub_task.params or "target-interval"` as well as type of operations?

danielmitterdorfer: I agree, we should also consider `target-throughput` and `target-interval` as before.

ebadyano:
@danielmitterdorfer Is this okay to merge? Thank you

esrally/chart_generator.py
```diff
@@ -1678,7 +1690,8 @@ def throttled_tasks(self):
                 # We should refactor the chart generator to make this classification logic more flexible so the user can specify
                 # which tasks / or types of operations should be used for which chart types.
                 if sub_task.operation.type == "search" or sub_task.operation.type == "composite" or sub_task.operation.type == "eql"\
-                    or sub_task.operation.type == "paginated-search" or sub_task.operation.type == "scroll-search":
```
danielmitterdorfer: We can simplify the checks for the operation type:

```python
if sub_task.operation.type in ["search", "composite", "eql", "paginated-search", "scroll-search"] ...
```

danielmitterdorfer (commented):
I left one suggestion but apart from that it looks fine.

danielmitterdorfer (approved):
Thank you! LGTM

PR #1280, opened 2021-06-07
Add support for ml_processing_time chart

ebadyano:


esrally/chart_generator.py
```diff
@@ -1503,6 +1590,18 @@ def generate_merge_time(chart_type, race_configs, environment):
 
     return structures
 
+def generate_ml_procesing_time(chart_type, race_configs, environment):
```
danielmitterdorfer, line 1593: * Nit (typo): `procesing` -> `processing`
* Nit: There should be two empty lines before each function (there are similar issues with the functions `generate_merge_count`, `generate_dashboard` and `generate_merge_time`)

esrally/chart_generator.py
```diff
@@ -1503,6 +1590,18 @@ def generate_merge_time(chart_type, race_configs, environment):
...
+    structures = []
+    if chart_type == BarCharts:
```
danielmitterdorfer, line 1595: IMHO this code should be independent of the chart type. We can instead implement the `ml_processing_time` method also for `BarCharts` but return `None`and check whether a value different from `None` has been returned by that method. If and only the value is not `None`, we'll append it to `structures`. This allows us to evolve the chart type independently of the generator code.

danielmitterdorfer (commented):
Thanks! I left a few comments.

danielmitterdorfer (approved):
Thanks for iterating. LGTM

PR #1284, opened 2021-06-16
Ensure max_connections value is logged correctly

### LINKED ISSUE #1278: max_connections value is misleadingly logged prior to being overridden
**Describe the feature**:
Currently `esrally` logs the output of  the `client_options` parameter at the time an [Elasticsearch client](https://github.com/elastic/rally/blob/master/esrally/client.py#L125) is created, for example:
```
2021-06-07 02:17:11,408 ActorAddr-(T|:62446)/PID:42037 esrally.client INFO Creating ES client connected to [{'host': '127.0.0.1', 'port': 39200}] with options [{'timeout': 180, 'max_connections': 2}]
```

The value displayed above for the `max_connections` is actually misleading. In the above log line, `max_connections` is equal to the number of simulated clients allocated per available core, yet this value is actually overridden later by [`AIOHttpConnection`](https://github.com/elastic/rally/blob/master/esrally/async_connection.py#L171) to a minimum of 256 per Elasticsearch client.
```
maxsize=max(256, kwargs.get("max_connections", 0)),
```

[The documentation is correct](https://esrally.readthedocs.io/en/stable/command_line_reference.html?highlight=max_connection#client-options) in explaining that the `max_connections` value is equal to the number of simulated clients, but _at least_ 256 connections. 

Therefore it is misleading for users to see this value being logged _before_ it is overridden.

b-deam:
With this commit we ensure that the value for max_connections per
    Elasticsearch client is correctly logged in a way that is not misleading
    for users. Prior to this, the value was misleadingly logged as equal to
    the number of simulated clients per worker.

Closes #1278

### DIFF
--- a/esrally/async_connection.py
+++ b/esrally/async_connection.py
@@ -168,7 +168,7 @@ class AIOHttpConnection(elasticsearch.AIOHttpConnection):
                          ssl_assert_fingerprint=ssl_assert_fingerprint,
                          # provided to the base class via `maxsize` to keep base class state consistent despite Rally
                          # calling the attribute differently.
-                         maxsize=max(256, kwargs.get("max_connections", 0)),
+                         maxsize=kwargs.get("max_connections", 0),
                          headers=headers,
                          ssl_context=ssl_context,
                          http_compress=http_compress,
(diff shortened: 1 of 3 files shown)


danielmitterdorfer (approved):
Thanks for the PR; works great! LGTM

Issue #1286, opened 2021-06-21
Transform metrics are red when they decrease.


przemekwitek:
When using:
```
esrally compare
```
on races with `transform-stats` reported, "Transform processing time", "Transform indexing time" and "Transform search time" are red when they decrease.
It should be that the decrease in time is seen as an improvement.

PR #1287, opened 2021-06-21
Do not treat increase in transform processing/indexing/search time as improvement

przemekwitek:
Fixes https://github.com/elastic/rally/issues/1286


danielmitterdorfer (approved):
Thanks for the PR! LGTM

PR #1288, opened 2021-06-22
add transform-stats operation type

hendrikmuhs:
add a transform-stats operation type with a optional condition. The condition can
be used to wait until transform has processed all data.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2327,14 +2327,14 @@ Meta-data
 This operation returns no meta-data.
 
 transform-stats
-~~~~~~~~~~~
+~~~~~~~~~~~~~~~
 
-With the operation type ``transform-stats`` you can call the `transform stats API https://www.elastic.co/guide/en/elasticsearch/reference/current/get-transform-stats.html`_.
+With the operation type ``transform-stats`` you can call the `transform stats API <https://www.elastic.co/guide/en/elasticsearch/reference/current/get-transform-stats.html>`_.
 
 Properties
 """"""""""
 
-* ``transform_id`` (mandatory): The id of the transform.
+* ``transform-id`` (mandatory): The id of the transform.
 * ``condition`` (optional, defaults to no condition): A structured object with the properties ``path`` and ``expected-value``. If the actual value returned by transform stats API is equal to the expected value at the provided path, this operation will return successfully. See below for an example how this can be used.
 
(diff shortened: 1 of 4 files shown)


esrally/driver/runner.py
```diff
@@ -2116,6 +2117,58 @@ def __repr__(self, *args, **kwargs):
...
+                "unit": "ops",
+                "condition": {
+                    "path": path,
+                    # avoid mapping issues in the ES metrics store by always rendering values as strings
+                    "actual-value": self._safe_string(actual_value),
+                    "expected-value": self._safe_string(expected_value)
+                },
+                # currently we only support "==" as a predicate but that might change in the future
+                "success": actual_value == expected_value,
+                "transform-stats": transform_stats
```
b-deam, line 2158: Is it always required to have the transform stats returned? The main concern here is around key/field explosion on the metrics store. 

Would it make sense to either not return it at all, or perhaps could it be returned only if a user passes in a specific parameter? e.g. the `BulkIndex` runner has a `detailed-results` param to determine whether or not we return detailed stats:
https://github.com/elastic/rally/blob/40334f18f92e82686c7a1b4440d53395767d2445/esrally/driver/runner.py#L462-L467

b-deam (commented):
Thanks for the PR @hendrikmuhs! I have a couple of comments, but it looks great.

esrally/driver/runner.py
hendrikmuhs: its not needed and I wasn't aware that this is pushed to the metric store. I removed it.

docs/track.rst
```diff
@@ -2326,6 +2326,41 @@ Meta-data
...
+
+This operation is :ref:`retryable <track_operations>`.
+
+Meta-data
+"""""""""
+
+* ``weight``: Always 1.
+* ``unit``: Always "ops".
+* ``success``: A boolean indicating whether the operation has succeeded.
+* ``transform-stats``: The stats for the given transform id.
```
b-deam, line 2362: As we no longer return the `transform-stats` meta-data, we should remove the docs reference.

b-deam (commented):
Thanks @hendrikmuhs - just a quick docs fix needed and I think we're good to go.

b-deam:
LGTM! Thanks for iterating @hendrikmuhs :)

b-deam (approved):

PR #1292, opened 2021-06-24
Bump Python 3.8 version for dev and CI to 3.8.10

danielmitterdorfer:
With this commit we bump the minimum patch version for Python 3.8 from
3.8.8 to 3.8.10. We do this in order to enable users that rely on pyenv
and use ARM-based Macs as 3.8.10 is the first patch release in the 3.8
series that provides build support for ARM-based Macs (see also
https://www.python.org/downloads/release/python-3810/) for more info.


dliappis (approved):
LGTM.

Once this is merged we should build a new release image.

danielmitterdorfer:
> Once this is merged we should build a new release image.

Thanks! I've published a new release image now.

PR #1293, opened 2021-06-24
fix download from s3/gs by moving http query parameter downstream

hendrikmuhs:
Fix a regression from #1159, the added no kpi parameter breaks downloads from s3/gs. The change adds a check of the protocol in `add_url_param_elastic_no_kpi` and only adds the parameter if its `http`

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -500,7 +500,7 @@ class ElasticsearchDistributionSupplier:
 
     def fetch(self):
         io.ensure_dir(self.distributions_root)
-        download_url = net.add_url_param_elastic_no_kpi(self.repo.download_url)
+        download_url = self.repo.download_url
         distribution_path = os.path.join(self.distributions_root, self.repo.file_name)
         self.logger.info("Resolved download URL [%s] for version [%s]", download_url, self.version)
         if not os.path.isfile(distribution_path) or not self.repo.cache:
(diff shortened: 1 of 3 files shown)


danielmitterdorfer (commented):
Thanks for spotting (and fixing) this! The code in the `net` module is used for downloads in general, such as downloading ES artifacts but also corpora. The logic of disabling KPI tracking makes sense for ES artifacts but not so much for the rest. May I therefore suggest that we really go with your other proposal of checking for the correct scheme before adding the request parameter? Here is a proposed change:

```diff
diff --git a/esrally/utils/net.py b/esrally/utils/net.py
index 4c0a868..7f2ad71 100644
--- a/esrally/utils/net.py
+++ b/esrally/utils/net.py
@@ -194,7 +194,11 @@ def download_http(url, local_path, expected_size_in_bytes=None, progress_indicat
         return expected_size_in_bytes

 def add_url_param_elastic_no_kpi(url):
-    return _add_url_param(url, {"x-elastic-no-kpi": "true"})
+    scheme = urllib3.util.parse_url(url).scheme
+    if scheme.startswith("http"):
+        return _add_url_param(url, {"x-elastic-no-kpi": "true"})
+    else:
+        return url
```

hendrikmuhs:
:+1: I missed that this is also used for downloading data files.

danielmitterdorfer (approved):
Thanks for the fix! LGTM. Let's wait for CI to finish, then I'll merge this.

PR #1294, opened 2021-06-29
Fix composable templates runner

b-deam:
The `put_index_template` method is not available on the `cluster` class,
so with this commit we change this to the correct `indices` class.

See the [Elasticsearch Python Client docs](https://elasticsearch-py.readthedocs.io/en/v7.10.1/api.html#elasticsearch.client.IndicesClient.put_index_template).


danielmitterdorfer (approved):
Good catch! Thanks for fixing this. LGTM

PR #1281, opened 2021-06-08
Format code with black and isort

pquentin:
This is a proposal to switch to black (and isort, as black does not sort imports yet.)

@danielmitterdorfer said this in #1276:

> black is on our radar but I fear that we won't have much capacity at the moment to work on it.

I'm not sure whether it was the actual pull request that was the issue, or the work needed to adapt (eg. everyone will have to configure their editors). My experience is that black allows to focus on the interesting parts of the code, which is highly valuable. Here's what Mike Bayer, creator and maintainer of SQLAlchemy [says about black](https://twitter.com/zzzeek/status/1398662789078650884):

> I would say that I can't think of any single tool in my entire programming career that has given me a bigger productivity increase by its introduction.   commits like https://github.com/sqlalchemy/sqlalchemy/commit/ebbbac0a76b3327a829864afb26ee1b7ff1dc780, Black did 90% of the work for me that previously we'd spend days formatting by hand

This time, lint passes (I was confused because lint fails on master with Python 3.9). The actual black run is in its own commit, so this pull request is easy to rebase. We can also tell git to ignore the black commit in `git blame`.

PR #1298, opened 2021-07-01
Use explicit config files for black and isort

danielmitterdorfer:
Using `pyproject.toml` for black and isort configuration has unwanted
side effects when Rally is installed in certain environments (and in
editable mode). While this should be addressed at a later point because
editable mode is about to removed anyway (see PEP-517 and PEP-518) we
will use explicitly specified config files for now.

Relates #1281
Relates https://github.com/pypa/pip/issues/6433

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,2 @@
+[settings]
+profile = black
--- a/Makefile
+++ b/Makefile
@@ -88,11 +88,11 @@ tox-env-clean:
 
 lint: check-venv
 	@find esrally benchmarks scripts tests it -name "*.py" -exec $(VEPYLINT) -j0 -rn --rcfile=$(CURDIR)/.pylintrc \{\} +
-	@. $(VENV_ACTIVATE_FILE); black --check esrally benchmarks scripts tests it
+	@. $(VENV_ACTIVATE_FILE); black --config=black.toml --check esrally benchmarks scripts tests it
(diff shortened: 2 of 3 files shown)


dliappis (approved):
LGTM thank you

pquentin:
Can you please clarify what the unwanted side effects are?

The problem with a `black.toml` file is that my editor won't see it, so I'd suggest renaming that file back to `pyproject.toml` (and still specifiy it explicitly if that helps).

danielmitterdorfer:
> Can you please clarify what the unwanted side effects are?

We have some internal tooling, for example to run our nightly benchmarks, that builds on top of Rally. It installs Rally master in editable mode. Due to the presence of `pyproject.toml`, `pip` started to prefer that file to `setup.py` and caused build failures ([example](https://elasticsearch-ci.elastic.co/view/All/job/elastic+elasticsearch+master+macrobenchmark-periodic-group-3/80/console)). 

There is definitely room for improvement in how we setup Rally in these scenarios but the most pragmatic solution to unblock us for now was to explicitly configure black and isort via dedicated config files so the current mechanism still works.

pquentin:
Thanks. I believe those lines in `pyproject.toml` are enough to fix the issue:

```
[build-system]
requires = ["setuptools >= 40.6.0", "wheel"]
build-backend = "setuptools.build_meta"
```

But I can't reproduce so I'm not sure. Is there a simple way to see if that fixes the issue?

PR #1304, opened 2021-07-15
Run integration tests with the default distro

danielmitterdorfer:
With this commit we enforce usage of the default distribution in our
integration tests. As this distribution is the default one, we should
also run our integration tests with it.


dliappis (approved):
LGTM thanks

danielmitterdorfer:
Thanks for the quick review!

Issue #1261, opened 2021-05-06
Support exiting of parallel block on first completed task


gingerwizard:
The parallel element provides a completed-by property but this only works if the task that should determine the exit condition is known upfront.

The proposal is that that we add a new property e.g. complete-on-first-task-completed. This would cause a parallel block to exit as soon as one task exits.

This is useful when running queries in parallel for example. As soon as one task exits the load reduces overall, invalidating all subsequent executions by other tasks. This would ensure a clean exit.

Note the naming above is a proposal by @danielmitterdorfer only and we're open to suggested better ones.

DJRickyB:
Has a mixed-type `completed-by` been considered?  an `int` for number of tasks to end before marking completed or a `string` for a named task? Having a single parameter vs two that are mutually exclusive makes the natural default behavior (all tasks complete) a little easier to grok, I would think

danielmitterdorfer:
> Has a mixed-type completed-by been considered?

No, but your proposal made me think whether we should allow the special value `any` for `completed-by` with the meaning that whichever task within a parallel block finishes first, completes the entire parallel block.

PR #1305, opened 2021-07-19
Allow parallel tasks to exit on any completion

b-deam:
With this commit we add the ability to specify a new `any` value
for the `completed-by` param for `parallel` task blocks. This 
allows the `parallel` block to be completed by whichever task is
first to complete successfully.

Closes https://github.com/elastic/rally/issues/1261

Tested with track:
```
{
  "description": "'any' testing ",
  "challenges": [
    {
      "name": "parallel-any",
      "description": "Track completed-by property",
      "schedule": [
        {
          "name": "outside-cluster-health",
          "operation": {
            "operation-type": "cluster-health"
          }
        },
        {
          "parallel": {
            "completed-by": "any",
            "tasks": [
              {
                "name": "check-cluster-health",
                "operation": {
                  "operation-type": "cluster-health"
                }
              },
              {
                "name": "run-node-stats",
                "operation": {
                  "operation-type": "node-stats"
                },
                "iterations": 300000000
              }
            ]
          }
        }
      ]
    }
  ]
}
```

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -593,7 +593,7 @@ All tasks in the ``schedule`` list are executed sequentially in the order in whi
 * ``time-period`` (optional, no default value if not specified): Allows to define a default value for all tasks of the ``parallel`` element.
 * ``warmup-iterations`` (optional, defaults to 0): Allows to define a default value for all tasks of the ``parallel`` element.
 * ``iterations`` (optional, defaults to 1): Allows to define a default value for all tasks of the ``parallel`` element.
-* ``completed-by`` (optional): Allows to define the name of one task in the ``tasks`` list, or the value ``any``. If a specific task name has been provided then as soon as the task has completed the whole ``parallel`` task structure is considered completed. If the value ``any`` is provided, then any task that is first to complete will render the ``parallel`` structure complete. If this property is not explicitly defined, the ``parallel`` task structure is considered completed as soon as all its subtasks have completed. A task is completed if and only if all associated clients have completed execution.
+* ``completed-by`` (optional): Allows to define the name of one task in the ``tasks`` list, or the value ``any``. If a specific task name has been provided then as soon as the named task has completed, the whole ``parallel`` task structure is considered completed. If the value ``any`` is provided, then any task that is first to complete will render the ``parallel`` structure complete. If this property is not explicitly defined, the ``parallel`` task structure is considered completed as soon as all its subtasks have completed (NOTE: this is _not_ true if ``any`` is specified, see below warning and example).
 * ``tasks`` (mandatory): Defines a list of tasks that should be executed concurrently. Each task in the list can define the following properties that have been defined above: ``clients``, ``warmup-time-period``, ``time-period``, ``warmup-iterations`` and ``iterations``.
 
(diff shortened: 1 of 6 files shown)


docs/track.rst
```diff
@@ -604,6 +604,49 @@ All tasks in the ``schedule`` list are executed sequentially in the order in whi
...
+              "parallel": {
+                "completed-by": "any",
+                "tasks": [
+                  {
+                    "name": "bulk-task-1",
+                    "operation": {
+                      "operation-type": "bulk",
+                      "bulk-size": 1000
+                    },
+                  "clients": 8
```
DJRickyB, line 633: nit - indentation
```suggestion
                    "clients": 8
```

docs/track.rst
```diff
@@ -604,6 +604,49 @@ All tasks in the ``schedule`` list are executed sequentially in the order in whi
...
 
+.. warning::
+
+    If the value ``any`` is provided for the ``completed-by`` parameter, the first completion of a client task in the ``parallel`` block will cause all remaining clients and tasks within the ``parallel`` block to immediately exit without waiting for completion.
+
+    For example, given the below track:
+
+        1. Both ``bulk-task-1`` and ``bulk-task-2`` execute in parallel
+        2. ``bulk-task-1`` Client 1/8 is first to complete its assigned partition of work
+        3. ``bulk-task-1`` will now cause the ``parallel`` task to complete and **not** wait for either the remaining 7/8 ``bulk-task-1``'s clients to complete, or for any of ``bulk-task-2``'s clients to complete
```
DJRickyB, line 615: isn't this contrary to
> A task is completed if and only if all associated clients have completed execution.

?

esrally/track/loader.py
```diff
@@ -1497,6 +1496,8 @@ def parse_parallel(self, ops_spec, ops, challenge_name):
                         f"'parallel' element for challenge '{challenge_name}' contains multiple tasks with the "
                         f"name '{completed_by}' marked with 'completed-by' but only task is allowed to match."
                     )
+                elif task.any_completes_parent:
+                    completion_task = "any"
```
danielmitterdorfer, line 1500: We use `completion_task` only as a flag whether there is a completion task so a specific name is actually not required. Can we refactor this to only set a flag instead in both cases (any task completes the parallel structure and specific task completes the parallel structure)?

danielmitterdorfer (commented):
Thanks for the PR! I left a few suggestions. IMHO also a test for the track loader is missing.

b-deam:
I _think_ I addressed the comments, can you take a look when you get a chance?

esrally/track/loader.py
```diff
@@ -1487,17 +1487,17 @@ def parse_parallel(self, ops_spec, ops, challenge_name):
                     )
 
         if completed_by:
-            completion_task = None
+            completion_task = False
```
danielmitterdorfer, line 1490: Nit: Based on the variable name it's not clear that this is a flag. How about `has_completion_task`?

danielmitterdorfer (approved):
Thanks for iterating. I left one nit but other than that LGTM. No need for another review round.

PR #1306, opened 2021-07-19
Fix deleting component/composable templates

gingerwizard:
This fixes an issue with the deletion of component and composable templates. This brings behaviour in line with the documentation. We now support deleting all component/composable templates defined in the track.

The implementation mirrors that of the creation of these templates.


DJRickyB (approved):
LGTM

PR #1308, opened 2021-07-29
always set default client timeout to 60

DJRickyB:
Currently, when no client options are defined, Rally uses a hard-coded default of 60 for the timeout value.  If client options are defined, but `timeout` specifically is omitted, then Rally uses the ES Client's default of 10 for the timeout value. 

This commit causes the default timeout behavior to be 60 seconds in all cases, to simplify behavior for the user.  We use 60 instead of the elasticsearch-py's default of 10 in order to avoid many timeout errors in e.g. bulks and refreshes when there is a lot of load on the benchmark target.

### DIFF
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -17,7 +17,7 @@ If you already have a cluster with data in it you can use the ``create-track`` s
 
 If you want to connect to a cluster with TLS and basic authentication enabled, for example via Elastic Cloud, also specify :ref:`--client-options <clr_client_options>` and change ``basic_auth_user`` and ``basic_auth_password`` accordingly::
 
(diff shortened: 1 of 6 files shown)


danielmitterdorfer:
As discussed offline, let's have a consistent default of 60 seconds for Rally in all cases. Historically, the default was 60 seconds because we wanted to ensure that heavier requests (like bulks) don't run into timeouts on less powerful machines. If we lower the default we'd risk timeouts in these cases which increase the error rate. It's easy to miss and has the potential to mistakenly skew benchmark results if people don't explicitly check the error rate and investigate.

danielmitterdorfer (commented):
With this alignment it will be simpler to reason about timeouts for users, thanks for tackling this. I left one comment but apart from that your change looks good. 

I see you've already updated the PR description but can you reword it to explain the motivation for the change more clearly (along the lines of "unaligned defaults cause confusion")? Keep in mind that Github does not copy the updated PR description to the commit message so you'd need to do this manually when merging your PR (I don't know how often I've fallen for this...). Also, the current PR title is fine but I'd like to offer "Always set default request timeout to 60 seconds" as an alternative title. IMHO it describes your change a bit clearer.

danielmitterdorfer (approved):
Thanks! LGTM

Issue #1309, opened 2021-08-03
Allow parametrization of Rally Elasticsearch datastore index settings


dliappis:
## What/Why

Currently when Rally is configured to use Elasticsearch as the datastore, the settings for the three indices that it creates/pushes results into are hard coded in index templates:

- `rally-metrics`: https://github.com/elastic/rally/blob/3974603b49cdd1bb282204bf83e78fd42192c7a7/esrally/resources/results-template.json#L4-L8
- `rally-results: https://github.com/elastic/rally/blob/3974603b49cdd1bb282204bf83e78fd42192c7a7/esrally/resources/results-template.json#L4-L8
- `rally-races`: https://github.com/elastic/rally/blob/3974603b49cdd1bb282204bf83e78fd42192c7a7/esrally/resources/races-template.json#L4-L7

The current setting for number of primary shards is `1` (and `1` replica, but that's due to Elasticsearch [index setting defaults](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings)).

The problem is that when large datastores can't leverage the power of several nodes and impacts performance when using Kibana dashboards and visualizations to analyze results.

## Proposal

Add a few [datastore properties](https://esrally.readthedocs.io/en/latest/configuration.html#reporting) when `datastore.type==elasticsearch`:

- `datastore.number_of_shards`
- `datastore.number_of_replicas`

that allow the user to change the default behavior.

As `number_of_replicas` (and `refresh_interval`) are dynamic settings, it's tempting to even skip from the configuration file, but perhaps there is value in offering this option for small single node datastores.

## Out of scope

In the future we'll likely leverage [datastreams](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams.html) and new-style [index templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html) as well as [ILM](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-lifecycle-management.html) to manage their lifecycle, but this is out of scope for this ticket.

PR #1310, opened 2021-08-09
Configure persistent datastore index settings

b-deam:
With this commit we add the ability for users to specify the number of
primary and replica shard counts for `rally-*` indices when using a 
persistent metrics store through the configuration of `reporting` settings 
in `rally.ini`.

Closes https://github.com/elastic/rally/issues/1309

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -69,7 +69,7 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.user``: Sets the name of the Elasticsearch user for the metrics store.
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
-* ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ```rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ``rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
 * ``datastore.number_of_replicas`` (default: 0): The number of replicas each primary shard has. Defaults to 0. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
 
 **Examples**
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -69,8 +69,8 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.user``: Sets the name of the Elasticsearch user for the metrics store.
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
-* ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ```rally-*`` indices should have. Any updates to this setting after the first run will only be applied to _new_ ``rally-*`` indices.
-* ``datastore.number_of_replicas`` (default: 0): The number of replicas each primary shard has. Defaults to 0. Any updates to this setting after the first run will only be applied to _new_ ``rally-*`` indices.
+* ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ```rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_replicas`` (default: 0): The number of replicas each primary shard has. Defaults to 0. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
 
 **Examples**
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -69,6 +69,8 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.user``: Sets the name of the Elasticsearch user for the metrics store.
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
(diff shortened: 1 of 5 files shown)


docs/configuration.rst
```diff
@@ -69,6 +69,8 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.user``: Sets the name of the Elasticsearch user for the metrics store.
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
+* ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ``rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_replicas`` (default: 0): The number of replicas each primary shard has. Defaults to 0. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
```
dliappis, line 73: We should also mention in `migrate.rst` that starting from version ... it's possible to configure the number of shards and replicas of Rally metrics indices, when an Elasticsearch store is in use.

esrally/metrics.py
```diff
@@ -250,7 +251,22 @@ def results_template(self):
...
+            # if we're using a persistent metrics store, then we should check and update the index templates settings
+            if self.cfg.opts("reporting", "datastore.type") == "elasticsearch":
+                return self._update_template(f)
+
+            else:
+                return f.read()
+
+    def _update_template(self, template_name):
+        template = json.load(template_name)
+        template["settings"]["index"]["number_of_shards"] = self.cfg.opts(
```
dliappis, line 263: somehow, referencing config parameters directly here feels (to me) slightly out of place.
I'd have defined them as instance variables instead e.g. `self._number_of_shards = self.cfg.opts("reporting", "datastore.number_of_shards", default_value=1, mandatory=False)` and `self._number_of_replicas = ...` for better readability and discoverability but obviously what you have here works fine too.

esrally/metrics.py
```diff
@@ -250,7 +251,22 @@ def results_template(self):
 
     def _read(self, template_name):
         with open("%s/resources/%s.json" % (self.script_dir, template_name), encoding="utf-8") as f:
-            return f.read()
+            # if we're using a persistent metrics store, then we should check and update the index templates settings
+            if self.cfg.opts("reporting", "datastore.type") == "elasticsearch":
```
dliappis, line 255: This class doesn't get called when there is an in-memory metrics store; the references are always from EsXYZStore i.e.  https://github.com/elastic/rally/blob/ede9670d50f1d0330baf09738a01e6535b17a49f/esrally/metrics.py#L862 / https://github.com/elastic/rally/blob/ede9670d50f1d0330baf09738a01e6535b17a49f/esrally/metrics.py#L1607 / https://github.com/elastic/rally/blob/ede9670d50f1d0330baf09738a01e6535b17a49f/esrally/metrics.py#L1700

So we don't need this conditional. Therefore, we can also move the code inside `_update_template()` here and reference the shard settings from instance variables, see the comment below.

dliappis (commented):
This looks good and works fine in tests. Left a few comments.

esrally/config.py
```diff
@@ -210,7 +210,7 @@ def exists(self, section, key):
 
     def config_present(self):
         """
-        :return: true iff a config file already exists.
+        :return: true if a config file already exists.
```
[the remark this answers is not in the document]
b-deam: Wow, TIL.

esrally/metrics.py
```diff
@@ -238,6 +238,7 @@ class IndexTemplateProvider:
 
     def __init__(self, cfg):
         self.script_dir = cfg.opts("node", "rally.root")
+        self.cfg = cfg
```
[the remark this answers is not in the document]
b-deam: I agree, `self._config` seems better fitting given the config is really only ever intended to be used privately.

esrally/metrics.py
b-deam: Good pick up!

b-deam:
I made the changes - but given that it _shouldn't_ be possible for an instance of `InMemoryMetricsStore` to have an `index_template_provider_class` attribute I wasn't sure on how to test this.

dliappis:
> I made the changes - but given that it _shouldn't_ be possible for an instance of `InMemoryMetricsStore` to have an `index_template_provider_class` attribute I wasn't sure on how to test this.

Additional settings like `datastore.host` simply get ignored when `datastore.type != elasticsearch` so I am not sure what you are aiming to test?

b-deam:
> > I made the changes - but given that it _shouldn't_ be possible for an instance of `InMemoryMetricsStore` to have an `index_template_provider_class` attribute I wasn't sure on how to test this.
> 
> Additional settings like `datastore.host` simply get ignored when `datastore.type != elasticsearch` so I am not sure what you are aiming to test?

Right - I guess I want to ensure that if any changes are made to the `InMemoryMetricsStore` in so that if it ever implements/utilises a `index_template_provider_class`, then I guess we should make sure that we catch that in unit tests because all of the logic around  is currently based on the assumption that it will only ever be executed in the context of a `EsMetricsStore`.

dliappis:
>  if any changes are made to the `InMemoryMetricsStore` in so that if it ever implements/utilises a `index_template_provider_class`

I don't think this will ever happen. @danielmitterdorfer do you feel we need a unit test for this unlikely scenario?

danielmitterdorfer:
> I don't think this will ever happen. @danielmitterdorfer do you feel we need a unit test for this unlikely scenario?

No, that will not happen. The index template provider is designed to be used only with an `EsMetricsStore`.

b-deam:
Thanks @danielmitterdorfer, @dliappis. 

I had the misconception that 'In-memory' == local metrics store - so it's obvious now that we'd never implement a template provider 🤦 .

I removed the unit test for the `InMemoryMetricsStore`. Will merge after CI passes.

dliappis (approved):
LGTM

PR #1311, opened 2021-08-11
Show race id in console

danielmitterdorfer:
Users can retrieve details about a race from the metrics store or via
Rally subcommands via the race id. While we log the race id at the
beginning of a race such an important piece of information should be
more visible to users. Therefore, we print the race id also at the
beginning of a race to the console.


dliappis (approved):
Very useful addition. LGTM.

PR #1312, opened 2021-08-13
Update rally-* index template settings to default to Elasticsearch defaults 

b-deam:
~~In https://github.com/elastic/rally/issues/1309 we added
configurable shard and replica shard counts to `rally.ini`.
In that commit, the default number of replicas was set to 0,
which is trappy given that Elasticsearch defaults to 1 replica.~~

~~This commit changes the default behaviour to add 1 replica to
the `rally-*` indices by default.~~

See comment: https://github.com/elastic/rally/pull/1312#issuecomment-899177300

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -69,8 +69,8 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.user``: Sets the name of the Elasticsearch user for the metrics store.
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
-* ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ``rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
-* ``datastore.number_of_replicas`` (default: 1): The number of replicas each primary shard has. Defaults to 1. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_shards`` (default: `Elasticsearch default value <https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings>`_): The number of primary shards that the ``rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_replicas`` (default: `Elasticsearch default value <https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings>`_): The number of replicas each primary shard has. Defaults to . Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
 
 **Examples**
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -70,7 +70,7 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
 * ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ``rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
-* ``datastore.number_of_replicas`` (default: 0): The number of replicas each primary shard has. Defaults to 0. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_replicas`` (default: 1): The number of replicas each primary shard has. Defaults to 1. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
 
 **Examples**
 
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -258,7 +258,7 @@ class IndexTemplateProvider:
                 if int(self._number_of_shards) < 1:
                     raise exceptions.SystemSetupError(
                         f"The setting: datastore.number_of_shards must be >= 1. Please "
-                        f"check the configuration in {self._config.config_file.location}/rally.ini"
+                        f"check the configuration in {self._config.config_file.location}"
                     )
                 template["settings"]["index"]["number_of_shards"] = int(self._number_of_shards)
             if self._number_of_replicas is not None:
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -255,9 +255,14 @@ class IndexTemplateProvider:
         with open("%s/resources/%s.json" % (self.script_dir, template_name), encoding="utf-8") as f:
             template = json.load(f)
             if self._number_of_shards is not None:
-                template["settings"]["index"]["number_of_shards"] = self._number_of_shards
+                if int(self._number_of_shards) < 1:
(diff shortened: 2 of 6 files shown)


dliappis:
As discussed offline let's change the logic to default to `None` and only attempt to override the number of shards and number of replicas if they've been explicitly set.
Additionally we'll need to remove the explicit `"number_of_shards":1` from the three template files, e.g. [here](https://github.com/elastic/rally/blob/27b97b5e645b546f837f5ff81d5fdd92624fe0c5/esrally/resources/results-template.json#L6). With these changes implemented, metric stores with a default `rally.ini` will use the Elasticsearch index defaults for the `rally-*` indices (i.e. `1p/1r`) and allow the user to override any of the two settings as needed.

b-deam:
As discussed, I've reworked the PR to meet the new requirements we discussed above ^^

The number of primary and replica shards are still configurable via `rally.ini`, but instead of explicitly settings defaults we fallback and allow Elasticsearch to set the defaults. We 'inject' these values into the respective rally metrics/races/results templates if a user has specified either or both `datastore.number_of_shards`/`datastore.number_of_replicas`.

In addition, we also remove the hardcoded `refresh_interval: 5s` setting to allow us to take advantage of [Elasticsearch's default refresh behaviour](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#dynamic-index-settings):
> `index.refresh_interval`
> How often to perform a refresh operation, which makes recent changes to the index visible to search. Defaults to 1s. Can be set to -1 to disable refresh. If this setting is not explicitly set, shards that haven’t seen search traffic for at least index.search.idle.after seconds will not receive background refreshes until they receive a search request. Searches that hit an idle shard where a refresh is pending will wait for the next background refresh (within 1s). This behavior aims to automatically optimize bulk indexing in the default case when no searches are performed. In order to opt out of this behavior an explicit value of 1s should set as the refresh interval.

docs/configuration.rst
```diff
@@ -70,7 +70,7 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
 * ``datastore.password``: Sets the password of the Elasticsearch user for the metrics store.
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
 * ``datastore.number_of_shards`` (default: 1): The number of primary shards that the ``rally-*`` indices should have. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
-* ``datastore.number_of_replicas`` (default: 0): The number of replicas each primary shard has. Defaults to 0. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
+* ``datastore.number_of_replicas`` (default: 1): The number of replicas each primary shard has. Defaults to 1. Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices.
```
dliappis, line 73: Should both of these be changed regarding the documents of the defaults to instead say:

> defaults to [Elasticsearch default value](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings)

esrally/metrics.py
```diff
@@ -254,8 +254,10 @@ def results_template(self):
     def _read(self, template_name):
         with open("%s/resources/%s.json" % (self.script_dir, template_name), encoding="utf-8") as f:
             template = json.load(f)
-            template["settings"]["index"]["number_of_shards"] = self._number_of_shards
-            template["settings"]["index"]["number_of_replicas"] = self._number_of_replicas
+            if self._number_of_shards is not None:
+                template["settings"]["index"]["number_of_shards"] = self._number_of_shards
+            if self._number_of_replicas is not None:
```
dliappis, line 259: Very good that you didn't `if self._number_of_shards` here (and above) because explicitly setting e.g. number of replicas to `0` would be evaluated as false and not set the value!

dliappis (commented):
Looks good, left a comment about making the docs about the default value more factual.

esrally/metrics.py
```diff
@@ -238,8 +238,8 @@ class IndexTemplateProvider:
 
     def __init__(self, cfg):
         self._config = cfg
-        self._number_of_shards = self._config.opts("reporting", "datastore.number_of_shards", default_value=1, mandatory=False)
-        self._number_of_replicas = self._config.opts("reporting", "datastore.number_of_replicas", default_value=0, mandatory=False)
+        self._number_of_shards = self._config.opts("reporting", "datastore.number_of_shards", default_value=None, mandatory=False)
+        self._number_of_replicas = self._config.opts("reporting", "datastore.number_of_replicas", default_value=None, mandatory=False)
```
dliappis, line 242: Let's use `int()` here as `ConfigParser()` always stores values [internally as strings](https://docs.python.org/3/library/configparser.html#supported-datatypes).

b-deam:
Alrighty - I think I've addressed the comments and added a couple new tests. Can you take a look?

esrally/metrics.py
```diff
@@ -254,8 +254,15 @@ def results_template(self):
     def _read(self, template_name):
         with open("%s/resources/%s.json" % (self.script_dir, template_name), encoding="utf-8") as f:
             template = json.load(f)
-            template["settings"]["index"]["number_of_shards"] = self._number_of_shards
-            template["settings"]["index"]["number_of_replicas"] = self._number_of_replicas
+            if self._number_of_shards is not None:
+                if int(self._number_of_shards) < 1:
+                    raise exceptions.SystemSetupError(
+                        f"The setting: datastore.number_of_shards must be >= 1. Please "
+                        f"check the configuration in {paths.rally_confdir()}/rally.ini"
```
dliappis, line 261: Note that due to undocumented parameter [configuration-name](https://github.com/elastic/rally/blob/aa1a89395783f712598d1a3bfd094a81ebeb1aa4/esrally/rally.py#L677) other ini files maybe references, e.g. in IT tests and elsewhere. Use `{self._config.config_file.location}` instead.

dliappis (changes requested):
This looks good, but we need to print the right ini file, honoring custom ini file invocations too.

esrally/metrics.py
b-deam: I ended up changing this to point to regex (`.*/rally.ini`) instead, is that acceptable here?

esrally/metrics.py
```diff
@@ -254,8 +254,15 @@ def results_template(self):
     def _read(self, template_name):
         with open("%s/resources/%s.json" % (self.script_dir, template_name), encoding="utf-8") as f:
             template = json.load(f)
-            template["settings"]["index"]["number_of_shards"] = self._number_of_shards
-            template["settings"]["index"]["number_of_replicas"] = self._number_of_replicas
+            if self._number_of_shards is not None:
+                if int(self._number_of_shards) < 1:
+                    raise exceptions.SystemSetupError(
+                        f"The setting: datastore.number_of_shards must be >= 1. Please "
+                        f"check the configuration in {config.ConfigFile().location}/rally.ini"
```
dliappis, line 261: This won't work, it will instantiate a new `ConfigFile()` object with `None` as `config_name` (see https://github.com/elastic/rally/blob/f4302fbbc38bf4e7c00cd43ac9b363f26e579196/esrally/config.py#L43). Therefore it'll always return `rally.ini`.

We need to use `self._config.config_file.location` as mentioned in the [previous comment](https://github.com/elastic/rally/pull/1312#discussion_r689351906).

dliappis (changes requested):
There's a small bug left.

dliappis (approved):
LGTM

PR #1313, opened 2021-08-13
Use bridge network in container mode

danielmitterdorfer:
With this commit we configure a bridge network that is only visible from
the IP address the Elasticsearch node should bind to.

### DIFF
--- a/tests/mechanic/provisioner_test.py
+++ b/tests/mechanic/provisioner_test.py
@@ -676,7 +676,7 @@ class DockerProvisionerTests(TestCase):
         docker_cfg = docker._render_template_from_file(docker.docker_vars(mounts={}))
 
         self.assertEqual(
-            f"""version: '2.2'
+            f'''version: '2.2'
 services:
   elasticsearch1:
     cap_add:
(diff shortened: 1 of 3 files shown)


tests/mechanic/provisioner_test.py
```diff
@@ -690,22 +692,28 @@ def test_provisioning_with_defaults(self, uuid4):
...
       timeout: 2s
-      retries: 10"""
-            % (data_dir, log_dir, heap_dump_dir),
+      retries: 10
+    networks:
+      - rally-es
+networks:
+  rally-es:
+    driver_opts:
+      com.docker.network.bridge.host_binding_ipv4: "{node_ip}\"""",
```
dliappis, line 708: not: to avoid this backslashing we can also change the f-string from `"""` to `'''`.

dliappis (approved):
LGTM

tests/mechanic/provisioner_test.py
danielmitterdorfer: Good point. Addressed in https://github.com/elastic/rally/pull/1313/commits/46786ddca5d67efd9ee8df3c7b05fe2872e2281b.

PR #1315, opened 2021-08-20
Fix load track behavior on multiple load driver machines

### LINKED ISSUE #1206: Track preparation can fail on multiple load generator machines with custom track processors
**Actual behavior**

**Steps to reproduce**:
 1. Implement a track that uses a track processor which adds instances of custom classes (not in Rally's default module load path) to the track in the `on_after_load_track` phase.
 2. Start a benchmark using `--load-driver-hosts`.
 3. Observe that Rally is hanging after starting the benchmark.

**Relevant logs**

In the actor system log we see:

```
p29529 I    Pending Actor request received for esrally.driver.driver.TrackPreparationActor reqs {'ip': '10.10.30.79'} from ActorAddr-(T|:42161)
p29529 Warn no system has compatible capabilities for Actor esrally.driver.driver.TrackPreparationActor
p29529 I    Requesting creation of esrally.driver.driver.TrackPreparationActor on remote admin ActorAddr-(T|10.10.30.79:1900)
p30298 I    completion error: ************* TransportIntent(ActorAddr-(T|10.10.30.79:32999)-=-BadPacketError-<class 'esrally.driver.driver.PrepareTrack'>-<esrally.driver.driver.PrepareTrack object at 0x7f04968b5fa0>-quit_0:04:59.841806)
```

This points us towards an issue during track preparation.

Furthermore we see in the logs:

```
ERR  OUCH!  Error deserializing received data: Traceback (most recent call last):
  File “/home/esbench/.local/lib/python3.8/site-packages/thespian/system/transport/TCPTransport.py”, line 1433, in _addedDataToIncoming
    rdata, extra = inc.data
  File “/home/esbench/.local/lib/python3.8/site-packages/thespian/system/transport/TCPTransport.py”, line 186, in data
    def data(self): return self._rData.completed()
  File “/home/esbench/.local/lib/python3.8/site-packages/thespian/system/transport/streamBuffer.py”, line 80, in completed
    return self._deserialize(self._buf), self._extra
ModuleNotFoundError: No module named ‘logs’
  (rdata=“”, extra=“”)
```

**Analysis**

The issue is that the actor system serializes the `PrepareTrack` message on the coordinator node, which has also loaded the track and sends it then to all hosts specified via `--load-driver-hosts`. However, the track including the custom classes has not been loaded yet, thus if the `track` object in the `PrepareTrack` message references a class that has not been loaded yet on that machine, deseralizing the message fails.

DJRickyB:
This commit fixes #1206 by causing the TrackPreparationActor and the Driver actor to register the local track repository/code prior to attempting to receive the loaded track as a serialized message.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -141,12 +141,10 @@ class StartWorker:
     def __init__(self, worker_id, config, track, client_allocations):
         """
         :param worker_id: Unique (numeric) id of the worker.
-        :param config: Rally internal configuration object.
         :param track: The track to use.
         :param client_allocations: A structure describing which clients need to run which tasks.
         """
         self.worker_id = worker_id
-        self.config = config
         self.track = track
         self.client_allocations = client_allocations
 
@@ -303,6 +301,7 @@ class DriverActor(actor.RallyActor):
         return self.createActor(Worker, targetActorRequirements=self._requirements(host))
 
     def start_worker(self, driver, worker_id, cfg, track, allocations):
+        self.send(driver, RallyConfig(cfg))
         self.send(driver, StartWorker(worker_id, cfg, track, allocations))
 
     def drive_at(self, driver, client_start_timestamp):
@@ -1116,11 +1115,14 @@ class Worker(actor.RallyActor):
         self.sample_queue_size = None
 
     @actor.no_retry("worker")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_RallyConfig(self, msg):
+        self.config = load_local_config(msg.config)
+
+    @actor.no_retry("worker")  # pylint: disable=no-value-for-parameter
     def receiveMsg_StartWorker(self, msg, sender):
         self.logger.info("Worker[%d] is about to start.", msg.worker_id)
         self.master = sender
         self.worker_id = msg.worker_id
-        self.config = load_local_config(msg.config)
         self.on_error = self.config.opts("driver", "on.error")
         self.sample_queue_size = int(self.config.opts("reporting", "sample.queue.size", mandatory=False, default_value=1 << 20))
         self.track = msg.track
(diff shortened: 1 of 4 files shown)


DJRickyB:
Further testing seems to indicate some silent hanging, I will circle back to this on Monday.  Feedback welcome on current state

esrally/driver/driver.py
```diff
@@ -71,18 +71,25 @@ class StartBenchmark:
     pass
 
 
+class RallyConfig:
+    """
+    Prompts loading of a track
```
dliappis, line 76: It's a bit unclear to me what this is actually doing, by just reading the docstring. I fail to understand how an object instantiating this that stores `cfg` prompts (as in causes/brings about?) the loading of a track. Is there anything short we could add here to make this easier to understand in the future without having to git blame and reference the issue https://github.com/elastic/rally/issues/1206?

dliappis (commented):
Thank you fixing this! I did an initial pass, looking only at the code, which looks good; left a few minor comments.

Will test a few scenarios, as the next step.

esrally/driver/driver.py
DJRickyB: Great point. I think I will go the "self-documenting" route here.  Other message classes are verbs, telling the actor what to do, this one would be better named "Bootstrap" and could in the future denote other activity actors should be doing before their intended work.

esrally/driver/driver.py
```diff
@@ -135,7 +142,6 @@ class StartWorker:
     def __init__(self, worker_id, config, track, client_allocations):
         """
         :param worker_id: Unique (numeric) id of the worker.
-        :param config: Rally internal configuration object.
```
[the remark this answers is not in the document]
DJRickyB: whoops! adding it back.  I had removed `config` from this message altogether in an earlier iteration

dliappis (approved):
I've now concluded tests, things seem to be working fine with large loaddriver clusters.
Also the latest commits after my previous comments are great.
LGTM.

PR #1319, opened 2021-08-31
Add support for listing nested tracks

### LINKED ISSUE #1086: Support Nested Tracks
Currently tracks can be nested and most functions work. This isn't by design e.g.

Consider the folder hierarchy

```
rally-tracks
  observability
    logging
    metrics
    apm
```

Most rally commands would work (including the ability to use the track) with the `track` parameter e.g.

`--track="observability/logging"`

Some commands such as `esrally list tracks` will not.

Do we want to support nested tracks? If so, where are the current deficiencies in support?

DJRickyB:
with this commit we add support for listing nested tracks in git-based track repositories.  Previously, `esrally list tracks` would only list top-level directories in a target track repo which contained a `track.json`.  This change causes full recursive search in the track repository, to better support the idea of "nested" track structures.

Fixes #1086

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -329,7 +329,13 @@ class GitTrackRepository:
 
     @property
     def track_names(self):
-        return filter(lambda p: os.path.exists(self.track_file(p)), next(os.walk(self.repo.repo_dir))[1])
+        retval = []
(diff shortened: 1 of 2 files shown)


ebadyano (approved):
LGTM, thank you for adding this

PR #1321, opened 2021-09-07
remove coverage dependency

DJRickyB:
This commit removes the rarely used `coverage` dependency from Rally, perhaps to be re-introduced at a later date


DJRickyB:
we should either go this route or #1322, looking for input from reviewers

dliappis (approved):
LGTM

I'd rather go this way if we need to unblock ci soon

DJRickyB:
actually, just realized both actions may still be good ideas anyway, open to that suggestion as well in the next couple of days

PR #1322, opened 2021-09-07
pin setuptools and wheel versions to working point-in-time

DJRickyB:
We encountered a breaking change by going with latest setuptools in our CI.  This commit pins the version of setuptools to the last known functioning version.

We should take this approach or the approach in #1321, and close the other.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ check-venv:
 	fi
 
 install-user: venv-create
-	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install --upgrade pip setuptools==58.0.1 wheel=0.37.0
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -56,7 +56,7 @@ check-venv:
 	fi
 
 install-user: venv-create
-	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install --upgrade pip setuptools wheel
+	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install --upgrade pip setuptools==58.0.1 wheel==0.37.0
```
pquentin, line 59: Can you please pin `pip` as well? In my experience that's the most actively developed one, and they regularly introduce breaking changes

DJRickyB:
Closing this in favor of a less reactive approach, now that we're past CI issues

PR #1326, opened 2021-09-09
Improve documentation in various ways

pquentin:
As I'm making progress through the docs, I'm seeing various little things that can be improved. Since you may not agree about all this, I split it into multiple commits to make it easy to cut things out.


docs/install.rst
```diff
@@ -112,7 +112,7 @@ To find the JDK, Rally expects the environment variable ``JAVA_HOME`` to be set
 Optional dependencies
 ---------------------
 
-S3 support is optional and can be installed using the ``s3`` extra. If you need S3 support, install ``esrally[s3]`` instead of just ``esrally``, but other than that follow the instructions below.
+S3 support is optional and can be installed using the ``esrally[s3]`` instead of just ``esrally``, but other than that follow the instructions below.
```
DJRickyB, line 115: imo this becomes less clear now, specifically the phrase `using ... esrally[s3] instead of just esrally`, and I'd prefer to revert 4fdcfef

pquentin: Agreed, reverted

DJRickyB (approved):
LGTM

Issue #1324, opened 2021-09-09
Collect master node during benchmark execution


b-deam:
It'd be handy to be able to regularly sample/collect the current elected master node during the execution of our benchmarks. This would mainly assist in performance analysis by allowing us to observe which node was acting as master throughout the duration of a benchmark.

[The ES client already exposes this via the `Cat` client](https://elasticsearch-py.readthedocs.io/en/v7.14.1/api.html?highlight=master%20node#elasticsearch.client.CatClient.master).

I propose we add either an internal, always on telemetry device to collect this on a regular sample throughout the benchmark. The functionality is similar to that implemented in `ClusterEnvironmentInfo`, however this is a static telemetry device that is only run once at the start of a benchmark.

pquentin:
Since the cat API is designed for human consumption, we're going to use the [nodes info API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html) instead. Is it enough to store the name of the master node, or should we also store its ip and host?

b-deam:
> Since the cat API is designed for human consumption, we're going to use the [nodes info API](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html) instead. Is it enough to store the name of the master node, or should we also store its ip and host?

I'm not sure that the nodes info API contains the current master, and instead just shows whether or not that node has been assigned the `master` _role_ (i.e. it _can_ be a master).

If we don't want to use the cat API, then we should fetch it directly from the cluster state [via `ClusterClient.state`](https://elasticsearch-py.readthedocs.io/en/v7.14.1/api.html?highlight=master%20node#elasticsearch.client.ClusterClient.state), e.g. `elasticsearch.client.cluster.master`. This will give you back a node ID that you'll need to use to lookup the actual node _name_.

PR #1330, opened 2021-09-15
Add master node telemetry device

pquentin:
When running benchmarks on clusters with multiple nodes, the master node
could perform worse due to the additional responsibilities it needs to
handle.

To identify this, we add a telemetry device to record the master node.
We don't expect this to change during a typical benchmark, which is why
the default sample interval is 30s. To compensate this large sample
interval, we also record that information on benchmark start/end.

I tested this feature with a short sample interval in test mode, checking that the information was indeed recorded in the test store:

```
esrally race --distribution-version=7.14.1 --car="4gheap" --track=geonames --test-mode --telemetry master-node-stats --telemetry-params="master-node-stats-sample-interval: 0.1"
http localhost:9200/rally-metrics-2021-09/_search\?size=1000 | jq . | grep master-node -c 
```

TODO:

- [x] #1324 mentioned that this device should be always on, how can I do that?
- [x] Is it OK to use those APIs to retrieve the information, or should we revert to the cat API which is much less verbose? (Thanks @b-deam for the help on this!)
- [x] I recorded the master at the start and end of a benchmark, is that OK?
- [x] Fix CI

### DIFF
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -30,6 +30,7 @@ You probably want to gain additional insights from a race. Therefore, we have ad
    searchable-snapshots-stats  Searchable Snapshots Stats  Regularly samples searchable snapshots stats
    shard-stats                 Shard Stats                 Regularly samples nodes stats at shard level
    data-stream-stats           Data Streams Stats          Regularly samples data streams stats
+   master-node-stats           Master Node Stats           Regularly determines the current master node name
 
    Keep in mind that each telemetry device may incur a runtime overhead which can skew results.
 
@@ -227,3 +228,19 @@ Example of recorded documents given two data streams in the cluster::
 Supported telemetry parameters:
 
 * ``data-stream-stats-sample-interval`` (default 10): A positive number greater than zero denoting the sampling interval in seconds.
+
+
+master-node-stats
+-----------------
+
+The master-node-stats telemetry device regularly reports the name of the master node by using:
+
+ 1. the `Cluster State API <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-state.html>`_, specifically `/_cluster/state/master_node` to determine the id of the master node
+ 2. the `Nodes info API <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html>`_ to determine its name
+
+Example of recorded document::
+
+   {
+     "name": "master-node-stats",
+     "node": "rally-0"
+   }
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -227,19 +227,3 @@ Example of recorded documents given two data streams in the cluster::
 Supported telemetry parameters:
 
 * ``data-stream-stats-sample-interval`` (default 10): A positive number greater than zero denoting the sampling interval in seconds.
-
-
-master-node-stats
------------------
-
-The master-node-stats telemetry device regularly reports the name of the master node by using:
-
- 1. the `Cluster State API <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-state.html>`_, specifically `/_cluster/state/master_node` to determine the id of the master node
- 2. the `Nodes info API <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-nodes-info.html>`_ to determine its name
-
-Example of recorded document::
-
-   {
-     "name": "master-node-stats",
-     "node": "rally-0"
-   }
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -30,7 +30,6 @@ You probably want to gain additional insights from a race. Therefore, we have ad
    searchable-snapshots-stats  Searchable Snapshots Stats  Regularly samples searchable snapshots stats
    shard-stats                 Shard Stats                 Regularly samples nodes stats at shard level
    data-stream-stats           Data Streams Stats          Regularly samples data streams stats
-   master-node-stats           Master Node Stats           Regularly determines the current master node name
 
    Keep in mind that each telemetry device may incur a runtime overhead which can skew results.
 
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -610,6 +610,7 @@ class Driver:
                 telemetry.TransformStats(telemetry_params, es, self.metrics_store),
                 telemetry.SearchableSnapshotsStats(telemetry_params, es, self.metrics_store),
                 telemetry.DataStreamStats(telemetry_params, es, self.metrics_store),
+                telemetry.MasterNodeStats(telemetry_params, es, self.metrics_store),
             ]
         else:
             devices = []
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -610,7 +610,7 @@ class Driver:
                 telemetry.TransformStats(telemetry_params, es, self.metrics_store),
                 telemetry.SearchableSnapshotsStats(telemetry_params, es, self.metrics_store),
                 telemetry.DataStreamStats(telemetry_params, es, self.metrics_store),
-                telemetry.MasterNodeStats(telemetry_params, es, self.metrics_store),
+                telemetry.MasterNodeStats(telemetry_params, es_default, self.metrics_store),
             ]
         else:
             devices = []
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -603,6 +603,7 @@ class Driver:
                 telemetry.JvmStatsSummary(es_default, self.metrics_store),
                 telemetry.IndexStats(es_default, self.metrics_store),
                 telemetry.MlBucketProcessingTime(es_default, self.metrics_store),
+                telemetry.MasterNodeStats(telemetry_params, es_default, self.metrics_store),
                 telemetry.SegmentStats(log_root, es_default),
                 telemetry.CcrStats(telemetry_params, es, self.metrics_store),
                 telemetry.RecoveryStats(telemetry_params, es, self.metrics_store),
@@ -610,7 +611,6 @@ class Driver:
                 telemetry.TransformStats(telemetry_params, es, self.metrics_store),
                 telemetry.SearchableSnapshotsStats(telemetry_params, es, self.metrics_store),
                 telemetry.DataStreamStats(telemetry_params, es, self.metrics_store),
-                telemetry.MasterNodeStats(telemetry_params, es_default, self.metrics_store),
             ]
         else:
             devices = []
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -1939,46 +1939,41 @@ class MasterNodeStats(TelemetryDevice):
     human_name = "Master Node Stats"
     help = "Regularly samples master node name"
 
-    def __init__(self, telemetry_params, clients, metrics_store):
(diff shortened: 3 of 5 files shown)


tests/telemetry_test.py
```diff
@@ -3664,3 +3669,78 @@ def test_stores_nothing_if_no_data_path(self, run_subprocess, metrics_store_clus
...
+        clients = {"default": Client(), "cluster_b": Client()}
+        cfg = create_config()
+        metrics_store = metrics.EsMetricsStore(cfg)
+        telemetry_params = {"master-node-stats-sample-interval": 0.001}
+        t = telemetry.MasterNodeStats(telemetry_params, clients, metrics_store)
+        t.on_benchmark_start()
+        t.on_benchmark_stop()
+
+        # the two clients record start and stop
+        assert record_count == 4
```
pquentin, line 3695: For what it's worth I tried using `@mock.patch` and `call_count` here but I was only seeing 2 calls, not 4. I still don't know why. I ended up switching to the pytest style which is more verbose but less magical.

pquentin:
> Is it OK to use those APIs to retrieve the information, or should we revert to the cat API which is much less verbose? (Thanks @b-deam for the help on this!)a_Run

@sethmlarson mentioned to me that we can use `format="json"` for the cat API, and it's stable, so maybe that's best in this case (only one call, less data to transfer). Will let you decide, switching is easy in any case.

esrally/telemetry.py
```diff
@@ -1926,3 +1927,104 @@ def detach_from_node(self, node, running):
...
+    def record(self):
+        """
+        Collect master node name and push to metrics store.
+        """
+        # pylint: disable=import-outside-toplevel
+        import elasticsearch
+
+        try:
+            state = self.client.cluster.state(metric="master_node")
+            info = self.client.nodes.info(node_id=state["master_node"], metric="os")
```
pquentin, line 2017: I don't care about the `os` metrics here, in fact the name is included in all cases, but there's no documented way to ask for "no metrics", and `os` seemed cheap to retrieve.

b-deam:
> > Is it OK to use those APIs to retrieve the information, or should we revert to the cat API which is much less verbose? (Thanks @b-deam for the help on this!)a_Run
> 
> @sethmlarson mentioned to me that we can use `format="json"` for the cat API, and it's stable, so maybe that's best in this case (only one call, less data to transfer). Will let you decide, switching is easy in any case.

IMO that `cat` approach is certainly easiest. If @sethmlarson says it's stable, then I'd trust him ;-), in either case I'd probably want to avoid extraneous API calls (e.g. to Node Info or others).

esrally/telemetry.py
b-deam: I think the `cat` API is the way to go here IMO. But in any case if we decide not to use the `cat` API I wonder if it's possible to retrieve the node IDs from existing metrics store metadata rather than making regular API calls? 

e.g. we collect some of this upfront here https://github.com/elastic/rally/blob/4355771a7d9fa027b83331afc2d5e669942b9a0e/esrally/telemetry.py#L1561-L1571)

esrally/driver/driver.py
```diff
@@ -610,6 +610,7 @@ def prepare_telemetry(self, es, enable):
                 telemetry.TransformStats(telemetry_params, es, self.metrics_store),
                 telemetry.SearchableSnapshotsStats(telemetry_params, es, self.metrics_store),
                 telemetry.DataStreamStats(telemetry_params, es, self.metrics_store),
+                telemetry.MasterNodeStats(telemetry_params, es, self.metrics_store),
```
b-deam, line 613: As above, if you want this to be an internal telemetry device that runs by default then you can just pass `es_default` in as below:
https://github.com/elastic/rally/blob/4355771a7d9fa027b83331afc2d5e669942b9a0e/esrally/driver/driver.py#L601-L603

docs/telemetry.rst
```diff
@@ -30,6 +30,7 @@ You probably want to gain additional insights from a race. Therefore, we have ad
    searchable-snapshots-stats  Searchable Snapshots Stats  Regularly samples searchable snapshots stats
    shard-stats                 Shard Stats                 Regularly samples nodes stats at shard level
    data-stream-stats           Data Streams Stats          Regularly samples data streams stats
+   master-node-stats           Master Node Stats           Regularly determines the current master node name
```
b-deam, line 33: I think the intention was to make this an internal telemetry device, so no need to expose any info of command line arguments here. See https://github.com/elastic/rally/blob/4355771a7d9fa027b83331afc2d5e669942b9a0e/esrally/telemetry.py#L1523 for an example.

b-deam (commented):
> Collect master node during benchmark execution #1324 mentioned that this device should be always on, how can I do that?

See comments around setting this as an `InternalTelemetryDevice`.

>  Is it OK to use those APIs to retrieve the information, or should we revert to the cat API which is much less verbose? (Thanks @b-deam for the help on this!)

As per comments RE: `cat` API :-)

> I recorded the master at the start and end of a benchmark, is that OK?

I'd say it's preferable to sample this semi-regularly. 30s seems like a good compromise, we don't expect master nodes to change much/if at all during a benchmark, and if it does that usually indicates some stability issues.

esrally/telemetry.py
pquentin: Resolving as I switched to the cat API. (For what it's worth, `add_metadata_for_node` seems to only be called at the end for the moment, but I haven't explored further.)

esrally/driver/driver.py
pquentin: I did that, thanks! But note that `es_default` vs. `es` is not about about internal/always-on vs. optional: https://elastic.slack.com/archives/C0JGTNK0S/p1631778220222700

docs/telemetry.rst
pquentin: Thanks, fixed

pquentin:
I addressed all review comments, please take another look

pquentin:
ℹ️ A seemingly unrelated test fails in a deterministic way on CI, both for Python 3.8 and Python 3.9, but I don't reproduce the failure locally yet.

<details>
<summary>Test failure output</summary>

```
16:26:23 ______ DriverTests.test_client_reaches_join_point_which_completes_parent _______
16:26:23 
16:26:23 self = <tests.driver.driver_test.DriverTests testMethod=test_client_reaches_join_point_which_completes_parent>
16:26:23 
16:26:23     def test_client_reaches_join_point_which_completes_parent(self):
16:26:23         target = self.create_test_driver_target()
16:26:23         d = driver.Driver(target, self.cfg, es_client_factory_class=DriverTests.StaticClientFactory)
16:26:23     
16:26:23         d.prepare_benchmark(t=self.track)
16:26:23         d.start_benchmark()
16:26:23     
16:26:23         self.assertEqual(0, len(d.workers_completed_current_step))
16:26:23     
16:26:23         d.joinpoint_reached(
16:26:23             worker_id=0,
16:26:23             worker_local_timestamp=10,
16:26:23             task_allocations=[driver.ClientAllocation(client_id=0, task=driver.JoinPoint(id=0, clients_executing_completing_task=[0]))],
16:26:23         )
16:26:23     
16:26:23         self.assertEqual(-1, d.current_step)
16:26:23         self.assertEqual(1, len(d.workers_completed_current_step))
16:26:23         # notified all drivers that they should complete the current task ASAP
16:26:23         self.assertEqual(4, target.complete_current_task.call_count)
16:26:23     
16:26:23         # awaiting responses of other clients
16:26:23         d.joinpoint_reached(
16:26:23             worker_id=1,
16:26:23             worker_local_timestamp=11,
16:26:23             task_allocations=[driver.ClientAllocation(client_id=1, task=driver.JoinPoint(id=0, clients_executing_completing_task=[0]))],
16:26:23         )
16:26:23     
16:26:23         self.assertEqual(-1, d.current_step)
16:26:23         self.assertEqual(2, len(d.workers_completed_current_step))
16:26:23     
16:26:23         d.joinpoint_reached(
16:26:23             worker_id=2,
16:26:23             worker_local_timestamp=12,
16:26:23             task_allocations=[driver.ClientAllocation(client_id=2, task=driver.JoinPoint(id=0, clients_executing_completing_task=[0]))],
16:26:23         )
16:26:23         self.assertEqual(-1, d.current_step)
16:26:23         self.assertEqual(3, len(d.workers_completed_current_step))
16:26:23     
16:26:23 >       d.joinpoint_reached(
16:26:23             worker_id=3,
16:26:23             worker_local_timestamp=13,
16:26:23             task_allocations=[driver.ClientAllocation(client_id=3, task=driver.JoinPoint(id=0, clients_executing_completing_task=[0]))],
16:26:23         )
16:26:23 
16:26:23 tests/driver/driver_test.py:226: 
16:26:23 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
16:26:23 esrally/driver/driver.py:752: in joinpoint_reached
16:26:23     self.move_to_next_task(workers_curr_step)
16:26:23 esrally/driver/driver.py:768: in move_to_next_task
16:26:23     m = self.metrics_store.to_externalizable(clear=True)
16:26:23 _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
16:26:23 
16:26:23 self = <esrally.metrics.InMemoryMetricsStore object at 0x7f0b7ce452e0>
16:26:23 clear = True
16:26:23 
16:26:23     def to_externalizable(self, clear=False):
16:26:23         docs = self.docs
16:26:23         if clear:
16:26:23             self.docs = []
16:26:23 >       compressed = zlib.compress(pickle.dumps(docs))
16:26:23 E       _pickle.PicklingError: Can't pickle <class 'unittest.mock.MagicMock'>: it's not the same object as unittest.mock.MagicMock
```
</details>

docs/telemetry.rst
```diff
@@ -226,4 +226,20 @@ Example of recorded documents given two data streams in the cluster::
...
\ No newline at end of file
+* ``data-stream-stats-sample-interval`` (default 10): A positive number greater than zero denoting the sampling interval in seconds.
+
+
+master-node-stats
+-----------------
+
+The master-node-stats telemetry device regularly reports the name of the master node by using:
+
+ 1. the `Cluster State API <https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-state.html>`_, specifically `/_cluster/state/master_node` to determine the id of the master node
```
b-deam, line 237: Nit: The docs here are outdated now since we just use the cat client. 

Though, I also wonder if we should even expose this in the docs given we intend for it to be an internal telemetry device? @danielmitterdorfer, thoughts?

b-deam (commented):
I tested it out and it works well! Thanks :-) Just one quick nit around the docs, and wondering whether we should expose it at all if we intend for it to be always on..

docs/telemetry.rst
danielmitterdorfer: I agree, we should skip that for internal telemetry devices.

pquentin: Thanks, I forgot to remove it when I updated the CLI output at the top of the page. Nice catch!

I removed it now.

b-deam (approved):
LGTM! This will be super handy when analysing benchmark results.

PR #1331, opened 2021-09-15
Remove noisy python-caches-clean output

pquentin:
Without `-prune`, find tries to recurse into the `__pycache__`
directories just after their deletion but fails.

This will avoid hundreds of lines of unnecessary output in jobs like https://elasticsearch-ci.elastic.co/job/elastic+rally+pull-request/285/consoleFull.


b-deam (approved):
LGTM!

PR #1336, opened 2021-09-20
Check benchmark stop at least every second in SamplerThread

### LINKED ISSUE #1334: SamplerThread can delay benchmarks unnecessarily
**Rally version** (get with `esrally --version`): master

**Invoked command**: `esrally race --distribution-version=7.14.1 --car=4gheap --track=geonames --test-mode --telemetry shard-stats --telemetry-params="shard-stats-sample-interval: 200"`

`telemetry.SamplerThread` currently has the following implementation, slightly edited for clarity:

```python3
class SamplerThread(threading.Thread):
    def __init__(self, recorder):
        threading.Thread.__init__(self)
        self.stop = False
        self.recorder = recorder

    def finish(self):
        self.stop = True
        self.join()

    def run(self):
        while not self.stop:
            self.recorder.record()
            time.sleep(self.recorder.sample_interval)
```

When the benchmark ends, the thread won't join until the sleep is over, which can be long in some cases, eg. the default `ShardStats` `sample_interval` is 60 seconds. This can be especially noticeable in test mode, used extensively in the integration tests.

**Steps to reproduce**:
 1. Run `esrally race --distribution-version=7.14.1 --car=4gheap --track=geonames --test-mode --telemetry shard-stats --telemetry-params="shard-stats-sample-interval: 200"`
 2. Run `esrally race --distribution-version=7.14.1 --car=4gheap --track=geonames --test-mode --telemetry shard-stats --telemetry-params="shard-stats-sample-interval: 1"`
 3. The first command should be roughly as fast as the second one.

**Proposed fix**: Never sleep more than one second to check `self.stop` more often. (If this code was async, we would simply cancel the task, which is bad idea with threads.)

pquentin:
The first commit does not change the logic, but extracts it and tests it. The second commit introduces the actual fix. See the issue for test instructions.

Closes #1334

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -164,9 +164,21 @@ class Sampler:
     def run(self):
         # noinspection PyBroadException
         try:
-            while not self.stop:
-                self.recorder.record()
-                self.sleep(self.recorder.sample_interval)
+            sleep_left = self.recorder.sample_interval
+            while True:
+                if sleep_left <= 0:
+                    self.recorder.record()
+                    sleep_left = self.recorder.sample_interval
+
+                if self.stop:
+                    break
+
+                sleep_seconds = sleep_left
+                if sleep_left >= 1:
+                    sleep_seconds = 1
+
+                self.sleep(sleep_seconds)
+                sleep_left -= sleep_seconds
         except BaseException:
             logging.getLogger(__name__).exception("Could not determine %s", self.recorder)
 
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -172,11 +172,8 @@ class Sampler:
 
                 if self.stop:
                     break
-
-                sleep_seconds = sleep_left
-                if sleep_left >= 1:
-                    sleep_seconds = 1
-
+                # check for self.stop at least every second
+                sleep_seconds = min(sleep_left, 1)
                 self.sleep(sleep_seconds)
                 sleep_left -= sleep_seconds
         except BaseException:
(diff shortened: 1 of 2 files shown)


esrally/telemetry.py
```diff
@@ -150,26 +150,49 @@ class InternalTelemetryDevice(TelemetryDevice):
...
+            sleep_left = self.recorder.sample_interval
+            while True:
+                if sleep_left <= 0:
+                    self.recorder.record()
+                    sleep_left = self.recorder.sample_interval
+
+                if self.stop:
+                    break
+
+                sleep_seconds = sleep_left
```
DJRickyB, line 176: may I suggest the following instead of this block, with the explainer comment:
```suggestion
                # check for self.stop at least every second
                sleep_seconds = min(sleep_left, 1)
```

esrally/telemetry.py
```diff
@@ -150,26 +150,49 @@ class InternalTelemetryDevice(TelemetryDevice):
...
+                if sleep_left >= 1:
```
DJRickyB, line 177:
```suggestion
```

esrally/telemetry.py
```diff
@@ -150,26 +150,49 @@ class InternalTelemetryDevice(TelemetryDevice):
...
+                    sleep_seconds = 1
```
DJRickyB, line 178:
```suggestion
```

esrally/telemetry.py
pquentin: That's much better, thanks!

DJRickyB (approved):
LGTM.  Thank you for this

PR #1337, opened 2021-09-21
Clarify definition of processing time

danielmitterdorfer:
With this commit we add specific documentation on the documentation
pages for the command line report and metrics records for "processing
time". Previously this metric has only been documented together with the
corresponding configuration setting `output.processingtime` and was
therefore hard to find.


pquentin (approved):
LGTM! Thanks for this.

PR #1338, opened 2021-09-21
pin pip, setuptools, and wheel versions

### LINKED ISSUE #1323: Pin pip/setuptools/wheel versions
## What/Why

Currently we upgrade `pip`, `setuptools` and `wheel` to the latest version in the corresponding Makefile target: https://github.com/elastic/rally/blob/08c0c202f7efdb9185030f1188a200ba7dddaa12/Makefile#L59

Always going for latest can cause issues as reported in https://github.com/elastic/rally/pull/1322 and for example break CI.

## Proposal

We'd like to pin the versions of these Python packages. Rather than doing that in the `Makefile` we'd rather define them in https://github.com/elastic/rally/blob/08c0c202f7efdb9185030f1188a200ba7dddaa12/.ci/variables.json#L1-L7 so that managing updates can be done in a central place.

DJRickyB:
with this commit we add fixed pip, setuptools, and wheel version numbers instead of always upgrading to latest.  resolves #1323

### DIFF
--- a/.ci/variables.json
+++ b/.ci/variables.json
@@ -1,4 +1,9 @@
 {
+  "prerequisite_versions": {
+    "PIP": "21.2.4",
+    "SETUPTOOLS": "58.0.4",
+    "WHEEL": "0.37.0"
+  },
   "python_versions": {
     "PY38": "3.8.10",
     "PY39": "3.9.2",
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ PY_BIN = python3
 PIP_WRAPPER = $(PY_BIN) -m pip
 export PY38 = $(shell jq -r '.python_versions.PY38' .ci/variables.json)
 export PY39 = $(shell jq -r '.python_versions.PY39' .ci/variables.json)
+export PIP_VERSION = $(shell jq -r '.prerequisite_versions.PIP' .ci/variables.json)
+export SETUPTOOLS_VERSION = $(shell jq -r '.prerequisite_versions.SETUPTOOLS' .ci/variables.json)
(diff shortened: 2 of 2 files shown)


.ci/variables.json
```diff
@@ -1,4 +1,9 @@
 {
+  "prerequisite_versions": {
+    "PIP": "21.2.4",
+    "SETUPTOOLS": "58.0.4",
+    "WHEEL": "0.37.0"
```
pquentin, line 5: I realize you're being consistent with `PY38` and `PY39` below, but what are your thoughts on using lowercase in that case? `PIP` seems weird, but maybe that's just me.

pquentin (commented):
Thanks for tackling this!

I should have realized this earlier, but there's actually a standard place to specify the setuptools and wheel version: pyproject.toml. It would look like this:

```
[build-system]
requires = ["setuptools==58.0.4", "wheel==0.37.0"]
build-backend = "setuptools.build_meta:__legacy__"
```

But that requires using pip exclusively to install Rally (not python setup.py) and we still have to specify the pip version.

But maybe we should do this as a second step?

.ci/variables.json
DJRickyB: I'm ambivalent.  They look more like constants to me when they're upper-cased

DJRickyB:
After discussion with @pquentin, this solution is timely but temporary.  We will prefer in the near future to instead use pyproject.toml for setuptools and wheel versions

pquentin (approved):
LGTM! Thanks for bearing with me :sweat_smile:

PR #1342, opened 2021-09-22
Drop support for Elasticsearch 6.x metrics store

### LINKED ISSUE #653: Drop support for metrics store versions before Elasticsearch 7.0.0
With #654 we introduce version-specific code in Rally to support metrics store versions before and after 7.0.0. We should drop support for older versions at some point to avoid version-specific code but we should wait at least a few months after 7.0.0 is out, maybe even until 8.0.0.

pquentin:
Thanks @danielmitterdorfer for preparing the ground here, it was easy to do. :) I tested that indexes created with the old templates on 7.x still work with the new code. I'm not sure how to test the 6.x -> 7.x upgrade path, though. Is it useful to test?

As usual, this can be reviewed commit by commit if you're so inclined.

Closes #653

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -4,6 +4,11 @@ Migration Guide
 Migrating to Rally 2.3.0
 ------------------------
 
+Support for Elasticsearch 6.x as a metrics store has been dropped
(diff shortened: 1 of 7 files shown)


DJRickyB (approved):
Looks good.  tried to find dangling instances where we use types and couldn't find any

pquentin:
Thanks for the review, will merge on Monday.

PR #1343, opened 2021-09-22
Re-introduce fix for listing nested tracks

DJRickyB:
This commit was previously reverted, due to a false positive in our testing. See #1086 for full detail.


danielmitterdorfer (approved):
LGTM

PR #1344, opened 2021-09-23
Bump Sphinx version

pquentin:
Let's see if Read the Docs reports a status now.


pquentin:
Notice that we have a GitHub check that points to the updated docs, which allows seeing that nothing significant has changed.

DJRickyB (approved):
Neat.  I like clicking through to the compiled/hosted docs from the check

PR #1345, opened 2021-09-29
Update Rally docker base image to Debian 11 and Python 3.8.12

### LINKED ISSUE #1339: Upgrade Docker base image
We're currently using Python 3.8.2 (and Debian 10) in our base image:

https://github.com/elastic/rally/blob/ec160e002750c0d29d29c7a1e997be98e3871bb0/docker/Dockerfiles/Dockerfile-release#L1

As this version is already dated we should look into upgrading to a newer version (and also use Debian 11).

dliappis:
Currently the Rally docker image is based on a Debian 10 base image and Python 3.8.2.

This commit upgrades both to a more recent version.

Closes #1339

### DIFF
--- a/docker/Dockerfiles/Dockerfile-dev
+++ b/docker/Dockerfiles/Dockerfile-dev
@@ -3,7 +3,7 @@
 # Install Rally from source inside a virtualenv
 ################################################################################
 
-FROM python:3.8.2-slim as builder
+FROM python:3.8.12-slim-bullseye as builder
 
 RUN apt-get -y update && \
     apt-get install -y curl git gcc && \
@@ -35,7 +35,7 @@ RUN pip3 install /rally
 # Add entrypoint
 ################################################################################
 
-FROM python:3.8.2-slim
+FROM python:3.8.12-slim-bullseye
 ARG RALLY_VERSION
 ARG RALLY_LICENSE
 ENV RALLY_RUNNING_IN_DOCKER True
--- a/docs/docker.rst
+++ b/docs/docker.rst
@@ -109,14 +109,14 @@ For example, after executing our earlier quickstart example ``docker run elastic
 
 To further examine the contents we can bind mount it from another image e.g.::
 
-    $ docker run --rm -i -v=96256462c3a1f61120443e6d69d9cb0091b28a02234318bdabc52b6801972199:/rallyvolume -ti python:3.8.2-slim /bin/bash
+    $ docker run --rm -i -v=96256462c3a1f61120443e6d69d9cb0091b28a02234318bdabc52b6801972199:/rallyvolume -ti python:3.8.12-slim-bullseye /bin/bash
     root@9a7dd7b3d8df:/# cd /rallyvolume/
     root@9a7dd7b3d8df:/rallyvolume# ls
     root@9a7dd7b3d8df:/rallyvolume/.rally# ls
     benchmarks  logging.json  logs	rally.ini
     # head -4 benchmarks/races/1d81930a-4ebe-4640-a09b-3055174bce43/race.json
     {
-     "rally-version": "1.2.1.dev0",
+     "rally-version": "2.2.1",
      "environment": "local",
      "race-id": "1d81930a-4ebe-4640-a09b-3055174bce43",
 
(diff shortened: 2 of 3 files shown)


docs/docker.rst
```diff
@@ -109,14 +109,14 @@ For example, after executing our earlier quickstart example ``docker run elastic
 
 To further examine the contents we can bind mount it from another image e.g.::
 
-    $ docker run --rm -i -v=96256462c3a1f61120443e6d69d9cb0091b28a02234318bdabc52b6801972199:/rallyvolume -ti python:3.8.2-slim /bin/bash
```
danielmitterdorfer, line 112: How about we keep the Rally version automatically  up-to-date?

```diff
 diff --git a/docs/docker.rst b/docs/docker.rst
index b58552f6..2bcc5105 100644
--- a/docs/docker.rst
+++ b/docs/docker.rst
@@ -107,16 +107,18 @@ For example, after executing our earlier quickstart example ``docker run elastic
     local               96256462c3a1f61120443e6d69d9cb0091b28a02234318bdabc52b6801972199


-To further examine the contents we can bind mount it from another image e.g.::
+To further examine the contents we can bind mount it from another image e.g.:

-    $ docker run --rm -i -v=96256462c3a1f61120443e6d69d9cb0091b28a02234318bdabc52b6801972199:/rallyvolume -ti python:3.8.2-slim /bin/bash
+.. parsed-literal:: :class: literal-block highlight
+
+    $ docker run --rm -i -v=96256462c3a1f61120443e6d69d9cb0091b28a02234318bdabc52b6801972199:/rallyvolume -ti python:3.8.12-slim-bullseye /bin/bash
     root@9a7dd7b3d8df:/# cd /rallyvolume/
     root@9a7dd7b3d8df:/rallyvolume# ls
     root@9a7dd7b3d8df:/rallyvolume/.rally# ls
     benchmarks  logging.json  logs     rally.ini
     # head -4 benchmarks/races/1d81930a-4ebe-4640-a09b-3055174bce43/race.json
     {
-     "rally-version": "1.2.1.dev0",
+     "rally-version": "\ |release|\ ",
      "environment": "local",
      "race-id": "1d81930a-4ebe-4640-a09b-3055174bce43",

@@ -129,9 +131,12 @@ Extending the Docker image
 --------------------------

 You can also create your own customized Docker image on top of the existing one.
-The example below shows how to get started::
+The example below shows how to get started:
+
+.. parsed-literal:: :class: literal-block highlight
+
+    FROM elastic/rally:\ |release|\

-    FROM elastic/rally:1.2.1
     COPY --chown=1000:0 rally.ini /rally/.rally/

 You can then build and test the image with::
```

danielmitterdorfer (commented):
Thank you! Looks fine to me; I left a suggestion  for the docs.

docs/docker.rst
dliappis: Thanks, this makes sense since we use it elsewhere as well. Added in https://github.com/elastic/rally/pull/1345/commits/cff0c490f6f466d7d66952b554c98b2d022afeda

docker/Dockerfiles/Dockerfile-dev
```diff
@@ -3,7 +3,7 @@
 # Install Rally from source inside a virtualenv
 ################################################################################
 
-FROM python:3.8.2-slim as builder
+FROM python:3.8.12-slim-bullseye as builder
```
pquentin, line 6: python:3.8-slim avoid having to specify a patch version and already defaults to bullseye:

```
docker run python:3.8-slim cat /etc/debian_version   
11.0
```

Any reason not to rely on the defaults here?

DJRickyB: local caching, in case I run docker tests in an environment where I already have an image tagged `3.8-slim` will prevent updating the patch version.  It's good practice to be explicit with things like this

dliappis: I agree about being explicit being the safer thing to do, thanks @DJRickyB

pquentin: Definitely best practice (I've used hadolint in the past to enforce this), was wondering if it made sense here. It does, thanks.

danielmitterdorfer (approved):
LGTM!

dliappis:
Thanks everyone for the quick reviews. Merging.

PR #1349, opened 2021-10-01
Allow running `make` commands with active virtualenv

pquentin:
Inside an active Python virtual environment, the `VIRTUAL_ENV` environment variable is set, pointing to the virtual environment folder (where bin/activate and bin/python are). Using `VIRTUAL_ENV` instead of `VENV_NAME` in the `Makefile` means that `make` will use a virtual environment when one is active. pyenv continues to be the preferred method, though.


pquentin:
Failed for an unrelated reason, retrying.

ebadyano (approved):
LGTM, thank you. tested locally.

PR #1351, opened 2021-10-01
Move version source of truth to esrally/_version.py

pquentin:
The main benefit is that we no longer have to import pkg_resources
inside esrally. This divides import time by 5 on my laptop. More
importantly, we can now install and use another version of
elasticsearch-py (say) without having to modify setup.py first.

I tested the following things:

 * `python -c 'import esrally; print(esrally.__version__)'`
 * `python setup.py --version`
 * `docker build -f docker/Dockerfiles/Dockerfile-dev .`
 * `make docs`
 * releasing, in uh, realistic conditions, sorry (see https://github.com/elastic/rally/commit/1ddb316c077b4dbd5373f7679eb10987d7762893 and https://github.com/elastic/rally/commit/1ddb316c077b4dbd5373f7679eb10987d7762893).


pquentin:
It took a bit of back and forth but Sphinx now has access to Rally's code inside Read the Docs. We can use `autodoc` and refer to specific parts of the code if we want to now.

.readthedocs.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+
+python:
+  version: "3.8"
```
dliappis, line 4: Is it possible instead of hardcoding it here, to reference a value from https://github.com/elastic/rally/blob/cca825568330b66a1ccc09172faba91ecca895ec/.ci/variables.json#L10 (like `MIN_PY_VER`)?

release.sh
```diff
@@ -59,7 +59,7 @@ git commit -a -m "Update changelog for Rally release $RELEASE_VERSION"
 
 # * Update version in `setup.py` and `docs/conf.py`
 echo "Updating release version number"
-echo "$RELEASE_VERSION" > version.txt
+echo "__version__ = \"$RELEASE_VERSION\"" > esrally/_version.py
```
dliappis, line 62: nit: to help readability and avoid backslashes one can also leverage `printf`:

`printf '__version__ = "%s"\n' $RELEASE_VERSION > esrally/_version.py`

dliappis (commented):
This looks good, thank you. I left some minor questions/comments.

I suggest that we don't set the milestone of `2.3.0` for this, as we intend to release a new version soon. So I am suggesting we merge after the upcoming release using a different milestone, as anyway there's already quite a backlog of changes for the next Rally version and we'd benefit from using this commit internally across our tooling for a bit longer to catch any edge/corner cases.

.readthedocs.yml
pquentin: I don't see how that would be possible unfortunately, this is just a static file

release.sh
pquentin: Thanks, fixed

dliappis (approved):
LGTM. As discussed let's not merge this until we've released 2.3.0

PR #1355, opened 2021-10-07
Support Python 3.10

DJRickyB:
With this commit we add CI for Python version 3.10 in order to ensure we are compatible going forward.  A couple of dependencies needed updating as well, so developers of Rally should have some notice if they are working with master

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,8 +111,10 @@ test: check-venv
 
 precommit: lint
 
+# checks min and max python versions
 it: check-venv python-caches-clean tox-env-clean
-	. $(VENV_ACTIVATE_FILE); tox
+	. $(VENV_ACTIVATE_FILE); tox -e py38
+	. $(VENV_ACTIVATE_FILE); tox -e py310
 
 it38: check-venv python-caches-clean tox-env-clean
 	. $(VENV_ACTIVATE_FILE); tox -e py38


DJRickyB:
This looks to be a red herring.  If I run this locally (indeed, on Mac) I see the following processes for `make it39`:

```
  501 65385 65384   0  8:37AM ttys001    0:00.36 /Users/rick.boyd/dev/rally/.venv/bin/python3 /Users/rick.boyd/dev/rally/.venv/bin/tox -e py39
  501 65449 65385   0  8:37AM ttys001    0:06.25 /Users/rick.boyd/dev/rally/.tox/py39/bin/python -m pip install --exists-action w .tox/.tmp/package/1/esrally-2.3.1.dev0.zip
```

So the "platform Python" in my virtualenv invokes `tox`, but tox uses the `py39` tox env python, which links to the pyenv-managed 3.9 installation directly:

```
(.venv) rick.boyd@Ricks-MBP night-rally % ls -lrt /Users/rick.boyd/dev/rally/.tox/py39/bin/python
lrwxr-xr-x  1 rick.boyd  staff  52 Oct 15 08:37 /Users/rick.boyd/dev/rally/.tox/py39/bin/python -> /Users/rick.boyd/.pyenv/versions/3.9.2/bin/python3.9
```

pquentin:
I think the pytest-asyncio release was not relevant, because they only fixed warnings in their own tess. The deprecation was in our own `test/__init__.py` file where we sometimes created a `Future` outside of an event loop. This happened in a few decorators.

 * One option was to put `run_async` before `@mock.patch("asyncio.sleep", return_value=None)`
 * But that seemed more fragile than taking advantage of the fact that `@mock.patch()` can handle async function just fine in Python 3.8+

We could probably get rid of `as_future()` entirely and use `AsyncMock` instead, but that's out of scope here.

setup.py
```diff
@@ -36,7 +36,7 @@ def str_from_file(name):
 long_description = str_from_file("README.rst")
 
 # tuples of (major, minor) of supported Python versions ordered from lowest to highest
-supported_python_versions = [(3, 8), (3, 9)]
+supported_python_versions = [(3, 8), (3, 9), (3, 10)]
```
danielmitterdorfer, line 39: Just as a thought (no need at all to address this here): As `.ci/variables.json` became our source of truth regarding Python versions I wonder whether we should pick up the supported versions from there (may require us to change the file format a bit).

tests/driver/runner_test.py
```diff
@@ -3274,7 +3274,7 @@ async def test_raw_with_timeout_and_opaqueid(self, es):
 class SleepTests(TestCase):
     @mock.patch("elasticsearch.Elasticsearch")
     # To avoid real sleeps in unit tests
-    @mock.patch("asyncio.sleep", return_value=as_future())
+    @mock.patch("asyncio.sleep", return_value=None)
```
danielmitterdorfer, line 3277: Is the `return_value` parameter a leftover? In https://github.com/elastic/rally/pull/1355#issuecomment-945419821 @pquentin  mentions:

> I ended up taking advantage of the fact that @mock.patch() can handle async function just fine in Python 3.8+

So I had expected that this (and related lines) don't specify the `return_value` parameter?

Makefile
```diff
@@ -118,6 +120,9 @@ it38: check-venv python-caches-clean tox-env-clean
 it39: check-venv python-caches-clean tox-env-clean
 	. $(VENV_ACTIVATE_FILE); tox -e py39
 
+it310: check-venv python-caches-clean tox-env-clean
```
danielmitterdorfer, line 123: This target (and `it39`) is missing from the phony targets. Can you please add both there? We should probably declare them as phony together with the task definition rather than at the end of the build file (could be a follow-up).

danielmitterdorfer (commented):
Looks pretty good already. I left a few minor comments.

tests/driver/runner_test.py
pquentin: Thanks, I'm not familiar with `@mock.patch` and had not realized that. Removed them now. :+1:

Makefile
```diff
@@ -135,4 +135,4 @@ release-checks: check-venv
 release: check-venv release-checks clean docs it
 	. $(VENV_ACTIVATE_FILE); ./release.sh $(release_version) $(next_version)
 
-.PHONY: install clean nondocs-clean docs-clean python-caches-clean tox-env-clean docs serve-docs test it it38 benchmark release release-checks prereq venv-create check-env
+.PHONY: install clean nondocs-clean docs-clean python-caches-clean tox-env-clean docs serve-docs test it it38 it39 310 benchmark release release-checks prereq venv-create check-env
```
danielmitterdorfer, line 138: Should this be `it310` instead of `310`?

pquentin: Oops, fixed, thanks.

setup.py
pquentin: Good idea, will open a follow-up pull request once this one is merged. :+1:

pquentin:
Do you have Python 3.10 installed? Is `python3.10` doing something? tox is complaining that it does not find it

pquentin:
Will merge this on Monday given we have one approval from someone who did not push code

danielmitterdorfer (approved):

PR #1362, opened 2021-10-13
Add timeout parameter to bulk operation

probakowski:
This change adds `timeout` parameter that can be used in `bulk` operation to define server-side timeout (for applying mappings updates, automatic index creation etc)


danielmitterdorfer (commented):
Thanks for your PR! The change itself looks fine but can you please add a unit-test that the `timeout` parameter is used by the client (see `BulkIndexRunnerTests`)? We should also add user-level docs for that parameter and it would be good to link to the respective section of the [Elasticsearch reference docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-query-params) as well there.

docs/track.rst
```diff
@@ -736,6 +736,7 @@ Properties
 * ``on-conflict`` (optional, defaults to ``index``): Determines whether Rally should use the action ``index`` or ``update`` on id conflicts.
 * ``recency`` (optional, defaults to 0): A number between [0,1] indicating whether to bias conflicting ids towards more recent ids (``recency`` towards 1) or whether to consider all ids for id conflicts (``recency`` towards 0). See the diagram below for details.
 * ``detailed-results`` (optional, defaults to ``false``): Records more detailed meta-data for bulk requests. As it analyzes the corresponding bulk response in more detail, this might incur additional overhead which can skew measurement results. See the section below for the meta-data that are returned.
+* ``timeout`` (optional, defaults to ``1m``): `Query parameter <https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-query-params>`_ defining period each action waits for the following operations: Automatic index creation, Dynamic mapping updates, Waiting for active shards
```
danielmitterdorfer, line 739: How about slightly rewording this to:

```
defines the `time period that Elasticsearch will wait per action <https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#docs-bulk-api-query-params>`_ until it has finished processing the following operations: automatic index creation, dynamic mapping updates, waiting for active shards.
```

probakowski: Yeah, it's better, I've updated docs per your suggestion

PR #1364, opened 2021-10-14
update elasticsearch dependency

Ro0tk1t:


pquentin:
@Ro0tk1t Thanks for the pull request! We appreciate your interest in Rally. We're definitely interested in fixing any issues with 7.15, as it contains many deprecations that we will have to fix before 8.0. Case in point, this pull request surfaced surprising issues that we'd like to fix.

However, we just upgraded from 7.10 to 7.14 last month, and upgrading the Python client requires additional work on our end to verify that we don’t have any accidental client-side bottlenecks. So we won't merge this pull request as it is, and would rather wait for 7.16 that is scheduled in December.

Are you interested in trying to figure out why the build fails? We'd be happy to merge 7.15 fixes right away, even if the actual client upgrade will have to wait.

Ro0tk1t:
@pquentin i don't known why i can't install ES server when use `make it39`
image

and should i replace version 7.6.0 to 7.15.1 ?

pquentin:
@Ro0tk1t I realize that this was probably more effort than you were looking for, but are you still interested in working on this? Can we help in any way?

Ro0tk1t:
i think i have no much time to do this, especially not familiar with this project。
i'm going to close the pr

pquentin:
Thanks for trying! That was useful: we found a few places where we can report errors more usefully.

Issue #1215, opened 2021-03-18
Remove scroll support in search operation


DJRickyB:
As of #1190, `scroll-search` is its own operation, and as such we should remove scroll support (aka the usage of the "pages" parameter) for search operations.

PR #1368, opened 2021-10-26
Call the correct functions for `scroll-search` and `paginated-search` operations

michaelbaamonde:
In the course of getting started on https://github.com/elastic/rally/issues/1215, I discovered that `scroll-search` operations don't default to the index (or data stream) defined at the top level of the track under the `indices` key. The `pmc` and `geonames` tracks expose this. But if you explicitly provide an `index` parameter to the operation (as `http_logs` does)[^1] it works...

...by accident, it turns out. After digging into it, the core issue seems to be that there was no parameter source registered for `scroll-search` or `paginated-search` operations. So, neither the track object's `indices` attribute (necessary for choosing a default `index` parameter for the operation) nor the `operation-type` parameter (necessary for choosing the right function to call) was getting passed down the chain. But, the way the control flow works for dispatching to the appropriate function in the in the `Query.__call__` method, it wound up just issuing a request body query with a `pages` parameter, which happens to be the legacy implementation of scroll searches.

So, this commit addresses this by doing three things:

1. Augmenting `SearchParamSource` to pass the `operation-type` it receives through to its callers.
2. Registering `SearchParamSource` for both scrolls and paginated queries.
3. Enforcing `operation-type` as a mandatory paramater when calling the `Query` runner (necessary for custom parameter sources to avoid unexpected behavior.)

_That said_, I do wonder if it'd make sense to have separate classes (perhaps subclasses of `Query`) for each of these three operations. Their logic is distinct, it's possible that they should have their own parameter sources, and dispatching on the operation type within one large `__call__` method feels a little indirect given that we already have a mechanism for explicitly registering runners on a per-operation basis. I worry there may be some other trappy behavior lurking in here somewhere, but I opted to fix the bug within the current implementation instead of doing a larger refactoring along those lines. Happy to hear any thoughts on that front.

[^1]: https://github.com/michaelbaamonde/rally-tracks/commit/b700eac059ab42246768f44d0e656caf144a0269 contains the relevant changes that sparked this investigation.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -968,7 +968,7 @@ Properties
 * ``body`` (mandatory): The query body.
 * ``response-compression-enabled`` (optional, defaults to ``true``): Allows to disable HTTP compression of responses. As these responses are sometimes large and decompression may be a bottleneck on the client, it is possible to turn off response compression.
 * ``detailed-results`` (optional, defaults to ``false``): Records more detailed meta-data about queries. As it analyzes the corresponding response in more detail, this might incur additional overhead which can skew measurement results. This flag is ineffective for scroll queries.
-* ``pages`` (optional): Number of pages to retrieve. If this parameter is present, a scroll query will be executed. If you want to retrieve all result pages, use the value "all".  See also the ``scroll-search`` operation type.
+* ``pages`` (optional, deprecated): Number of pages to retrieve. If this parameter is present, a scroll query will be executed. If you want to retrieve all result pages, use the value "all". This parameter is deprecated and will be replaced with the ``scroll-search`` operation in a future release.
 * ``results-per-page`` (optional):  Number of documents to retrieve per page. This maps to the Search API's ``size`` parameter, and can be used for scroll and non-scroll searches. Defaults to ``10``
 
 Example::
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -2213,6 +2213,7 @@ class ScheduleHandle:
         :return: A generator for the corresponding parameters.
         """
         self.task_allocation = task_allocation
+        self.operation_type = task_allocation.task.operation.type
         self.sched = sched
         self.task_progress_control = task_progress_control
         self.runner = runner
@@ -2243,6 +2244,11 @@ class ScheduleHandle:
     def after_request(self, now, weight, unit, request_meta_data):
         self.sched.after_request(now, weight, unit, request_meta_data)
 
+    def params_with_operation_type(self):
+        p = self.params.params()
+        p.update({"operation-type": self.operation_type})
+        return p
+
     async def __call__(self):
         next_scheduled = 0
         if self.task_progress_control.infinite:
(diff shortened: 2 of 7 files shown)


michaelbaamonde:

michaelbaamonde:
@danielmitterdorfer @DJRickyB The unrelated test failures are resolved now, so this is ready for another review. Thanks!

DJRickyB:
thanks.  now that `operation-type` is not mandatory, can you undo changes to the tests here?

michaelbaamonde:
> now that `operation-type` is not mandatory, can you undo changes to the tests here?

It's not mandatory in the sense that we don't require custom parameter sources to provide an operation type, but it is required that calls to the `Query` runner include one, somehow. The runner needs to have some awareness of the operation type in order to avoid the bug that motivated this PR initially.

That said, your comment did make me consider another problem, which is that defaulting to `search` as the operation type in the absence of a parameter source providing one means that custom parameter sources for `scroll-search` or `paginated-search` operations won't work as expected; they'll wind up ultimately calling the wrong function. So, this PR fixes the bug in the default case, but it's still lurking there for more advanced usage.

Ultimately, we really do rely on a 1:1 mapping between operation types and runners, and the registration mechanism provides a straightforward way to look up the exact runner for a given operation. The `Query` runner breaks this model in a way that forces us to _tell it_ what type of operation it needs to execute. Parameter sources give us a way to do that, but this is a bit of a hack on my part, and it entails a tighter coupling between parameter sources and runners than I think is intended.

With that said, I just pushed a commit that passes the `operation-type` to the runner by essentially injecting it into the params returned by the parameter source, but in a way that doesn't require any changes to the parameter source itself. This preserves the contract between parameter sources and runners, at least from the track author's perspective. It's another hack, though, which brings me back to:

> I worry there may be some other trappy behavior lurking in here somewhere, but I opted to fix the bug within the current implementation instead of doing a larger refactoring

I think it may be safer and quicker to refactor the `Query` runner into three separate classes instead of iterating further on workarounds. If this PR is mergeable after this latest commit, we can defer that, but if not, I think the time would be better spent on the more fundamental refactoring.

What do you think @DJRickyB @danielmitterdorfer?

DJRickyB:
It's the end of my workday, I want to look at this deeper tomorrow, but I'll note that the "1:1" question came up in the course of #1190, when at the time it was deemed not necessary.  As far as I can tell you've bumped up against the limitations of that idea and I'm curious about @danielmitterdorfer's thoughts as well.

danielmitterdorfer:
> I think it may be safer and quicker to refactor the Query runner into three separate classes instead of iterating further on workarounds. If this PR is mergeable after this latest commit, we can defer that, but if not, I think the time would be better spent on the more fundamental refactoring.

@michaelbaamonde and I had a chat on another channel about this. tl;dr as long as none of our tracks break with this change, we can go ahead and merge this and then start working on the refactoring in a separate PR immediately after.

michaelbaamonde:
I ran all of our tracks and found no issues, so I'll merge.

PR #1373, opened 2021-10-29
Revert "Retrieve master node using cat API"

pquentin:
This reverts commit fbcf365671fcb9279cca735d02d56efc347381a3. The _cat APIs are not meant for programmatic use, can be slower and don't have any backwards compatibility guarantees.

I had this commit ready in [`fbcf365` (#1330)](https://github.com/elastic/rally/pull/1330/commits/fbcf365671fcb9279cca735d02d56efc347381a3), and I just learned about _cat APIs lack of guarantees, so I figured out it would make sense to revert it.

Tested like this:

```
esrally race --distribution-version=7.14.1 --car="4gheap" --track=geonames --test-mode --telemetry master-node-stats --telemetry-params="master-node-stats-sample-interval: 0.1"
http localhost:9200/rally-metrics-2021-10/_search\?size=1000 | jq . | grep master-node -c  # should return more than 0 values
```

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2022,7 +2022,8 @@ class MasterNodeStatsRecorder:
         import elasticsearch
 
         try:
-            info = self.client.cat.master(format="json")
+            state = self.client.cluster.state(metric="master_node")
+            info = self.client.nodes.info(node_id=state["master_node"], metric="os")
         except elasticsearch.TransportError:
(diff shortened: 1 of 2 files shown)


DJRickyB:
@elasticmachine run tests

ebadyano (approved):
LGTM, thank you!

PR #1384, opened 2021-11-15
Stop installing Rally twice in CI

### LINKED ISSUE #1369: Stop installing Rally twice in CI
Here's how our CI currently works.

1. First, Jenkins launches [.ci/build.sh](https://github.com/elastic/rally/blob/24dee9ed704e5c98fc2999a665853ba4b2941586/.ci/build.sh) which runs the following Makefile commands:

        make prereq  # installs Python versions using pyenv
        make install  # installs Rally 
        make precommit  # pylint, black and isort
        make it  # runs tox -e py38,py310

2. Thanks to the above, tox runs unit tests and integration tests for Python 3.8 and 3.10

    But it does not make sense to install Rally in the Makefile, because it will only be installed for Python 3.8. In fact, until Python 3.10 support that's the only place where we installed Rally, and were only testing Python 3.8, not 3.9. So, now, as a quick fix to get the Python 3.10 PR in, we added `deps = .[develop]` in tox.

However, we now install Rally twice. I considered dropping tox because it didn't do much, but creating the correct virtual environments is still useful.

Instead, I'm now convinced now that **dropping `make install` from .ci/build.sh is the correct thing to do**.

uparamonau-es:
Closes #1369

Remove unwanted ''make install'' call.

### DIFF
--- a/.ci/build.sh
+++ b/.ci/build.sh
@@ -49,7 +49,6 @@ function build {
   eval "$(pyenv virtualenv-init -)"
 
   make prereq
-  make install
   make precommit
   make it
 }


pquentin:
This fails because we need the `make install` to run lint. It would be cleaner to use tox to run lint, but it's OK in the Makefile too, and at least the installation is not useless. Closing, thanks for trying this.

PR #1385, opened 2021-11-16
Update chart-generator for release security by default

ebadyano:


esrally/chart_generator.py
```diff
@@ -1507,7 +1507,7 @@ def add_configs(race_configs_per_lic, flavor_name="oss", lic="oss", track_name=N
     def add_race_configs(license_configs, flavor_name, track_name):
         if chart_type == BarCharts:
             # Only one license config, "basic", is present in bar charts
```
pquentin, line 1509: Should this be updated?

```suggestion
            # Only one license config, "trial", is present in bar charts
```

ebadyano:
@pquentin thank you for the fast review!

PR #1387, opened 2021-11-19
Switch to native pytest in tests/utils

pquentin:
There are two different commits, which should help with the review.

### DIFF
--- a/tests/utils/convert_test.py
+++ b/tests/utils/convert_test.py
@@ -36,4 +36,4 @@ class TestToBool:
         for value in values:
             with pytest.raises(ValueError) as exc:
                 convert.to_bool(value)
-            assert exc.value.args[0] == "Cannot convert [%s] to bool." % value
+            assert exc.value.args[0] == f"Cannot convert [{value}] to bool."
--- a/tests/utils/convert_test.py
+++ b/tests/utils/convert_test.py
@@ -15,14 +15,12 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from unittest import TestCase
-
 import pytest
 
 from esrally.utils import convert
 
 
-class ToBoolTests(TestCase):
+class TestToBool:
     def test_convert_to_true(self):
         values = ["True", "true", "Yes", "yes", "t", "y", "1", True]
         for value in values:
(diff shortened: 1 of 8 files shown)


tests/utils/convert_test.py
```diff
@@ -17,23 +17,25 @@
...
+            assert convert.to_bool(value) is False
 
     def test_cannot_convert_invalid_value(self):
         values = ["Invalid", None, []]
         for value in values:
-            with self.assertRaises(ValueError, msg="Expect [%s] of type [%s] to fail to be converted." % (str(value), type(value))) as ctx:
+            with pytest.raises(ValueError) as ctx:
                 convert.to_bool(value)
-            self.assertEqual("Cannot convert [%s] to bool." % value, ctx.exception.args[0])
+            assert "Cannot convert [%s] to bool." % value == ctx.exception.args[0]
```
danielmitterdorfer: We could also switch to f-strings here?

tests/utils/io_test.py
```diff
@@ -47,24 +47,24 @@ def mock_red_hat(path):
 
 class IoTests(TestCase):
     def test_normalize_path(self):
-        self.assertEqual("/already/a/normalized/path", io.normalize_path("/already/a/normalized/path"))
-        self.assertEqual("/not/normalized", io.normalize_path("/not/normalized/path/../"))
-        self.assertEqual(os.path.expanduser("~"), io.normalize_path("~/Documents/.."))
+        assert io.normalize_path("/already/a/normalized/path") == "/already/a/normalized/path"
+        assert io.normalize_path("/not/normalized/path/../") == "/not/normalized"
+        assert os.path.expanduser("~") == io.normalize_path("~/Documents/..")
```
danielmitterdorfer, line 51: Should we reverse the order here so the call to `io.normalize_path` comes first? Then this is consistent with the other assertions.

danielmitterdorfer (approved):
I left two minor comments. Please feel free to address them but no need for another review round in any case. LGTM

tests/utils/io_test.py
pquentin: Nice catch, thanks

tests/utils/convert_test.py
pquentin: My intention is to this with pyupgrade at some point. Fixed here, though!

PR #1388, opened 2021-11-23
Initial approach to refactoring the `Query` runner.

michaelbaamonde:
I'm usually not one for draft PRs, but I wanted to get this work-in-progress out there for some thoughts before the US holiday.

This is a follow-up to https://github.com/elastic/rally/pull/1368 and my intention is to ultimately close out https://github.com/elastic/rally/issues/1215 with it (although, we could maintain backwards compatibility with this PR and make the breaking change in another.)

The commit message lays out the rationale, but the main change is splitting the currently overloaded `Query` runner into three separate runners: `Search`, `PaginatedSearch`, and `ScrollSearch`.

I'm specifically curious about folks' thoughts on two implementation details:

1. Introducing the `Query` mixin
2. Converting `params` dicts into `namedtuples` in `Query._query_params()`.

The thinking behind (1) is that while there is arguably an "is-a" relationship between these three new runners and the concept of a `Query`, making them subclasses just doesn't seem worth it. It does seem worth it to minimize duplication, however, and they have enough in common vis-a-vis constructing parameters and issuing transport requests that a mixin felt pretty natural to me.

The thinking behind (2) is that we have to do quite a bit of mutation of parameters in between requests for paginated searches and scroll searches, and the immutability of a `namedtuple` forces us to be defensive about copying, which I like. Having named fields also improves readability and--at least for me and my sketchy short-term memory--makes it easier to reason about what's happening. I haven't done any microbenchmarking or anything to that effect, but my assumption is that there won't be enough of a performance hit here to worry about. Please let me know if that's a poor assumption!

Finally, I intend to add better docstrings and comments. If it's hard to follow without, don't worry about it. It's part of why I marked it as a draft, and please feel free to wait until it's truly ready for review if that's easier. Thanks!

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -24,7 +24,7 @@ import re
 import sys
 import time
 import types
-from collections import Counter, OrderedDict
+from collections import Counter, OrderedDict, namedtuple
 from copy import deepcopy
 from enum import Enum
 from functools import total_ordering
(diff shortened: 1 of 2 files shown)


michaelbaamonde:
Totally didn't mark it as a Draft. Bad muscle memory. Will redo.

Issue #1180, opened 2021-02-10
Make composite tasks extensible


danielmitterdorfer:
Rally is currently strict what operation-types are allowed in a composite context because not every operation would work properly out of the box. However, it is also not possible to use custom runners in a composite context. We should therefore add a mechanism for custom runners to indicate whether they support composite contexts.

pquentin:
We should probably do that for built-in runners, too, because maintaining a list is brittle anyway. See #1391 for context.

PR #1391, opened 2021-11-29
Allow paginated-search in CompositeContext

pquentin:
This is the documented way to run paginated searches.

Unfortunately, the functionality is tested but at a lower-level that does not involve the `Composite` class. I considered fixing the tests but this is a short-time fix: we agreed offline that the long-term fix should be to let the runner tell whether it supports composite, following the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -2320,6 +2320,7 @@ class Composite(Runner):
         self.supported_op_types = [
             "open-point-in-time",
             "close-point-in-time",
+            "paginated-search",
             "search",
             "raw-request",
             "sleep",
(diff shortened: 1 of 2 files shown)


michaelbaamonde:
> the long-term fix should be to let the runner tell whether it supports composite

Sounds good. This seems to relate to https://github.com/elastic/rally/issues/1180, which specifically pertains to custom runners. Should we augment that issue to also apply to built-in runners?

pquentin:
Great catch, thanks @michaelbaamonde! Commented there.

Issue #1379, opened 2021-11-09
Improve performance by explicitly specifying the base in int() conversions


dliappis:
As discussed [here](https://stackoverflow.com/a/21234576) string conversions to int can be unnecessarily costly if the base is not explicitly specified.

As can be seen using a simple microbenchmark[^1] there is a 10% slowdown when not explicitly specifying the integer base[^2].

We should can our codebase for invocations of `int()`, especially around hot code paths, and replace them with the base explicitly specified.

[^1]: Source for the microbenchmark

    ```python
    $ cat microbench_int.py

    import pytest
    
    dec_string = "6823456194"
    
    
    @pytest.mark.benchmark(
        group="timestamps",
        warmup="on",
        warmup_iterations=10000,
        disable_gc=True,
    )
    def test_int(benchmark):
        benchmark(int, dec_string)
    
    
    @pytest.mark.benchmark(
        group="timestamps",
        warmup="on",
        warmup_iterations=10000,
        disable_gc=True,
    )
    def test_int_base10(benchmark):
        benchmark(int, dec_string, 10)
    ```

[^2]: Microbenchmark results

    ```
    $ pytest microbench_int.py --benchmark-max-time=10

    microbench_int.py::test_int PASSED                                                                                                                                                                                                                                                                                    [ 50%]
    microbench_int.py::test_int_base10 PASSED                                                                                                                                                                                                                                                                             [100%]


    --------------------------------------------------------------------------------- benchmark 'timestamps': 2 tests ---------------------------------------------------------------------------------
    Name (time in ns)          Min                   Max                Mean            StdDev              Median               IQR               Outliers  OPS (Mops/s)            Rounds  Iterations
    ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
    test_int_base10        91.6800 (1.0)        678.3899 (1.0)      101.2257 (1.0)      4.8681 (1.0)      100.6899 (1.0)      4.7102 (1.68)    159578;14535        9.8789 (1.0)      967118         100
    test_int              100.6899 (1.10)     1,370.4103 (2.02)     110.9220 (1.10)     6.7785 (1.39)     110.3099 (1.10)     2.7998 (1.0)      19958;30411        9.0153 (0.91)     866476         100
    ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

    Legend:
[...]

PR #1397, opened 2021-12-08
Specify base-10 for int() for known strings and bytes

inqueue:
With this commit, we explicitly specify base-10 for string and bytes `int()` conversions. Specifying the base is known to boost performance for `int()` conversions as mentioned in https://github.com/elastic/rally/issues/1379. This optimization can only be 
applied when we know the input to be a string or bytes.

While I would have liked to have added the optimization to more code paths, there is a limitation when we cannot guarantee non-numeric inputs. Python will throw a `TypeError` when attempting to convert a non-string with an explicit base, therefore we only designate base-10 when we know for certain the input will be a non-numeric type. A full review of every invocation of `int()` in the code base revealed we cannot always make that guarantee. 

Closes https://github.com/elastic/rally/issues/1379

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -106,7 +106,7 @@ def wait_for_pidfile(pidfilename, timeout=60, clock=time.Clock):
                 buf = f.read()
                 if not buf:
                     raise EOFError
-                return int(buf)
+                return int(buf, 10)
         except (FileNotFoundError, EOFError):
             time.sleep(0.5)
 
--- a/esrally/mechanic/mechanic.py
+++ b/esrally/mechanic/mechanic.py
@@ -50,7 +50,7 @@ def install(cfg):
     sources = not distribution
     build_type = cfg.opts("mechanic", "build.type")
     ip = cfg.opts("mechanic", "network.host")
-    http_port = int(cfg.opts("mechanic", "network.http.port"))
+    http_port = int(cfg.opts("mechanic", "network.http.port"), 10)
     node_name = cfg.opts("mechanic", "node.name")
     master_nodes = cfg.opts("mechanic", "master.nodes")
     seed_hosts = cfg.opts("mechanic", "seed.hosts")
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -60,7 +60,7 @@ class ExitStatus(Enum):
 
 def create_arg_parser():
     def positive_number(v):
-        value = int(v)
+        value = int(v, 10)
         if value <= 0:
             raise argparse.ArgumentTypeError(f"must be positive but was {value}")
         return value
--- a/esrally/utils/console.py
+++ b/esrally/utils/console.py
@@ -110,7 +110,7 @@ def init(quiet=False, assume_tty=True):
     #
     # Set a proper width (see argparse.HelpFormatter)
     try:
-        int(os.environ["COLUMNS"])
+        int(os.environ["COLUMNS"], 10)
     except (KeyError, ValueError):
         # noinspection PyBroadException
         try:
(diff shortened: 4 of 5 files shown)


dliappis (commented):
Thank you for raising this PR and going through the code base for potential improvements based on #1379.

Looking at the files here, it seems to me that none of these uses are related to runners or operations in the hot code path that can potentially cause loaddriver bottlenecks; they are rather used in setup tasks, before the benchmark starts, like downloading the corpora, configuring/starting elasticsearch and parsing cli arguments.

Therefore, I don't think they add value. If we don't have any uses of `int()` in hot code paths, I am leaning towards closing this PR and the corresponding issue. WDYT?

pquentin (commented):
I'm a bit on the fence here. Explicit is better than implicit, but this is also more likely to raise questions than anything else, and the performance benefit is zero: I'd lean towards rejecting this.

By the way, Jason, do you have examples of cases where we don't always send strings to `int()`?

dliappis:
It occurred to me that we could at least document this best practice (among others in the future perhaps) in a section that is linked in all the warnings callouts we currently have for [custom parameter sources](https://esrally.readthedocs.io/en/latest/adding_tracks.html?highlight=profiling#custom-parameter-sources), [custom runners](https://esrally.readthedocs.io/en/latest/adding_tracks.html?highlight=profiling#custom-runners) and [custom schedulers](https://esrally.readthedocs.io/en/latest/adding_tracks.html?highlight=profiling#custom-schedulers). WDYT?

michaelbaamonde:
> we could at least document this best practice (among others in the future perhaps) in a section that is linked in all the warnings callouts

Take all of my +1s @dliappis. 

Expanding this a bit, I'd love a section in the docs that explains what we mean by the "hot path"/"performance-critical code-path", but also more generally a high-level overview of why this is so important in the first place. Reference material on client-side bottlenecks would be great: why they're so crucial to avoid, how they manifest themselves, what we do in Rally to avoid them, etc. You can piece some of this together with what's documented already, but a conceptual overview would help a lot.

inqueue:
> at least document this best practice (among others in the future perhaps) in a section that is linked in all the warnings callouts

I like the idea of a dedicated section explaining the hot path which outlines best practices around it. While I appreciate the warning to be careful, it would be useful to know how and why.

> By the way, Jason, do you have examples of cases where we don't always send strings to `int()`?

Input parameters could be a string from CLI arguments or read as a string or integer from JSON: 
https://github.com/elastic/rally/blob/605da7649ede37a41df17ceca55c04b2cde859e8/esrally/driver/runner.py#L852

Input is read from `rally.ini` (string), but the default value is an integer:
https://github.com/elastic/rally/blob/605da7649ede37a41df17ceca55c04b2cde859e8/esrally/driver/driver.py#L639
https://github.com/elastic/rally/blob/605da7649ede37a41df17ceca55c04b2cde859e8/esrally/driver/driver.py#L1135

I agree we are not gaining any value from this change. For a few cases, there may be some refactoring we could do to avoid `int()` entirely and/or set the base explicitly when we know we have the string (CLI args, `rally.ini`). My vote is to leave things how they are for now. WDYT?

pquentin:
> I like the idea of a dedicated section explaining the hot path which outlines best practices around it. While I appreciate the warning to be careful, it would be useful to know how and why.

Would you like to work on this?

> Input parameters could be a string from CLI arguments or read as a string or integer from JSON. [...] My vote is to leave things how they are for now. WDYT?

Thanks for capturing the issue. The reason I asked is that mypy won't be happy about it eventually, so we might as well fix if by making sure a given value is always a string or always an int. But leavings things as they are currently is also fine, and in any case we all agree that we can close this pull request.

inqueue:
> Would you like to work on this?

Sure, I will work on the docs section about the hot path. Closing this PR.

pquentin:
I opened a new issue to discuss the docs improvement: https://github.com/elastic/rally/issues/1421

PR #1402, opened 2021-12-14
Tell pip to install latest Rally on newer Pythons

pquentin:
Currently, running `pip install esrally` on Python 3.10 installs Rally
1.4.0. Installing the latest version is much more likely to be helpful.

:warning: Please rebase instead of squash due to the lint commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,13 +91,13 @@ tox-env-clean:
 	rm -rf .tox
 
 lint: check-venv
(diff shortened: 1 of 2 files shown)


michaelbaamonde:
Why does this happen?

pquentin:
Because we only added `python_requires` in Rally 1.4.1

michaelbaamonde:
Ahh, got it. Thanks.

michaelbaamonde (approved):
LGTM

pquentin:
Sorry for the terse explanation above, by the way! I did not want to imply this was obvious, I certainly scratched my head quite a bit when I got 1.4.0 with Python 3.10.

Issue #1346, opened 2021-09-30
Document conversion of certutil-generated certs to PEM format for use by OpenSSL


pquentin:
As the Elastic Stack moves to increased security, using TLS certificates becomes more and more prelavent. TLS certificates are usually stored either with the PEM format or the PKCS#12 format, and we currently have a mismatch:

 * Elasticsearch recommends using the PKCS#12 format which is common in Java: https://www.elastic.co/guide/en/elasticsearch/reference/current/certutil.html#_examples_5
 * Rally currently only supports the PEM format, as it's written in Python, which [is tied to OpenSSL](https://www.python.org/dev/peps/pep-0644/) and uses OpenSSL's [SSL_CTX_load_verify_locations](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_load_verify_locations.html) to load certificates, which only supports PEM.

If we would want Rally to support PKCS#12, we would have to use [cryptography to transform PKCS#12 to PEM ](https://blog.cyberreboot.org/using-pkcs-12-formatted-certificates-in-python-fd98362f90ba) inside Rally's [TLS support code](https://github.com/elastic/rally/blob/21e844eabc430fb870df9d4a209e7c869784e095/esrally/client.py#L130-L189). The https://github.com/pyca/cryptography project is well established and well regarded in the Python world, but that would be an additional dependency of a few megabytes. (It also requires a Rust compiler to compile from source, but they have various wheels to support common architectures, so that's not a concern.) Until there are other use cases or higher demand, we won't include that new dependency in Rally.

However, we can document how to [use the openssl CLI to convert a certificate generated as documented in the Elasticsearch docs for use in Rally](https://stackoverflow.com/questions/15144046/converting-pkcs12-certificate-into-pem-using-openssl). This is what this issue is about! Please ask any questions you have if you want to tackle it.

PR #1403, opened 2021-12-15
Document exporting certificates from PKCS#12 keystores

inqueue:
This commit adds documentation for converting certificates in PKCS#12 formatted keystores to PEM for use by Rally. The document includes examples for a CA certificate, node private key & certificate which could be used for client authentication.

Closes https://github.com/elastic/rally/issues/1346


docs/command_line_reference.rst
```diff
@@ -674,11 +674,33 @@ Enable it with ``use_ssl:true``.
 **TLS/SSL Certificate Verification**
 
 Server certificate verification is controlled with the ``verify_certs`` boolean. The default value is `true`. To disable use ``verify_certs:false``.
-If ``verify_certs:true``, Rally will attempt to verify the certificate provided by Elasticsearch. If they are private certificates, you will also need to supply the private CA certificate using ``ca_certs:'/path/to/cacert.pem'``.
+If ``verify_certs:true``, Rally will attempt to verify the certificate provided by Elasticsearch. If they are private certificates, you will also need to supply the private CA certificate in PEM format using ``ca_certs:'/path/to/cacert.pem'``.
 
 You can also optionally present client certificates, e.g. if Elasticsearch has been configured with ``xpack.security.http.ssl.client_authentication: required`` (see also `Elasticsearch HTTP TLS/SSL settings <https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html#http-tls-ssl-settings>`_).
 Client certificates can be presented regardless of the ``verify_certs`` setting, but it's strongly recommended to always verify the server certificates.
 
+**CA & Client Certificate Format**
```
pquentin, line 682: I still have to test this, but can you please move the instructions after then "TLS/SSL Examples" section? Otherwise it isn't clear why we're talking about client_key and client_cert.

inqueue: > can you please move the instructions after then "TLS/SSL Examples" section?

👍🏼 Will do. I had actually considered the examples section initially, then thought it might be better to have client_cert and client_key before the examples.

pquentin: Ah, I see what you mean, but for better or worse, those examples are the only part where we currently define them. (Another option is to define them properly in the "Certificate Verification" section.)

That way, we can keep separate the **definition** of the option from their **conversion** from PKCS#12.

docs/command_line_reference.rst
```diff
@@ -674,11 +674,33 @@ Enable it with ``use_ssl:true``.
...
+
+Rally supports PEM format for CA and client certificates. Certificates in PKCS#12 formatted keystores will need to be exported to PEM format in order for Rally to use them. The ``openssl`` command can be used to export certificates from a PKCS#12 keystore. In the following example, PKCS#12 file ``elastic-certificates.p12`` was generated by the Elasticsearch ``elasticsearch-certutil`` command.
+
+* Export CA certificate:
+
+ ::
+
+   openssl pkcs12 -in elastic-certificates.p12 -cacerts -out cacert.pem
```
pquentin, line 690: Let's use the default `elasticsearch-certutil` file name?
```suggestion
   openssl pkcs12 -in elastic-stack-ca.p12 -cacerts -out cacert.pem
```

pquentin (changes requested):
Have you tested that these files are readable by OpenSSL? For example, using `openssl verify`. I don't think that's the case, because they contain "Bag attributes". Also, cacert.pem was a private key for me, not a certificate.

docs/command_line_reference.rst
inqueue: `elastic-certificates.p12` is the default name of the single-instance bundle generated by `elasticsearch-certutil` and it always includes the CA certificate and not the CA private key. It was probably the wrong choice. The `-cacerts` option only applies to instance bundles generated by certutil, so pulling it from the CA bundle is a little simpler:

`openssl pkcs12 -in elastic-stack-ca.p12 -out cacert.pem`

I reason the CA bundle file name would be the most likely to remain the same and a user knowing their way around `openssl` can export the CA cert however they want. I will change it.

docs/command_line_reference.rst
inqueue: Adjustments are in https://github.com/elastic/rally/pull/1403/commits/ee28cd159f7e5f5a42affea6c80cf251be51f5bd.

docs/command_line_reference.rst
pquentin: I'm not sure I'm following. Where are you seeing `elastic-certificates.p12`? With the 7.16 and 8.0.0-beta1 Docker images, `elasticsearch-certutil ca` generates a `elastic-stack-ca.p12` file.

Also, have you tested that Rally can use those PEM files correctly?

inqueue: > I'm not sure I'm following. Where are you seeing `elastic-certificates.p12`? With the 7.16 and 8.0.0-beta1 Docker images, `elasticsearch-certutil ca` generates a `elastic-stack-ca.p12` file.

`elastic-certificates.p12` is the default file produced when executing the `cert` subcommand for `elasticsearch-certutil`. See the example at https://www.elastic.co/guide/en/elasticsearch/reference/current/certutil.html#_examples_5.

It was worded this way to encourage the use of a node certificate vs. the CA certificate for clients. And good point, I will definitely verify the exported certificates work with Rally correctly.

Please let me know if anything is unclear.

inqueue: I can confirm certificates exported using the `openssl` commands in this PR work correctly with Rally without further modification. The only consideration probably deserving further documentation is `verify_certs:true` since the hostname used in `--target-hosts` must match the one assigned to the node's instance certificate to complete server verification. In the default case for a single node, `elasticsearch-certutil` assigns the instance name "instance". 

```
openssl x509 -in client_cert.pem -noout -subject
subject= /CN=instance
```

All of this can get much more complex and I want to keep these docs as simple as possible to serve as a starting point for those looking to use server certificate verification and client authentication. That said, a better option and practice for a Rally cert/key would be to generate one specifically for benchmarking:

```sh
bin/elasticsearch-certutil cert --ca elastic-stack-ca.p12 --dns rally --pem --name rally
```

I look forward to updating the PR pending your feedback @pquentin.

pquentin: Sorry for the delay in reviewing, and thanks for iterating!

I agree that the idea is just to give instructions for the simplest use case, we don't have to document more complex cases. If documenting `--name rally` helps, then we can do it, but otherwise leave it like this.

While this now looks good, I have not managed to test this, because I can't make Elasticsearch work with p12 files. https://www.elastic.co/guide/en/elasticsearch/reference/current/configuring-tls-docker.html is actually using PEM (turns out there's a `--pem` flag!) and https://www.elastic.co/guide/en/elasticsearch/reference/7.16/security-basic-setup-https.html suggests using `elasticsearch-certutil http` instead, and requires more fiddling than I'd like.

What steps have you followed?

pquentin (changes requested):
One small nit, otherwise this looks good, thanks a lot for the instructions! I was able to make this work in Rally too.

docs/command_line_reference.rst
```diff
@@ -685,6 +685,28 @@ Client certificates can be presented regardless of the ``verify_certs`` setting,
...
 
+**Converting PKCS#12 Files to PEM Format**
+
+Rally supports PEM format for CA and client certificates. Certificates in PKCS#12 formatted keystores will need to be exported to PEM format in order for Rally to use them. The ``openssl`` command can be used to export certificates from a PKCS#12 keystore. In the following example, PKCS#12 file ``elastic-stack-ca.p12`` is generated by the Elasticsearch ``elasticsearch-certutil ca`` command and contains only the CA certificate and private key.
+
+* Export CA certificate:
+
+ ::
+
+   openssl pkcs12 -in elastic-stack-ca.p12 -out cacert.pem
```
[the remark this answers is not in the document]
inqueue: I missed it since openssl on macOS behaves slightly differently than Linux when an export password is not supplied. Sorry for that, I've added `-nokeys` to correct it.

docs/command_line_reference.rst
```diff
@@ -674,7 +674,7 @@ Enable it with ``use_ssl:true``.
 **TLS/SSL Certificate Verification**
 
 Server certificate verification is controlled with the ``verify_certs`` boolean. The default value is `true`. To disable use ``verify_certs:false``.
-If ``verify_certs:true``, Rally will attempt to verify the certificate provided by Elasticsearch. If they are private certificates, you will also need to supply the private CA certificate using ``ca_certs:'/path/to/cacert.pem'``.
+If ``verify_certs:true``, Rally will attempt to verify the certificate provided by Elasticsearch. If they are private self-signed certificates, you will also need to supply the private self-signed CA certificate in PEM format using ``ca_certs:'/path/to/cacert.pem'``.
```
pquentin, line 677: The wording was incorrect, even before this PR. The server (Elastic) does not present a self-signed a certificate, it presents a server certificate that was signed by a root CA. And yes, all root CA are self-signed, but I think "private" is a better description than "self-signed".

```suggestion
If ``verify_certs:true``, Rally will attempt to verify the certificate provided by Elasticsearch. If this certificate is signed by a private certificate authority (CA), you will also need to supply that CA in PEM format using ``ca_certs:'/path/to/cacert.pem'``.
```

You don't have to agree, please merge with or without this change. It's fine to merge a doc change on Friday.

pquentin (approved):
Thank you for this effort! Hopefully it will help users with PKCS#12 certificates (even though --pem is a thing).

LGTM!

docs/command_line_reference.rst
inqueue: I like the revised wording, thank you!

PR #1405, opened 2021-12-20
Add track dependency support

### LINKED ISSUE #1396: Allow Tracks to Declare Additional Dependencies
If track code would require additional libraries to be installed, it would be good for them to be able to declare them.

I propose we extend the Track Registry concept to allow for additional dependencies, for example:

**track.py**
```python
def register(registry):
  registry.register_dependency("numpy", "==1.21.4")
```

Which would be calling...
```python
def register_dependency(library, version_string):
```

...and would invoke the runtime-local `pip` in-code, in the fashion mentioned [in the pip docs](https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program) (as a `subprocess`).

Potential goblins:
1. I'm not sure whether we should try to perform an uninstall of any additional libraries after the race ends (or on any failure, or whatever)
2. Registry invocations are re-run in every worker process, due to `load_track` being used as a way to ensure all Actors have the track code classloaded.  We would likely only want to run `pip` on the first call per `DriverActor`, so some check for the library's existence (and compliance with the `version_string`) is appropriate

DJRickyB:
Closes #1396.
This PR adds a `dependencies` section to the Track specification. Track dependencies are `pip install`-able targets, which are sent to the newly-established `.rally/lib` directory, which is on the `sys.path` for all workers, and is cleaned between benchmark invocations.

This functionality is to support more complex track code (such as in Track Processors, Custom Runners, or Custom Parameter Sources), which may require additional python packages to be installed prior to runtime.


esrally/track/loader.py
```diff
@@ -1115,6 +1135,24 @@ def meta_data(self):
...
+    """
+    Installs track-specific dependencies to the local runtime
+    """
+
+    def __init__(self, install_dependencies):
+        self.install_dependencies = install_dependencies
+
+    def __call__(self, name):
+        if self.install_dependencies:
+            console.info(f"Installing track dependency {name}")
```
pquentin, line 1148: Is it expected that I'm not seeing this the the CI log? https://elasticsearch-ci.elastic.co/job/elastic+rally+pull-request/24/consoleFull

pquentin (changes requested):
Installing dependencies like this seems a bit off, but the code looks good!

I don't think keeping this undocumented is a good idea though. Since we're going to use this, we should document it, if only to be able to refer to it ourselves. We can always mark it as experimental, as we do in https://esrally.readthedocs.io/en/stable/cluster_management.html.

esrally/track/loader.py
```diff
@@ -1115,6 +1135,24 @@ def meta_data(self):
...
+            try:
+                with open(os.path.join(paths.logs(), "dependency.log"), "ab") as install_log:
+                    subprocess.check_call([sys.executable, "-m", "pip", "install", f"{name}"], stdout=install_log, stderr=install_log)
```
dliappis, line 1151: One problem I see with this way of installation, is that we support multiple ways of installing Rally:

While for development environments we expect most people to use a venv, [we document](https://esrally.readthedocs.io/en/latest/install.html#installing-rally) (and have uses of, e.g. in nightly envs) installing Rally using the [user mode scheme](https://docs.python.org/3/install/index.html#alternate-installation-the-user-scheme). These [user installs](https://pip.pypa.io/en/stable/user_guide/#user-installs), are done using the `--user` flag.

Doing a `python3 -mpip install` here will not honor any user setups and in those cases end up installing the package in unexpected locations.

I don't think it's a big problem determining if `esrally` has been installed with `--user`; we could leverage the information in https://docs.python.org/3/library/site.html#site.USER_SITE or maybe there are even simpler ways.

esrally/track/loader.py
DJRickyB: What I know is that this will get muted by `--quiet`, but what I don't know currently is how all the ITs seem to have quiet output, since only a few actually specify it

esrally/track/loader.py
DJRickyB: IMO supportability concerns such as you outline here (and worse, the ones we aren't anticipating), as well as--for lack of a better word--stupidity concerns that might fundamentally change how this feature is implemented and consumed after trying it in the real world are both the reasons I'd go the extreme `TrackPreparator` route and leave this undocumented.

@pquentin I think your point is valid, indeed this is exactly what an experimental feature designation is for, but there's a precedent for the *really* shaky (AKA evolving) stuff that keeps us from even documenting how to use it until it's had some time to harden and evolve. If this is not compelling to you, though, I'll go the documented-but-unsupported route as you previously suggested

pquentin: I *think* I understand your concerns, but I still don't agree, sorry. :( In Rally, I'd say documentation is as import as tests. Would you say "this is experimental, let's skip the tests"?

It's also simple to add, you only need a short new section at the bottom of docs/adding_tracks.rst with a warning plus a code block with the call to `register.register_dependency` you have in the tests. This should not prevent us from fearlessly updating/removing this feature.

We can ask others to act as tie breakers if you'd prefer?

DJRickyB: Will write up the feature, and tag it as experimental.  Thanks

pquentin:
What is the status of this? As far as I can tell the tests pass, the failure is a transient one.

@elasticmachine please run tests

DJRickyB:
@pquentin The failure seems legit. I will fix the test, document the feature, then re-request your review when it's ready.

DJRickyB:
Reviewers, I've removed the "experimental" designation from the description and ":internal" from the labels. Our incubation period/process for experimental features is not well-defined, and I think bringing one more thing out of experimental state without a defined process just adds to our team's considerable tech debt.

esrally/paths.py
```diff
@@ -41,6 +41,10 @@ def install_root(cfg=None):
     return os.path.join(races_root(cfg), install_id)
 
 
+def libs():
+    return os.path.join(rally_confdir(), "lib")
```
pquentin, line 45: nit: this should probably be called "libs" or "dependencies" not just "lib".

The issue in putting this in ~/.rally/lib is that it will affect all tracks and will cause havoc when multiple tracks start using this feature. The other options I see are the track directory (but we need to tell git to ignore the libs) and the race directory.

docs/track.rst
```diff
@@ -3162,3 +3181,26 @@ You can also specify a number of clients on sub tasks explicitly (by default, on
...
+    "git+https://github.com/elastic/apm-agent-python@main",
+    "eland<8.0.0"
+  ]
+
+
+Rally will manage the installation of the libraries, given the environmentally-available ``pip`` and the target directory ``~/.rally/lib``
+
+.. warning::
+
+    Because modules will be loaded in the same environment as ``esrally`` itself, conflicts with Rally and its own dependencies are possible with the usage of Track Dependencies.
```
pquentin, line 3205: We need to document that we're using a different directory (maybe say where it is?) and mention that it will have precedence over other environments as we're using `sys.path.insert(0)`.

pquentin (changes requested):
Thanks, I still need to test but have a number of concerns. I believe being careful is warranted as dependency issues are hard to track in Python and many things can go wrong that will be hard for non-Python experts to fix.

esrally/paths.py
DJRickyB: The code fully wipes and re-creates the directory before `sys.path.insert(0)` to avoid collisions with other dependencies

pquentin: Oh, I see, not sure how I missed it. Can we also document this?

docs/track.rst
DJRickyB: addressed in line 3243 above. the specifics of using `sys.path.insert(0)` feels like implementation detail leak, specifically since the `libs` directory should not be loaded if nothing is declared for dependencies

docs/track.rst
```diff
@@ -126,17 +126,18 @@ We recommend the following path:
...
+* :ref:`indices <track_indices>`
+* :ref:`templates <track_templates>`
+* :ref:`data-streams <track_data_streams>`
+* :ref:`composable-templates <track_composable_templates>`
+* :ref:`component-templates <track_component_templates>`
+* :ref:`corpora <track_corpora>`
+* :ref:`operations <track_operations>`
+* :ref:`schedule <track_schedule>`
+* :ref:`challenge(s) <track_challenge>`
+* :ref:`dependencies <track_dependencies>`
```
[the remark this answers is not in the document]
pquentin: dependencies is now under "Example" -- is that intentional?

esrally/track/loader.py
```diff
@@ -189,15 +191,40 @@ def load_track(cfg):
...
 
 
-def _load_single_track(cfg, track_repository, track_name):
+def _install_dependencies(dependencies):
+    def _cleanup():
+        # fully destructive is fine, we only allow one Rally to run at a time and we will rely on the pip cache for download caching
+        shutil.rmtree(paths.libs(), onerror=_trap)
+
+    def _trap(function, path, exc_info):
+        logging.debug("Failed to clean up [%s] with [%s]", path, function, exc_info=True)
```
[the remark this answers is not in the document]
pquentin: Do we need `_trap` at all? (I'm truly wondering, this is not an ask to remove it)

docs/track.rst
pquentin: While `sys.path.insert(0)` is an implementation detail, what are your thoughts on just mentioning that the declared `dependencies` will take precedence over already install libraries?

docs/track.rst
```diff
@@ -2940,6 +2959,30 @@ The following meta data is always returned:
...
+.. _track_dependencies:
+
+dependencies
+............
+
+If your track code requires additional Python dependencies at runtime (for instance, for code included in :ref:`custom runners <adding_tracks_custom_runners>` or :ref:`parameter sources <adding_tracks_custom_param_sources>`) you can declare pip-installable packages in your track's ``dependencies`` section.
+``dependencies`` is an array of ``pip install`` targets, which Rally will manage when the benchmark begins::
+
+  "dependencies": [
+    "pytoml",
```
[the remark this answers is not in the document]
DJRickyB: lack of development is why I thought it was a good library to use in the example without a pinned version

esrally/track/loader.py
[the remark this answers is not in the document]
DJRickyB: I like the idea of contextualizing errors for user consumption, though that idea that I needed added context was prior to documenting where dependencies would be installed. I still think it's good to have (note we're commenting here on an older rev, this currently throws an error)

esrally/track/loader.py
```diff
@@ -189,15 +191,41 @@ def load_track(cfg):
...
+def _install_dependencies(dependencies):
+    def _cleanup():
+        # fully destructive is fine, we only allow one Rally to run at a time and we will rely on the pip cache for download caching
+        shutil.rmtree(paths.libs(), onerror=_trap)
+
+    def _trap(function, path, exc_info):
+        logging.exception("Failed to clean up [%s] with [%s]", path, function, exc_info=True)
+        raise exceptions.SystemSetupError(f"Unable to clean [{paths.libs()}]. See Rally log for more information.")
+
+    for dependency in dependencies:
```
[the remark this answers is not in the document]
DJRickyB: this is a great idea. i will re-tool and test this change

esrally/track/loader.py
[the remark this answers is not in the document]
pquentin: Yes I realize this now throws an error. Fair enough, resolving.

docs/track.rst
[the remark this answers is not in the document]
pquentin: I see you also use it in tests to do something real! I think just importing it was enough. But indeed it's a good idea in the it test to use a library that will never be a dependency of Rally. Resolving.

esrally/track/loader.py
```diff
@@ -189,15 +191,41 @@ def load_track(cfg):
...
     repo = track_repo(cfg)
-    return _load_single_track(cfg, repo, repo.track_name)
+    return _load_single_track(cfg, repo, repo.track_name, install_dependencies)
 
 
-def _load_single_track(cfg, track_repository, track_name):
+def _install_dependencies(dependencies):
+    def _cleanup():
+        # fully destructive is fine, we only allow one Rally to run at a time and we will rely on the pip cache for download caching
+        shutil.rmtree(paths.libs(), onerror=_trap)
```
pquentin, line 200: I've used it and `console.info` would be nice here, in the same way that "Installing track dependency X" is nice

pquentin (approved):
Thank you! And sorry for the delay.

michaelbaamonde:
Late to the party, but I'll take a look at this today.

pquentin (approved):
Thanks! LGTM.

PR #1406, opened 2022-01-05
Revert "Use explicit config files for black and isort"

pquentin:
This reverts commit 9c4505f7f3b64d61e9e102049ff82c2b586a8ab3. The
problem with the pyproject.toml file was that it enabled PEP 517, but we
don't want it because editable installs don't work as PEP 660 is not
implemented yet.

However, having a pyproject.toml file is the only practical way of
configuring black, and we can opt out of PEP 517 using
`--no-use-pep517`.

A few months later, now that our internal tooling is ready, we can have a pyproject.toml file again. This way, we won't have to configure black line length manually in our editors.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,0 @@
-[settings]
-profile = black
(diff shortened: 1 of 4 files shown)


DJRickyB (approved):
LGTM

Issue #1409, opened 2022-01-07
🐞 parse error if args contains multi colons when run command `esrally race`


Kua-Fu:
### I find a little bug🐞 ,  when I run such command 

```
esrally race --pipeline=benchmark-only --track=http_logs --challenge=index-data --target-hosts=xxx:443 --client-options="use_ssl:true,verify_certs:true,basic_auth_user:'elastic',basic_auth_password:'12345abcde:'"
```

### the error message like that 

```
2022-01-07 07:16:03,253 -not-actor-/PID:32347 esrally.rally ERROR A fatal error occurred while running subcommand [race].
Traceback (most recent call last):
  File "rally/lib/python3.9/site-packages/esrally/rally.py", line 964, in dispatch_sub_command
    configure_connection_params(arg_parser, args, cfg)
  File "rally/lib/python3.9/site-packages/esrally/rally.py", line 888, in configure_connection_params
    client_options = opts.ClientOptions(args.client_options, target_hosts=target_hosts)
  File "rally/lib/python3.9/site-packages/esrally/utils/opts.py", line 184, in __init__
    self.parse_options()
  File "rally/lib/python3.9/site-packages/esrally/utils/opts.py", line 193, in parse_options
    self.parsed_options = to_dict(
  File "rally/lib/python3.9/site-packages/esrally/utils/opts.py", line 86, in to_dict
    return default_parser(csv_to_list(arg))
  File "rally/lib/python3.9/site-packages/esrally/utils/opts.py", line 206, in normalize_to_dict
    return {TargetHosts.DEFAULT: {**default_client_map, **kv_to_map(arg)}}
  File "rally/lib/python3.9/site-packages/esrally/utils/opts.py", line 72, in kv_to_map
    k, v = kv.split(":")
ValueError: too many values to unpack (expected 2)
```

### Cause of problem ⚠️

as my password contains colons, so when split with `:` will generate error

![image](https://user-images.githubusercontent.com/13880225/148515017-2b874a3b-6234-4799-974c-006f45a2f1a6.png)

### Solution ✅

![image](https://user-images.githubusercontent.com/13880225/148515242-590f5712-d31e-4508-b877-2f965e20521c.png)

pquentin:
Thank you for your interest in Rally!

However, I don't think the proposed fix adds much value, sorry. Instead, to handle special characters, send a JSON file to `--client-options`, that will work with any special character: https://esrally.readthedocs.io/en/stable/command_line_reference.html#id2

[...]

PR #1411, opened 2022-01-07
🐞 parse error if args contains multi colons, detail in issue #1409

Kua-Fu:


pquentin:
Thank for your interested and going as as far as submitting a pull request! However, I'll be closing this, see the explanation in https://github.com/elastic/rally/issues/1409.

PR #1418, opened 2022-01-17
Derive correct artifact name for ARM architecture

danielmitterdorfer:
To download an Elasticsearch artifact, Rally needs to resolve the CPU
architecture of the current system to download the correct distribution
tarball. This does not work currently on Macs with ARM M1 chips. These
are reported as `arm64` but the Elasticsearch artifact uses `aarch64` to
refer to this CPU architecture. Consequently, downloading an artifact on
ARM-based Macs fails unless the user specifies `--target-arch=aarch64`.

With this commit we detect this case and ensure that the correct
artifact can be downloaded. We intentionally change the architecture
specifier only in the component that is responsible for deriving the
download URL for artifacts but keep referring to this CPU architecture
as `arm64` in other parts of the code, just as it is reported by the OS.

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -243,7 +243,8 @@ class TemplateRenderer:
         if arch is not None:
             self.arch = arch
         else:
-            self.arch = sysstats.cpu_arch().lower()
+            derived_arch = sysstats.cpu_arch().lower()
+            self.arch = "aarch64" if derived_arch == "arm64" else derived_arch
 
     def render(self, template):
         substitutions = {"{{VERSION}}": self.version, "{{OSNAME}}": self.os, "{{ARCH}}": self.arch}
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -244,6 +244,7 @@ class TemplateRenderer:
             self.arch = arch
         else:
             derived_arch = sysstats.cpu_arch().lower()
+            # Elasticsearch artifacts for Apple Silicon use "aarch64" as the CPU architecture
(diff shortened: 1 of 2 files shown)


esrally/mechanic/supplier.py
```diff
@@ -243,7 +243,8 @@ def __init__(self, version, os_name=None, arch=None):
         if arch is not None:
             self.arch = arch
         else:
-            self.arch = sysstats.cpu_arch().lower()
+            derived_arch = sysstats.cpu_arch().lower()
+            self.arch = "aarch64" if derived_arch == "arm64" else derived_arch
```
pquentin, line 247: Maybe a small comment would explain this confusing line?

```suggestion
            derived_arch = sysstats.cpu_arch().lower()
            # Elasticsearch artifacts for Apple Silicon use "aarch64" as the CPU architecture
            self.arch = "aarch64" if derived_arch == "arm64" else derived_arch
```

pquentin (approved):
Thanks! Please merge with or without my suggestion.

esrally/mechanic/supplier.py
dliappis: Isn't it the other way around, i.e. Linux reports `aarch64` but Apple `arm64`?

pquentin: Apple does report `arm64`, but the Apple Silicon URL is `https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.16.3-darwin-aarch64.tar.gz`, which is why Rally needs to use `aarch64` when the OS reports `arm64`

danielmitterdorfer: Thanks for the suggestion. I've committed your comment as well. Once CI is green, I'll merge it.

PR #1422, opened 2022-01-28
Rename 'indices stats API' to 'index stats API'

pquentin:
It was [renamed in 2019](https://github.com/elastic/elasticsearch/pull/46322) and is documented as "index stats" since 7.3: https://www.elastic.co/guide/en/elasticsearch/reference/7.3/indices-stats.html. (The URL did not change, but the title did.)

I'm not changing the `node-stats-include-indices` to avoid having a lengthy migration. Should I have done that?


dliappis (approved):
LGTM thank you!

Issue #1430, opened 2022-02-03
Failing test: test_eventdata_daily_volume[in-memory-it] 


b-deam:
The `test_eventdata_daily_volume[in-memory-it]` test seems to fail somewhat randomly with exit code `16384` (WTF?):

https://github.com/elastic/rally/blob/7dac7cb00c9d9199c198d8bf1be25d268766eaaf/it/distribution_test.py#L134-L138

We should probably look into why.

pquentin:
So, why exit code `16384`? It's because the integration tests use `os.system()` whose return code is opaque. `os.waitstatus_to_exitcode` can be used to retrieve the actual exit code, which is 64 in that case. This is what Rally returns  when the result is `ExitStatus.ERROR`.

https://github.com/elastic/rally/blob/846b8e404b4b4c324cc95e6621f4992d0023da81/esrally/rally.py#L1071-L1074

So that's not really helpful. The only logs we currently get are:

```
07:22:50 it/distribution_test.py::test_eventdata_daily_volume[in-memory-it] 
07:22:50     ____        ____
07:22:50    / __ \____ _/ / /_  __
07:22:50   / /_/ / __ `/ / / / / /
07:22:50  / _, _/ /_/ / / / /_/ /
07:22:50 /_/ |_|\__,_/_/_/\__, /
07:22:50                 /____/
07:22:50 
07:22:56 [INFO] Race id is [ca1c4c24-8a14-487a-a6ce-c618aa1704dd]
07:22:56 [ERROR] Cannot race. Worker [0] has exited prematurely.
07:22:56 
07:22:56 Getting further help:
07:22:56 *********************
07:22:56 * Check the log files in /var/lib/jenkins/workspace/elastic+rally+pull-request/.rally/logs for errors.
07:22:56 * Read the documentation at https://esrally.readthedocs.io/en/latest/.
07:22:56 * Ask a question on the forum at https://discuss.elastic.co/tags/c/elastic-stack/elasticsearch/rally.
07:22:56 * Raise an issue at https://github.com/elastic/rally/issues and include the log files in /var/lib/jenkins/workspace/elastic+rally+pull-request/.rally/logs.
07:22:56 
07:22:56 -------------------------------
07:22:56 [INFO] FAILURE (took 6 seconds)
07:22:56 -------------------------------
```

Also not helpful! We definitely need to show the Rally logs in case of failure. (And note that this isn't specific to `test_eventdata_daily_volume`, I've seen this with other tests.)

[...]

PR #1432, opened 2022-02-03
Show Rally log on it test failure

pquentin:
Relates #1430


pquentin:
This is still a work in progress, besides the syntax errors I need to figure out how to only print the new lines from the log. Or ask Rally to log somewhere else entirely for each new it test, possibly.

pquentin:
Agreed, closing for now, even if I like seeing the Rally log next to the failure.

PR #1433, opened 2022-02-11
minor change to docs

DJRickyB:


inqueue (approved):
This is the right change to clarify bulk `action_and_meta_data` is to not be included in the corpora document count when it is included in the documents file. LGTM

PR #1437, opened 2022-02-16
Support csv values as JSON

dliappis:
Support JSON (via a filename, ending in .json, or as a JSON inline
string) for all cli arguments accepting csv.

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -976,16 +976,21 @@ Examples, assuming that two clusters have been specified with ``--target-hosts``
 .. WARNING::
    If you use ``client-options`` you must specify options for **every** cluster name defined with ``target-hosts``. Rally will raise an error if there is a mismatch.
 
-Command line parameters accepting comma-separated values (csv)
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Command line parameters accepting comma-separated values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Parameters that can accept comma-separated values such as ``--car``, ``--telemetry``, ``--include-tasks`` etc. can also accept a JSON array.
 This can either be defined in a file ending in ``.json`` or passed as an inline JSON string.
 
 Examples:
 
+* comma-separated values::
+
+    --car="4gheap,trial-license"
+
 * json file: ``--car="car.json"``::
 
+    $ cat car.json
     ["4gheap", "trial-license"]
 
 * json inline string::
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -975,3 +975,19 @@ Examples, assuming that two clusters have been specified with ``--target-hosts``
 
 .. WARNING::
    If you use ``client-options`` you must specify options for **every** cluster name defined with ``target-hosts``. Rally will raise an error if there is a mismatch.
+
+Command line parameters accepting comma-separated values (csv)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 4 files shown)


docs/command_line_reference.rst
```diff
@@ -975,3 +975,19 @@ Examples, assuming that two clusters have been specified with ``--target-hosts``
 
 .. WARNING::
    If you use ``client-options`` you must specify options for **every** cluster name defined with ``target-hosts``. Rally will raise an error if there is a mismatch.
+
+Command line parameters accepting comma-separated values (csv)
```
pquentin, line 979: I must be weird, but seeing "csv" immediately makes me think about CSV files that can be separated by newlines, and we're not supporting newlines.

```suggestion
Command line parameters accepting comma-separated values
```

docs/command_line_reference.rst
```diff
@@ -975,3 +975,19 @@ Examples, assuming that two clusters have been specified with ``--target-hosts``
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Parameters that can accept comma-separated values such as ``--car``, ``--telemetry``, ``--include-tasks`` etc. can also accept a JSON array.
+This can either be defined in a file ending in ``.json`` or passed as an inline JSON string.
+
+Examples:
```
pquentin, line 985: Should we also give a comma-separated example?

esrally/utils/opts.py
```diff
@@ -17,13 +17,25 @@
 
 import difflib
 import json
+import re
 
 from esrally.utils import io
 
+# detect (very simplistically) json that starts with an array
+RE_JSON_ARRAY_START = re.compile(r'^(\s*\[\s*\]|(\s*\[\s*".*))')
```
pquentin, line 25: The parentheses confused me. If we want to keep them, let's use them for each group, like `(...)|(...)`? But we can also leave them out:

```suggestion
# detect (very simplistically) json that starts with an empty array or array of strings
RE_JSON_ARRAY_START = re.compile(r'^\s*\[\s*\]|\s*\[\s*".*')
```

pquentin (changes requested):
Thank you! This will improve things. I've added a few inline comments.

esrally/utils/opts.py
dliappis: > If we want to keep them, let's use them for each group, like `(...)|(...)`?

Yes the intention was to have `(...)|(...)` but the end parens of the first group wrongly ended up at the end up of the expression.

Btw, I contemplated on using `r"""` and having line separated entries, but it felt a bit of an overkill for a regex that is supposed to just check those two things.

I feel the parens add value to emphasize that we are checking two different things, so I'd go for

`r'^(\s*\[\s*\])|(\s*\[\s*".*)'`

pquentin: Sounds good!

docs/command_line_reference.rst
dliappis: We already do for each individual option, e.g. for [car](https://github.com/elastic/rally/blame/c1bba05b7bbed682063bd7cde8f34af32c2d2cd2/docs/car.rst#L35), but I guess you feel that here it'd make sense to present all possibilities as a convenience (even if it adds some docs duplication)?

pquentin: Right, that's my point, aiming for completeness seems useful. We could even link to this section from the individual options, but that may be too much effort.

I'll trust your judgment on this one, you can leave things as is too.

esrally/utils/opts.py
dliappis: Fixed in 837bd15ba7890f1ba055324855f08d0014e20e98

docs/command_line_reference.rst
dliappis: Fixed in 837bd15ba7890f1ba055324855f08d0014e20e98

docs/command_line_reference.rst
dliappis: Fixed in 837bd15ba7890f1ba055324855f08d0014e20e98

pquentin (approved):
Thanks! LGTM.

Issue #1382, opened 2021-11-11
Add native support for SQL and EQL queries


Luegg:
For benchmarking SQL and EQL queries we're currently using custom runners (see [here for EQL](https://github.com/elastic/rally-tracks/blob/master/eql/track.py) and [here for SQL](https://github.com/elastic/rally-tracks/pull/216)).

As @danielmitterdorfer [pointed out](https://github.com/elastic/rally-tracks/pull/216#discussion_r747310975), there is some potential to improve the benchmarks by discarding the response instead of parsing it (as usually done by the ES client).

Luegg:
Additionally to a native "sql" operation type it would also be very useful to have a "sql-scroll" operation analogous to [scroll-search](https://esrally.readthedocs.io/en/latest/track.html#scroll-search).

michaelbaamonde:
Thanks for raising this @Luegg. We'd definitely like to limit the need for custom runners as much as possible, especially in our default tracks. I'm working in this general area of the Rally code right now, so I'm assigning myself, but if someone else wants to give it a shot, please feel free.

PR #1438, opened 2022-02-17
Introduce `sql` operation type

Luegg:
Addresses #1382 for SQL.

~The operation type is named `sql-query` to avoid naming conflicts with the custom operation type defined in https://github.com/elastic/rally-tracks/blob/master/sql/track.py~ The operation type is named `sql` like the custom operation type defined in https://github.com/elastic/rally-tracks/blob/master/sql/track.py.

The operation performs pagination if the `pages` parameter is set.

Because `json` is the only format that allows cursor extraction from the response body (instead of the response header), the operation does currently not support overriding the `format` parameter.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2907,7 +2907,7 @@ the `SQL search API <https://www.elastic.co/guide/en/elasticsearch/reference/cur
 Properties
 """"""""""
 
-* ``body``: The request body passed to the _sql API with the initial request. Must contain the "query" key. Subsequent pagination requests will only include the cursor in the request body.
+* ``body`` (mandatory): The request body passed to the _sql API with the initial request. Must contain the "query" key. Subsequent pagination requests will only include the cursor in the request body.
 * ``pages``: (optional, defaults to 1) Number of pages to retrieve at most for this search. If a query yields fewer results than the specified number of pages it will fail with an exception.
 
 **Example**
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2898,6 +2898,48 @@ Meta-data
 
 This operation returns no meta-data.
 
+sql-query
+~~~~~~~~~~~~~~~~~~~
+
+Executes an SQL query and optionally paginates through subsequent pages. The ``body`` property will directly be passed to
+the `SQL search API <https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-search-api.html>`_.
+
+Properties
+""""""""""
+
+* ``body``: The request body passed to the _sql API with the initial request. Must contain the "query" key. Subsequent pagination requests will only include the cursor in the request body.
+* ``pages``: (optional, defaults to 1) Number of pages to retrieve at most for this search. If a query yields fewer results than the specified number of pages it will fail with an exception.
+
+**Example**
+
+Run an SQL query and fetch the first 10 pages with 100 records each::
+
+    {
+      "operation": {
+        "operation-type": "sql-query",
+        "name": "select-station-names",
+        "body": {
+          "query": "SELECT station.name FROM \"weather-data-2016\"",
+          "page_timeout": "1000ms",
+          "request_timeout": "1000ms",
+          "fetch_size": 100
+        },
+        "pages": 10
+      },
+      "target-throughput": 10,
+      "clients": 2,
+      "warmup-iterations": 50,
+      "iterations": 100
+    }
+
+Meta-data
+"""""""""
+
+The following meta data is always returned:
+
+* ``weight``: The number of fetched pages. Should always equal to the ``pages`` parameter.
+* ``unit``: The unit in which to interpret ``weight``. Always "ops".
+
 Examples
 ========
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2898,7 +2898,7 @@ Meta-data
 
 This operation returns no meta-data.
 
-sql-query
+sql
 ~~~~~~~~~~~~~~~~~~~
 
 Executes an SQL query and optionally paginates through subsequent pages. The ``body`` property will directly be passed to
@@ -2916,7 +2916,7 @@ Run an SQL query and fetch the first 10 pages with 100 records each::
 
     {
       "operation": {
-        "operation-type": "sql-query",
+        "operation-type": "sql",
         "name": "select-station-names",
         "body": {
           "query": "SELECT station.name FROM \"weather-data-2016\"",
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -56,7 +56,7 @@ def register_default_runners():
     register_runner(track.OperationType.DeleteAsyncSearch, DeleteAsyncSearch(), async_runner=True)
     register_runner(track.OperationType.OpenPointInTime, OpenPointInTime(), async_runner=True)
     register_runner(track.OperationType.ClosePointInTime, ClosePointInTime(), async_runner=True)
-    register_runner(track.OperationType.SqlQuery, SqlQuery(), async_runner=True)
+    register_runner(track.OperationType.Sql, Sql(), async_runner=True)
 
     # This is an administrative operation but there is no need for a retry here as we don't issue a request
     register_runner(track.OperationType.Sleep, Sleep(), async_runner=True)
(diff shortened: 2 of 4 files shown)


michaelbaamonde:
> The operation type is named sql-query to avoid naming conflicts with the custom operation type defined in https://github.com/elastic/rally-tracks/blob/master/sql/track.py

Right now, the `sql` track defines a custom runner for an operation type named `sql` on the `master`, `7.14`, and `7.12` branches of `rally-tracks`. This native operation type is meant to supersede that custom one, meaning the custom one will be removed once this native one is available in Rally. 

So, I'm thinking we should come up with a way to name _this PR's operation type_ `sql`, as it's just a nicer name. But, we'd need to do so without accidentally breaking the current `sql` track, which runs nightly with a build of `rally/master`, or breaking users of released versions of Rally.

  There's a bit of a race condition across repos, so I'm thinking that the cleanest way to approach this would be to:

  1. Issue a PR against `rally-tracks/master` that renames the custom `sql` operation type something else (maybe `custom-sql`).
  2. Change the name of the operation type in this PR to `sql`.
  3. Merge the `rally-tracks` PR from (1) and backport it to `7.12` and `7.14`.
  4. Once this Rally PR is approved and merged, issue a PR against `rally-tracks/master` that removes the `custom-sql` runner and instead references the native `sql` runner now available in Rally.
  5. At the time of the next Rally release, migrate the `7.12` and `7.14` branches of `rally-tracks` to the native `sql` runner by backporting the PR merged in (4). This delayed backport ensures that we don't break users of the currently-released version of Rally, which does not have the `sql` operation type.

This way, we avoid the naming collision, get a cleaner name for the native operation type, and avoid the possibility of accidentally breaking `rally-tracks` during the time between merging this PR that introduces the native operation and merging the `rally-tracks` PR that migrates to using it. There's other ways we could do it, but this method checks all the boxes and minimizes the potential for breakage and confusion. For example, I wouldn't want to be in a state where `rally-tracks/7.14` defines a custom runner called `sql` while `rally/master` also does.

  Curious what you think about this @DJRickyB @inqueue . Maybe I'm over-complicating it.

DJRickyB:
Devil's advocate thoughts:

1. I don't know for certain that this collision is problematic, in other words it's possible that registering the custom operation continues to work in the context of the SQL track, and we can unregister it once this is supported natively.
2. I think `sql-query` is a fine name for the op. We use `scroll-search` for the scrolling runner that wraps queries, and who's to say other SQL-related operations might not get introduced later
3. step 5 doesn't prevent older Rally version users from breaking on the SQL track, but there's not any way I can see of avoiding that anyway.

Those said, if it's generally agreed that `sql` is a better name and that it's necessary to change the custom operation name, then the cleanest way to do it would be as you've outlined.

michaelbaamonde:
> I think sql-query is a fine name for the op. We use scroll-search for the scrolling runner that wraps queries, and who's to say other SQL-related operations might not get introduced later

I have a mild preference for `sql`, but really just want to make sure that we name the operation what we really want it to be named as opposed to something that we come up with just to get around the potential collision.

Luegg:
I ran a quick experiment and it looks like rally is just doing fine with the conflict and picks the custom runner. I guess this is the ideal outcome for this purpose and eliminates step 1.

I'm gonna rename the operation-type to sql then 👍

michaelbaamonde:
Okay, I think that's fine. In that case, once this PR is merged, I'd suggest that we remove the custom runner in the `sql` track on `rally-tracks/master`, but preserve it on `7.12` and `7.14` until the next Rally release. We're moving to a model where we guarantee that that `rally-tracks` branches corresponding to Elasticsearch versions are guaranteed to work with the latest released version of Rally, while `rally-tracks/master` is only guaranteed to be compatible with `rally/master`. So, this sequence of PRs is in that spirit.

I'll have an actual review of the code ready today. Thanks for your patience @Luegg!

docs/track.rst
```diff
@@ -2898,6 +2898,48 @@ Meta-data
...
+sql
+~~~~~~~~~~~~~~~~~~~
+
+Executes an SQL query and optionally paginates through subsequent pages. The ``body`` property will directly be passed to
+the `SQL search API <https://www.elastic.co/guide/en/elasticsearch/reference/current/sql-search-api.html>`_.
+
+Properties
+""""""""""
+
+* ``body``: The request body passed to the _sql API with the initial request. Must contain the "query" key. Subsequent pagination requests will only include the cursor in the request body.
```
michaelbaamonde, line 2910:
```suggestion
* ``body`` (mandatory): The request body passed to the _sql API with the initial request. Must contain the "query" key. Subsequent pagination requests will only include the cursor in the request body.
```

michaelbaamonde: Let's mark this as mandatory. We do that throughout these docs for required params.

esrally/driver/runner.py
```diff
@@ -2437,6 +2438,41 @@ def __repr__(self, *args, **kwargs):
...
 
 
+class Sql(Runner):
+    """
+    Executes an SQL query and optionally paginates through subsequent pages.
+    """
+
+    async def __call__(self, es, params):
+        body = mandatory(params, "body", self)
+        mandatory(body, "query", str(self) + ".body")
```
michaelbaamonde, line 2448: This is interesting. I'm not sure that we have another case of a nested parameter that's mandatory. This function call will wind up surfacing an error message like this if the `body` doesn't contain a `query` key:

```
Parameter source for operation 'sql.body' did not provide the mandatory parameter 'query'.
```
That's a little confusing. `query` is a required key nested under the required `body` param, so we'd want the error message to make that clear and say something like:

```
Parameter source for operation 'sql' did not provide the mandatory parameter 'body.query'.
```
The `mandatory` function can't really express this, since it only tests for the existence of of a top-level key. I'd suggest that we do the check here instead like this:

```suggestion
        if body.get("query") is None:
            raise exceptions.DataError(
                f"Parameter source for operation 'sql' did not provide the mandatory parameter 'body.query'. "
                f"Add it to your parameter source and try again."
            )
```

tests/driver/runner_test.py
```diff
@@ -4913,6 +4913,90 @@ async def test_delete_ilm_policy_without_request_params(self, es):
...
+        assert ctx.value.message == "Result set has been exhausted before all pages have been fetched, 1 page(s) remaining."
+
+        es.transport.perform_request.assert_has_calls(
+            [
+                mock.call("POST", "/_sql", body=params["body"]),
+                mock.call("POST", "/_sql", body={"cursor": self.default_response["cursor"]}),
+            ]
+        )
+
+
```
michaelbaamonde, line 4999: We should add a test that ensures that an exception is raised if the `body` and `body.query` parameters aren't included. Something like this would do it:

```suggestion
    @mock.patch("elasticsearch.Elasticsearch")
    @run_async
    async def test_mandatory_body_param(self, es):
        sql_runner = runner.Sql()
        params = {
            "operation-type": "sql",
            "pages": 3,
        }

        with pytest.raises(exceptions.DataError) as exc:
            await sql_runner(es, params)
        assert exc.value.args[0] == (
            "Parameter source for operation 'sql' did not provide the mandatory parameter 'body'. "
            "Add it to your parameter source and try again."
        )
        
    @mock.patch("elasticsearch.Elasticsearch")
    @run_async
    async def test_mandatory_query_in_body_param(self, es):
        sql_runner = runner.Sql()
        params = {
            "operation-type": "sql",
            "body": {},
            "pages": 3,
        }

        with pytest.raises(exceptions.DataError) as exc:
            await sql_runner(es, params)
        assert exc.value.args[0] == (
            "Parameter source for operation 'sql' did not provide the mandatory parameter 'body.query'. "
            "Add it to your parameter source and try again."
        )
```

michaelbaamonde (commented):
This looks good overall! I tested locally with a version of the `sql` track without its custom runner, and it worked great.

I added some suggestions regarding validation of required params and corresponding tests and a minor nitpick on the docs. Assuming you and CI are both happy with those suggested changes, this will be good to merge.

We may want to create a parameter source for this class in the future, but I don't see it as a blocker for merging this PR.

Luegg:
Thanks a lot for your suggestions @michaelbaamonde! I applied all of them and just run the linter & fmt to make CI happy.

I will then go ahead with removing the custom `sql` runner from rally tracks.

michaelbaamonde (approved):
LGTM, thanks!

pquentin:
Thanks for the pytest conversion!

PR #1444, opened 2022-02-22
Add Ingest Processor Line Chart to Chart Generator

DJRickyB:
Here we take advantage of the recent addition of the ingest telemetry device and add a longitudinal chart to the chart generator.  Here is the output for `solutions/logs` when we add `ingest` to its charts list:
image


b-deam (approved):
LGTM

PR #1446, opened 2022-02-24
Activate the Makefile venv in a single place

pquentin:
It makes it easier to read and less error-prone.


pquentin:
It's a make limitation, the source has to be in the same process, which isn't the case with my changes.

PR #1447, opened 2022-02-24
Stop leading users away from finding errors in docs

pquentin:
There's a link to the metrics store page below and the first part of the
section does not need a metrics store. This confused me multiple times.

### DIFF
--- a/docs/recipes.rst
+++ b/docs/recipes.rst
@@ -215,7 +215,7 @@ This behavior can also be changed, by invoking Rally with the :ref:`--on-error <
 
 	esrally race --track=geonames --on-error=abort
 	
-Errors can also be investigated if you have configured a :doc:`dedicated Elasticsearch metrics store </configuration>`.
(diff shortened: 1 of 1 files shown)


docs/recipes.rst
```diff
@@ -215,7 +215,7 @@ This behavior can also be changed, by invoking Rally with the :ref:`--on-error <
 
 	esrally race --track=geonames --on-error=abort
 	
-Errors can also be investigated if you have configured a :doc:`dedicated Elasticsearch metrics store </configuration>`.
+Errors can also be investigated: this is what the next section is about.
```
DJRickyB, line 218: I'm not sure we need this line at all

DJRickyB (approved):
Looks fine. I'd just as soon delete it, I don't think docs need transitional language

docs/recipes.rst
pquentin: Agreed, I removed it

PR #1449, opened 2022-03-02
Add "revisions" chart type to chart generator

Luegg:
Adds the ability to generate dashboards with a revisions table. E.g. given the race config

```
[
  {
    "track": "sql",
    "placement": 1,
    "flavors": [
      {
        "name": "default",
        "licenses": [
          {
            "name": "trial",
            "configurations": [
              {
                "name": "sql-1node",
                "label": "sql",
                "charts": [
                  "query",
                  "gc",
                  "io",
                  "revisions"
                ],
                "challenge": "sql",
                "car": "4gheap",
                "x-pack": [
                  "security"
                ]
              }
            ]
          }
        ]
      }
    ]
  }
]
```

the command `esrally generate charts --configuration-name=nightly --chart-spec-path=../night-rally/night_rally/resources/race-configs-group-2.json --chart-type=time-series --output-path=../nightly-sql.ndjson` will produce a dashboard with the following table:

image

The `BarCharts` chart type could in theory use the same implementation for the revisions table. But since bar charts are used for comparing releases, the revision table is not particularly helpful in these dashboards.

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -572,6 +572,10 @@ class BarCharts:
         # TBD
         return None
 
+    @staticmethod
+    def revisions_table(title, environment, race_config):
+        return None
+
 
 class TimeSeriesCharts:
     @staticmethod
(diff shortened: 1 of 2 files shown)


esrally/chart_generator.py
```diff
@@ -1292,6 +1296,83 @@ def ingest(environment, race_config, title):
...
+                        "listeners": {},
+                    }
+                ),
+                "uiStateJSON": "{}",
+                "description": "revisions",
+                "version": 1,
+                "kibanaSavedObjectMeta": {
+                    "searchSourceJSON": json.dumps(
+                        {
+                            "index": "13969350-badf-11ea-af86-7d06bde52cfd",
```
pquentin, line 1359:
```suggestion
                            "index": str(uuid.uuid4()),
```

pquentin (changes requested):
Wow, thanks! Won't be able to try it today but noticed a hardcoded UUID

esrally/chart_generator.py
Luegg: Oh, I forgot to point that out, I wanted to ask wether there is a way to work around the UUID. It's the id of the `rally-races-*` index pattern on elasticsearch-benchmark-analytics.elastic.co. It's the only way I found to reference the index pattern. For `rally-results-*` this has been solved by changing the id of the index pattern to `rally-results-*` (see for example https://github.com/elastic/rally/blob/8b01f3c853564a6d48337b4d3eb5b81698024be1/esrally/chart_generator.py#L180 ).

One way to resolve this would be to also change the id of the `rally-races-*` index pattern but I'm unsure wether this would have other side effects (or maybe create a copy).

pquentin: That would be nice, and I'm not too worried about side effects. But how can this be done at all? I don't know where to find the ids of index patterns. Or change them.

Luegg: I managed to recreate the index pattern with the `rally-races-*` id by exporting the saved object and re-importing it with the new id. Accordingly, we can use the nicer id here.

pquentin (approved):
Thanks, this works well! I've discussed with @dliappis and we decided to enable this for all charts given how useful this is. I also fixed a small Python issue due to one of our dependencies updating and will merge when the tests pass.

Luegg:
that's great to hear :) I also thought about adding a race timestamp column that shows the race time in the format accepted by `night-rally-admin add annotation` (e.g. "20220322T180125Z") but that's not so straight forward and would probably require a runtime field in the `rally-races-*` index pattern to do the date formatting.

PR #1450, opened 2022-03-02
Switch default number alignment to decimal

nik9000:


nik9000:
I had no idea rally could support "decimal" alignment. It's what I've always wanted. And I could have just turned it on. And I never knew it. But! Should it be the default? I certainly like it better. I tend to manually reformat data from rally to be aligned this way.

nik9000:
If we don't merge this and you stumble on this in the future and want to align decimals the magic invocation is ` --report-numbers-align decimal`

nik9000:
And it looks like:
```
|           tsdb container.id doc values |        |    19.1982      |     kB |
|                tsdb container.id total |        |    37.1816      |     kB |
|                  tsdb _tsid doc values |        |    37.2734      |     kB |
|                       tsdb _tsid total |        |    37.2734      |     kB |
| tsdb kubernetes.pod.uid inverted index |        |    25.1602      |     kB |
|     tsdb kubernetes.pod.uid doc values |        |    26.3398      |     kB |
|          tsdb kubernetes.pod.uid total |        |    51.5         |     kB |
|             tsdb _source stored fields |        |   358.075       |     kB |
|                     tsdb _source total |        |   358.075       |     kB |

```

pquentin:
@danielmitterdorfer did not like the `decimal` alignment in #1202:

> "IMHO that makes the layout a bit more "noisy" so I'd rather not enable this by default. Maybe we can make this configurable though."

I'm happy with this change, since I think I prefer it and given the older way will still be available. I'd like to know what others think. @elastic/es-perf 

(BTW it's weird that we don't just settle on a given amount of significant figures instead of making the number of decimals seemingly dependent on the magnitude on the value. But I have no idea how the code works.)

nik9000:
If we settled on one or two decimal places it'd make the problem moot,
yeah. Settling in three significant figures or something would be better, I
think. If we did that we could still want the decimal alignment.

dliappis:
> I'm happy with this change, since I think I prefer it and given the older way will still be available. I'd like to know what others think.

IMHO we aren't striving to provide spreadsheet like flexibility in the tabular output of Rally; it should be focused on providing the quickest way for interpretation of results by *humans*. For all other cases, incl. further analysis of metrics (and accuracy) an ES metric store should be preferred.

> (BTW it's weird that we don't just settle on a given amount of significant figures instead of making the number of decimals seemingly dependent on the magnitude on the value. But I have no idea how the code works.)

This is a hard one and obviously largely a matter of taste. I'd personally like consistency to 3 decimal places and alignment on decimal, however, for larger magnitude results like indexing throughput, seeing something like `52123.456 docs/s` would be a bit pointless/unnecessarily verbose. Therefore, also combining my thinking above, I think I am good with this PR as is, i.e. decimal alignment and only show more decimals where it makes sense (ideally at maximum 3 decimal points).

pquentin (changes requested):
Thanks! Looks like the consensus is to do this. Can you please update the command line reference docs and maybe set `self.numbers_align` to `decimal` in reporter_test.py?

nik9000:
> 52123.456 docs/s

It's not related to this PR, but I think this'd be better as

> 52000 docs/s

At least, better for my human eyes. 3 significant figures.

Also I'm :+1: on not worrying too much about spreadsheet like stuff. The report is so that people can read it. Or copy it into github. So other people can read it.

I'll make the requested changes soon!

nik9000:
Ready for another round @pquentin !

tests/reporter_test.py
```diff
@@ -29,20 +29,33 @@ def setup_method(self, method):
 
         self.metrics_header = ["Metric", "Task", "Baseline", "Contender", "Diff", "Unit", "Diff %"]
         self.metrics_data = [
-            ["Min Throughput", "index", "17300", "18000", "700", "ops/s", "4.04%"],
-            ["Median Throughput", "index", "17500", "18500", "1000", "ops/s", "5.71%"],
-            ["Max Throughput", "index", "17700", "19000", "1300", "ops/s", "7.34%"],
+            ["Min Throughput", "index", 17300, 18000, 700, "ops/s", "4.04%"],
+            ["Median Throughput", "index", 17500, 18500, 1000, "ops/s", "5.71%"],
+            ["Max Throughput", "index", 17700, 19000, 1300, "ops/s", "7.34%"],
+            ["Bubblegum", "", 1.7, 1.9, 0.2, "bubbles/flavonoid", "11.71"],
```
pquentin, line 35:
```suggestion
            ["Bubblegum", "", 1.7, 1.9, 0.2, "bubbles/flavonoid", "11.71%"],
```

(this will require changing the tests below)

tests/reporter_test.py
```diff
@@ -29,20 +29,33 @@ def setup_method(self, method):
...
         ]
-        self.numbers_align = "right"
 
-    def test_formats_as_markdown(self):
+    def test_formats_as_markdown_empty(self):
+        self.numbers_align = "ignored"
```
pquentin, line 39: Can you please avoid using a class member here? Either by fully inlining it in the function call or just calling it `numbers_align` without self. That will be one less thing to worry about if we want to run tests in parallel one day. :)

pquentin (changes requested):
Looks good, thanks for adding tests! We're nearly there.

tests/reporter_test.py
nik9000: Sure! I'm all for contained tests. For what it's worth in junit-land it spins up entirely new instances of the test class for each run. I have no idea what pytest does though. *and* junit doesn't run tests for the same class in parallel. At least ours doesn't. Either way. Sure!

nik9000:
Fixes pushed!

pquentin (approved):
Thanks!

PR #1428, opened 2022-02-01
Add disk usage telemetry device

nik9000:
Relates #1289

PR #1452, opened 2022-03-03
Default disk-usage-stats telemetry indices

nik9000:
When I made the `disk-usage-stats` telemetry device in #1428 there
wasn't a way to use have it collect telemetry from the indices defined
in `track.json`. This adds such a way.

### DIFF
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -290,9 +290,9 @@ disk-usage-stats
 
 The disk-usage-stats telemetry device runs the `_disk_usage <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-disk-usage.html>`_ API after the track has completed and adds the disk used of each field to the report.
 
-Required telemetry parameters:
+Supported telemetry parameters:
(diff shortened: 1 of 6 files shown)


pquentin (approved):
Thanks! This is an awesome addition, thanks for spotting that it could be done that way!

Let's merge this today otherwise the nightlies won't measure anything.

pquentin:
We hit https://github.com/elastic/rally/issues/1430.

nik9000:
:+1:

I have maybe an hour before I have to go out for a bit. It looks like the failing test is the docker distribution trying to run the geonames track in test mode. I'll see if I can reproduce it. In the mean time I've merged master and will push that.

nik9000:
Or maybe it's a known transient thing. OK then!

nik9000:
If I don't merge this in the next hour and the tests look good feel tree to merge. I'll get it when I get back if not.

nik9000:
Thanks for the merge @pquentin !

PR #1453, opened 2022-03-08
Run disk usage telemetry one index at a time

nik9000:
Elasticsearch can only queue 16 shards worth of disk usage analysis at a
time. If the track creates more than that many shard the disk usage
telemetry would get shard failures back. This helps with that by running
the indices one at a time.

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2258,19 +2258,30 @@ class DiskUsageStats(TelemetryDevice):
         import elasticsearch
 
         indices = self.telemetry_params.get("disk-usage-stats-indices", ",".join(self.index_names))
+        found = False
         for index in indices.split(","):
             self.logger.debug("Gathering disk usage for [%s]", index)
             try:
                 response = self.client.transport.perform_request("POST", f"/{index}/_disk_usage", params={"run_expensive_tasks": "true"})
             except elasticsearch.RequestError:
-                msg = "A transport error occurred while collecting disk usage"
+                msg = f"A transport error occurred while collecting disk usage for {index}"
                 self.logger.exception(msg)
                 raise exceptions.RallyError(msg)
+            except elasticsearch.NotFoundError:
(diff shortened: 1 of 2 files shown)


nik9000:
@DJRickyB pointed out that this is likely to fail if there are indices declared but not created. I'm not really sure how that can happen in rally, but if it can it'd fail here. I'm not sure what to do in that case though.

nik9000:
> @DJRickyB pointed out that this is likely to fail if there are indices declared but not created. I'm not really sure how that can happen in rally, but if it can it'd fail here. I'm not sure what to do in that case though.

We've talked some and decided that so long as any index is found we'll call that a success. But if all listed indices fail we'll fail the request. This is pretty much how it used to work. Only now we'll be iterating in rally.

nik9000:
@DJRickyB I've pushed an update that ignores 404s.

nik9000:
> @DJRickyB I've pushed an update that ignores 404s.

So long as one index doesn't 404.

b-deam (approved):
LGTM.

Nit: I did notice the use of some camel case for variables in the telemetry device (unrelated to this PR), but this isn't a huge deal and I don't think you need to change them as part of this PR ([see PEP8 style guide here](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)): https://github.com/elastic/rally/blob/78f1793f26ca07678e0d14c94b24068b5b19bb7b/esrally/telemetry.py#L2289-L2290

b-deam:
(Tested with `srally race --track http_logs \
     --telemetry disk-usage-stats  --test-mode --kill-running-processes`)

nik9000:
Old habits die hard. I'd love a lint rule catch myself. I'll have a look at
the camel case later today in another PR.

pquentin:
@dliappis tested all tracks in test mode, and that worked well (except security for an unrelated reason). Thank you!

PR #1455, opened 2022-03-09
Pick up data streams in disk-usage-stats telemetry

nik9000:
Some of our tracks just declare data streams.

### DIFF
--- a/tests/telemetry_test.py
+++ b/tests/telemetry_test.py
@@ -4150,16 +4150,21 @@ class TestIngestPipelineStats:
 
 class TestDiskUsageStats:
     @mock.patch("elasticsearch.Elasticsearch")
-    def test_uses_indices_param_by_default(self, es):
+    def test_uses_indices_and_data_streams_by_default(self, es):
         cfg = create_config()
         metrics_store = metrics.EsMetricsStore(cfg)
         tc = TransportClient(response={"_shards": {"failed": 0}})
         es = Client(transport_client=tc)
-        device = telemetry.DiskUsageStats({}, es, metrics_store, ["foo", "bar"])
+        device = telemetry.DiskUsageStats({}, es, metrics_store, index_names=["foo", "bar"], data_stream_names=["baz", "bork"])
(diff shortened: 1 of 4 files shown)


tests/telemetry_test.py
```diff
@@ -4150,24 +4150,31 @@ def test_logs_warning_on_missing_stats(self, metrics_put_value_cluster_level, me
 
 class TestDiskUsageStats:
     @mock.patch("elasticsearch.Elasticsearch")
-    def test_uses_indices_param_by_default(self, es):
+    def test_uses_indices_and_data_streams_by_default(self, es):
```
DJRickyB, line 4153: can you split this case, please? Per the docs, we can only use one or the other 
>The indices section contains a list of all indices that are used by this track. Cannot be used if the data-streams section is specified.

tests/telemetry_test.py
```diff
@@ -4150,24 +4150,31 @@ def test_logs_warning_on_missing_stats(self, metrics_put_value_cluster_level, me
...
 
     @mock.patch("elasticsearch.Elasticsearch")
     def test_uses_indices_param_if_specified(self, es):
         cfg = create_config()
         metrics_store = metrics.EsMetricsStore(cfg)
         tc = TransportClient(response={"_shards": {"failed": 0}})
         es = Client(transport_client=tc)
-        device = telemetry.DiskUsageStats({"disk-usage-stats-indices": "foo,bar"}, es, metrics_store, ["baz"])
+        device = telemetry.DiskUsageStats(
+            {"disk-usage-stats-indices": "foo,bar"}, es, metrics_store, index_names=["baz"], data_stream_names=["not_this"]
```
DJRickyB, line 4176: Likewise would be good here to only use index_names or data_stream_names

DJRickyB (changes requested):

DJRickyB (approved):
Looks good. Excited to get this for our data stream-based benchmarks also

nik9000:
Nice!

PR #1460, opened 2022-03-17
Add disk usage chart

nik9000:
This teaches rally how to build per-field disk usage charts based on the
`disk-usage-stats` telemetry device.

### DIFF
--- a/esrally/chart_generator.py
+++ b/esrally/chart_generator.py
@@ -1082,6 +1082,7 @@ class TimeSeriesCharts:
                 "kibanaSavedObjectMeta": {"searchSourceJSON": '{"query":"*","filter":[]}'},
             },
         }
+
     @staticmethod
     def query(environment, race_config, q, iterations):
         metric = "latency"
@@ -1463,7 +1464,6 @@ def generate_disk_usage(chart_type, race_configs, environment):
             structures.append(chart_type.disk_usage(title, environment, race_config))
 
     return structures
-
 
 
 def generate_gc(chart_type, race_configs, environment):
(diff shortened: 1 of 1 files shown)


esrally/chart_generator.py
```diff
@@ -1003,6 +1007,82 @@ def io(title, environment, race_config):
...
+                "drop_last_bucket": 0,
+                "time_field": "race-timestamp",
+                "type": "timeseries",
+                "filter": f"name:disk_usage_total {env_filter}",
+                "annotations": [
+                    {
+                        "fields": "message",
+                        "template": "{{message}}",
+                        "index_pattern": "rally-annotations",
+                        "query_string": f'((NOT _exists_:track) OR track:"{race_config.track}") AND ((NOT _exists_:chart) OR chart:io) '
```
DJRickyB, line 1058: `chart` in this query should reference `disk_usage`, not `io`

esrally/chart_generator.py
```diff
@@ -1003,6 +1007,82 @@ def io(title, environment, race_config):
...
+        }
+
+        return {
+            "id": str(uuid.uuid4()),
+            "type": "visualization",
+            "attributes": {
+                "title": title,
+                "visState": json.dumps(vis_state),
+                "uiStateJSON": "{}",
+                "description": "io",
```
DJRickyB, line 1080: It's in my mind to go through all these and make them more helpful. Since you should change this from `io` anyway, can you provide a short description here about the data/vis?

DJRickyB (changes requested):

esrally/chart_generator.py
nik9000: I've added a super short description. I could do a longer thing if you'd like, but I'm kind of close to the project so my tiny blurb feels fine to me. Limited vision. If you have an idea for a better description I'd be happy to take it.

nik9000:
> LGTM. I believe `make format` will fix your line length linter failure, then if CI passes we are all good here

I had to pull the query out to a variable. linter couldn't split the line.

PR #1461, opened 2022-03-21
Add 'root' attribute to Track class

DJRickyB:
This change adds a `root` attribute to the Track class, in order to provide an absolute path to original track files to any consumers, such as custom parameter sources.

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -1122,6 +1122,7 @@ class TrackSpecificationReader:
 
     def __init__(self, track_params=None, complete_track_params=None, selected_challenge=None, source=io.FileSource):
         self.name = None
+        self.base_path = None
         self.track_params = track_params if track_params else {}
         self.complete_track_params = complete_track_params
         self.selected_challenge = selected_challenge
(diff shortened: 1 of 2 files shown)


esrally/track/loader.py
```diff
@@ -1128,8 +1128,9 @@ def __init__(self, track_params=None, complete_track_params=None, selected_chall
         self.source = source
         self.logger = logging.getLogger(__name__)
 
-    def __call__(self, track_name, track_specification, mapping_dir):
+    def __call__(self, track_name, track_specification, mapping_dir, spec_file=None):
         self.name = track_name
+        self.base_path = os.path.dirname(os.path.abspath(spec_file)) if spec_file else None
```
dliappis, line 1133: should we add `self.base_path = None` in the constructor above? (like we do for `self.name`)

dliappis (approved):
LGTM! I left a question of small importance.

esrally/track/loader.py
DJRickyB: yes, great point. I think this is a typical pylint violation that we're currently excluding actually

PR #1464, opened 2022-03-23
Parse SQL cursor only when needed

Luegg:
The recent regression in the `select_star` task of the `sql` track revealed the cost of JSON parsing for extracting the cursor is quite significant for large responses. In the `select_star` case, the response size is ~500kb. For the other queries it is typically around 20kb and less of an issue.

This PR ensures that the cursor is only extracted if the `pages` task parameter is > 1.

The issue can also be reproduced locally:

Before the fix:
```
(.venv) ➜  rally git:(45e43e2) esrally race --pipeline=benchmark-only --target-host=127.0.0.1:39200 --track-path=~/benchmarks/rally-tracks/sql --challenge=sql --race-id=$(uuidgen) --include-tasks=select_star

...
|                                                 Min Throughput | select_star |     6.58        |  ops/s |
|                                                Mean Throughput | select_star |     6.82        |  ops/s |
|                                              Median Throughput | select_star |     6.88        |  ops/s |
|                                                 Max Throughput | select_star |     6.93        |  ops/s |
|                                        50th percentile latency | select_star |  9167.63        |     ms |
|                                        90th percentile latency | select_star | 14534.9         |     ms |
|                                        99th percentile latency | select_star | 15632.2         |     ms |
|                                       100th percentile latency | select_star | 15781           |     ms |
|                                   50th percentile service time | select_star |    45.6969      |     ms |
|                                   90th percentile service time | select_star |    55.9674      |     ms |
|                                   99th percentile service time | select_star |    85.2662      |     ms |
|                                  100th percentile service time | select_star |    95.9319      |     ms |
|                                                     error rate | select_star |     0           |      % |
...
```

with the fix:
```
(.venv) ➜  rally git:(45e43e2) esrally race --pipeline=benchmark-only --target-host=127.0.0.1:39200 --track-path=~/benchmarks/rally-tracks/sql --challenge=sql --race-id=$(uuidgen) --include-tasks=select_star --on-error=abort --kill-running-processes

...
|                                                 Min Throughput | select_star |  10.01        |  ops/s |
|                                                Mean Throughput | select_star |  10.02        |  ops/s |
|                                              Median Throughput | select_star |  10.02        |  ops/s |
|                                                 Max Throughput | select_star |  10.03        |  ops/s |
|                                        50th percentile latency | select_star |  61.4802      |     ms |
|                                        90th percentile latency | select_star |  65.5847      |     ms |
|                                        99th percentile latency | select_star |  79.2033      |     ms |
|                                       100th percentile latency | select_star |  86.5243      |     ms |
|                                   50th percentile service time | select_star |  60.3462      |     ms |
|                                   90th percentile service time | select_star |  64.4287      |     ms |
|                                   99th percentile service time | select_star |  78.549       |     ms |
|                                  100th percentile service time | select_star |  85.3458      |     ms |
|                                                     error rate | select_star |   0           |      % |
...
```

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2910,6 +2910,11 @@ Properties
 * ``body`` (mandatory): The request body passed to the _sql API with the initial request. Must contain the "query" key. Subsequent pagination requests will only include the cursor in the request body.
 * ``pages``: (optional, defaults to 1) Number of pages to retrieve at most for this search. If a query yields fewer results than the specified number of pages it will fail with an exception.
 
+    .. note::
+
+        If ``pages`` is greater than 1 the response body will be parsed to extract ``cursor`` for subsequent pagination requests.
+        This JSON parsing can be expensive and might have a significant impact on the measured latency if the response body is too large.
+
 **Example**
 
 Run an SQL query and fetch the first 10 pages with 100 records each::


esrally/driver/runner.py
```diff
@@ -2464,18 +2464,17 @@ async def __call__(self, es, params):
...
+        while pages > 0:
+            cursor = parse(r, ["cursor"]).get("cursor")
+
+            if not cursor:
+                raise exceptions.DataError(
+                    "Result set has been exhausted before all pages have been fetched, {} page(s) remaining.".format(pages)
+                )
 
-        while cursor and pages > 0:
             r = await es.transport.perform_request("POST", "/_sql", body={"cursor": cursor})
```
dliappis, line 2475: Should we clarify in the [docs](https://esrally.readthedocs.io/en/latest/track.html#sql) for this runner that when pages > 1 and the response is large, json parsing can be expensive and affect results (latency)?

dliappis (commented):
Looks good, left a comment because we are pushing the problem further down when pages > 1 and I think we should warn users about it.

esrally/driver/runner.py
Luegg: good point! I added a note to the docs.

dliappis (approved):
LGTM

PR #1471, opened 2022-04-01
Introduce field-caps operation-type

dnhatn:
The field-caps API can be found at https://www.elastic.co/guide/en/elasticsearch/reference/current/search-field-caps.html.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2972,7 +2972,7 @@ Retrieve `the capabilities of fields among indices <https://www.elastic.co/guide
 Properties
 """"""""""
 
-* ``index`` (optional, defaults to ``_all``): An `index pattern <https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#api-multi-index>`_ that defines which indices should be targeted by this operation.
+* ``index`` (optional, defaults to ``_all``): An `index pattern <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_ that defines which indices should be targeted by this operation.
 * ``fields`` (optional, default to ``*``): Comma-separated list of fields to retrieve capabilities for.
 * ``index_filter`` (optional): An index_filter to limit this operation to target only indices that match this index_filter
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2533,12 +2533,16 @@ Properties
 
 The ``composite`` operation only supports the following operation-types:
 
+* ``open-point-in-time``
+* ``close-point-in-time``
+* ``search``
+* ``paginated-search``
 * ``raw-request``
 * ``sleep``
-* ``search``
 * ``submit-async-search``
 * ``get-async-search``
 * ``delete-async-search``
+* ``field-caps``
 
 **Examples**
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -2978,7 +2978,7 @@ Properties
 
 **Example**
 
-Retrieve the capabilities of fields starting with `message` of the `log-*` indices::
+Retrieve the capabilities of fields starting with ``message`` of the ``log-*`` indices::
 
     {
       "operation": {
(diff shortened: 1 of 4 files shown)


docs/track.rst
```diff
@@ -2964,6 +2964,41 @@ The following meta data is always returned:
...
 
+field-caps
+~~~~~~~~~~~~~~~~~~~
+
+Retrieve `the capabilities of fields among indices. <https://www.elastic.co/guide/en/elasticsearch/reference/current/search-field-caps.html>`_.
+
+Properties
+""""""""""
+
+* ``index`` (optional, defaults to `_all`): An `index pattern <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_ that defines which indices should be targeted by this operation.
```
pquentin, line 2975:
```suggestion
* ``index`` (optional, defaults to ``_all``): An `index pattern <https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html>`_ that defines which indices should be targeted by this operation.
```

Have you considered linking to https://www.elastic.co/guide/en/elasticsearch/reference/current/api-conventions.html#api-multi-index instead? Landing on a page with only two links is a bit unsettling, but it may be the best we have.

docs/track.rst
```diff
@@ -2964,6 +2964,41 @@ The following meta data is always returned:
...
+* ``fields`` (optional, default to `*`): Comma-separated list of fields to retrieve capabilities for.
+* ``index_filter`` (optional): An index_filter to limit this operation to target only indices that match this index_filter
+
+**Example**
+
+Retrieve the capabilities of fields starting with `message` of the `log-*` indices::
```
pquentin, line 2981:
```suggestion
Retrieve the capabilities of fields starting with ``message`` of the ``log-*`` indices::
```

esrally/driver/runner.py
```diff
@@ -2482,6 +2483,26 @@ def __repr__(self, *args, **kwargs):
...
+    async def __call__(self, es, params):
+        index = params.get("index", "_all")
+        fields = params.get("fields", "*")
+        body = params.get("body", {})
+        index_filter = params.get("index_filter")
+        if index_filter:
+            body["index_filter"] = index_filter
+        request_params = params.get("request-params")
+        await es.field_caps(index=index, body=body, fields=fields, params=request_params)
+
```
pquentin, line 2501: We're missing the return value here. If you look at the runner class, all runner need to return the number of operations that were done. This is important for the summary and reported metrics.

Have you tested this operation in an actual track?

```suggestion
        await es.field_caps(index=index, body=body, fields=fields, params=request_params)
        
        return {"weight": 1, "unit": "ops", "success": True}
```

Can you also please change the tests so that they check that this was returned?

dnhatn: What should we return here? We only need to measure the latency of these requests.

docs/track.rst
dnhatn: Good suggestion. I just copied this from other runners.

dnhatn:
@pquentin Thank you for your review. I've applied your suggestions. I also pushed https://github.com/elastic/rally/pull/1471/commits/5229307369c3fabfe58e3cc8c69b79aacc082c0b to make the field-caps a composite runner as we will use field-caps and search requests together in tracks.

esrally/driver/runner.py
pquentin: For the record, we need to return success and number of operations, per the suggestion you applied.

docs/track.rst
dnhatn: I've reverted to using the previous link to be consistent with other runners. If we decide to change this link, I think we need to fix other links too.

pquentin (approved):
Thanks! LGTM

dnhatn:
@pquentin Thank you for reviewing :).

PR #1475, opened 2022-04-11
Add raw-request op type for latency charts

inqueue:
Add the `raw-request` operation type to the chart generator for latency charts.


DJRickyB (approved):
LGTM. Thank you

Issue #1487, opened 2022-05-10
[Bug] Composable templates getting silently filtered


DJRickyB:
In https://github.com/elastic/rally/blob/master/esrally/track/params.py#L447, we assert that if and only if a provided template has the top-level key `"template"` is it a [composable template](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-template.html). This is not true per the API, and also generally may not be maintainable to try to check spec of things before allowing to send them to Elasticsearch (and fail, and log an error).

It's my belief this is not intentional, particularly as it causes templates provided in the `composable-templates` entry of a track and without the top-level `template` field get silently filtered, the task fails with a weird error referencing `request_start` and there's no information in the log about what happened.

This logic should be more straightforward, providing whatever template(s) I give it as the correct parameter for the `create-composable-template` task. 

As a hint, I think this may mean getting rid of the `TemplateParamSource` ABC and just spelling out each of its subclasses independently, but I haven't looked that hard at this (as you can tell by my half-hearted attempt in #1486  😄 )

PR #1489, opened 2022-05-16
Not every composable template has a top-level `template`

cavokz:
Fixes https://github.com/elastic/rally/issues/1487


DJRickyB (approved):
LGTM

PR #1492, opened 2022-05-17
Address wrapped templates

cavokz:
This allows to use integration assets dumped by `elastic-package` as-is, which enclose an outer meta data layer otherwise rejected when sent to Elastisearch.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -283,6 +283,7 @@ Each composable template in this list consists of the following properties:
 * ``index-pattern`` (mandatory): Index pattern that matches the composable template. This must match the definition in the template file.
 * ``delete-matching-indices`` (optional, defaults to ``true``): Delete all indices that match the provided index pattern if the template is deleted.
 * ``template`` (mandatory): Composable template file name.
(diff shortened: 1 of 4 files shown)


DJRickyB (commented):
Thank you for this. Please add a description to the issue (we are doing this because `elastic-package dump` produces wrapped objects, I think that's fine to mention here), and some documentation to the track reference docs

DJRickyB (approved):
This LGTM with one nit. I don't think we should work to support a leading `.` in the `template-path` (for the root element), but I think one is implied by the wording `JSON path` in the docs/docstrings. I know this was my wording, but if you agree I think it's better documented as the `JSON field inside the file content that contains the template`. That way, we support the dot-notation in the way that you've coded but there's no reason for anyone to plug a leading `.`

With just a docs change to this effect, I don't see a need for further review, but please feel free to re-request

DJRickyB:
@elasticmachine run tests

DJRickyB:
@elasticmachine update branch

PR #1496, opened 2022-05-20
Fix use_ssl: True on Python 3.10 (second try)

pquentin:
Rally is a client, so its purpose is to authenticate servers. This means
that we should use ssl.Purpose.SERVER_AUTH instead of CLIENT_AUTH, with
or without client certs. This is super confusing, see
https://github.com/python/cpython/issues/73996. Anyway, in 3.10 making this mistake isn't possible
anymore due https://github.com/python/cpython/pull/26646.
    
To make sure this does not break in the future, we also add tests for
the IP and client certs cases.


inqueue (approved):
LGTM, thanks!

PR #1498, opened 2022-05-20
Bump upper tested version to 8.4.0

pquentin:
Untested, let's see the results.


pquentin:
After failing to bisect this, I finally realized the bug is not about the Elasticsearch version but the version of the track that is chosen by the Elasticsearch version.

The assertions introduced in https://github.com/elastic/rally-tracks/pull/156/files#r574499734 don't seem to work at all. Even though @dliappis tested them explicitly it's easy to miss that you need to use `--enable-assertions` which is disabled by default. @nik9000 Did you know about `--enable-assertions`?

```
esrally race --track=http_logs --test-mode --enable-assertions --distribution-version=7.12.0
```

And since those changes have not been backported to earlier versions of the track, 7.11.0 appears to work. I'll look into the track to see if we can use assertions that make sense.

pquentin:
Nevermind, the issue is specific to test mode. Running http_logs with test-mode and assertions will break.

pquentin:
Thanks for testing this locally. At least this pull request slightly improves the situation! :smile: 

I think we want to keep testing 6.8.0 though, but we can skip it on M1 Mac in a future PR?

michaelbaamonde:
@elasticmachine update branch

pquentin:
(I had to bump to 8.4 as nyc_taxis uses a new mode mapping on the main branch)

Issue #1052, opened 2020-08-25
Fail Gracefully when Additional Telemetry Stats are Unavailable


gingerwizard:
Reported via https://discuss.elastic.co/t/es-jvm-use-zgc-esrally-cant-work/246253

**Describe the feature**:

This may require further investigation to identify potential areas where Rally could be more robust to missing JSON keys in ES responses. The case above occurs when using ZGC and ES is unable to provide GC statistics through the node stats api.

sanjun1995:
Hi，gingerwizard
By commenting out some code, I can now test es cluster with ZGC using esrally.

at File /Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/esrally/telemetry.py

at def on_benchmark_stop(self):
 956                # young_gc_time = max(jvm_stats_end["young_gc"] - jvm_stats_start["young_gc"], 0)
 957                # old_gc_time = max(jvm_stats_end["old_gc"] - jvm_stats_start["old_gc"], 0)
 959                 # total_young_gen_collection_time += young_gc_time
 960                 # total_old_gen_collection_time += old_gc_time
 962                 # self.metrics_store.put_value_node_level(node_name, "node_young_gen_gc_time", young_gc_time, "ms")
 963                 # self.metrics_store.put_value_node_level(node_name, "node_old_gen_gc_time", old_gc_time, "ms")
 978         # self.metrics_store.put_value_cluster_level("node_total_young_gen_gc_time", total_young_gen_collection_time, "ms")
 979         # self.metrics_store.put_value_cluster_level("node_total_old_gen_gc_time", total_old_gen_collection_time, "ms")

at def jvm_stats(self):
 996            # old_gen_collection_time = gc["old"]["collection_time_in_millis"]
 997            # young_gen_collection_time = gc["young"]["collection_time_in_millis"]
 999                # "young_gc": young_gen_collection_time,
1000               # "old_gc": old_gen_collection_time,

If there is no mistake, just comment out these lines of code.

Although this is not a good way, the rest does not affect the test.

Because es does not support ZGC, esrally's code is still compatible with the interface data of young GC and old GC in the past, which leads to the current problem.

PR #1416, opened 2022-01-14
Add Ingest Pipeline Telemetry device

b-deam:
With this commit we had a Ingest Pipeline stats telemetry device. This 
telemetry device utilises the node stats API, but rather than sampling 
these statistics throughout the benchmark (i.e. node stats telemetry 
device) we collect an output at the beginning and end, and report only the 
delta.

The metrics store contains metrics at the cluster, node, pipeline, and processor levels under `rally-metrics*`, and summarised cluster level results under `rally-results*`.

**TODO**:
- [x] Add documentation

<details>
<summary>Sample output</summary>

```diff
$ esrally race --track-path=rally/tracks/ingest-pipelines --telemetry ingest-pipeline-stats --kill-running-processes

    ____        ____
   / __ \____ _/ / /_  __
  / /_/ / __ `/ / / / / /
 / _, _/ /_/ / / / /_/ /
/_/ |_|\__,_/_/_/\__, /
                /____/

[INFO] Race id is [4121d7ec-aee4-405e-a596-84dfa03a89e8]
[INFO] Preparing for race ...
[INFO] Racing on track [ingest-pipelines], challenge [bulk-to-pipeline-with-1001-failures] and car ['defaults'] with version [8.0.0-SNAPSHOT].

Running put-pipeline                                                           [100% done]
Running bulk                                                                   [100% done]

------------------------------------------------------
    _______             __   _____
   / ____(_)___  ____ _/ /  / ___/_________  ________
  / /_  / / __ \/ __ `/ /   \__ \/ ___/ __ \/ ___/ _ \
 / __/ / / / / / /_/ / /   ___/ / /__/ /_/ / /  /  __/
/_/   /_/_/ /_/\__,_/_/   /____/\___/\____/_/   \___/
------------------------------------------------------
            
|                                                         Metric |   Task |       Value |   Unit |
|---------------------------------------------------------------:|-------:|------------:|-------:|
|                     Cumulative indexing time of primary shards |        |   0.0370333 |    min |
|             Min cumulative indexing time across primary shards |        |   0.0370333 |    min |
|          Median cumulative indexing time across primary shards |        |   0.0370333 |    min |
|             Max cumulative indexing time across primary shards |        |   0.0370333 |    min |
|            Cumulative indexing throttle time of primary shards |        |           0 |    min |
|    Min cumulative indexing throttle time across primary shards |        |           0 |    min |
| Median cumulative indexing throttle time across primary shards |        |           0 |    min |
[...]

PR #1499, opened 2022-05-24
Flexible support of GC collectors in telemetry.

j-bennet:
Support collectors other than `young` and `old` when collecting GC stats in telemetry.

Fixes https://github.com/elastic/rally/issues/1052.


pquentin (approved):
Thank you, this works perfectly! This is a very polished contribution, it was a pleasure to review. Per our process, not merging, feel free to do it yourself.

If you're looking for a follow up task, since we now store ZGC information in the metrics store, should we go as far as adding the data in the race report and comparison report? This would involve modifying metrics.py and reporter.py, see https://github.com/elastic/rally/pull/1416 for an example.

Issue #1459, opened 2022-03-17
Running `esrally` without any arguments crashes


nik9000:
```
$ esrally
Traceback (most recent call last):
  File "/home/nik9000/Code/Elastic/rally/.venv/bin/esrally", line 33, in <module>
    sys.exit(load_entry_point('esrally', 'console_scripts', 'esrally')())
  File "/home/nik9000/Code/Elastic/rally/esrally/rally.py", line 1035, in main
    console.init(quiet=args.quiet)
AttributeError: 'Namespace' object has no attribute 'quiet'
```

I was hoping for the list of subcommands.

PR #1500, opened 2022-05-24
When running rally with no options, print help and exit.

j-bennet:
### Fixes https://github.com/elastic/rally/issues/1459.

When running `esrally` with no options, print help and exit.

Another option would be to specify `required=True` in `add_subparsers`, to enforce user choosing a subparser. But in that case, the message is less helpful:

```
usage: esrally [-h] [--version] [--quiet] [--offline] {race,list,info,create-track,generate,compare,download,install,start,stop} ...
esrally: error: the following arguments are required: subcommand
```

so I opted for printing help instead:

```
> esrally
usage: esrally [-h] [--version] [--quiet] [--offline] {race,list,info,create-track,generate,compare,download,install,start,stop} ...

    ____        ____
   / __ \____ _/ / /_  __
  / /_/ / __ `/ / / / / /
 / _, _/ /_/ / / / /_/ /
/_/ |_|\__,_/_/_/\__, /
                /____/

 You Know, for Benchmarking Elasticsearch.

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  --quiet               Suppress as much as output as possible (default: false).
  --offline             Assume that Rally has no connection to the Internet (default: false).

subcommands:
  {race,list,info,create-track,generate,compare,download,install,start,stop}
    race                Run a benchmark
    list                List configuration options
    info                Show info about a track
    create-track        Create a Rally track from existing data
    generate            Generate artifacts
    compare             Compare two races
    download            Downloads an artifact
    install             Installs an Elasticsearch node locally
    start               Starts an Elasticsearch node locally
    stop                Stops an Elasticsearch node locally

Find out more about Rally at https://esrally.readthedocs.io/en/latest/
```

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -661,7 +661,6 @@ def create_arg_parser():
     )
 
     for p in [
-        parser,
         list_parser,
         race_parser,
         compare_parser,
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -661,6 +661,7 @@ def create_arg_parser():
     )
 
     for p in [
+        parser,
         list_parser,
         race_parser,
         compare_parser,
@@ -1031,6 +1032,10 @@ def main():
 
     arg_parser = create_arg_parser()
     args = arg_parser.parse_args()
+
+    if args.subcommand is None:
+        arg_parser.print_help()
+        sys.exit(0)
 
     console.init(quiet=args.quiet)
     console.println(BANNER)
--- a/it/basic_test.py
+++ b/it/basic_test.py
@@ -1,0 +1,35 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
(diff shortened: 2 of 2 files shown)


esrally/rally.py
```diff
@@ -661,6 +661,7 @@ def add_track_source(subparser):
     )
 
     for p in [
+        parser,
```
pquentin, line 664: Is this actually needed given the `sys.exit(0)` call?

pquentin (changes requested):
Thank you! This is a small addition that will really help new users. Left one nit.

esrally/rally.py
j-bennet: Not anymore, you're right. I think I put this in place before I added the early exit. Removed.

pquentin (approved):
Thanks! LGTM.

Normally I'd ask you to add a test so that this does not breaks again in the future but that's not really practical with argparse and switching to click/typer isn't trivial.

j-bennet:
@pquentin 

> Normally I'd ask you to add a test so that this does not breaks again in the future but that's not really practical with argparse and switching to click/typer isn't trivial.

I saw there are integration tests. I added one - perhaps better than nothing?

PR #1501, opened 2022-05-25
metrics.rst: minor formatting correction

inqueue:
Correct minor text formatting issue as viewed at https://esrally.readthedocs.io/en/stable/metrics.html#task-operation-operation-type.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -95,7 +95,7 @@ task, operation, operation-type
 
 ``operation`` is the name of the operation (as specified in the track file) that ran when this metric has been gathered. It will only be set for metrics with name ``latency`` and ``throughput``.
 
-``operation-type`` is the more abstract type of an operation. During a race, multiple queries may be issued which are different operations but they all have the same ``operation-type`` (Search). For some metrics, only the operation type matters, e.g. it does not make any sense to attribute the CPU usage to an individual query but instead attribute it just to the operation type.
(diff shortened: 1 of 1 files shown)


docs/metrics.rst
```diff
@@ -95,7 +95,7 @@ task, operation, operation-type
 
 ``operation`` is the name of the operation (as specified in the track file) that ran when this metric has been gathered. It will only be set for metrics with name ``latency`` and ``throughput``.
 
-``operation-type`` is the more abstract type of an operation. During a race, multiple queries may be issued which are different ``operation``s but they all have the same ``operation-type`` (Search). For some metrics, only the operation type matters, e.g. it does not make any sense to attribute the CPU usage to an individual query but instead attribute it just to the operation type.
+``operation-type`` is the more abstract type of an operation. During a race, multiple queries may be issued which are different operations but they all have the same ``operation-type`` (Search). For some metrics, only the operation type matters, e.g. it does not make any sense to attribute the CPU usage to an individual query but instead attribute it just to the operation type.
```
pquentin, line 98:
```suggestion
``operation-type`` is the more abstract type of an operation. During a race, multiple queries may be issued which are different operations but they all have the same ``operation-type`` (search). For some metrics, only the operation type matters, e.g. it does not make any sense to attribute the CPU usage to an individual query but instead attribute it just to the operation type.
```

While we're at it, why not capitalize `search` properly?

pquentin (approved):
Thanks! LGTM

Issue #1468, opened 2022-03-24
Fail fast and suggest the user to run in `--offline` if no internet


dliappis:
While working on https://github.com/elastic/rally/pull/1467 we figured that the current mode of operation which a) validates internet connectivity by
attempting a successful connection to github.com and b) if not true, runs Rally in `offline` mode

is doing too much implicitly and might be difficult to troubleshoot if the intention is to run normally but there is a small github outage.

Instead an alternative would be that if all retries have failed, we should raise an error and advise the user to specify `--offline`, if that's what they intend.

PR #1503, opened 2022-05-27
Explicit offline mode

j-bennet:
### Closes https://github.com/elastic/rally/issues/1468.

In case there's no internet connection, exit and suggest using `--offline` mode, as opposed to falling back to offline mode without asking.

Added an integration test as well.

### DIFF
--- a/esrally/utils/process.py
+++ b/esrally/utils/process.py
@@ -28,11 +28,16 @@ def run_subprocess(command_line):
     return os.system(command_line)
 
 
-def run_subprocess_with_output(command_line):
+def run_subprocess_with_output(command_line, env_vars=None):
     logger = logging.getLogger(__name__)
     logger.debug("Running subprocess [%s] with output.", command_line)
     command_line_args = shlex.split(command_line)
-    with subprocess.Popen(command_line_args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT) as command_line_process:
+    kwargs = {}
+    if env_vars:
+        cmd_env = os.environ.copy()
+        cmd_env.update(env_vars)
+        kwargs["env"] = cmd_env
+    with subprocess.Popen(command_line_args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs) as command_line_process:
         has_output = True
         lines = []
         while has_output:
--- a/it/__init__.py
+++ b/it/__init__.py
@@ -189,7 +189,7 @@ class TestCluster:
 
 
 class EsMetricsStore:
-    VERSION = "7.6.0"
+    VERSION = os.getenv("it_es_version", "7.6.0")
 
     def __init__(self):
         self.cluster = TestCluster("in-memory-it")
--- a/it/basic_test.py
+++ b/it/basic_test.py
@@ -33,3 +33,11 @@ def test_run_with_help(cfg):
     output = process.run_subprocess_with_output(cmd)
     expected = "usage: esrally [-h] [--version]"
     assert expected in "\n".join(output)
+
+
+@it.rally_in_mem
+def test_run_without_http_connection(cfg):
+    cmd = it.esrally_command_line_for(cfg, "list races")
+    output = process.run_subprocess_with_output(cmd, {"http_proxy": "http://non-existent-proxy-server.com"})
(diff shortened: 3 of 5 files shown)


it/__init__.py
```diff
@@ -189,7 +189,7 @@ def __str__(self):
 
 
 class EsMetricsStore:
-    VERSION = "7.6.0"
+    VERSION = os.getenv("it_es_version", "7.6.0")
```
j-bennet, line 192: This allows using a different ES version when running integration tests locally.

esrally/utils/process.py
```diff
@@ -28,11 +28,16 @@ def run_subprocess(command_line):
...
     logger = logging.getLogger(__name__)
     logger.debug("Running subprocess [%s] with output.", command_line)
     command_line_args = shlex.split(command_line)
-    with subprocess.Popen(command_line_args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT) as command_line_process:
+    kwargs = {}
+    if env_vars:
+        cmd_env = os.environ.copy()
+        cmd_env.update(env_vars)
+        kwargs["env"] = cmd_env
+    with subprocess.Popen(command_line_args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, **kwargs) as command_line_process:
```
pquentin, line 40: nit: Do we really need kwargs here?

```suggestion
    env = None
    if env_vars:
        env = os.environ.copy()
        env.update(env_vars)
    with subprocess.Popen(command_line_args, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env) as command_line_process:
```

it/__init__.py
pquentin: Good idea. I'm curious: what did you need that for?

A few comments:

 * This only applies to the metrics store, and not `DISTRIBUTIONS`: the name should reflect that
 * To follow existing convention around env vars, can we also use caps and maybe prefix the environment variable with `RALLY`?

That would give `RALLY_IT_METRICS_ES_VERSION` which is maybe a bit too much. What do you think?

pquentin: (I now realize that Rally uses lower case for `http_proxy` but everything else is upper case, such as `RALLY_HOME` and `RALLY_RUNNING_IN_DOCKER`)

it/basic_test.py
```diff
@@ -33,3 +33,11 @@ def test_run_with_help(cfg):
     output = process.run_subprocess_with_output(cmd)
     expected = "usage: esrally [-h] [--version]"
     assert expected in "\n".join(output)
+
+
+@it.rally_in_mem
+def test_run_without_http_connection(cfg):
+    cmd = it.esrally_command_line_for(cfg, "list races")
+    output = process.run_subprocess_with_output(cmd, {"http_proxy": "http://non-existent-proxy-server.com"})
```
pquentin, line 41: Can you please use `http://invalid` instead of `http://non-existent-proxy-server.com`? It's reserved so no one will be able to register it: https://en.wikipedia.org/wiki/.invalid (Unlikely, but this happened to me in the past: I was using a GCS bucket name that did not exist in tests, eventually someone used it and it broke our CI.)

pquentin (changes requested):
This works great, thank you! I've asked a few minor changes.

it/__init__.py
j-bennet: ...or maybe I can leave this part as it was. It's probably not important what version of Elasticsearch is being used by metrics store.

j-bennet: Reverted the change.

it/basic_test.py
j-bennet: Changed.

it/basic_test.py
```diff
@@ -33,3 +33,11 @@ def test_run_with_help(cfg):
     output = process.run_subprocess_with_output(cmd)
     expected = "usage: esrally [-h] [--version]"
     assert expected in "\n".join(output)
+
+
+@it.rally_in_mem
+def test_run_without_http_connection(cfg):
+    cmd = it.esrally_command_line_for(cfg, "list races")
```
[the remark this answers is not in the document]
pquentin: Thanks for opening that issue!

PR #1505, opened 2022-05-31
Instantiate SubClient only with kwargs

pquentin:
When SubClient could only be used to mock the node stats API, this was
clear, but it's no longer the case. Thankfully most usage was using
keyword arguments already, so it could be enforced with low effort.

We apply this to the similar Client and TransportClient classes too.

I noticed this when reviewing #1499 from @j-bennet.


DJRickyB (approved):
Sorry for the delay. LGTM

PR #1502, opened 2022-05-26
Display ZGC collector stats in report

j-bennet:
Following up in https://github.com/elastic/rally/pull/1499.

This adds ZGC stats to the race report and comparison report.

I'm not very fond of all the hardcoded metric names, this can probably be improved.

Related PR: https://github.com/elastic/rally-teams/pull/74.

PR #1507, opened 2022-06-01
Remove unused analyze.py script

pquentin:
It can be used to generate image charts from race statistics (useful for docs and blog posts) but is currently broken and is blocking #1502.

I decided to continue linting the scripts/ directory even though there's no Python scripts there currently, but I could be convinced otherwise.


DJRickyB:
> I decided to continue linting the scripts/ directory even though there's no Python scripts there currently
Seems fine

DJRickyB (approved):
LGTM

PR #1510, opened 2022-06-03
Upgrade ES client to 8.2.0

### LINKED ISSUE #1350: Upgrade Elasticsearch client to 8.0.0
_Note: This is not actionable at the moment but acts as a reference of things to consider when upgrading the Elasticsearch client to 8.0.0._

Looking at the [Elasticsearch client roadmap](https://github.com/elastic/elasticsearch-py/issues/1696) a few items stand out but specifically https://github.com/elastic/elasticsearch-py/issues/1680 will be tricky: The client will not allow to pass a `body` but rather requires to specify the body in a more structured manner. Initially using `body` will only be deprecated but we need to think about alternatives.

michaelbaamonde:
This PR upgrades Rally's Elasticsearch Python client library dependency to `v8.2.0` from `v7.14.0`. The changes are pretty substantial. I've tried to make the commits reasonably-sized and atomic, so I suggest reviewing by commit. Nevertheless, some (especially those touching unit tests) are large and mostly consist of migrating from positional arguments to keyword arguments. 

The most important change is contained in e7d83cc60b890f7a1526958f808535e36f9bd126. The Python client has moved all transport logic into a separate library, https://github.com/elastic/elastic-transport-python. Low-level transport-related details have changed substantially, and our customizations within `esrally/client.py` and `esrally/async_connection.py` have had to change accordingly. It's a large diff, but it should all be transparent to the user.

I've also had to override the client's default behavior for verifying that a given server is authentically Elasticsearch. See the commit message (and comments in the code) for details, but the gist is that the client as of `v8.0.0` does this check exclusively via a header that's only available in ES `7.14.0`+. Previous client versions fell back to other mechanisms for verifying authenticity absent this header. Since Rally must support all versions of ES >= `6.8.0`, I've resurrected this prior behavior.

Otherwise, most changes are fairly mechanical. 

A few notes:

- I've tested this with every maintained track that I have access to. Even so, potential for breakage is very high, and I intend to do more. Please feel free to do the same!
- I've issued a [rally-tracks PR](https://github.com/elastic/rally-tracks/pull/271) and an [eventdata PR ](https://github.com/elastic/rally-eventdata-track/pull/113) to update some runners that currently fail with this PR, and will cause integration tests to fail.
- Same as above, but for a few internal tracks.
- There are some TODOs that you'll see in comments, particularly around exception handling, which has changed considerably. What I have right now works, but I don't consider it robust enough yet. I'm actively working on this. Any thoughts on this in particular would be helpful.
- The [`EsClientFactory` class](url) has gotten a bit too big, so I am happy to reorganize that code.
- There are a handful of additional tests that I have in mind that would be useful to add, but could probably be done in follow-up PRs. But if anything seems glaringly missing, please let me know.

Closes #1350.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -954,8 +954,12 @@ class Query(Runner):
                         took = props.get("took", 0)
                         all_results_collected = (size is not None and hits < size) or hits == 0
                     else:
-                        r = await es.transport.perform_request(
-                            "GET", "/_search/scroll", body={"scroll_id": scroll_id, "scroll": "10s"}, params=request_params, headers=headers
+                        r = await es.perform_request(
+                            method="GET",
+                            path="/_search/scroll",
+                            body={"scroll_id": scroll_id, "scroll": "10s"},
+                            params=request_params,
+                            headers=headers,
                         )
                         props = parse(r, ["timed_out", "took"], ["hits.hits"])
                         timed_out = timed_out or props.get("timed_out", False)
@@ -1011,7 +1015,7 @@ class Query(Runner):
             components.append(doc_type)
         components.append("_search")
         path = "/".join(components)
-        return await es.transport.perform_request("GET", "/" + path, params=params, body=body, headers=headers)
+        return await es.perform_request(method="GET", path=f"/{path}", params=params, body=body, headers=headers)
 
     def _query_headers(self, params):
         # reduces overhead due to decompression of very large responses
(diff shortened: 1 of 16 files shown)


esrally/driver/runner.py
```diff
@@ -1533,17 +1534,14 @@ async def __call__(self, es, params):
...
-                    f"/_xpack/ml/datafeeds/{datafeed_id}",
-                    body=body,
-                )
-            else:
-                raise e
+            await es.perform_request(
+                method="PUT",
+                path=f"/_xpack/ml/datafeeds/{datafeed_id}",
+                body=body,
+            )
```
pquentin, line 1544: Can you please mention that this is still needed for Elasticsearch 6.8? That when we drop support for it we'll remove it.

And somehow mention that this applies to all other datafeeds classes.

michaelbaamonde:
I'm going to break this up into a sequence of smaller PRs, so closing this one.

PR #1512, opened 2022-06-08
Prepare switch to Elasticsearch Python client 8.2.0

pquentin:
For compatibility with version 8.x of the Elasticsearch Python client. #1510 does it all in one go but I can't review that given the number of lines and the different concerns. This pull request contains self-sufficient commits for mechanical changes to reduce the scope of #1510. I'll be happy to fix the conflicts.

 * Switch from `es.xpack.ml` to `es.ml` (already possible with 7.14)
 * Switch from `es.transport.perform_request` to `es.perform_request` (required an adapter method)
 * Call Elasticsearch Python client with keyword arguments (and automatically convert path to url in `perform_request` calls)

### DIFF
--- a/esrally/client.py
+++ b/esrally/client.py
@@ -237,6 +237,7 @@ class EsClientFactory:
 
         class RallySyncElasticsearch(elasticsearch.Elasticsearch):
             def perform_request(self, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


michaelbaamonde (approved):
Looks good to me, thanks.

Issue #1506, opened 2022-05-31
Rally checks for internet connection even when it's not needed


j-bennet:
**Rally version** (get with `esrally --version`): 2.5.0.dev0

**Description of the problem including expected versus actual behavior**:

Rally unnecessarily checking Internet connection for commands such as:

* list [races|other things]
* compare --baseline x --contender y
* generate [charts]
* possibly others?

**Steps to reproduce**:

Simulate being offline by providing an invalid proxy URL. Invoke a command that should work offline:

`http_proxy=http://invalid esrally list races`

Observe a warning:

```
[WARNING] No Internet connection detected.
```

This check should only apply to commands that need internet connection.

PR #1514, opened 2022-06-08
Exclude some commands from online check.

j-bennet:
Some of the commands don't need Rally to be online.

Fixes https://github.com/elastic/rally/issues/1506.


pquentin (changes requested):
Fantastic that you're tackling this, thanks! Three notes:

* This breaks an integration test that was relying on `esrally list tracks` to test the internet connection behavior.
* Maybe we don't need tests for this in two places, but I think keeping the unit test to completely test the logic is a nice touch. Up to you. If you keep the test however, please mock the `net.has_internet_connection` call too, or running unit tests will get three seconds slower for me, which would be ironic as I run tests more than offline Rally commands.
* Is this still a draft? Do you expect review on drafts?

j-bennet:
@pquentin This is a draft because I wanted to get all tests passing before requesting a review. Yes, I'd rather remove the integration tests and replace them with unit tests where we can, in this case we can. Integration tests are slow.

j-bennet:
@pquentin

> Maybe we don't need tests for this in two places

What do you think about removing some of the integration tests, because the new unit test replaces them? I think that these can go away:

* https://github.com/elastic/rally/blob/28eea6c10dc158128dbc3cf06ed0de463312e434/it/proxy_test.py#L72-L94
* https://github.com/elastic/rally/blob/28eea6c10dc158128dbc3cf06ed0de463312e434/it/basic_test.py#L39-L43

j-bennet:
@pquentin Tests now pass, `net.has_internet_connection` is mocked, and I marked the PR as ready for review.

You'll notice that I removed `test_run_without_http_connection` in `it/basic_tests.py`, because the new unit test covers that.

I didn't remove `it/proxy_test.py`, because it seems like a lot of work went into these integration tests, but TBH, I think they are not necessary. What do you think?

esrally/rally.py
```diff
@@ -51,6 +51,8 @@
 from esrally.tracker import tracker
 from esrally.utils import console, convert, io, net, opts, process, versions
 
+OFFLINE_SUBCOMMANDS = ("compare", "list", "start", "stop", "generate")
```
pquentin, line 54: So this is complicated:

 * `start` can pull Docker images,
 * `stop` writes to the metrics store which can be a remote cluster.
 * `list races` and `list tracks` support fetching the data from a remote metrics cluster
 * `list tracks`, `list cars` and `list elasticsearch-plugins` need to fetch data from GitHub

So the commands that are truly offline are: compare, list telemetry, list pipelines, and generate.

esrally/rally.py
```diff
@@ -898,6 +900,21 @@ def configure_reporting_params(args, cfg):
     cfg.add(config.Scope.applicationOverride, "reporting", "numbers.align", args.report_numbers_align)
 
 
+def ensure_internet_connection(args, cfg, logger):
```
pquentin, line 903: For what it's worth, I'm not really happy with the name, there are two cases where no Internet connection is needed at all, but I haven't found a better name! :shrug:

pquentin (changes requested):
Thanks! Unfortunately, digging deeper it turns out that many commands do need Internet connection in special cases. I mentioned this inline. (To be honest I'm not sure why we need to [look before leaping](https://docs.python.org/3/glossary.html#term-lbyl), we could just try the command and report a connection error if it failed due to a bad connection...)

Regarding unit tests vs. integration tests, this is highly subjective. A few comments:

 1. The effort made for those integration tests is not a consideration, code is code, and removing code is always nice.
 2. But then, integration tests actually test the functionality end-to-end, eg. they exercise the `has_internet_connection` function that is otherwise mocked
 3. We also don't have any unit test that test the proxy case?
 4. And yes, it tests are so slow that I don't run them locally.

So I would tend towards leaving proxy_test.py. What do you think?

esrally/rally.py
j-bennet: 👍  Thanks, that's good to know. I'll make changes.

j-bennet:
> To be honest I'm not sure why we need to [look before leaping](https://docs.python.org/3/glossary.html#term-lbyl), we could just try the command and report a connection error if it failed due to a bad connection...

Same here, I'm starting to wonder if we need this check at all. It feels like we should be able to do without this check, or at least, if it's necessary for some reason, do it lazily as needed. I'm going to look closer at the code and figure out what's. the point of the offline mode. There's probably something I'm not seeing.

esrally/rally.py
j-bennet: Agree. I struggled with this one. I'll think about it some more.

j-bennet:
@pquentin 

> 2. But then, integration tests actually test the functionality end-to-end, eg. they exercise the has_internet_connection function that is otherwise mocked

Good point.

> 3. We also don't have any unit test that test the proxy case?

Ah. In that case, we should definitely keep them.

> So I would tend towards leaving proxy_test.py. What do you think?

I agree.

I looked through the code to figure out what the offline mode actually does. A couple of things:

1. `RallyRepository` (teams or tracks) won't attempt to update itself.
2. A custom `TrackProcessor` with a `downloader` will raise an error when `download` is called.

The offline feature was designed for corporate environments, where benchmarking clients are not able to access the Internet.

I don't see a good reason to keep the method `net.check_internet_connection`. By default, it checks if we have connection to `github.com`, or user can assign a custom `probing.url` in config, for example if their teams and tracks repos are hosted on the internal network. But why an extra check? We might as well try and fail when updating these repos.

At this point, I think I would:

1. remove `probing.url` and `net.check_internet_connection`
2. if `offline` mode is specified, rally shouldn't attempt to update repos, whether they are internal or not (and in any case, there's no good way to tell if they're internal)
3. if `offline` mode is not specified, rally should try updating repos, and fail on error.

What do you think? This would change the PR, but it makes more sense to me.

pquentin:
I do prefer #1517, but asked other to chime in first.

PR #1515, opened 2022-06-09
Replace os.system with subprocess.call

pquentin:
See https://docs.python.org/3/library/subprocess.html#replacing-os-system for more details. I think the default semantics are better.

One big advantage is that the exit code is not opaque with `subprocess.call`, which has slowed us down in https://github.com/elastic/rally/issues/1430.


michaelbaamonde (approved):
Nice, thanks.

Issue #1448, opened 2022-03-01
Prevent benchmark locks due to actor system internal message queue getting full


dliappis:
In various cases, e.g. when using lots of clients, or collecting lots of metrics via telemetry devices, we have witnessed hanging benchmarks. In summary, the observation is that either when too many log messages or too many metrics are sent, the benchmark may end up hanging.

Example error message can be (from `~/.rally/logs/actor-system-internal.log`):

```
...
2021-08-04 20:48:38.779581 p7907 Warn Transmit attempt from ActorAddr-(T|:41611) to ActorAddr-(T|:40337) timed out, returning PoisonPacket
2021-08-04 20:48:38.779778 p7907 I    completion error: ************* TransportIntent(ActorAddr-(T|:40337)-=-SENDSTS_FAILED-WAITSLOT-<class 'esrally.driver.driver.UpdateSamples'>-<esrally.driver.driver.UpdateSamples object at 0x7fb337558d30>-quit_0:00:00)
2021-08-04 20:48:38.779914 p7907 I    bad internal file descriptor!
2021-08-04 20:48:38.779968 p7907 I    wrecv 1168 is bad
2021-08-04 20:48:38.780125 p7907 I    bad internal file descriptor!
2021-08-04 20:48:38.780180 p7907 I    wrecv 1168 is bad
2021-08-04 20:48:38.780243 p7907 Warn Transmit attempt from ActorAddr-(T|:41611) to ActorAddr-(T|:40337) timed out, returning PoisonPacket
2021-08-04 20:48:38.780369 p7907 I    completion error: ************* TransportIntent(ActorAddr-(T|:40337)-=-SENDSTS_FAILED-WAITSLOT-<class 'esrally.driver.driver.JoinPointReached'>-<esrally.driver.driver.JoinPointReached object at 0x7fb337dd2610>-quit_0:00:00)
```

whereas the actual `~/.rally/logs/rally.log` for the same actor shows:

```
2021-08-04 20:43:36,649 ActorAddr-(T|:41611)/PID:7907 esrally.driver.driver INFO Total run duration: 16205.111997 seconds.
2021-08-04 20:43:36,650 ActorAddr-(T|:41611)/PID:7907 esrally.driver.driver INFO Total time to shutdown asyncgens: 0.000340 seconds.
2021-08-04 20:43:36,652 ActorAddr-(T|:41611)/PID:7907 esrally.driver.driver INFO Total time to close transports: -0.001832 seconds.
[...]

PR #1516, opened 2022-06-13
Create an ES client per simulated client instead of per worker.

michaelbaamonde:
This is a prerequisite for #1067 and also indirectly relates to #1350.

As [discussed](https://github.com/elastic/rally/issues/1067#issuecomment-1130566298) in the context of API keys, runners shouldn't have to concern themselves with ES transport options; they should just be provided ES client objects that already have the right options set. This PR is the first step in that direction.

Previously, an ES client was created per worker and shared among all of the worker's simulated clients. Now, each worker will instead create distinct ES client objects per simulated client that the worker is responsible for. This unique ES client object is what ultimately gets passed to the runner(s) invoked by that simulated client.

In lower-level terms, a worker is basically represented by an `AsyncIoAdapter` object, while a client basically maps to an `AsyncExecutor` object. An `AsyncIoAdapter` will now create an ES client per `client_id` that it's responsible for, and pass it to the `AsyncExecutor` that's instantiated for that `client_id`. The `AsyncExecutor` ultimately invokes the runner(s) that issue requests via the ES client.

This isn't the only possible implementation, though. As the [discussion linked above](https://github.com/elastic/rally/issues/1067#issuecomment-1130566298) details, version `8.0`+ of the ES Python client offers a new `.options()` method on clients for [overriding transport options](https://www.elastic.co/guide/en/elasticsearch/client/python-api/master/migration.html#migration-options) on a _per-request_ basis. It doesn't mutate the client, but rather returns a new client object with the provided transport options set.

We originally planned on using this new API, either by upgrading our client dependency (which we've paused for now) or by implementing `.options()` ourselves in Rally today. However, after getting into the details more, it seems that `.options()` isn't strictly necessary right now, and may actually make implementation more difficult and risky. Here's why:

  1. While the `.options()` method is meant for _request-level overrides_, creating clients with transport options defined in the first place is still the normal way of doing things. "Overriding default transport settings at request time" is one way to conceptualize what we're trying to do here, but it's still perfectly valid for workers to just instantiate ES clients upfront with the transport options they already know they'll need. Now, at a lower level, individual _runners_ may have a genuine need for overriding _those_ options (e.g. setting a `request_timeout` or `ignore_status`), but it's not necessary at the level of simulated clients in order to support something like client-specific authentication. The downside could be the overhead of creating potentially many more ES client objects, but...

  2. ...`.options()` will do that, anyway. As mentioned, `.options()` doesn't mutate the object it's called on, but rather [creates a new object of its type](https://github.com/elastic/elasticsearch-py/blob/v8.2.0/elasticsearch/_async/client/__init__.py#L500). And leads to the real trickiness, which is that Rally's async ES client objects aren't just vanilla instances of `AsyncElasticsearch`. They [also](https://github.com/elastic/rally/blob/80d4d8ff193878c5885801ab3362d0f4454f441a/esrally/client/asynchronous.py#L276) inherit from [`RequestContextHolder`](https://github.com/elastic/rally/blob/80d4d8ff193878c5885801ab3362d0f4454f441a/esrally/client/context.py#L59) and are therefore stateful. My initial experiments with porting `.options()` to Rally quickly resulted in benchmarks hanging forever. I have a sense of why this is happening, but in light of (1) above, it doesn't seem worth the time or risk to pursue, at least not right now. We will need to figure this out eventually for the client upgrade, however, since `.options()` is how individual runners will need to set options like `request_timeout`, `ignore_status`, as mentioned.

So, to sum up, implementing `.options()` ourselves doesn't seem to offer obvious advantages at the moment, and making it work would probably require refactoring the code that Rally relies on for measuring timings and scheduling task execution, which is risky to say the least.

Regarding the potential overhead of creating more clients, I'm curious to hear input on this from the likes of @DJRickyB and @dliappis. If we merge as-is, we'll quickly get some feedback from the nightlies, but I'd think that if this is going to cause problems, it will be with very high client counts. I ran a logging indexing benchmark with 1,000 bulk indexing clients with both `master` and this branch of Rally and found no evidence of this change causing a client-side bottleneck. It's just one data point, so I'm happy to flesh out a larger number of test cases if we feel that's in order, but wanted to get the team's thoughts before investing a bunch of time in that.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -1662,14 +1662,10 @@ class AsyncIoAdapter:
                 es[cluster_name] = client.EsClientFactory(cluster_hosts, all_client_options[cluster_name]).create_async()
             return es
 
-        # Properly size the internal connection pool to match the number of expected clients but allow the user
-        # to override it if needed.
-        client_count = len(self.task_allocations)
-        es = es_clients(self.cfg.opts("client", "hosts").all_hosts, self.cfg.opts("client", "options").with_max_connections(client_count))
-
         self.logger.info("Task assertions enabled: %s", str(self.assertions_enabled))
         runner.enable_assertions(self.assertions_enabled)
 
+        clients = []
         aws = []
         # A parameter source should only be created once per task - it is partitioned later on per client.
         params_per_task = {}
@@ -1679,6 +1675,8 @@ class AsyncIoAdapter:
                 param_source = track.operation_parameters(self.track, task)
                 params_per_task[task] = param_source
             schedule = schedule_for(task_allocation, params_per_task[task])
+            es = es_clients(self.cfg.opts("client", "hosts").all_hosts, self.cfg.opts("client", "options"))
+            clients.append(es)
             async_executor = AsyncExecutor(
                 client_id, task, schedule, es, self.sampler, self.cancel, self.complete, task.error_behavior(self.abort_on_error)
             )
@@ -1693,8 +1691,9 @@ class AsyncIoAdapter:
             await asyncio.get_event_loop().shutdown_asyncgens()
             shutdown_asyncgens_end = time.perf_counter()
             self.logger.info("Total time to shutdown asyncgens: %f seconds.", (shutdown_asyncgens_end - run_end))
-            for e in es.values():
-                await e.transport.close()
+            for c in clients:
+                for es in c.values():
+                    await es.close()
             transport_close_end = time.perf_counter()
             self.logger.info("Total time to close transports: %f seconds.", (shutdown_asyncgens_end - transport_close_end))
 


pquentin:
Thanks for the detailed explanation! I like the approach but would like to hear from others too.

michaelbaamonde:
We ran several tests with both this branch and `master` to determine whether this change introduces any signs of a client-side bottleneck at higher client counts. We did not find any evidence to that effect, but, unfortunately, we discovered that even `master` has issues scaling beyond ~2k clients on a single load driver due to https://github.com/elastic/rally/issues/1448 in some cases.

There’s no reason in principle to expect this change to significantly affect resource consumption on the load driver in terms of memory usage, file handles, etc. So while we don’t have corroboration at very high client counts, it seems that this change aligns with those expectations and doesn’t do worse than `master`.

So, we will merge this change and look into the scalability issues separately.

Issue #580, opened 2018-09-27
Add client-options to integration tests


danielmitterdorfer:
#578 shows that we should increase coverage of our integration test suite. At the moment we test without any custom track parameters but we should probably add support for at least HTTP compression and SSL.

karthikayan4u:
Hello @danielmitterdorfer , I wish to work this. Can you share more details as I am new to open source.

michaelbaamonde:
@karthikayan4u Apologies for the long delay in getting back to you. If you're still interested, adding an integration test that uses the `http_compress` client option would be great. It's [documented as an example ](https://esrally.readthedocs.io/en/stable/command_line_reference.html#client-options)client option. 

In terms of implementation, [this would be a good reference point](https://github.com/elastic/rally/blob/master/it/distribution_test.py#L32). I don't think it's necessary to test [all tracks](https://github.com/elastic/rally/blob/master/it/__init__.py#L34) with the option added, however. Even just one would be an improvement.

PR #1520, opened 2022-06-15
Create and use a unique ES API key for each simulated client

michaelbaamonde:
This PR introduces the `create_api_keys_per_client` client option. If `true`, the coordinating load driver will create a unique API key per logical client after the benchmark's allocation matrix is created, but before any task execution begins. For any given client, its generated API key will be used for authentication for all of the tasks assigned to it. Upon benchmark completion, the coordinating load driver will delete all API keys that it created initially.

Basic auth credentials are required to create API keys at the start of the benchmark and delete them at the end. We do intend to support using a "global API key" for these administrative operations (see https://github.com/elastic/rally/issues/1067#issuecomment-1127823875) but that will be a follow-up.

Here is an example CLI invocation that you can use to test:

```sh
esrally race --distribution-version=8.2.0 --car="defaults,trial-license,x-pack-security" --client-options="use_ssl:true,verify_certs:false,basic_auth_user:'rally',basic_auth_password:'rally-password',create_api_key_per_client:true" --track=geonames --test-mode
```

### DIFF
--- a/esrally/client/factory.py
+++ b/esrally/client/factory.py
@@ -308,8 +308,12 @@ def delete_api_keys(es, ids, max_attempts=5):
     """
     logger = logging.getLogger(__name__)
 
-    def raise_exception(failed_ids):
-        raise exceptions.RallyError(f"Could not delete API keys with the following IDs: {failed_ids}")
+    def raise_exception(failed_ids, cause=None):
+        msg = f"Could not delete API keys with the following IDs: {failed_ids}"
+        if cause is not None:
+            raise exceptions.RallyError(msg) from cause
+        else:
+            raise exceptions.RallyError(msg)
 
     # Before ES 7.10, deleting API keys by ID had to be done individually.
     # After ES 7.10, a list of API key IDs can be deleted in one request.
@@ -363,10 +367,10 @@ def delete_api_keys(es, ids, max_attempts=5):
                 logger.debug("Got status code [%s] on attempt [%s] of [%s]. Sleeping...", e.status_code, attempt, max_attempts)
                 time.sleep(1)
             else:
-                raise_exception(remaining)
+                raise_exception(remaining, cause=e)
         except Exception as e:
             if attempt < max_attempts:
                 logger.debug("Got error on attempt [%s] of [%s]. Sleeping...", attempt, max_attempts)
                 time.sleep(1)
             else:
-                raise_exception(remaining)
+                raise_exception(remaining, cause=e)


esrally/client/factory.py
```diff
@@ -238,3 +266,76 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+    """
+    logger = logging.getLogger(__name__)
+
+    for attempt in range(1, max_attempts + 1):
+        # pylint: disable=import-outside-toplevel
+        import elasticsearch
+
+        try:
+            api_key = es.security.create_api_key({"name": f"rally-client-{client_id}"})
+            return api_key
```
pquentin, line 288:
```suggestion
            return es.security.create_api_key({"name": f"rally-client-{client_id}"})
```

esrally/client/factory.py
```diff
@@ -238,3 +266,76 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+
+    # Before ES 7.10, deleting API keys by ID had to be done individually.
+    # After ES 7.10, a list of API key IDs can be deleted in one request.
+    current_version = versions.Version.from_string(es.info()["version"]["number"])
+    minimum_version = versions.Version.from_string("7.10.0")
+
+    def legacy_delete(es, ids):
+        while ids:
+            es.security.invalidate_api_key({"id": ids[0]})
+            ids.pop(0)
```
pquentin, line 313: Why not a for loop?

```suggestion
        for id in ids:
            es.security.invalidate_api_key({"id": id})
```

esrally/client/factory.py
```diff
@@ -238,3 +266,76 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+
+    def delete(es, ids):
+        es.security.invalidate_api_key({"ids": ids})
+
+    if current_version >= minimum_version:
+        func = delete
+    else:
+        func = legacy_delete
```
pquentin, line 321: nit: the functions seem simple enough to be inlined here

esrally/driver/driver.py
```diff
@@ -539,6 +542,16 @@ def num_cores(cfg):
     return int(cfg.opts("system", "available.cores", mandatory=False, default_value=multiprocessing.cpu_count()))
 
 
+ApiKey = collections.namedtuple("ApiKey", ["id", "secret"])
```
pquentin, line 545: Should we use a dataclass for this too?

esrally/driver/driver.py
```diff
@@ -710,10 +736,21 @@ def start_benchmark(self):
...
+                    worker_client_contexts = {}
                     for client_id in clients:
                         client_allocations.add(client_id, self.allocations[client_id])
                         self.clients_per_worker[client_id] = worker_id
-                    self.target.start_worker(worker, worker_id, self.config, self.track, client_allocations)
+                        client_context = ClientContext(client_id=client_id, parent_worker_id=worker_id)
+
+                        if create_api_keys:
+                            resp = self.create_api_key(self.default_sync_es_client, client_id)
+                            client_context.api_key = ApiKey(resp["id"], resp["api_key"])
```
pquentin, line 747:
```suggestion
                            client_context.api_key = ApiKey(id=resp["id"], secret=resp["api_key"])
```

pquentin (changes requested):
This looks really great! I yet have to try it, but expect to only leave nits.

esrally/driver/driver.py
michaelbaamonde: I thought about that, but since API keys are immutable, I opted to use a namedtuple instead. Do you think there's a significant advantage to using a dataclass instead?

esrally/client/factory.py
michaelbaamonde: It's because I want to include any IDs that we were unable to delete in the exception's message. Popping IDs from the list after they're deleted means that it will only contain un-deleted API keys, making it easy to report them in case of failure. We could of course do this with a for loop and some additional bookkeeping for tracking what IDs haven't been deleted if it's clearer.

esrally/client/factory.py
```diff
@@ -238,3 +266,75 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+        try:
+            func(es, ids)
+            return True
+        except elasticsearch.TransportError as e:
+            if attempt < max_attempts:
+                logger.debug("Got status code [%s] on attempt [%s] of [%s]. Sleeping...", e.status_code, attempt, max_attempts)
+                time.sleep(1)
+            else:
+                raise exceptions.RallyError(f"Could not delete API keys with the following IDs: {ids}") from e
+        except BaseException as e:
```
pquentin, line 335: You want to catch `Exception` here,  not `BaseException`, otherwise you will eat Ctrl-C

esrally/driver/driver.py
pquentin: No significant advantage, I find named tuples a bit obscure and am not sure if [the performance issues were solved](https://lwn.net/Articles/731423/). Immutable is a nice advantage, though so let's keep things like this.

esrally/client/factory.py
pquentin: I see. OK I think that's fine, and inlining the functions will make this clearer.

esrally/client/factory.py
michaelbaamonde: Good catch!

esrally/client/factory.py
michaelbaamonde: Is https://github.com/elastic/rally/pull/1520/commits/5b86c7b43487a615deedd44e1874b7851ab6acd2 something like you had in mind?

esrally/driver/driver.py
michaelbaamonde: Done in https://github.com/elastic/rally/pull/1520/commits/d9d7ec5010ac540246360f5a053b9ae76c69a90e

esrally/client/factory.py
```diff
@@ -238,3 +266,74 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+                es.security.invalidate_api_key({"id": ids[0]})
+                ids.pop(0)
+
+    for attempt in range(max_attempts + 1):
+        # pylint: disable=import-outside-toplevel
+        import elasticsearch
+
+        try:
+            delete(es, ids)
+            return True
```
pquentin, line 327: Sorry I wasn't clear, here's what I meant by inlining:

```suggestion
    for attempt in range(max_attempts + 1):
        # pylint: disable=import-outside-toplevel
        import elasticsearch

        try:
            if current_version >= minimum_version:
                es.security.invalidate_api_key({"ids": ids})
            else:
                while ids:
                    es.security.invalidate_api_key({"id": ids[0]})
                    ids.pop(0)
            return True
```

Also I don't think this matter but I for what it's worth using `pop(0)` makes the while loop accidentally quadratic. You would have to use a dequeue or pop items from the end to makes this linear. https://docs.python.org/3/tutorial/datastructures.html#using-lists-as-stacks

michaelbaamonde: :facepalm: Ha, I don't know why my brain really wanted to give those functions names.

Re: pop, yeah it had crossed my mind. Probably doesn't matter, but can't hurt to prevent it anyway.

esrally/client/factory.py
```diff
@@ -238,3 +266,74 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+    Deletes the provided list of API key IDs.
+
+    :param es: Elasticsearch client to use for connecting.
+    :param ids: List of API key IDs to delete.
+    :param max_attempts: The maximum number of attempts to delete the API keys.
+    :return: True iff all provided key IDs were successfully deleted.
+    """
+    logger = logging.getLogger(__name__)
+
+    # Before ES 7.10, deleting API keys by ID had to be done individually.
```
michaelbaamonde, line 310: We could actually get around this by deleting all API keys associated with the `basic_auth_user` that was supplied as a client option. We'd call `invalidate_api_key` with a `username` instead of `ids`/`id`. The API for this has existed since the introduction of API keys in 7.0, so we'd need to version-specific logic.

However, Rally would delete all of the user's API keys, not just the API keys that Rally created. If there are pre-existing API keys, this could be a problem. Benchmarking a cluster with any kind of pre-existing state is generally a bad idea, but I wouldn't want Rally to do something like invalidate a bunch of API keys just because a user made a mistake. So, if we want to go this route, we could first double-check that there are no existing API keys and fail with a helpful message if any are found, in addition to being clear about this in the docs.

It's a little funny that this component of the PR is perhaps the trickiest bit, which got me thinking about potential simplifications. What do you think @pquentin?

pquentin: I think the current code is fine, I mostly commented on details. Rally has a tradition of not messing with the cluster it's running on (https://esrally.readthedocs.io/en/stable/faq.html#will-rally-destroy-my-existing-indices), so let's keep it that way.

esrally/client/factory.py
```diff
@@ -238,3 +266,74 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+    :return: A dict with at least the following keys: ``id``, ``name``, ``api_key``.
+    """
+    logger = logging.getLogger(__name__)
+
+    for attempt in range(1, max_attempts + 1):
+        # pylint: disable=import-outside-toplevel
+        import elasticsearch
+
+        try:
+            return es.security.create_api_key({"name": f"rally-client-{client_id}"})
```
pquentin, line 287: Should we log about the fact that we're creating API keys?

pquentin (commented):
I'd like to reiterate that I'm a big fan of the amount of care that went into this pull requests and its tests.

I have a final issue: I can't get Rally to show me the error message that you carefully crafted. If I leave out basic auth credentials or don't ask for x-pack then I only get a LaunchError in my console and an unprintable RallyError object in the logs. I should see the error messages instead. (This might be unrelated to this pull request. If yes, we can fix it in another one.)

esrally/client/factory.py
michaelbaamonde: Agreed, I prefer this approach, too.

michaelbaamonde:
> I can't get Rally to show me the error message that you carefully crafted. If I leave out basic auth credentials or don't ask for x-pack then I only get a LaunchError in my console and an unprintable RallyError object in the logs. I should see the error messages instead. (This might be unrelated to this pull request. If yes, we can fix it in another one.)

@pquentin I think we may not always fail particularly gracefully _in general_ if the combination of client options and cars provided are somehow invalid (implementing #580 would help uncover some of these scenarios). But in this PR's case, here's how I've forced the main failure modes that are specific to API keys that we're trying to catch:

**Basic auth missing**

  Invocation:

  ```sh
  esrally race --distribution-version=8.2.0 --car="defaults,trial-license,x-pack-security" --client-options="create_api_key_per_client:true" --track=geonames --test-mode
  ```

<details><summary>Output</summary>
<p>

```
    [INFO] Race id is [3a64c6f4-287c-4311-b249-e5d0462c46c4]
    [INFO] Preparing for race ...
    Basic auth credentials are required in order to create API keys.
    Missing basic auth client options are: ['basic_auth_user', 'basic_auth_password']
    Read the documentation at https://esrally.readthedocs.io/en/latest/command_line_reference.html#client-options
    [ERROR] Cannot race. Traceback (most recent call last):
      File "/home/baamonde/code/elastic/rally/esrally/actor.py", line 92, in guard
        return f(self, msg, sender)
      File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 272, in receiveMsg_PrepareBenchmark
        self.coordinator.prepare_benchmark(msg.track)
      File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 677, in prepare_benchmark
        es_clients = self.create_es_clients()
      File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 605, in create_es_clients
        es[cluster_name] = self.es_client_factory(cluster_hosts, cluster_client_options).create()
      File "/home/baamonde/code/elastic/rally/esrally/client/factory.py", line 123, in __init__
        raise exceptions.SystemSetupError(
    esrally.exceptions.SystemSetupError: You must provide the 'basic_auth_user' and
      'basic_auth_password' client options in addition to
      'create_api_key_per_client' in order to create client API keys.
```
</p>
</details>

**Basic auth incomplete (password missing)**

  Invocation:

  ```sh
  esrally race --distribution-version=8.2.0 --car="defaults,trial-license,x-pack-security" --client-options="create_api_key_per_client:true,basic_auth_user:rally" --track=geonames --test-mode
  ```

<details><summary>Output</summary>
<p>

```
[INFO] Race id is [b93464a5-105a-484a-acdb-05b527958cfe]
[INFO] Preparing for race ...
Basic auth credentials are required in order to create API keys.
Missing basic auth client options are: ['basic_auth_password']
Read the documentation at https://esrally.readthedocs.io/en/latest/command_line_reference.html#client-options
[ERROR] Cannot race. Traceback (most recent call last):
  File "/home/baamonde/code/elastic/rally/esrally/actor.py", line 92, in guard
    return f(self, msg, sender)
  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 272, in receiveMsg_PrepareBenchmark
    self.coordinator.prepare_benchmark(msg.track)
  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 677, in prepare_benchmark
    es_clients = self.create_es_clients()
  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 605, in create_es_clients
    es[cluster_name] = self.es_client_factory(cluster_hosts, cluster_client_options).create()
  File "/home/baamonde/code/elastic/rally/esrally/client/factory.py", line 123, in __init__
    raise exceptions.SystemSetupError(
esrally.exceptions.SystemSetupError: You must provide the 'basic_auth_user' and
  'basic_auth_password' client options in addition to
  'create_api_key_per_client' in order to create client API keys.
```
</p>
</details>

**Security not enabled**

  Invocation:

  ```sh
  esrally race --distribution-version=8.2.0 --client-options="create_api_key_per_client:true,basic_auth_user:'rally',basic_auth_password:'rally-password'" --track=geonames --test-mode --kill-running-processes
  ```

<details><summary>Output</summary>
<p>

```
[INFO] Race id is [e7cedccf-6508-4767-97a7-1e28c2d076f1]
[INFO] Preparing for race ...
[INFO] Racing on track [geonames], challenge [append-no-conflicts] and car ['defaults'] with version [8.2.0].

[ERROR] Cannot race. Traceback (most recent call last):
  File "/home/baamonde/code/elastic/rally/esrally/client/factory.py", line 287, in create_api_key
    return es.security.create_api_key({"name": f"rally-client-{client_id}"})
  File "/home/baamonde/code/elastic/rally/.venv/lib/python3.8/site-packages/elasticsearch/client/utils.py", line 168, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
  File "/home/baamonde/code/elastic/rally/.venv/lib/python3.8/site-packages/elasticsearch/client/security.py", line 117, in create_api_key
    return self.transport.perform_request(
  File "/home/baamonde/code/elastic/rally/.venv/lib/python3.8/site-packages/elasticsearch/transport.py", line 458, in perform_request
    raise e
  File "/home/baamonde/code/elastic/rally/.venv/lib/python3.8/site-packages/elasticsearch/transport.py", line 419, in perform_request
    status, headers_response, data = connection.perform_request(
  File "/home/baamonde/code/elastic/rally/.venv/lib/python3.8/site-packages/elasticsearch/connection/http_urllib3.py", line 277, in perform_request
    self._raise_error(response.status, raw_data)
  File "/home/baamonde/code/elastic/rally/.venv/lib/python3.8/site-packages/elasticsearch/connection/base.py", line 330, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(
elasticsearch.exceptions.TransportError: TransportError(405, 'Incorrect HTTP method for uri [/_security/api_key] and method [PUT], allowed: [POST]', 'Incorrect HTTP method for uri [/_security/api_key] and method [PUT], allowed: [POST]')

During handling of the above exception, another exception occurred:

  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 657, in create_api_key
    api_key = client.create_api_key(es, client_id)
  File "/home/baamonde/code/elastic/rally/esrally/client/factory.py", line 291, in create_api_key
    raise exceptions.SystemSetupError(
esrally.exceptions.SystemSetupError: Got status code 405 when attempting to create API keys. Is Elasticsearch Security enabled?

During handling of the above exception, another exception occurred:

  File "/home/baamonde/code/elastic/rally/esrally/actor.py", line 92, in guard
    return f(self, msg, sender)
  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 277, in receiveMsg_StartBenchmark
    self.coordinator.start_benchmark()
  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 750, in start_benchmark
    resp = self.create_api_key(self.default_sync_es_client, client_id)
  File "/home/baamonde/code/elastic/rally/esrally/driver/driver.py", line 664, in create_api_key
    raise exceptions.SystemSetupError(e.message)
esrally.exceptions.SystemSetupError: Got status code 405 when attempting to
create API keys. Is Elasticsearch Security enabled?
```
</p>
</details>

What did you try? If there's a scenario that's API-key specific that we can handle better, let's do it in this PR. If it's a more generic issue with invalid client options, a follow-up sounds good.

michaelbaamonde:

tests/client/factory_test.py
```diff
@@ -455,6 +486,128 @@ def test_ssl_error(self, es):
...
+        ids = ["foo", "bar", "baz"]
+        es.security.invalidate_api_key.side_effect = [
+            elasticsearch.TransportError(503, "Service Unavailable"),
+            elasticsearch.TransportError(401, "Unauthorized"),
+            Exception("Whoops!"),
+        ]
+
+        with pytest.raises(exceptions.RallyError, match=re.escape(f"Could not delete API keys with the following IDs: {ids}")):
+            client.delete_api_keys(es, ids)
+
```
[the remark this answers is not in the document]
michaelbaamonde: This actually raised a more substantial issue in my mind, which I've addressed in https://github.com/elastic/rally/pull/1520/commits/8eec4b167f934fff73b59e56c6ba9f234503b0f9:

It's possible for the 7.10.0+ version of the deletion code to fail silently if we rely just on exceptions to catch errors. This is because it's basically a bulk request, which means that the response can contain both successful and unsuccessful deletions but still have an HTTP 200 status code. That commit handles that scenario, modifies the logic for the "legacy" code, and refactors the relevant tests.

esrally/client/factory.py
```diff
@@ -238,3 +266,111 @@ def wait_for_rest_layer(es, max_attempts=40):
...
+                        raise_exception(remaining)
+            return True
+
+        except elasticsearch.TransportError as e:
+            if attempt < max_attempts:
+                logger.debug("Got status code [%s] on attempt [%s] of [%s]. Sleeping...", e.status_code, attempt, max_attempts)
+                time.sleep(1)
+            else:
+                raise_exception(remaining, cause=e)
+        except Exception as e:
```
pquentin, line 371: Should we only catch `RallyError` here? I think for something else like say a `KeyError` there's no point in retrying.

esrally/client/factory.py
pquentin: So this function is now super complex and hard to follow, which made me rethink this question. I still don't want to remove all keys and still want to take advantage of the bulk request, so I've tried to understand how we could improve the situation.

I think we are mixing two concerns here:

 * handle both the 7.10- and 7.10+ cases (ironically the latter became much more complex)
 * handle retries

I think we could have a `issue_delete_keys_call(ids, version)` function (with a better name?) that would encapsulate the versioning logic. And then we could handle retries the same way in both cases.

The retry itself is the most complex part. I'm not sure why we think sleeping will help? Not even sure why we think retries would help on deletion failure actually. I would expect that unless we hit a TransportError, then the issue is with the request, so retrying would be futile? It's slightly ridiculous we're calling `raise_exception()` four times.

Anyway, that's a lot of typing just to say that well, this works and is well tested, so let's ship it!

pquentin (approved):
This works great, thanks! Ship it with or without the change to the except clause. And ignore my other comment. :)

PR #1521, opened 2022-06-16
Retry incomplete HTTP downloads

### LINKED ISSUE #1504: Retry incomplete HTTP downloads
We have an internal track that downloads many files individually with `esrally.track.loader.Downloader.download` that eventually calls `esrally.utils.net.download` which uses urllib3 to download the data. This `download` function checks that we downloaded all the bytes as specified in the Content-Length header. If not, it simply fails:

```
esrally.exceptions.DataError: Download of [~/.rally/benchmarks/data/solutions/logs/system-syslog-logs/document-50.json.bz2] is corrupt. Downloaded [1866599] bytes but [26703576] bytes are expected. Please retry.     
```

In that instance, the data was downloaded from https://rally-tracks.elastic.co/observability/logging/system/infra-stats/system.syslog/raw/document-50.json.bz2. This is a proxy maintained by Elastic, and apparently sometimes it serves us incomplete results.

But why isn't urllib3 covering this for us? https://blog.petrzemek.net/2018/04/22/on-incomplete-http-reads-and-the-requests-library-in-python/ has all the details. The 3.0 branch of requests has died since then, but thankfully we use urllib3 directly, and this post taught me that there is an undocumented flag in urllib3 to cover our use case: https://github.com/urllib3/urllib3/pull/949. (It will become the default in urllib3 v2).

So it appears that simply setting `enforce_content_length` and remove the custom checks will fix our issue.

pquentin:
Currently when we did not download enough data, we just abort the
benchmark. Instead, let's ask urllib3 to retry it per our retry policy.

This does not change the existing Rally logic around Content-Length as
it's possible to specify the expected length and I don't want this fix
to change other behavior.

Closes #1504 (it was a good first issue but we hit this a lot recently)

### DIFF
--- a/esrally/utils/net.py
+++ b/esrally/utils/net.py
@@ -174,9 +174,9 @@ def download_from_bucket(blobstore, url, local_path, expected_size_in_bytes=None
 
 
 def download_http(url, local_path, expected_size_in_bytes=None, progress_indicator=None):
-    with __http().request("GET", url, preload_content=False, retries=10, timeout=urllib3.Timeout(connect=45, read=240)) as r, open(
-        local_path, "wb"
-    ) as out_file:
+    with __http().request(
+        "GET", url, preload_content=False, enforce_content_length=True, retries=10, timeout=urllib3.Timeout(connect=45, read=240)
+    ) as r, open(local_path, "wb") as out_file:
         if r.status > 299:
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM thanks!

PR #1522, opened 2022-06-21
Fix pyenv install on Apple Silicons Macs

pquentin:
This requires upgrading the patch versions of Python we use. I simply used the latest as listed here: https://www.python.org/downloads/

https://stackoverflow.com/questions/71577626/issues-installing-python-3-8-10-on-macos-12-3-monterey

I have not run any benchmarks given how unlikely a client-side performance regression is here. Instead, I intend to look at nightly benchmarks closely tomorrow if this gets merged.


.ci/variables.json
```diff
@@ -5,9 +5,9 @@
     "WHEEL": "0.37.0"
   },
   "python_versions": {
-    "PY38": "3.8.10",
-    "PY39": "3.9.2",
-    "PY310": "3.10.0",
-    "MIN_PY_VER": "3.8.10"
+    "PY38": "3.8.13",
+    "PY39": "3.9.13",
+    "PY310": "3.10.5",
```
dliappis, line 10: As of now, if one install's `pyenv` on an M1 using `brew` (which is the first recommendation in the [pyenv docs](https://github.com/pyenv/pyenv#homebrew-in-macos) on how to install `pyenv` on macOS), it pulls pyenv version `2.3.1` which doesn't know how to install Python `3.10.5`:

```

pyenv install --skip-existing 3.10.5
python-build: definition not found: 3.10.5

See all available versions with `pyenv install --list'.

If the version you need is missing, try upgrading pyenv:

  brew update && brew upgrade pyenv
make: *** [prereq] Error 2
```

```
bash-5.1$ pyenv --version
pyenv 2.3.1
bash-5.1$ pyenv install --list | grep 3\\\.10
  3.10.0
  3.10-dev
  3.10.1
  3.10.2
  3.10.3
  3.10.4
  miniconda-3.10.1
  miniconda3-3.10.1
```

Unfortunately lots of users will prefer to install pyenv with homebrew unless we modify our docs to point to installing pyenv using the [git checkout](https://github.com/pyenv/pyenv#basic-github-checkout) as e.g. done on Linux.

Any chance of just switching to `3.10.4` until we fix the issue in our docs?

dliappis (commented):
Left a comment

.ci/variables.json
pquentin: Thanks for spotting this! Thankfully this is the only version that is affected. I don't think we should update our docs, installing pyenv through homebrew is really convenient.

Downgrading to 3.10.4 is fine as the fix we're interested in was introduced in 3.10.3: https://stackoverflow.com/a/71779344/481584

Do we need someone with a M1 Mac to confirm?

dliappis (approved):
LGTM

.ci/variables.json
pquentin: Thank you!

PR #1524, opened 2022-06-22
create-notice: update links for repos that switched to main

pquentin:
While the old links still work, I think it makes sense to update to the new versions?


michaelbaamonde (approved):
LGTM. Thanks for all these cleanups!

PR #1525, opened 2022-06-22
docs/conf.py: rename master_doc to root_doc

pquentin:
This was renamed in the default configuration since Sphinx 4.0 (we
currently use 4.2.0).

It will also be one less thing that will come up when grepping for master.


michaelbaamonde (approved):
LGTM

PR #1526, opened 2022-06-22
Add runner for paging through composite aggregations

ywelsch:
Adds a "composite-agg" runner that allows paginating through composite aggregations. This allows benchmarking realistic use of composite aggs better (e.g. ML transforms) as users of composite aggs are typically paginating through all pages.

Relates https://github.com/elastic/elasticsearch/issues/70035


esrally/driver/runner.py
```diff
@@ -891,6 +910,114 @@ async def _search_after_query(es, params):
...
+                else:
+                    # body needs to be un-mutated for the next iteration (preferring to do this over a deepcopy at the start)
+                    body.pop("pit", None)
+                    composite_agg_body.pop("after", None)
+                    break
+
+            return results
+
+        def select_aggs(obj):
+            if isinstance(obj, dict):
```
michaelbaamonde, line 973: I think we can refactor the `isinstance` checks in this function (as well as in `paths_to_composite_agg`, `resolve_composite_agg`, and `tree_copy_composite_agg` below.)  As far as I can tell--and let me know if I'm missing something--everywhere `isinstance` is called in these functions, the variable whose type is being inspected can only ever be a `dict` or `None`. 

This is because `body` is what's being passed through the chain of function calls, and it can only ever be a `dict`. Along the way, we can also get intermediate `None` return values. I think that covers all cases, unless I'm misreading.

If so, this function could look something like:

```python
def select_aggs(obj):
    if obj is not None:
        return obj.get("aggs", None) or obj.get("aggregations", None)
    else:
        return None
```

esrally/driver/runner.py
```diff
@@ -891,6 +910,114 @@ async def _search_after_query(es, params):
...
+                if "aggs" in obj:
+                    return obj["aggs"]
+                if "aggregations" in obj:
+                    return obj["aggregations"]
+            return None
+
+        def paths_to_composite_agg(obj, parent_key_path):
+            aggs = select_aggs(obj)
+            paths = []
+            if isinstance(aggs, dict):
```
michaelbaamonde, line 983: Similar to above, I think we can refactor this to:

```python
def paths_to_composite_agg(obj, parent_key_path):
    aggs = select_aggs(obj)
    paths = []
    if aggs is not None:
        for key, subobj in aggs.items():
            if subobj.get("composite") is not None:
                paths = paths + [parent_key_path + [key]]
            else:
                paths = paths + paths_to_composite_agg(subobj, parent_key_path + [key])
    return paths
```

esrally/driver/runner.py
```diff
@@ -891,6 +910,114 @@ async def _search_after_query(es, params):
...
+                    subobj = aggs[key]
+                    if isinstance(subobj, dict):
+                        if "composite" in subobj.keys():
+                            if isinstance(subobj["composite"], dict):
+                                paths = paths + [parent_key_path + [key]]
+                    paths = paths + paths_to_composite_agg(subobj, parent_key_path + [key])
+            return paths
+
+        def resolve_composite_agg(obj, key_path):
+            if isinstance(obj, dict):
```
michaelbaamonde, line 994: A similar refactoring as the above two examples would work here, too.

esrally/driver/runner.py
```diff
@@ -891,6 +910,114 @@ async def _search_after_query(es, params):
...
+                    return obj["composite"]
+                else:
+                    aggs = select_aggs(obj)
+                    if isinstance(aggs, dict):
+                        if key_path[0] in aggs:
+                            return resolve_composite_agg(aggs[key_path[0]], key_path[1:])
+            raise Exception("Could not find composite agg - parser inconsistency")
+
+        def tree_copy_composite_agg(obj, key_path):
+            if isinstance(obj, dict):
```
michaelbaamonde, line 1005: And here, as well.

esrally/driver/runner.py
```diff
@@ -891,6 +910,114 @@ async def _search_after_query(es, params):
...
+            # explicitly convert to int to provoke an error otherwise
+            total_pages = sys.maxsize if params.get("pages", "all") == "all" else int(mandatory(params, "pages", self))
+            for page in range(1, total_pages + 1):
+                if pit_op:
+                    pit_id = CompositeContext.get(pit_op)
+                    body["pit"] = {"id": pit_id, "keep_alive": "1m"}
+
+                paths_to_composite = paths_to_composite_agg(body, [])
+                if not paths_to_composite or len(paths_to_composite) != 1:
+                    raise Exception("Unique path to composite agg required")
```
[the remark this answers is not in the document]
ywelsch: sure, fixed in 2f868ba

esrally/driver/runner.py
ywelsch: There's two types of functions here. The first ones are `paths_to_composite_agg` (and the `select_aggs` function that is being used within), which figures out whether the given user-defined request body contains a composite agg definition. The parsing here cannot assume for the request body to be well-formed, and is rather lenient in the parsing (i.e. double-checks that we have dicts in the right place) so that it can later throw a proper error stating that it couldn't find the composite agg. If we were to just assume that some of these definitions would be dicts, we would give a cryptic Python error message accessing a non-dict in the wrong way.

For the above I prefer the current parsing code.

There are two more functions, `resolve_composite_agg` and `tree_copy_composite_agg`, where we use as input whatever was derived via `paths_to_composite_agg`, and hence we can assume that the right structure and types are in place. I made the code in those functions [less defensive](https://en.wikipedia.org/wiki/Defensive_programming), as `paths_to_composite_agg` has already checked the right shape of dicts.

I hope this addresses your concerns.

michaelbaamonde: Sure, this makes sense. Thanks!

michaelbaamonde (approved):
LGTM. Thanks!

ywelsch:
Thanks @michaelbaamonde!

PR #1527, opened 2022-06-22
Make esrally.utils.git branch agnostic

pquentin:
This will force callers to be explicit with the branch they're using
which in turn will make the gradual transition to main easier.

### DIFF
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -62,31 +62,30 @@ def fetch(src, remote="origin"):
 
 
 @probed
-def checkout(src_dir, branch="master"):
+def checkout(src_dir, *, branch):
     if process.run_subprocess_with_logging("git -C {0} checkout {1}".format(io.escape_path(src_dir), branch)):
         raise exceptions.SupplyError("Could not checkout [%s]. Do you have uncommitted changes?" % branch)
 
 
 @probed
-def rebase(src_dir, remote="origin", branch="master"):
-    checkout(src_dir, branch)
+def rebase(src_dir, remote="origin", *, branch):
+    checkout(src_dir, branch=branch)
     if process.run_subprocess_with_logging("git -C {0} rebase {1}/{2}".format(io.escape_path(src_dir), remote, branch)):
         raise exceptions.SupplyError("Could not rebase on branch [%s]" % branch)
 
 
 @probed
-def pull(src_dir, remote="origin", branch="master"):
+def pull(src_dir, remote="origin", *, branch):
     fetch(src_dir, remote)
-    rebase(src_dir, remote, branch)
+    rebase(src_dir, remote, branch=branch)
 
 
 @probed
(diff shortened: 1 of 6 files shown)


esrally/utils/git.py
```diff
@@ -62,31 +62,30 @@ def fetch(src, remote="origin"):
...
-def checkout(src_dir, branch="master"):
+def checkout(src_dir, *, branch):
     if process.run_subprocess_with_logging("git -C {0} checkout {1}".format(io.escape_path(src_dir), branch)):
         raise exceptions.SupplyError("Could not checkout [%s]. Do you have uncommitted changes?" % branch)
 
 
 @probed
-def rebase(src_dir, remote="origin", branch="master"):
-    checkout(src_dir, branch)
+def rebase(src_dir, remote="origin", *, branch):
```
dliappis, line 71: Should we force the use of kwargs for `remote` as well (i.e. move `*` before `remote`?)

esrally/utils/git.py
```diff
@@ -62,31 +62,30 @@ def fetch(src, remote="origin"):
...
+    checkout(src_dir, branch=branch)
     if process.run_subprocess_with_logging("git -C {0} rebase {1}/{2}".format(io.escape_path(src_dir), remote, branch)):
         raise exceptions.SupplyError("Could not rebase on branch [%s]" % branch)
 
 
 @probed
-def pull(src_dir, remote="origin", branch="master"):
+def pull(src_dir, remote="origin", *, branch):
```
dliappis, line 78: similar question as above

esrally/utils/git.py
```diff
@@ -62,31 +62,30 @@ def fetch(src, remote="origin"):
...
     fetch(src_dir, remote)
-    rebase(src_dir, remote, branch)
+    rebase(src_dir, remote, branch=branch)
 
 
 @probed
-def pull_ts(src_dir, ts):
+def pull_ts(src_dir, ts, remote="origin", *, branch):
```
dliappis, line 84: similar question as above

dliappis (commented):
Thanks, this makes sense. I left one question.

esrally/utils/git.py
pquentin: I considered it but did not want to clutter the code unnecessarily or make this pull request bigger. Now that I reconsider it again I guess it would be closer to the way we use the git cli?

I don't have a strong opinion here, I can do it if you prefer

dliappis: > Now that I reconsider it again I guess it would be closer to the way we use the git cli?

Yeah that's how I am thinking as well.

dliappis: We discussed offline, @pquentin will tackle making `remote` a mandatory kw arg without a default value (as done in `branch` here) in a follow up PR.

dliappis (approved):
LGTM

PR #1529, opened 2022-06-22
Make git.clone remote kwarg-only

pquentin:
This is the discussed follow-up to https://github.com/elastic/rally/pull/1527.


dliappis (approved):
LGTM thank you. Now the code reads as if we were typing the git commands in CLI which I find a great readability/extensibility improvement.

PR #1541, opened 2022-07-07
Update opts.py to allow for 'None' timeout-value in client_options

MoBoo:
Rally sets a default timeout (60s) for every http api call to elasticsearch.
For long running queries this timeout introduces some problems, because rally will fail with a client timeout (which is only traceable via the logs). To work around this timeout issues, one must set the timeout to an arbitrary high value to account for any possible timeouts.

Rally actually supports the following three options for values set in the `--client-options` parameter: `int`, `float` or `boolean`. For the timeout value an additional `None` is required to allow not specifing a timeout at all.
This is due to the underlying `urllib3` which requires either an `int`, `float` or `None` value as a timeout as shown by the following error message: `.../urllib3/util/timeout.py": ValueError: Timeout value connect was None, but it must be an int, float or None.`
So far only `int` and `float` are usable from the commandline because `None`cannot be passed via the  `--client-options` parameter.

This pull request fixes this issue by added the needed functionality to the `opts.py` module. A `to_none()` function is added which parses the String `"None"` to the pythonic `None` and is called in the `kv_to_map()` function which parses the `--client-options` parameters.

Edit: signed contributor agreement

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -670,8 +670,9 @@ We support the following data types:
 * Strings: Have to be enclosed in single quotes. Example: ``ca_certs:'/path/to/CA_certs'``
 * Numbers: There is nothing special about numbers. Example: ``sniffer_timeout:60``
 * Booleans: Specify either ``true`` or ``false``. Example: ``use_ssl:true``
+* None: Specify ``None`` without quotes.``'None'`` will be treated as string. Example: ``timeout:None``
 
-Default value: ``timeout:60`` (applies any time ``timeout`` is not specified)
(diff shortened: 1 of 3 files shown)


pquentin (changes requested):
Thanks! I like the idea. Can you please sign the CLA, document this and add a test in `tests/utils/opts_test.py`?

pquentin (approved):
Thanks for iterating so quickly! This now looks good to me.

We don't merge on Fridays to avoid breaking our nightly benchmarks (https://elasticsearch-benchmarks.elastic.co/), so we'll merge on Monday.

b-deam:
Thanks again @MoBoo !

PR #1546, opened 2022-07-25
Run lint with pre-commit

pquentin:
pre-commit is faster because it only applies to modified files and runs in parallel. It's also more convenient because it runs on commits, which ensures you don't forget lint.


michaelbaamonde (approved):
This is cool. Worked well for me locally. Thanks!

PR #1551, opened 2022-07-28
Switch from setuptools to hatchling

pquentin:
Python packaging is evolving quite rapidly and there are now [multiple alternatives to setuptools](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Additionally:

 * setuptools does not support editable installs yet, which required us to add `--no-use-pep517` in various places
 * setuptools still requires a setup.py file (as you can see in https://github.com/elastic/rally/pull/1548) but we'd like to move to a fully declarative model
 * setuptools is slower than other backends and it produces confusing output
 * setuptools broke Rally a few times in the past, which we solved by pinning it

I hesitated between flit (minimal and simple, did not want to be the default recommendation) and hatch (more popular, doing more things, actively maintained). The main difference for our use case is how we select files that go in the distribution, and hatch is the first that I was able to get to work. All other parts are standardized, so switching is going to be easy until we start using more hatch features.

I tested:
 * installing Rally in development mode (`make install`), running it with ./rally
 * running tests, including integration tests
   * note tox now uses isolated builds, so the cleaning step is not needed anymore
 * running a fake release
 * building a Docker image

Before merging, we'll have to:
 * merge https://github.com/elastic/rally/pull/1548
 * wait until our internal tooling, including CI, is ready


michaelbaamonde:
@elasticmachine update branch

michaelbaamonde (approved):
LGTM. I tested locally and also with some internal tooling that installs Rally from source using this branch.

The merge conflict is minor BTW.

pquentin:
I merged from master and fixed the conflicts. I also reinstated `tox-env-clean` because it turns out it's still needed.

This pull request is now even smaller.

PR #1553, opened 2022-08-01
Run unit and integration tests in separate tox environments

michaelbaamonde:
This will facilitate parallelizing CI jobs in the future. At the moment, it's essentially a no-op; the various `it` targets in the `Makefile` will run all of the same tests as before.


dliappis (approved):
LGTM!

PR #1555, opened 2022-08-01
Add an integration test for static responses.

michaelbaamonde:
This adds an integration test for the static responses client option, using the example from [the docs](https://esrally.readthedocs.io/en/stable/command_line_reference.html#client-options).

Relates to https://github.com/elastic/rally/issues/580.


inqueue (approved):
LGTM

PR #1559, opened 2022-08-09
Install Rally in PEP 517 mode

pquentin:
Simply having a pyproject.toml file with a build-system entry as done in https://github.com/elastic/rally/pull/1548 means that `--no-use-pep517` is going to fail. This pull request integrates the relevant part of https://github.com/elastic/rally/pull/1551 to stop doing that.

We need to merge this before the nightly benchmarks run.


DJRickyB (approved):
LGTM. working in both an existing and fresh venv

dliappis (approved):
LGTM. Good to see that we don't rely on setuptools anymore.

pquentin:
> LGTM. Good to see that we don't rely on setuptools anymore.

We still rely on setuptools right now, but it's installed by pip itself and encoded in a single place (pyproject.toml).

dliappis:
> We still rely on setuptools right now, but it's installed by pip itself and encoded in a single place (pyproject.toml).

Ah yes thanks for the clarification, I forgot the previous PR https://github.com/elastic/rally/pull/1548/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R2

Issue #1562, opened 2022-08-10
Implement cross-cluster runners


b-deam:
Use cases involving multiple clusters like Cross Cluster Search (CCS) or Cross Cluster Replication (CCR) require various additional runners in order for tracks to properly set them up.

We should implement runners in Rally for;

1. [Configuring 'local' and 'remote' cluster (these terms are relative) seeds](https://www.elastic.co/guide/en/elasticsearch/reference/current/remote-clusters-connect.html)
2. [Add support for 'local' and 'remote' clusters to 'follow' indices](https://www.elastic.co/guide/en/elasticsearch/reference/current/xpack-ccr.html)

b-deam:
Upon review, multi-cluster benchmarks are almost always going to involve complex enough setups and architectures that it will be difficult to abstract cleanly via generalised multi-cluster runner(s). [Our docs suggest as much too](https://esrally.readthedocs.io/en/latest/command_line_reference.html#id1). 

Closing this for now.

PR #1563, opened 2022-08-15
Fix multi-cluster attribute check for wrapped runners

b-deam:
During testing for https://github.com/elastic/rally-tracks/pull/293 I discovered that that the `DeleteDataStream` runner was not multi-cluster aware, and that our previous attribute check would miss any wrapped (i.e. with `Retry()`) multi-cluster aware runner.

With this commit we make the DeleteDataStream runner multi-cluster aware
by default, and fix an existing bug where the multi-cluster attribute of
wrapped runners were not checked properly.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -121,22 +121,17 @@ def enable_assertions(enabled):
 
 
 def register_runner(operation_type, runner, **kwargs):
-    def is_multi_cluster(runner):
-        if getattr(runner, "multi_cluster", False):
-            return True
-        if hasattr(runner, "delegate"):
-            if getattr(runner.delegate, "multi_cluster", False):
-                return True
-
     logger = logging.getLogger(__name__)
     async_runner = kwargs.get("async_runner", False)
     if isinstance(operation_type, track.OperationType):
         operation_type = operation_type.to_hyphenated_string()
+
     if not async_runner:
         raise exceptions.RallyAssertionError(
             "Runner [{}] must be implemented as async runner and registered with async_runner=True.".format(str(runner))
         )
-    if is_multi_cluster(runner):
+
+    if getattr(runner, "multi_cluster", False):
         if "__aenter__" in dir(runner) and "__aexit__" in dir(runner):
             if logger.isEnabledFor(logging.DEBUG):
                 logger.debug("Registering runner object [%s] for [%s].", str(runner), str(operation_type))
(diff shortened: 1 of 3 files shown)


esrally/driver/runner.py
```diff
@@ -121,17 +121,22 @@ def enable_assertions(enabled):
 
 
 def register_runner(operation_type, runner, **kwargs):
+    def is_multi_cluster(runner):
+        if getattr(runner, "multi_cluster", False):
+            return True
+        if hasattr(runner, "delegate"):
+            if getattr(runner.delegate, "multi_cluster", False):
+                return True
```
pquentin, line 129:
```suggestion
        if hasattr(runner, "delegate"):
            return hasattr(runner.delegate, "multi_cluster")
        return False
```

esrally/driver/runner.py
```diff
@@ -121,17 +121,22 @@ def enable_assertions(enabled):
...
+
     logger = logging.getLogger(__name__)
     async_runner = kwargs.get("async_runner", False)
     if isinstance(operation_type, track.OperationType):
         operation_type = operation_type.to_hyphenated_string()
-
```
pquentin, line 128: The blank lines here help clarify that this is not an `elif`

pquentin (commented):
Thanks! I'm wondering about the general direction here. Looking at #488 it looks like multi-cluster support was meant for custom runners that wanted to do different operations for each cluster.

Here you want to do the same operation for all clusters, which begs the following questions:

 1. What if you don't actually want to delete data streams in all clusters?
 2. This does not generalize very well: are we going to add 100 lines of code every time we want an operation to work an all clusters?

It seems to me that we can address both those concerns by adding a new operation say `all-clusters` that will run a given operation on all clusters. Maybe I'm over thinking it: I'd like to know what @dliappis thinks, as he added #488.

(Also, if this was split in two commits or two PRs, we could have merged the first part! It's always suspicious when a PR or commit contains the word "and".)

b-deam:
> Thanks! I'm wondering about the general direction here. Looking at #488 it looks like multi-cluster support was meant for custom runners that wanted to do different operations for each cluster.
> 
> Here you want to do the same operation for all clusters, which begs the following questions:
> 
> 1. What if you don't actually want to delete data streams in all clusters?
> 2. This does not generalize very well: are we going to add 100 lines of code every time we want an operation to work an all clusters?
> 
> It seems to me that we can address both those concerns by adding a new operation say `all-clusters` that will run a given operation on all clusters. Maybe I'm over thinking it: I'd like to know what @dliappis thinks, as he added #488.

I agree with these comments, and therefore reverted the `DeleteDataStream` changes in https://github.com/elastic/rally/pull/1563/commits/bd11ad3390f2b98f0e457d0edc19e1ba424436e0 and updated https://github.com/elastic/rally/issues/1562 with more details on more general approach to handling multi-cluster support for runners in the future. For now (to unblock me) I'll add the required logic to the custom `FollowIndex` runner merged via https://github.com/elastic/rally-tracks/pull/293.

> (Also, if this was split in two commits or two PRs, we could have merged the first part! It's always suspicious when a PR or commit contains the word "and".)

Agreed and apologies, I kept the 'bug' fix in https://github.com/elastic/rally/pull/1563/commits/287fbc07f39a05fd608c4daa8dcaf93986949455. I've renamed the PR title to match.

esrally/driver/runner.py
```diff
@@ -130,8 +137,7 @@ def register_runner(operation_type, runner, **kwargs):
         raise exceptions.RallyAssertionError(
             "Runner [{}] must be implemented as async runner and registered with async_runner=True.".format(str(runner))
         )
-
```
pquentin, line 133: I would also keep the empty line here :)

esrally/driver/runner.py
```diff
@@ -121,6 +121,13 @@ def enable_assertions(enabled):
 
 
 def register_runner(operation_type, runner, **kwargs):
+    def is_multi_cluster(runner):
+        if getattr(runner, "multi_cluster", False):
```
pquentin, line 125: Should we use `hasattr` here too?

pquentin (approved):
Thanks! LGTM. I left two small comments, feel free to merge after addressing them.

pquentin (approved):
Thanks! LGTM.

PR #1564, opened 2022-08-22
Introduce rally-tracks compatibility testing

michaelbaamonde:
This PR adds the infrastructure necessary to run `rally-tracks` integration tests (see https://github.com/elastic/rally-tracks/pull/289) from within the Rally repository, both locally and--more importantly--as a PR check via CI.

We [modify and extend](https://github.com/elastic/rally/pull/1564/files#diff-64a87f6e18e0f986ce1182a302e810ba67da5432ded7a1e695bcbc54ed995646) the `pytest-rally` plugin so that its behavior and defaults make sense when run from within the Rally repo as opposed to a track repo. This enables Rally developers and CI jobs to test Rally changes against arbitrary revisions of local track repositories, using arbitrary versions of Elasticsearch. 

In the default case, it simply ensures that changes being made to Rally do not break the `master` branch of `rally-tracks`. It does this by executing whatever tests are contained in the `it` subdirectory of `~/.rally/benchmarks/tracks/default`. By default, it uses a build of the main branch of ES from source (i.e. `--revision=current` in Rally terms) as the benchmark candidate.

To run these tests, execute `pytest it/track_repo_compatibility/` from the root of the Rally repo after running `make install`, or run `make rally-tracks-compat` to run them within a `tox` environment.

**Default behavior**

Here's an example of what happens by default, but we'll limit the example to just one test (for brevity of output) and add some extra logging to more clearly see what the `pytest-rally` plugin is doing:

`pytest it/track_repo_compatibility --log-cli-level=INFO -k metricbeat`

```
===================================================================================== test session starts ======================================================================================
platform linux -- Python 3.8.13, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /home/baamonde/code/elastic/rally/.venv/bin/python3
cachedir: .pytest_cache
benchmark: 3.2.2 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rally: track-repository=/home/baamonde/.rally/benchmarks/tracks/default, track-revision=master
rootdir: /home/baamonde/code/elastic/rally, configfile: pyproject.toml
plugins: benchmark-3.2.2, rally-0.0.1, asyncio-0.18.1, anyio-3.6.1, httpserver-1.0.4
asyncio: mode=strict
collecting ...
------------------------------------------------------------------------------------- live log collection --------------------------------------------------------------------------------------
INFO     pytest_rally.rally:rally.py:110 Running command: [esrally list tracks --track-repository="/home/baamonde/.rally/benchmarks/tracks/default" --track-revision="master" --configuration-name="pytest"]
collected 68 items / 67 deselected / 1 selected

test_all_tracks_and_challenges.py::TestTrackRepository::test_autogenerated[metricbeat-append-no-conflicts]
---------------------------------------------------------------------------------------- live log setup ----------------------------------------------------------------------------------------
INFO     pytest_rally.elasticsearch:elasticsearch.py:84 Installing Elasticsearch: [esrally install --quiet --http-port=19200 --node=rally-node --master-nodes=rally-node --car=4gheap,trial-license,x-pack-ml --seed-hosts="127.0.0.1:19300" --revision=current]
INFO     pytest_rally.elasticsearch:elasticsearch.py:93 Starting Elasticsearch: [esrally start --runtime-jdk=bundled --installation-id=a14708f6-d2c0-49e1-aaa2-dd60a2acaf9d --race-id=b50f7204-3f70-4724-bb91-3e0050d0e2e0]
---------------------------------------------------------------------------------------- live log call -----------------------------------------------------------------------------------------
INFO     pytest_rally.rally:rally.py:144 Running command: [esrally race --track="metricbeat" --challenge="append-no-conflicts" --track-repository="/home/baamonde/.rally/benchmarks/tracks/default" --track-revision="master" --configuration-name="pytest" --enable-assertions --kill-running-processes --on-error="abort" --pipeline="benchmark-only" --target-hosts="127.0.0.1:19200" --test-mode]
PASSED                                                                                                                                                                                   [100%]
-------------------------------------------------------------------------------------- live log teardown ---------------------------------------------------------------------------------------
INFO     pytest_rally.rally:rally.py:91 Removing Rally config from [/home/baamonde/.rally/rally-pytest.ini]
INFO     pytest_rally.elasticsearch:elasticsearch.py:104 Stopping Elasticsearch: [esrally stop --installation-id=a14708f6-d2c0-49e1-aaa2-dd60a2acaf9d]


============================================================================== 1 passed, 67 deselected in 33.14s ===============================================================================

```

You can see that we've applied the defaults for the `--track-repository` and `--track-revision` options in the first few lines of the output:

`rally: track-repository=/home/baamonde/.rally/benchmarks/tracks/default, track-revision=master`

And that these make their way into the `esrally` commands that `pytest-rally` generates:

```
INFO     pytest_rally.rally:rally.py:144 Running command: [esrally race --track="metricbeat" --challenge="append-no-conflicts" --track-repository="/home/baamonde/.rally/benchmarks/tracks/default" --track-revision="master" --configuration-name="pytest" --enable-assertions --kill-running-processes --on-error="abort" --pipeline="benchmark-only" --target-hosts="127.0.0.1:19200" --test-mode]`
```

**Overriding defaults**

For local development, overriding these defaults could be useful. For example, if you're working on changes to *both* Rally and rally-tracks locally, and want to test with a released version of ES, you could do something like this:

```sh
pytest it/track_repo_compatibility/ \
    --log-cli-level=INFO \
    --track-repository=/home/baamonde/code/elastic/rally-tracks \
    --track-revision=ci \
    --distribution-version=8.3.2 \
    -k metricbeat
```

The pytest session header will reflect these changes:

```
rally: track-repository=/home/baamonde/code/elastic/rally-tracks, track-revision=ci
```

The plugin will pass along the `--distribution-version` to install the ES fixture:

```
INFO     pytest_rally.elasticsearch:elasticsearch.py:84 Installing Elasticsearch: [esrally install --quiet --http-port=19200 --node=rally-node --master-nodes=rally-node --car=4gheap,trial-license,x-pack-ml --seed-hosts="127.0.0.1:19300" --distribution-version=8.3.2]
```

And race commands will provide the correct `--track-repository` and `--track-revision`:

```
INFO     pytest_rally.rally:rally.py:144 Running command: [esrally race --track="metricbeat" --challenge="append-no-conflicts" --track-repository="/home/baamonde/code/elastic/rally-tracks" --track-revision="ci" --configuration-name="pytest" --enable-assertions --kill-running-processes --on-error="abort" --pipeline="benchmark-only" --target-hosts="127.0.0.1:19200" --test-mode]
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ install-user: venv-create
 install: install-user
 	# Also install development dependencies
 	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install -e .[develop]
-#	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install git+https://github.com/elastic/pytest-rally.git
+	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install git+https://github.com/elastic/pytest-rally.git
 
 
 clean: nondocs-clean docs-clean
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ install-user: venv-create
 install: install-user
 	# Also install development dependencies
 	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install -e .[develop]
-	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install git+https://github.com/elastic/pytest-rally.git
+#	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install git+https://github.com/elastic/pytest-rally.git
 
 
 clean: nondocs-clean docs-clean
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,9 @@ it39: check-venv python-caches-clean tox-env-clean
 it310: check-venv python-caches-clean tox-env-clean
 	. $(VENV_ACTIVATE_FILE); tox -e py310-it
 
+rally-tracks-compat: check-venv python-caches-clean tox-env-clean
+	. $(VENV_ACTIVATE_FILE); tox -e rally-tracks-compat
+
 check-all: lint test it
 
 benchmark: check-venv
--- a/Makefile
+++ b/Makefile
@@ -67,6 +67,8 @@ install-user: venv-create
 install: install-user
 	# Also install development dependencies
 	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install -e .[develop]
+	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install git+https://github.com/elastic/pytest-rally.git
+
 
 clean: nondocs-clean docs-clean
 
--- a/it/track_repo_compatibility/conftest.py
+++ b/it/track_repo_compatibility/conftest.py
@@ -16,6 +16,8 @@
 # under the License.
 
 import os
+import shlex
+import subprocess
 
 import pytest
 
@@ -24,7 +26,7 @@ RALLY_CONFIG_DIR = os.path.join(RALLY_HOME, ".rally")
 RALLY_TRACKS_DIR = os.path.join(RALLY_CONFIG_DIR, "benchmarks", "tracks", "default")
 
 
-@pytest.hookimpl
+@pytest.hookimpl(tryfirst=True)
 def pytest_addoption(parser):
     group = parser.getgroup("rally")
     group.addoption(
@@ -43,9 +45,19 @@ def pytest_addoption(parser):
     )
 
 
-@pytest.hookimpl
+@pytest.hookimpl(tryfirst=True)
 def pytest_cmdline_main(config):
     repo = config.option.track_repository
+    if not os.path.isdir(repo):
+        if repo == RALLY_TRACKS_DIR:
+            try:
+                # this will perform the initial clone of rally-tracks
+                subprocess.run(shlex.split("esrally list tracks"), capture_output=True, check=True)
+            except subprocess.CalledProcessError as e:
+                raise AssertionError(f"Unable to list tracks in {repo}") from e
+        else:
+            raise AssertionError(f"Directory {repo} does not exist.")
+
     test_dir = config.option.track_repo_test_dir
     config.args.append(os.path.join(repo, test_dir))
 
--- a/it/track_repo_compatibility/conftest.py
+++ b/it/track_repo_compatibility/conftest.py
@@ -23,8 +23,8 @@ import pytest
 
 RALLY_HOME = os.getenv("RALLY_HOME", os.path.expanduser("~"))
 RALLY_CONFIG_DIR = os.path.join(RALLY_HOME, ".rally")
-RALLY_TRACKS_DIR = os.path.join(RALLY_CONFIG_DIR, "benchmarks", "tracks", "default")
-
+TRACK_REPO_PATH = os.path.join(RALLY_CONFIG_DIR, "benchmarks", "tracks", "rally-tracks-compat")
+REMOTE_TRACK_REPO = "https://github.com/elastic/rally-tracks"
 
 @pytest.hookimpl(tryfirst=True)
 def pytest_addoption(parser):
@@ -32,8 +32,8 @@ def pytest_addoption(parser):
     group.addoption(
         "--track-repository",
         action="store",
-        default=RALLY_TRACKS_DIR,
-        help=("Path to a local track repository\n" f"(default: {RALLY_TRACKS_DIR})"),
+        default=TRACK_REPO_PATH,
+        help=("Path to a local track repository\n" f"(default: {TRACK_REPO_PATH})"),
     )
     group.addoption("--track-revision", action="store", default="master", help=("Track repository revision to test\n" "default: `master`"))
     group.addoption(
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -67,6 +67,8 @@ install-user: venv-create
 install: install-user
 	# Also install development dependencies
 	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install -e .[develop]
+	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install git+https://github.com/elastic/pytest-rally.git
```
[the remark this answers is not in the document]
pquentin: I think it's a bug in pip. In any case, I think we should upload `pytest-rally` on PyPI since it's not expected to change much and the resulting experience would be nicer.

it/track_repo_compatibility/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+import subprocess
+
+import pytest
+
+RALLY_HOME = os.getenv("RALLY_HOME", os.path.expanduser("~"))
+RALLY_CONFIG_DIR = os.path.join(RALLY_HOME, ".rally")
+RALLY_TRACKS_DIR = os.path.join(RALLY_CONFIG_DIR, "benchmarks", "tracks", "default")
+
+
+@pytest.hookimpl(tryfirst=True)
```
pquentin, line 29: Do we need those `hookimp` decorators? Looking at examples of `pytest_addoption`, `pytest_cmdline_main` and `pytest_report_header`, I do not see that used.

it/track_repo_compatibility/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+def pytest_addoption(parser):
+    group = parser.getgroup("rally")
+    group.addoption(
+        "--track-repository",
```
pquentin, line 33: I don't understand how `--track-repository` affects `esrally list tracks`?

it/track_repo_compatibility/conftest.py
```diff
@@ -0,0 +1,67 @@
...
+
+import os
+import shlex
+import subprocess
+
+import pytest
+
+RALLY_HOME = os.getenv("RALLY_HOME", os.path.expanduser("~"))
+RALLY_CONFIG_DIR = os.path.join(RALLY_HOME, ".rally")
+RALLY_TRACKS_DIR = os.path.join(RALLY_CONFIG_DIR, "benchmarks", "tracks", "default")
```
pquentin, line 26: Should we consider another location in other to not affect and be affected by changes and checkouts in the default tracks git repository?

michaelbaamonde: I was originally intending to have these tests mimic a clean installation of Rally. By default, if a user installs Rally, it will clone `rally-tracks` into `~/.rally/benchmarks/tracks/default` the first time an `esrally` command is run that requires a track repository to be available on disk.

I think this makes sense for CI, but I can see the argument for instead explicitly cloning into a distinct, non-default location for these tests, especially locally.

[This commit ](https://github.com/elastic/rally/pull/1564/commits/257794baaf8cb4228196cc63c8ff7172bf2da86e)implements the following control flow:

- If `--track-repository` is provided directly, we make sure that the directory actually exists on disk, raising an exception if it doesn't.
- If `--track-repository` isn't provided, we default to `~/.rally/benchmarks/tracks/rally-tracks-compat`, cloning `https://github.com/elastic/rally-tracks` there if the directory doesn't yet exist.

Re: your suggestions [here](https://github.com/elastic/rally/pull/1564#discussion_r953452173), shelling out to `git` does seem easier for cloning, so I went with that.

I considered making other things configurable (the remote repo URL, an option to perform an initial clone into a custom track repository path, etc.) but I think this approach serves CI and local development well enough in their default cases.

What do you think?

Makefile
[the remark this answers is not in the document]
michaelbaamonde: Yeah, I looked into it a bit and it does see to be a bug in `pip`. Definitely agree on uploading the plugin to PyPI once it's stable.

michaelbaamonde:
Thanks for the review @pquentin! I believe I've addressed your comments, so this is ready for another look whenever you're free.

pquentin (approved):
Thanks, this looks great now! Haven't tested the CI configuration though.

michaelbaamonde:
> Haven't tested the CI configuration though

It's a little cumbersome (especially for jobs triggered by PRs) unless you're conversant in the local Jenkins workflow we use internally, which maybe you are already. We can sync on this if you'd like.

it/track_repo_compatibility/conftest.py
DJRickyB: On one hand, I don't want to conflict with people's local configurations. On the other hand, I don't want additional checkouts present on my local in order to isolate testing properly.

I propose creating a `rally-compat.ini` which has a comment at the header `# Automatically managed - do not touch` and running compatibility tests via `--configuration-name compat`. This would specify the canonical upstream `default.url = https://github.com/elastic/rally-tracks` (as well as our `teams` and `source` URLs) in a place that doesn't need to be touched by users and would better avoid users shooting themselves in the foot. Is this feasible?

DJRickyB: Mike and I discussed this offline. This is not a huge concern compared to the trappy nature of trying to isolate it perfectly. I agree with him that we could just move forward with this approach as-is

DJRickyB (approved):
LGTM

it/track_repo_compatibility/conftest.py
pquentin: Yeah I also think the rally-tracks-compat checkout is the best compromise here. Thanks!

Issue #1484, opened 2022-05-09
use_ssl: True broken on Python 3.10 and later


pquentin:
@lhuet reported us an issue where the following command:

```
esrally race --track http_logs --pipeline=benchmark-only --target-hosts="<snip>.elastic-cloud.com:9243" \
   --client-options="timeout:60,use_ssl:true,verify_certs:false,basic_auth_user:'elastic',basic_auth_password:'mypassword'"
```

failed like this:

```
elasticsearch.exceptions.SSLError: ConnectionError(Cannot create a client socket with a PROTOCOL_TLS_SERVER context
(_ssl.c:801)) caused by: SSLError(Cannot create a client socket with a PROTOCOL_TLS_SERVER context (_ssl.c:801))
```

The issue is in esrally/client:py:

https://github.com/elastic/rally/blob/609a043223d0fb7e180f40def4d568064078e672/esrally/client.py#L138-L142

Even though literally everyone thinks the comment is true at first, it's wrong. For a client socket, `ssl.Purpose.SERVER_AUTH` should be used, with or without client certs. Everyone agrees that `ssl.Purpose` is super confusing, see https://github.com/python/cpython/issues/73996. Anyway, in 3.10 making this mistake isn't possible anymore: https://github.com/python/cpython/pull/26646.

However, just using `SERVER_AUTH` is not enough, because it activates more checks. When fixing this we need to make sure that:

 * SSLContext attributes are changed in the correct order (eg.disabling verification is only possibly if hostname checking is disabled)
 * client certs still work
 * SSL connections to IP address still work (we will probably need https://github.com/elastic/elastic-transport-python/pull/81)
 * bonus points for catching all of this in tests (https://gist.github.com/dliappis/97d013ab6e0e3250fdd6cbfba8e58a5c is great - can it be part of IT tests? Or maybe we can use trustme + a simple http server)

PR #1565, opened 2022-08-24
Fix `make lint` on Python 3.10

pquentin:
In practice this means upgrading pylint which was failing on 3.9 and 3.10.

I would like to use Python 3.10 as my daily driver to avoid discovering issues like https://github.com/elastic/rally/issues/1484 only through our users, but I could not run `make lint` on Python 3.10 which made this impractical.


michaelbaamonde:
The `precommit` [check failed](https://elasticsearch-ci.elastic.co/job/elastic+rally+pull-request+precommit/13/console) because the similarities checker found several violations of [`duplicate-code`](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/duplicate-code.html).

Do you want to disable this for now?

pquentin:
For now I'm more concerned by the fact that I'm not seeing those errors locally! Do you see them when running `make lint`?

dliappis:
> For now I'm more concerned by the fact that I'm not seeing those errors locally! Do you see them when running `make lint`?

I also don't see the above mentioned errors when I run this PR locally with a clean venv using `make lint` or `make precommit`.
I am using `Python 3.8.13`.

Kind of unrelated, `make install` reports:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
hatchling 1.6.0 requires pluggy>=1.0.0, but you have pluggy 0.13.1 which is incompatible.
```

pquentin:
I ended up just disabling that check like I've done in esbench as it's of low value anyway.

pquentin:
> Kind of unrelated, `make install` reports:
> 
> ```
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> hatchling 1.6.0 requires pluggy>=1.0.0, but you have pluggy 0.13.1 which is incompatible.
> ```

Thanks! I opened https://github.com/elastic/rally/pull/1567 to address this.

michaelbaamonde (approved):
LGTM

PR #1567, opened 2022-08-26
Bump develop dependencies

pquentin:
This will fix the pluggy incompatibility between hatchling and pytest.

As noticed here: https://github.com/elastic/rally/pull/1565#issuecomment-1228286217

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -41,7 +41,7 @@ templates_path = ["_templates"]
 # source_suffix = ['.rst', '.md']
 source_suffix = ".rst"
 root_doc = "index"
-language = None
+language = "en"
 
 CI_VARS = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..", ".ci", "variables.json")
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -90,19 +90,19 @@ develop = [
     "boto3==1.18.46",
     # tests
     "ujson",
-    "pytest==6.2.5",
-    "pytest-benchmark==3.2.2",
-    "pytest-asyncio==0.18.1",
-    "tox==3.14.0",
-    "sphinx==4.2.0",
+    "pytest==7.1.2",
+    "pytest-benchmark==3.4.1",
+    "pytest-asyncio==0.19.0",
+    "pytest-httpserver==1.0.5",
+    "tox==3.25.0",
+    "sphinx==5.1.1",
     "furo==2022.06.21",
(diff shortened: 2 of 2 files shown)


docs/conf.py
```diff
@@ -41,7 +41,7 @@
 # source_suffix = ['.rst', '.md']
 source_suffix = ".rst"
 root_doc = "index"
-language = None
+language = "en"
```
pquentin, line 44: This is required by Sphinx 5

pyproject.toml
```diff
@@ -90,19 +90,19 @@ develop = [
...
-    "sphinx==4.2.0",
+    "pytest==7.1.2",
+    "pytest-benchmark==3.4.1",
+    "pytest-asyncio==0.19.0",
+    "pytest-httpserver==1.0.5",
+    "tox==3.25.0",
+    "sphinx==5.1.1",
     "furo==2022.06.21",
-    "twine==1.15.0",
+    "twine==4.0.1",
```
pquentin, line 100: No breaking changes between those two versions

pyproject.toml
```diff
@@ -90,19 +90,19 @@ develop = [
...
     "wheel==0.37.1",
-    "github3.py==1.3.0",
+    "github3.py==3.2.0",
```
pquentin, line 102: This is only used by `changelog.py` and it continues to work.

pyproject.toml
```diff
@@ -90,19 +90,19 @@ develop = [
     "boto3==1.18.46",
     # tests
     "ujson",
-    "pytest==6.2.5",
-    "pytest-benchmark==3.2.2",
-    "pytest-asyncio==0.18.1",
-    "tox==3.14.0",
-    "sphinx==4.2.0",
+    "pytest==7.1.2",
+    "pytest-benchmark==3.4.1",
```
pquentin, line 94: `make benchmark` continues to work

pquentin (commented):
I've left comments about places that are not exercised by CI and explain the docs/conf.py change.

pyproject.toml
pquentin: However it notices a bug in benchmarks/driver/runner_test.py where we call bulk_index but this is an async function and it's never awaited so the benchmark is bogus. pytest-benchmark does not support asyncio though: https://github.com/ionelmc/pytest-benchmark/issues/66

michaelbaamonde (approved):
LGTM. Thanks!

PR #1572, opened 2022-09-05
[docs] Mention that create-snapshot is an administrative operation

dliappis:
Docs are currently missing that create-snapshot is an administrative operation.


pquentin (changes requested):
Thanks, great catch! I checked all snapshot operations, and `restore-snapshot` is also an administrative operation that is not reported as such. Can you please include it too?

dliappis:
> Thanks, great catch! I checked all snapshot operations, and `restore-snapshot` is also an administrative operation that is not reported as such. Can you please include it too?

Thanks for checking. Added it in b13679b02985fafb0c92b9cadeed402a759df7ad and adjusted PR description.

pquentin (approved):
Thanks! LGTM.

PR #1574, opened 2022-09-07
Include a new downsampling operation

salvatore-campagna:
The new downsampling operation allows Rally to hit the Elasticsearch  `/<source-index>/_downsample/<target-index>` endpoint to start a downsampling operation on the source index. Expected parameters include:

* `fixed_interval`: the aggregation granularity using the same interval definition used by date histograms
* `source-index`: the index to downsample applying the aggregation
* `target-index`: the index created as a result of the aggregation on the timestamp field

Example: aggregate the index `test-source-index` on the timestamp field using a 1 minute interval and store the result in a new index called `test-target-index`.

```
{
  "name": "downsample",
  "operation": {
    "operation-type": "downsample",
    "fixed-interval": "1m",
    "source-index": "test-source-index",
    "target-index": "test-target-index"
  }
}
```

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3070,13 +3070,13 @@ Properties
 
 **Example**
 
-Run a downsampling operation aggregating data in the source index (test-source-index) and creating a new target index (test-target-index) applying an aggregation
+Executes a downsampling operation aggregating data in the source index (test-source-index) and creating a new target index (test-target-index) applying an aggregation
 interval of 1 minute on the @timestamp field::
 
     {
-      "name": "downsampling",
+      "name": "downsample",
       "operation": {
-      "operation-type": "downsampling",
+      "operation-type": "downsample",
       "body": {
         "fixed-interval": "1m",
         "source-index": "test-source-index",
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3065,10 +3065,9 @@ Executes a downsampling operation on an index producing a new index whose data i
 Properties
 """"""""""
 
-* ``fixed_interval``: The aggregation interval key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`.
-* ``source-index``: The index containing data to aggregate which includes a ``@timestamp`` field. Note that this index should be marked read-only prior to the execution of this operation.
-* ``target-index``: Tne new target index created by the downsampling operation and including aggregated data.
-
+* ``fixed_interval`` (optional, defaults to ``1h``): The aggregation interval key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`.
+* ``source-index`` (optional, defaults to ``tsdb-index``): The index containing data to aggregate which includes a ``@timestamp`` field. Note that this index should be marked read-only prior to the execution of this operation.
+* ``target-index`` (optional, defaults to ``{source-index}-{fixed-interval}``): Tne new target index created by the downsampling operation and including aggregated data.
 
 **Example**
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3056,6 +3056,41 @@ The following meta data is always returned:
 * ``weight``: The number of fetched pages. Should always equal to the ``pages`` parameter.
 * ``unit``: The unit in which to interpret ``weight``. Always "ops".
 
+downsampling
+~~~~~~~~~~~~~~~~~~~
+
+Executes a downsampling operation on an index producing a new index whose data is aggregated on the @timestamp field.
+
+
+Properties
+""""""""""
+
+* ``body`` (mandatory): The ``body`` property must contain the ``fixed_interval`` key defined as in 
+`https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`
+the ``source-index`` and the ``target-index``. The source index is an existing index which includes a ``@timestamp`` field. Write operations
+to the source index must be disabled. Tne target index is a new index that is created by the downsamplig operation and including aggregated data.
+
+**Example**
+
+Run a downsampling operation aggregating data in the source index (test-source-index) and creating a new target index (test-target-index) applying an aggregation
+interval of 1 minute on the @timestamp field.
+
+  {
+    "name": "downsampling",
+    "operation": {
+    "operation-type": "downsampling",
+    "body": {
+      "fixed-interval": "1m",
+      "source-index": "test-source-index",
+      "target-index": "tsdb-target-index"
+    }
+  }
+
+Meta-data
+"""""""""
+
+The operation returns no meta-data.
+
 field-caps
 ~~~~~~~~~~~~~~~~~~~
 
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3065,27 +3065,24 @@ Executes a downsampling operation on an index producing a new index whose data i
 Properties
 """"""""""
 
-* ``body`` (mandatory): The ``body`` property must contain the ``fixed_interval`` key defined as in 
-`https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`
-the ``source-index`` and the ``target-index``. The source index is an existing index which includes a ``@timestamp`` field. Write operations
-to the source index must be disabled. Tne target index is a new index that is created by the downsamplig operation and including aggregated data.
+* ``body`` (mandatory): The ``body`` property must contain the ``fixed_interval`` key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals` the ``source-index`` and the ``target-index``. The source index is an existing index which includes a ``@timestamp`` field. Write operations to the source index must be disabled. Tne target index is a new index that is created by the downsamplig operation and including aggregated data.
 
 
 **Example**
 
 Run a downsampling operation aggregating data in the source index (test-source-index) and creating a new target index (test-target-index) applying an aggregation
-interval of 1 minute on the @timestamp field.
+interval of 1 minute on the @timestamp field::
 
-{
-  "name": "downsampling",
-  "operation": {
-  "operation-type": "downsampling",
-  "body": {
-    "fixed-interval": "1m",
-    "source-index": "test-source-index",
-    "target-index": "tsdb-target-index"
-  }
-}
+    {
+      "name": "downsampling",
+      "operation": {
+      "operation-type": "downsampling",
+      "body": {
+        "fixed-interval": "1m",
+        "source-index": "test-source-index",
+        "target-index": "tsdb-target-index"
+      }
+    }
 
 Meta-data
 """""""""
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3056,7 +3056,7 @@ The following meta data is always returned:
 * ``weight``: The number of fetched pages. Should always equal to the ``pages`` parameter.
 * ``unit``: The unit in which to interpret ``weight``. Always "ops".
 
-downsampling
+downsample
 ~~~~~~~~~~~~~~~~~~~
 
 Executes a downsampling operation on an index producing a new index whose data is aggregated on the @timestamp field.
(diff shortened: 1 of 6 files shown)


esrally/driver/runner.py
```diff
@@ -2683,6 +2684,49 @@ def __repr__(self, *args, **kwargs):
...
+            )
+
+        target_index = body.get("target-index")
+        if target_index is None:
+            raise exceptions.DataError(
+                "Parameter source for operation 'downsampling' did not provide the mandatory parameter 'body.target-index'. "
+                "Add it to your parameter source and try again."
+            )
+
+        path = "/" + source_index + "/_rollup/" + target_index
```
salvatore-campagna, line 2718: I will need to change the endpoint to include `_downsample` instead of `_rollup` as soon as PR https://github.com/elastic/elasticsearch/pull/89809 is merged.

docs/track.rst
```diff
@@ -3056,6 +3056,39 @@ The following meta data is always returned:
 * ``weight``: The number of fetched pages. Should always equal to the ``pages`` parameter.
 * ``unit``: The unit in which to interpret ``weight``. Always "ops".
 
+downsampling
```
DJRickyB, line 3059: What about `downsample`? Similarly to how we call our query operation `search` as opposed to `searching`

docs/track.rst
```diff
@@ -3056,6 +3056,39 @@ The following meta data is always returned:
...
+~~~~~~~~~~~~~~~~~~~
+
+Executes a downsampling operation on an index producing a new index whose data is aggregated on the @timestamp field.
+
+
+Properties
+""""""""""
+
+* ``body`` (mandatory): The ``body`` property must contain the ``fixed_interval`` key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals` the ``source-index`` and the ``target-index``. The source index is an existing index which includes a ``@timestamp`` field. Write operations to the source index must be disabled. Tne target index is a new index that is created by the downsamplig operation and including aggregated data.
```
DJRickyB, line 3068: Most of the time we abstract common or critical `body` elements for the user to explicit parameters. If the `body` has required keys, it's likely better to just parameterize the operation instead of providing the body verbatim. This also gives the operation advantage over just using a `raw-request`.

esrally/driver/runner.py
```diff
@@ -60,6 +60,7 @@ def register_default_runners():
     register_runner(track.OperationType.ClosePointInTime, ClosePointInTime(), async_runner=True)
     register_runner(track.OperationType.Sql, Sql(), async_runner=True)
     register_runner(track.OperationType.FieldCaps, FieldCaps(), async_runner=True)
+    register_runner(track.OperationType.Downsampling, Downsampling(), async_runner=True)
```
DJRickyB, line 63: Should the operation be considered administrative? Maybe I misunderstand but this is not something we would want multiple users doing, or performed multiple times, correct?

esrally/driver/runner.py
DJRickyB: FYI we tend to prefer f-strings for this type of code
```suggestion
        path = f"/{source_index}/_rollup/{target_index}"
```

docs/track.rst
salvatore-campagna: I was thinking about that and at the end I chose to have the body because I had the feeling that having the body was "more consistent" with other operations. But I see your point here and I will get rid of it.

esrally/driver/runner.py
salvatore-campagna: This is something I am not sure. What is the criterion that we use to define an operation administrative? The expectation for this operation is that it is executed by means of an ILM action/policy (downsample the index when it reaches a certain size or after some amount of time). It might also be executed manually but I think those are more exceptional cases like for debugging purposes. So maybe these are good reasons to have the operation as an administrative operation?

Does it change something when the operation is executed?

salvatore-campagna: Note that this is an operation that we want to benchmark...so I guess if we make it administrative we don't want to retry.

docs/track.rst
```diff
@@ -3056,6 +3056,39 @@ The following meta data is always returned:
 * ``weight``: The number of fetched pages. Should always equal to the ``pages`` parameter.
 * ``unit``: The unit in which to interpret ``weight``. Always "ops".
 
+downsample
+~~~~~~~~~~~~~~~~~~~
```
DJRickyB, line 3060: nit
```suggestion
~~~~~~~~~~
```

docs/track.rst
```diff
@@ -3056,6 +3056,39 @@ The following meta data is always returned:
...
+
+Executes a downsampling operation on an index producing a new index whose data is aggregated on the @timestamp field.
+
+
+Properties
+""""""""""
+
+* ``fixed_interval``: The aggregation interval key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`.
+* ``source-index``: The read-only index containing data to aggregate which includes a ``@timestamp`` field.
+* ``target-index``: Tne new target index created by the downsamplig operation and including aggregated data.
```
DJRickyB, line 3070:
```suggestion
* ``target-index``: Tne new target index created by the downsampling operation and including aggregated data.
```

docs/track.rst
```diff
@@ -3056,6 +3056,39 @@ The following meta data is always returned:
...
+~~~~~~~~~~~~~~~~~~~
+
+Executes a downsampling operation on an index producing a new index whose data is aggregated on the @timestamp field.
+
+
+Properties
+""""""""""
+
+* ``fixed_interval``: The aggregation interval key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`.
+* ``source-index``: The read-only index containing data to aggregate which includes a ``@timestamp`` field.
```
DJRickyB, line 3069:
```suggestion
* ``source-index``: The index containing data to aggregate which includes a ``@timestamp`` field. Note that this index should be marked read-only prior to the execution of this operation.
```

salvatore-campagna:

salvatore-campagna:

esrally/driver/runner.py
```diff
@@ -1370,6 +1370,7 @@ async def __call__(self, es, params):
         try:
             for index_name in indices:
                 if not only_if_exists:
+                    request_params["ignore_unavailable"] = "true"
```
salvatore-campagna, line 1373: I added this line to the `DeleteIndex` operation to avoid a failure...

salvatore-campagna:
The CI is failing with
`XFAIL (index deletion bug under investigation) `

esrally/driver/runner.py
DJRickyB: We will prefer handling via #1580 as it doesn't change client/operation behavior

esrally/track/params.py
```diff
@@ -838,6 +838,19 @@ def params(self):
         return parsed_params
 
 
+class DownsampleParamSource(ParamSource):
+    def __init__(self, track, params, **kwargs):
+        super().__init__(track, params, **kwargs)
+        self._fixed_interval = params.get("fixed-interval", "1h")
+        self._source_index = params.get("source-index", "tsdb-index")
```
DJRickyB, line 845: Does a static default make sense here? We have a helper function for this purpose to extract a default target from whatever track we are running. Note: this would require the source parameter name be either `index` or `data-stream` (or support both, if it's supported by the API?)

```suggestion
        self._source_index = params.get_target(track, params)
```

salvatore-campagna: `source-index` and `target-index` is the same name we use in the docs about the downsampling API. I thought that keeping the naming consistent helps because they match and it is easier to understand what they refer to if you read the API docs.

DJRickyB: Got it. I'd still recommend this change, perhaps setting `params["index"] = params.get("source-index")` to use the function as-is, or replicating its functionality in-line to get a dynamic default

salvatore-campagna: The thing is...this API involves two indices...one that we read and one that we write. Having a parameter called `index`, in my opinion, does not make clear whether `index` is the source or the target.

DJRickyB: This would be transparent to the end user. just suggesting 
```suggestion
        # alias provided user param for get_target function
        params["index"] = params.get("source-index")
        self._source_index = params.get_target(track, params)
```

DJRickyB: or else, as I mentioned, just re-implement the logic for a default target from the function to here

salvatore-campagna: Moreover I see that `get_target` would support a parameter called `indices` which suggests this operation can be applied to multiple indices together. This is definitely not the case. A downsampling operation can only operate on a single source inde and on a single target index. Also data streams are not supported (unless at some point we decide some way to select the backing index in the data stream we would like to use as the source index).

DJRickyB: `indices` in this context is not a parameter but rather the track element, for example: https://github.com/elastic/rally-tracks/blob/master/tsdb/track.json#L21.

I think for some better information I'll reference the docs where this function is used elsewhere. From [Track Reference#search](https://esrally.readthedocs.io/en/stable/track.html#search):
```
index (optional): An [index pattern](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html) that defines which indices or data streams should be targeted by this query. Only needed if the indices or data-streams section contains more than one index or data stream respectively. Otherwise, Rally will automatically derive the index or data stream to use. If you have defined multiple indices or data streams and want to query all of them, just specify "index": "_all".
```

In this operation's case it would be more
```
source-index (optional): The index containing data to aggregate which includes a ``@timestamp`` field. Note that this index should be marked read-only prior to the execution of this operation. Only needed if the indices section contains more than one index. Otherwise, Rally will automatically derive the index or data stream to use.
```

esrally/track/params.py
```diff
@@ -838,6 +838,19 @@ def params(self):
         return parsed_params
 
 
+class DownsampleParamSource(ParamSource):
+    def __init__(self, track, params, **kwargs):
+        super().__init__(track, params, **kwargs)
+        self._fixed_interval = params.get("fixed-interval", "1h")
+        self._source_index = params.get_target(track, params)
```
salvatore-campagna, line 845: this should be `get_target` not `params.get_target`.

docs/track.rst
```diff
@@ -3056,6 +3056,39 @@ The following meta data is always returned:
...
+~~~~~~~~~~
+
+Executes a downsampling operation on an index producing a new index whose data is aggregated on the @timestamp field.
+
+
+Properties
+""""""""""
+
+* ``fixed_interval`` (optional, defaults to ``1h``): The aggregation interval key defined as in `https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#fixed_intervals`.
+* ``source-index`` (optional, defaults to ``tsdb-index``): The index containing data to aggregate which includes a ``@timestamp`` field. Note that this index should be marked read-only prior to the execution of this operation.
```
DJRickyB, line 3069:
```suggestion
``source-index`` (optional): The index containing data to aggregate which includes a ``@timestamp`` field. Note that this index should be marked read-only prior to the execution of this operation. If there is only one index defined in the ``indices`` of the track definition, that will be used as the default.
```

tests/track/params_test.py
```diff
@@ -2772,3 +2772,41 @@ def test_force_merge_all_params(self):
...
+            params={"source-index": "tsdb", "fixed-interval": "1m"},
+        )
+
+        p = source.params()
+
+        assert p["fixed-interval"] == "1m"
+        assert p["source-index"] == "tsdb"
+        assert p["target-index"] == "tsdb-1m"
+
+    def test_downsample_empty_params(self):
```
DJRickyB, line 2802: Not sure what this test buys us? Checked with a debugger and captured the value of p:
```python
{'fixed-interval': '1h', 'source-index': None, 'target-index': 'None-1h', 'request-timeout': None, 'headers': None, 'opaque-id': None}
```

Might be better instead to have the Track constructor call include an `indices` param with value like `[{"name": "test-source-index", "body": "index.json"}]` and assert that source-index gets provided as `test-source-index`

tests/track/params_test.py
```diff
@@ -2772,3 +2772,41 @@ def test_force_merge_all_params(self):
...
+        source = params.DownsampleParamSource(
+            track.Track(name="unit-test"),
+            params={},
+        )
+
+        p = source.params()
+
+        assert p["fixed-interval"] != ""
```
DJRickyB, line 2810: This should be the documented default, and should need changing if the default changes

DJRickyB (changes requested):
Still have 3 "unresolved conversations" on this PR, relating to documentation update and hardening testing the defaults

PR #1578, opened 2022-09-13
Correctly retry HTTP downloads after incomplete reads

pquentin:
The [previous change] only helped urllib3 *notice* incomplete reads, but it could not retry given this was a streaming download. This commits implements the actual retries.

[previous change]: https://github.com/elastic/rally/issues/1504

### DIFF
--- a/esrally/utils/net.py
+++ b/esrally/utils/net.py
@@ -218,10 +218,8 @@ def download_http(url, local_path, expected_size_in_bytes=None, progress_indicat
         except urllib3.exceptions.ProtocolError as exc:
             if i == HTTP_DOWNLOAD_RETRIES:
                 raise
-            if isinstance(exc.args[1], urllib3.exceptions.IncompleteRead):
-                logger.warning("Retrying after %s", exc)
-                continue
-            raise
+            logger.warning("Retrying after %s", exc)
+            continue
 
 
 def add_url_param_elastic_no_kpi(url):
--- a/esrally/utils/net.py
+++ b/esrally/utils/net.py
@@ -182,7 +182,10 @@ def download_from_bucket(blobstore, url, local_path, expected_size_in_bytes=None
     return expected_size_in_bytes
 
 
-def download_http(url, local_path, expected_size_in_bytes=None, progress_indicator=None):
+HTTP_DOWNLOAD_RETRIES = 10
+
+
+def _download_http(url, local_path, expected_size_in_bytes=None, progress_indicator=None):
     with _request(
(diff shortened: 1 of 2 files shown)


pquentin:
@elasticmachine run rally/rally-tracks-compat please

esrally/utils/net.py
```diff
@@ -207,6 +210,18 @@ def download_http(url, local_path, expected_size_in_bytes=None, progress_indicat
...
 
+def download_http(url, local_path, expected_size_in_bytes=None, progress_indicator=None):
+    logger = logging.getLogger(__name__)
+    for i in range(HTTP_DOWNLOAD_RETRIES + 1):
+        try:
+            return _download_http(url, local_path, expected_size_in_bytes, progress_indicator)
+        except urllib3.exceptions.ProtocolError as exc:
+            if i == HTTP_DOWNLOAD_RETRIES:
+                raise
+            logger.warning("Retrying after %s", exc)
```
dliappis, line 221: Should we have a delay in between retries? Alternatively we could consider using something like https://pypi.org/project/tenacity/?

dliappis (commented):
Looking forward to this stability improvement. Left a question.

esrally/utils/net.py
pquentin: I considered introducing a new dependency but thought this was not needed. I did not introduce a delay because I only saw connectivity issues (connection reset, incomplete reads) and not because the server was overloaded. Maybe a simple 1 second sleep could be enough?

dliappis: Agreed on a simple sleep; I'd go for `5s`, but up to you.

pquentin: Done, please take another look.

dliappis (approved):
LGTM

esrally/utils/net.py
```diff
@@ -207,6 +211,19 @@ def download_http(url, local_path, expected_size_in_bytes=None, progress_indicat
         return expected_size_in_bytes
 
 
+def download_http(url, local_path, expected_size_in_bytes=None, progress_indicator=None, *, sleep=time.sleep):
+    logger = logging.getLogger(__name__)
+    for i in range(HTTP_DOWNLOAD_RETRIES + 1):
```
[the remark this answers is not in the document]
dliappis: nvm I see why below. Ignore the comment.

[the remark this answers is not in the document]
pquentin: For reference, the reason is that `range(10)` would only do 9 retries (1 try + 9 retries).

PR #1579, opened 2022-09-13
Use narrower exception to avoid pylint ignore

pquentin:
getheader() cannot throw an exception and now always returns a string so the only possible exception is ValueError.

This is a small thing I noticed when working on #1578.


pquentin:

DJRickyB (approved):
LGTM, thank you

PR #1580, opened 2022-09-13
Remove buggy indices.exists() calls from code

DJRickyB:
Recently we've had a few errors surface in Rally executions when indices/data-streams are deleted. The errors would be reported as `connection timed out` for the `DELETE /<index>` call, but when inspected (via a packet capture) the server was shown to have responded correctly and in a timely fashion to Rally.

It turns out that in our version of `elasticsearch-py`, `indices.exists()` calls were handled via GET rather than HEAD. It's unclear to me why this causes an issue, but any of the following fixes the issue when targeting very large indices with the `delete-index` operation:
* Reduce the body size of the index (e.g. reduce the number of mappings)
* Set `only-if-exists` to `false` in the operation (to avoid the `exists()` call)
* Change the `exists()` call to a `get()` call.

I'm led to believe the bytes from the body are left in the event loop and not dequeued properly by the client, hence blocking somehow the response from the very next transaction.

This PR changes from using the `exists()` method to using the `get()` method to determine an index's existence.

A test is updated to show the difference in behavior under-the-hood.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1631,13 +1631,11 @@ Properties
 
 If you want it to delete all data streams that have been declared in the ``data-streams`` section, you can specify the following properties:
 
-* ``only-if-exists`` (optional, defaults to ``true``): Defines whether a data stream should only be deleted if it exists.
 * ``request-params`` (optional): A structure containing any request parameters that are allowed by the delete index API. Rally will not attempt to serialize the parameters and pass them as is. Always use "true" / "false" strings for boolean parameters (see example below).
 
 If you want it to delete one specific data stream (pattern) instead, you can specify the following properties:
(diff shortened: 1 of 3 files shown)


nik9000 (commented):
It's ok by me! What a wild ride. Probably worth a comment about why using `get` instead of `exists` and maybe even an assertion that you are on the un-fixed version of the client. Once you upgrade I think you want to revert this commit, right?

DJRickyB:
@elasticmachine  run rally/rally-tracks-compat

michaelbaamonde (approved):
LGTM. Thanks so much for chasing this down.

Another point in favor of prioritizing the client upgrade.

Issue #1582, opened 2022-09-15
Document valid conditions for making assertions


ywelsch:
The documentation (see https://esrally.readthedocs.io/en/stable/recipes.html?highlight=debug#checking-queries-and-responses) does not describe what fields are available and what conditions are supported in assertions on responses.

For example, I tried using assertions on a raw-request, which was just silently ignored by the system.

Further, it is unclear what fields are available and what comparison operators you can use.

ywelsch:
Additionally, rally currently silently ignores assertions (just logging a DEBUG message), maybe this should be treated as a hard failure (see https://github.com/elastic/rally/blob/067b27cbc9d807d4c37ec17fc066c461d8fba6d9/esrally/driver/runner.py#L399)

PR #1584, opened 2022-09-22
Add support for assertions on raw requests

ywelsch:
Relates #1582


pquentin (approved):
Thanks, this looks good to me! I dislike boolean arguments to change the behavior of functions, but modifying props to become nested sounded harder.

I did change the parameter to be keyword only and renamed it, I think this makes the `check_assertion()` calls clearer.

ywelsch:
Thank you, @pquentin!

PR #1586, opened 2022-09-27
Fix supplier tests for aarch64

b-deam:
Because `TemplateRenderer.arch` defaults to the arch of the system
invoking the tests, when invoked on an aarch64 system, the various 
`Car` instances would throw exceptions [^1] because they were 
missing the architecture specific build commands added in 
https://github.com/elastic/rally-teams/pull/75

This commit fixes the existing tests to default to `x86_64`,
and adds two new `aarch64` tests.

[^1]:  (e.g. `esrally.exceptions.SystemSetupError: Car "default" requires config key "system.build_command.arch"`)


pquentin (approved):
Thanks, this looks good to me!

I find it sad that Elasticsearch forces us to have so many conditions just because x64 is the default arch and is left out from the build command and generated tarball. Well, that's the way it is.

PR #1587, opened 2022-09-27
Refactor `pull_revision` to `pull`

b-deam:
The method name `pull_revision` is not consistent with its behaviour, it
only fetches a specific revision from the remote source, and does not 
'pull' it into the local checkout.

This commit makes the behaviour of the method consistent with its name,
by removing the `pull_revision` method, and instead replacing with with
a call to `pull`.


pquentin (commented):
So pull_revision used to do git fetch + git checkout. This is what pull is supposed to do, right?

With the new revision, you're adding a rebase, which is equivalent to git pull --rebase instead, which is not standard git pull.

Did you have any issues with the existing code?

pquentin (changes requested):
We chatted offline, and the difference here is that git pull does a merge by default, which we're not doing here. It has not made a difference until now because `pull_revision` is only expecting a commit hash, not a branch. But we would like to extend Rally to support that, and indeed adding a rebase or a merge adds support for branches without breaking the commit hash use case. What needs to be done now:

 * document that --revision supports branches now
 * fix the test_pull_ts and test_pull_revision tests that use `has_calls` instead of `assert_has_calls` which means they were not asserting anything and are broken
 * decide between merge or rebase

pquentin (approved):
Thanks for iterating! LGTM.

PR #1588, opened 2022-09-27
Fail benchmark if assertion could not run

pquentin:
When you 1/ enable assertions and 2/ have assertions in your track but 3/ the operation is returning an unsupported type, we used to only issue a debug log! I believe that we should instead complain loudly. This is technically breaking but will only report valid issues so I think it makes sense to include in the next minor version of Rally.

This was initially reported by @ywelsch in https://github.com/elastic/rally/issues/1582#issuecomment-1247798560.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -404,7 +404,7 @@ class AssertingRunner(Runner, Delegator):
                     props = parse(return_value, assertion["property"])
                     self.check_assertion(op_name, assertion, props, props_expand_keys=False)
             else:
-                self.logger.debug("Skipping assertion check in [%s] as [%s] does not return a dict.", op_name, repr(self.delegate))
+                raise exceptions.DataError(f"Cannot check assertion in [{op_name}] as [{repr(self.delegate)}] does not return a dict.")
         return return_value
 
     def __repr__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


ywelsch (approved):
Thanks for picking this up.

DJRickyB (approved):
LGTM. This feels like it's not just sensible but correct, and I'm glad you've addressed it; agree it should go in the next minor

esrally/driver/runner.py
```diff
@@ -404,7 +404,7 @@ async def __call__(self, *args):
                     props = parse(return_value, assertion["property"])
                     self.check_assertion(op_name, assertion, props, props_expand_keys=False)
             else:
-                self.logger.debug("Skipping assertion check in [%s] as [%s] does not return a dict.", op_name, repr(self.delegate))
+                raise exceptions.DataError(f"Cannot check assertion in [{op_name}] as [{repr(self.delegate)}] does not return a dict.")
```
DJRickyB, line 407: The only nit I have here is that perhaps this could say `did not return a dict`, as it could be that the operation *typically* returns a valid parseable response, but in this instance returned a stacktrace or something

pquentin: Thanks, I fixed that. Well it could also return JSON encoded as bytes in a BytesIO but I did not want to overcomplicate the error message.

PR #1589, opened 2022-09-28
Fix ijson deprecation warning

### LINKED ISSUE #1533: `DeprecationWarning: ijson works by reading bytes, but a string reader has been given instead` returned when running `make test`
While running `make test` (on Linux at least) we observe the following warning:

```
tests/driver/runner_test.py: 31 warnings
  /home/dl/source/elastic/rally/.venv/lib/python3.8/site-packages/ijson/compat.py:48: DeprecationWarning: 
  ijson works by reading bytes, but a string reader has been given instead. This
  probably, but not necessarily, means a file-like object has been opened in text
  mode ('t') rather than binary mode ('b').
  
  An automatic conversion is being performed on the fly to continue, but on the
  other hand this creates unnecessary encoding/decoding operations that decrease
  the efficiency of the system. In the future this automatic conversion will be
  removed, and users will receive errors instead of this warning. To avoid this
  problem make sure file-like objects are opened in binary mode instead of text
  mode.
  
    warnings.warn(_str_vs_bytes_warning, DeprecationWarning)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```

pquentin:
Like the json module from the standard library, ijson expect bytes. While the signature of `parse()` is type hinted correctly, types are not enforced and many tests were passing a StringIO instance incorrectly. This produced deprecation warnings locally (but not on CI for unknown reasons).

This closes #1533. As you can see in #1533, the original warning was not helpful. I had an eureka moment today: I modified the ijson code to use a higher [stacklevel](https://docs.python.org/3/library/warnings.html#warnings.warn) so that the warning would warn about Rally code, not ijson code. Turns out the problem was in tests. I could not resist taking 20 minutes to fix the issue at that point. Sorry.

I also reported the issue upstream: https://github.com/ICRAR/ijson/issues/79

### DIFF
--- a/tests/driver/driver_test.py
+++ b/tests/driver/driver_test.py
@@ -1373,7 +1373,7 @@ class TestAsyncExecutor:
         task_start = time.perf_counter()
         es.new_request_context.return_value = self.StaticRequestTiming(task_start=task_start)
 
-        es.bulk = mock.AsyncMock(return_value=io.StringIO('{"errors": false, "took": 8}'))
+        es.bulk = mock.AsyncMock(return_value=io.BytesIO(b'{"errors": false, "took": 8}'))
 
         params.register_param_source_for_name("driver-test-param-source", DriverTestParamSource)
         test_track = track.Track(name="unittest", description="unittest track", indices=None, challenges=None)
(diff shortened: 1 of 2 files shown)


b-deam:
> I could not resist taking 20 minutes to fix the issue at that point. Sorry.

Please don't ever apologise for fixing an issue! 😄 

> This produced deprecation warnings locally (but not on CI for unknown reasons).

Also not produced on my M1 MacBook.

b-deam (approved):
The changes work for me locally, but I also never saw the deprecation warning, so take that as you will.

(I merged master to get https://github.com/elastic/rally/pull/1586)

dliappis (approved):
Unfortunately, I can't reproduce https://github.com/elastic/rally/issues/1533 anymore on latest `master`. This PR ofc, works fine too, so LGTM.

PR #1591, opened 2022-09-29
Support checking out remote branches

b-deam:
Unfortunately I inadvertently broke building Elasticsearch from a specific commit hash from remote sources in https://github.com/elastic/rally/pull/1587.

This PR reinstates the `pull_revision` function, refactors the existing `is_commit_hash` to instead validate whether the argument is a valid SHA1 hash (5-40 characters, to allow shorthand), and adds `git.checkout_remote`, which checkouts a remote branch (note the checkout `{remote}/{branch}`)

I have tested this locally with `rally.ini` (note that we always assume Elasticsearch has a remote repo available):
```
[source]
remote.repo.url = https://github.com/b-deam/elasticsearch.git
```

And invocations:
```
esrally build --revision="test-branch" --target-os=linux --target-arch="x86_64" --source-build-method=docker
esrally build --revision="3a8dc92da20" --target-os=linux --target-arch="x86_64" --source-build-method=docker
```

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -596,7 +596,7 @@ You can specify the revision in different formats:
 * ``--revision=current``: Use the current revision (i.e. don't alter the local source tree).
 * ``--revision=abc123``: Where ``abc123`` is some git revision hash.
 * ``--revision=@2013-07-27T10:37:00Z``: Determines the revision that is closest to the provided date. Rally logs to which git revision hash the date has been resolved and if you use Elasticsearch as metrics store (instead of the default in-memory one), :doc:`each metric record will contain the git revision hash also in the meta-data section </metrics>`.
-* ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will fetch and checkout the remote branch.
+* ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will rebase your local checkout on the latest commit from the specified branch.
 
 Supported date format: If you specify a date, it has to be ISO-8601 conformant and must start with an ``@`` sign to make it easier for Rally to determine that you actually mean a date.
 
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -595,6 +595,7 @@ You can specify the revision in different formats:
 * ``--revision=latest``: Use the HEAD revision from origin/main.
 * ``--revision=current``: Use the current revision (i.e. don't alter the local source tree).
 * ``--revision=abc123``: Where ``abc123`` is some git revision hash.
+* ``--revision=v8.4.0``: Where ``v8.4.0`` is some git tag.
 * ``--revision=@2013-07-27T10:37:00Z``: Determines the revision that is closest to the provided date. Rally logs to which git revision hash the date has been resolved and if you use Elasticsearch as metrics store (instead of the default in-memory one), :doc:`each metric record will contain the git revision hash also in the meta-data section </metrics>`.
 * ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will fetch and checkout the remote branch.
 
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -596,7 +596,7 @@ You can specify the revision in different formats:
 * ``--revision=current``: Use the current revision (i.e. don't alter the local source tree).
 * ``--revision=abc123``: Where ``abc123`` is some git revision hash.
 * ``--revision=@2013-07-27T10:37:00Z``: Determines the revision that is closest to the provided date. Rally logs to which git revision hash the date has been resolved and if you use Elasticsearch as metrics store (instead of the default in-memory one), :doc:`each metric record will contain the git revision hash also in the meta-data section </metrics>`.
-* ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will rebase your local checkout on the latest commit from the specified branch.
+* ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will fetch and checkout the remote branch.
 
 Supported date format: If you specify a date, it has to be ISO-8601 conformant and must start with an ``@`` sign to make it easier for Rally to determine that you actually mean a date.
 
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -201,7 +201,7 @@ def _supply_requirements(sources, distribution, plugins, revisions, distribution
                 except KeyError:
                     # maybe we can use the catch-all revision (only if it's not a git revision)
                     plugin_revision = revisions.get("all")
-                    if not plugin_revision or SourceRepository.is_branch(plugin_revision):
+                    if not plugin_revision or SourceRepository.is_commit_hash(plugin_revision):
                         raise exceptions.SystemSetupError("No revision specified for plugin [%s]." % plugin.name)
                     logging.getLogger(__name__).info(
                         "Revision for [%s] is not explicitly defined. Using catch-all revision [%s].", plugin.name, plugin_revision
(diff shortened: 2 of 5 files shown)


pquentin (approved):
I made a few improvements to make the diff smaller between the parent of the last PR and this one, which allowed me to review this holistically. This now looks good to me, except for two small things:

 * I think the docs are now wrong because we no longer rebase when the revision is a branch
 * is_branch should probably have a probed decorator because it uses -C, but if fixing the test is difficult then don't bother because I think I'm going to kill probed soon

b-deam:
> I made a few improvements to make the diff smaller between the parent of the last PR and this one, which allowed me to review this holistically. This now looks good to me, except for two small things:
> 
> * I think the docs are now wrong because we no longer rebase when the revision is a branch
> * is_branch should probably have a probed decorator because it uses -C, but if fixing the test is difficult then don't bother because I think I'm going to kill probed soon

Many thanks for the improvements! Having a few days over the weekend to sleep on it, my original `name-rev` approach had a few flaws. 

The most major being that if a developer pushed a new branch (i.e. doesn't exist in Rally's existing local checkout) and then specified that new branch name at runtime, the `is_branch` check would return False, and we'd end up taking the commit hash route (i.e. `pull_revision`). This is trappy because it _will_ fetch and checkout the new branch as a remote tracking branch, meaning any subsequent commits to the 'new' remote branch will not be reflected. 

I took a step back and realised that although a commit hash is a technically a valid branch name, no one sane does this, and moreover GitHub actually prevents you from doing this anyway:
```
$ git checkout -b 30b52a48011d54cc591cc3427f01bfe1b6fd1e73
$ git push --set-upstream origin 30b52a48011d54cc591cc3427f01bfe1b6fd1e73
Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
remote: error: GH002: Sorry, branch or tag names consisting of 40 hex characters are not allowed.
remote: error: Invalid branch or tag name "30b52a48011d54cc591cc3427f01bfe1b6fd1e73"
To https://github.com/b-deam/elasticsearch.git
 ! [remote rejected]         30b52a48011d54cc591cc3427f01bfe1b6fd1e73 -> 30b52a48011d54cc591cc3427f01bfe1b6fd1e73 (pre-receive hook declined)
error: failed to push some refs to 'https://github.com/b-deam/elasticsearch.git'
```

I think the new approach is significantly simplified!

esrally/mechanic/supplier.py
```diff
@@ -725,7 +726,8 @@ def _update(self, revision):
 
     @classmethod
     def is_commit_hash(cls, revision):
-        return revision != "latest" and revision != "current" and not revision.startswith("@")
```
pquentin, line 723: So we're calling `is_commit_hash` in `_supply_requirements`, did you test that this new version will not affect behavior?

esrally/mechanic/supplier.py
```diff
@@ -725,7 +726,8 @@ def _update(self, revision):
...
+        sha1_pattern = re.compile("[0-9a-f]{5,40}")
```
pquentin, line 729: I'm uncomfortable with allowing short commit hashes, if your branch is named "facefeed" for example things are going to break. https://en.wikipedia.org/wiki/Hexspeak

Maybe calling `git show-ref --heads` and looking at the exit code is more robust?

pquentin (changes requested):
I've added more thoughts on the new way to detect a commit hash.

esrally/mechanic/supplier.py
b-deam: This change doesn't affect behaviour, but I did discover that the approach doesn't account for users specifying a tag name (e.g. `--revision=v7.12.0`).

esrally/mechanic/supplier.py
b-deam: > I'm uncomfortable with allowing short commit hashes, if your branch is named "facefeed" for example things are going to break. https://en.wikipedia.org/wiki/Hexspeak

Would you feel more comfortable if we upped the the short commit hash minimum to 11/12 (from 5)?

> Maybe calling git show-ref --heads and looking at the exit code is more robust?

Unfortunately this also doesn't account for tags either. I need to brainstorm a bit to work out what to do here.

b-deam: We discussed this offline, but I reverted the regex approach in https://github.com/elastic/rally/pull/1591/commits/ecab3d6f3a18909ae92db90ff61c9d89558e6e91, and went with the suggested `git show-ref` approach in https://github.com/elastic/rally/pull/1591/commits/5902fdff7c9fcfdebe251c542b1716093e631aaa.

From my testing https://github.com/elastic/rally/pull/1591/commits/5902fdff7c9fcfdebe251c542b1716093e631aaa allows us to properly support commit hashes, branch names, and tags.

docs/command_line_reference.rst
```diff
@@ -596,7 +596,7 @@ You can specify the revision in different formats:
 * ``--revision=current``: Use the current revision (i.e. don't alter the local source tree).
 * ``--revision=abc123``: Where ``abc123`` is some git revision hash.
 * ``--revision=@2013-07-27T10:37:00Z``: Determines the revision that is closest to the provided date. Rally logs to which git revision hash the date has been resolved and if you use Elasticsearch as metrics store (instead of the default in-memory one), :doc:`each metric record will contain the git revision hash also in the meta-data section </metrics>`.
-* ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will rebase your local checkout on the latest commit from the specified branch.
+* ``--revision=my-branch-name``: Where ``my-branch-name`` is the name of an existing branch name. If you are using a :ref:`remote source repository <configuration_source>`, Rally will fetch and checkout the remote branch.
```
pquentin, line 599: Should we mention tags?

tests/utils/git_test.py
```diff
@@ -32,6 +32,49 @@ def test_is_git_working_copy(self):
...
+
+        assert not git.is_branch(src, remote="origin", identifier=branch)
+
+    @mock.patch("esrally.utils.process.run_subprocess_with_logging")
+    @mock.patch("esrally.utils.process.exit_status_as_bool")
+    def test_is_not_branch_commit_hash(self, mock_exit_status_as_bool, run_subprocess_with_logging):
+        run_subprocess_with_logging.return_value = 0
+        src = "/src"
+        branch = "3694a07"
+        # True, True is for @Probed on is_branch, and fetch
```
pquentin, line 73:
```suggestion
        # True, True is for @probed on is_branch, and fetch
```

esrally/utils/git.py
```diff
@@ -48,6 +48,23 @@ def is_working_copy(src):
     return os.path.exists(src) and os.path.exists(os.path.join(src, ".git"))
 
 
+@probed
+def is_branch(src_dir, remote, identifier):
+    fetch(src_dir, remote=remote)
```
pquentin, line 53: This is very implicit, you would not expect a function called `is_branch` to affect the state by fetching. I also don'tlike that we're calling it twice in the tag/commit hash case, which can be slow when you're far away from the servers. Is there a reasonable way to avoid both issues?

esrally/utils/git.py
```diff
@@ -48,6 +48,23 @@ def is_working_copy(src):
...
+    show_ref_cmd = f"git -C {src_dir} show-ref {identifier}"
+
+    # if we get an non-zero exit code, we know that the identifier is not a branch (local or remote)
+    if not process.exit_status_as_bool(lambda: process.run_subprocess_with_logging(show_ref_cmd)):
+        return False
+
+    # it's possible the identifier could be a tag, so we explicitly check that here
+    ref = process.run_subprocess_with_output(show_ref_cmd)
+    if "refs/tags" in ref[0]:
+        return False
```
pquentin, line 63: The double call here is a bit sad, but I agree that our current process abstractions enforce it. We could add a wrapper around subprocess.run(), but I think that's not needed as this is a fast local call anyway.

pquentin (changes requested):
Thank you, we're nearly there.

docs/command_line_reference.rst
b-deam: Fixed in https://github.com/elastic/rally/pull/1591/commits/16faa6b4c8f1e0ab564d0f23a28584c7565e58d6

esrally/utils/git.py
b-deam: Addressed in https://github.com/elastic/rally/pull/1591/commits/af49a7002ecf21fbe7805a88acceed649d86888d

esrally/utils/git.py
b-deam: ++, I didn't want to add a new process function in this PR. I raised https://github.com/elastic/rally/issues/1595 to track.

pquentin (approved):
Thanks! I have not tested myself but @b-deam shared this test plan which looks solid:
- providing a branch name (already local or not) correctly fetches new commits
- providing a tag (already local or not) follows the same code path as the revisions
- revisions continue to work as expected

One big reason with the struggle here is that we're mocking git when we could have been using it in our test suite.

b-deam:
> Thanks! I have not tested myself but @b-deam shared this test plan which looks solid:
> 
> * providing a branch name (already local or not) correctly fetches new commits
> * providing a tag (already local or not) follows the same code path as the revisions
> * revisions continue to work as expected
> 
> One big reason with the struggle here is that we're mocking git when we could have been using it in our test suite.

I raised https://github.com/elastic/rally/issues/1596 to track.

PR #1594, opened 2022-10-02
Feature/format units (Issue #1419)

### LINKED ISSUE #1419: Formatting units should be conditional
When we report statistics at the end of a race, their units are hardcoded:
https://github.com/elastic/rally/blob/0b1d1317e3afe85ba89531a767868d6bc490bc28/esrally/reporter.py#L314-L319

Instead, it would be nice to have the unit formatting change depending on the value. For example, units such as time should scale depending on their value making it easier to read (milliseconds > seconds > minutes > hours).

anthonyg876:
Updated the reporter to update the lines of units if they were to large or small for units such as seconds, mins, etc.

Closes #1419

### DIFF
--- a/esrally/reporter.py
+++ b/esrally/reporter.py
@@ -367,7 +367,7 @@ class SummaryReporter:
                 if (v < 1.0):
                     v *= 60
                     u = "s"
-                elif (v > 1.0):
+                elif (v > 60.0):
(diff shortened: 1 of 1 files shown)


pquentin:

pquentin:
@elasticmachine  /test

PR #1597, opened 2022-10-13
Add additional filter options to rally list races

ebadyano:
- Added extra files to display for listing races ( `ES Version`, `Revision` and `Rally Version`).
- Removed `Track Parameters` from being displayed, as for some races (e.g. logs) it's very long and makes it hard to understand the output. 
- Added new filters for list races: `--track`, `--name` and `--to-date`, `--from-date`

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -160,8 +160,8 @@ def create_arg_parser():
         default=None,
     )
     list_parser.add_argument(
-        "--name",
-        help="Show only records from with corresponding 'name' user tag",
+        "--benchmark-name",
+        help="Show only records from with corresponding 'name' or 'benchmark-name' user tag",
         default=None,
     )
     list_parser.add_argument(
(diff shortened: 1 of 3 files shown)


esrally/rally.py
```diff
@@ -144,6 +154,28 @@ def add_track_source(subparser):
         help="Limit the number of search results for recent races (default: 10).",
         default=10,
     )
+    list_parser.add_argument(
```
michaelbaamonde, line 157: With the way things are structured right now, all `list` commands will accept these new race-specific arguments, but can't do anything useful with them. It's not a huge deal, but seeing them in the `--help` output for something like `esrally list tracks --help` could add some confusion.

That said, there's already precedent for this. `--limit` only really applies to races, but you can pass it to any other `list` subcommand.

Could you perhaps change the help text for the new arguments to indicate that they're intended to be used for races only? Something like `Show only races for this track` in this case, for example.

michaelbaamonde (changes requested):
I tested out the basic functionality with both a local and remote metrics store and it worked well! This is a very nice improvement, thanks.

In addition to my comments on adjusting the help text, I think we should add a unit test for the new filtering logic contained in `metrics.py`.

ebadyano:
@michaelbaamonde thank you for your review. I added unit tests for the new filter logic. Thank you!

michaelbaamonde (approved):
LGTM!

PR #1599, opened 2022-10-22
add `--user-tags` as alias for `--user-tag`

### LINKED ISSUE #1577: Alias `--user-tags` to `--user-tag`
Rally currently supports custom user tags via the [--user-tag](https://esrally.readthedocs.io/en/stable/command_line_reference.html?highlight=user-tag#user-tag) cli option.
This doesn't convey the concept that several tags can be user.

In order to not break BWC, I suggest we add an alias `--user-tags` acting as
a synonym to `--user-tag`.

pakio:
This PR closes #1577, adding a track option `--user-tags` as an alias for `--user-tag`. 

==== checklist ====

* Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? - yes
* Have you followed the [contributor guidelines](https://github.com/elastic/rally/blob/master/CONTRIBUTING.md)? - yes
* Have you run `make check-all` successfully? - yes
* Did you choose a [descriptive title and description](https://chris.beams.io/posts/git-commit/) for your PR? - yes
* (Only for maintainers) Did you apply appropriate labels and a milestone? - skip

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -902,20 +902,18 @@ Rally usually installs and launches an Elasticsearch cluster internally and wipe
 .. note::
    This option does only affect clusters that are provisioned by Rally. More specifically, if you use the pipeline ``benchmark-only``, this option is ineffective as Rally does not provision a cluster in this case.
 
-``user-tags/user-tag(deprecated)``
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+``user-tags``
+~~~~~~~~~~~~~
 
 This is only relevant when you want to run :doc:`tournaments </tournament>`. You can use this flag to attach an arbitrary text to the meta-data of each metric record and also the corresponding race. This will help you to recognize a race when you run ``esrally list races`` as you don't need to remember the concrete timestamp on which a race has been run but can instead use your own descriptive names.
 
-The required format is ``key`` ":" ``value``. You can choose ``key`` and  ``value`` freely.
+The required format is ``key`` ":" ``value``. You can choose ``key`` and  ``value`` freely. You can also specify multiple tags. They need to be separated by a comma.
 
 **Example**
 
  ::
 
    esrally race --track=pmc --user-tags="intention:github-issue-1234-baseline,gc:cms"
-
-You can also specify multiple tags. They need to be separated by a comma.
 
 **Example**
 
@@ -933,6 +931,9 @@ When you run ``esrally list races``, this will show up again::
     20160518T112341Z  pmc                         append-no-conflicts  defaults  disk:SSD,data_node_count:4
 
 This will help you recognize a specific race when running ``esrally compare``.
+
+.. note::
+   This option used to be named `--user-tag` without an s, which was confusing as multiple tags are supported. While users are now encouraged to use `--user-tags` for clarity, Rally will continue to honor `--user-tag` in the future to avoid breaking backwards-compatibility.
 
 ``indices``
 ~~~~~~~~~~~
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -258,7 +258,7 @@ Example JSON file::
 
 All track parameters are recorded for each metrics record in the metrics store. Also, when you run ``esrally list races``, it will show all track parameters::
 
-    Race Timestamp    Track    Track Parameters          Challenge            Car       User Tag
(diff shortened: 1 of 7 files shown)


esrally/rally.py
```diff
@@ -651,6 +651,7 @@ def add_track_source(subparser):
     )
     race_parser.add_argument(
         "--user-tag",
+        "--user-tags",
```
pakio, line 654: didn't change the order to keep backward compatibility, in order to be able to set tag by env var `USER_TAG`.

docs/command_line_reference.rst
```diff
@@ -913,21 +913,21 @@ The required format is ``key`` ":" ``value``. You can choose ``key`` and  ``valu
...
 
  ::
 
-   esrally race --track=pmc --user-tag="disk:SSD,data_node_count:4"
+   esrally race --track=pmc --user-tags="disk:SSD,data_node_count:4"
 
 
 
-When you run ``esrally list races``, this will show up again::
+Wesrally list raceshen you run ``esrally list races``, this will show up again::
```
dliappis, line 928: I think this was introduced by accident?

docs/metrics.rst
```diff
@@ -108,7 +108,7 @@ Rally captures also some meta information for each metric record:
 * Node name: If Rally provisions the cluster, it will choose a unique name for each node.
 * Source revision: We always record the git hash of the version of Elasticsearch that is benchmarked. This is even done if you benchmark an official binary release.
 * Distribution version: We always record the distribution version of Elasticsearch that is benchmarked. This is even done if you benchmark a source release.
-* Custom tag: You can define one custom tag with the command line flag ``--user-tag``. The tag is prefixed by ``tag_`` in order to avoid accidental clashes with Rally internal tags.
+* Custom tags: You can define one custom tags with the command line flag ``--user-tags``. The tags are prefixed by ``tag_`` in order to avoid accidental clashes with Rally internal tags.
```
dliappis, line 111:
```suggestion
* Custom tags: You can define custom tags with the command line flag ``--user-tags``. The tags are prefixed by ``tag_`` in order to avoid accidental clashes with Rally internal tags.
```

docs/migrate.rst
```diff
@@ -446,7 +446,7 @@ If you need lap functionality, the following shell script can be used instead::
 
     for lap in $(seq 1 ${RALLY_LAPS})
     do
-      esrally --pipeline=benchmark-only --user-tag lap:$lap
+      esrally --pipeline=benchmark-only --user-tags lap:$lap
```
dliappis, line 449: It would be very nice if we could document at the top of this file that `--user-tag` is now deprecated in favor of `--user-tags`. The next version can be found in https://github.com/elastic/rally/blob/master/esrally/_version.py#L1 so it'll be effective from `2.6.1`.

dliappis (changes requested):
Thank you for raising this!

I left some comments. One thing that is missing IMHO is that we should emit a "Deprecation warning" when `--user-tag` is use in the CLI. That way people who don't happen to check the migration guide or read docs will be reminded that they should be switching away.

dliappis:
> Thank you for raising this!
> 
> I left some comments. One thing that is missing IMHO is that we should emit a "Deprecation warning" when `--user-tag` is use in the CLI. That way people who don't happen to check the migration guide or read docs will be reminded that they should be switching away.

I discussed with @pquentin and instead of doing what I proposed here, we could just keep the functionality for both i.e. no need for a depreciation warning and allow both for the foreseeable future. Given the large usage of Rally this would probably be the least disruptive change for users.

docs/migrate.rst
dliappis: As mentioned in https://github.com/elastic/rally/pull/1599#issuecomment-1288801857 maybe it's just enough to mention that starting with 2.6.1 `--user-tags` has been introduced as a "synonym" to `--user-tag`.

pakio: oh, sorry I didn't notice this was a migration guide.
I think that might be better to be in the migration guide section for v2.6.1, so I simply reverts this change.

docs/command_line_reference.rst
pakio: unintended, thank you for the catch 🙏

dliappis:

dliappis (approved):
Sorry for the delay, this looks great thank you. I've triggered now CI tests, if they are green, I'll merge it in.

pquentin:

PR #1601, opened 2022-10-31
Remove useless pylint suppressions

pquentin:
pylint improves over times, which means some suppressions are not necessary. It helps limiting the pylint clutter we have.


michaelbaamonde (approved):
LGTM, thanks!

Issue #1596, opened 2022-10-05
we should replace mock in the git_test with real git operations


b-deam:
The existing ITs and unit test coverage for our `git` module didn't stop me from breaking an existing workflow in https://github.com/elastic/rally/pull/1587, and with the merging of https://github.com/elastic/rally/pull/1591, there's even more logic to dictate the different git commands Rally invokes.

Given the relative speed of the various git commands, we should replace the unit test mocks with the real deal.

PR #1602, opened 2022-10-31
Remove mocks from git tests

b-deam:
We learned in https://github.com/elastic/rally/pull/1587 that perhaps we were overusing 'mocks', as we didn't catch a subtle bug I introduced. This change adds approx ~20s to the total test runtime when invoking real git commands (~45s, up from ~25s). 

With this commit we remove all usage
of mocks in the `git` module tests, instead
taking the approach of copying the local Rally
checkout to a temp location and performing
a series of git operations.

Closes https://github.com/elastic/rally/issues/1596

### DIFF
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -22,6 +22,11 @@ from esrally import exceptions
 from esrally.utils import io, process
 
 
+def with_retry(command):
+    # git has no native concept of retries
+    return f'/bin/bash -c "for i in {{1..5}}; do {command} && s=0 && break || s=1 && sleep 1; done; (exit $s)"'
+
+
 def probed(f):
     def probe(src, *args, **kwargs):
         # Probe for -C
@@ -67,13 +72,13 @@ def is_branch(src_dir, identifier):
 def clone(src, *, remote):
     io.ensure_dir(src)
     # Don't swallow subprocess output, user might need to enter credentials...
-    if process.run_subprocess_with_logging("git clone %s %s" % (remote, io.escape_path(src))):
+    if process.run_subprocess_with_logging(with_retry("git clone %s %s" % (remote, io.escape_path(src)))):
         raise exceptions.SupplyError("Could not clone from [%s] to [%s]" % (remote, src))
 
 
 @probed
 def fetch(src, *, remote):
-    if process.run_subprocess_with_logging("git -C {0} fetch --prune --tags {1}".format(io.escape_path(src), remote)):
+    if process.run_subprocess_with_logging(with_retry("git -C {0} fetch --prune --tags {1}".format(io.escape_path(src), remote))):
         raise exceptions.SupplyError("Could not fetch source tree from [%s]" % remote)
 
 
@@ -92,7 +97,7 @@ def checkout_branch(src_dir, remote, branch):
 @probed
 def rebase(src_dir, *, remote, branch):
     checkout(src_dir, branch=branch)
-    if process.run_subprocess_with_logging("git -C {0} rebase {1}/{2}".format(io.escape_path(src_dir), remote, branch)):
+    if process.run_subprocess_with_logging(with_retry("git -C {0} rebase {1}/{2}".format(io.escape_path(src_dir), remote, branch))):
         raise exceptions.SupplyError("Could not rebase on branch [%s]" % branch)
 
 
@@ -106,7 +111,7 @@ def pull(src_dir, *, remote, branch):
 def pull_ts(src_dir, ts, *, remote, branch):
     fetch(src_dir, remote=remote)
     clean_src = io.escape_path(src_dir)
-    rev_list_command = f'git -C {clean_src} rev-list -n 1 --before="{ts}" --date=iso8601 {remote}/{branch}'
+    rev_list_command = with_retry(f'git -C {clean_src} rev-list -n 1 --before="{ts}" --date=iso8601 {remote}/{branch}')
     revision = process.run_subprocess_with_output(rev_list_command)[0].strip()
     if process.run_subprocess_with_logging("git -C {0} checkout {1}".format(clean_src, revision)):
         raise exceptions.SupplyError("Could not checkout source tree for timestamped revision [%s]" % ts)
@@ -114,7 +119,7 @@ def pull_ts(src_dir, ts, *, remote, branch):
 
 @probed
 def checkout_revision(src_dir, *, revision):
-    if process.run_subprocess_with_logging("git -C {0} checkout {1}".format(io.escape_path(src_dir), revision)):
+    if process.run_subprocess_with_logging(with_retry("git -C {0} checkout {1}".format(io.escape_path(src_dir), revision))):
         raise exceptions.SupplyError("Could not checkout source tree for revision [%s]" % revision)
 
 
@@ -134,7 +139,9 @@ def branches(src_dir, remote=True):
     if remote:
         # alternatively: git for-each-ref refs/remotes/ --format='%(refname:short)'
         return _cleanup_remote_branch_names(
-            process.run_subprocess_with_output("git -C {src} for-each-ref refs/remotes/ --format='%(refname:short)'".format(src=clean_src))
+            process.run_subprocess_with_output(
+                with_retry("git -C {src} for-each-ref refs/remotes/ --format='%(refname:short)'".format(src=clean_src))
+            )
         )
     else:
         return _cleanup_local_branch_names(
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -90,7 +90,7 @@ def checkout(src_dir, *, branch):
 
 @probed
 def checkout_branch(src_dir, remote, branch):
-    if process.run_subprocess_with_logging("git -C {0} checkout {1}/{2}".format(io.escape_path(src_dir), remote, branch)):
+    if process.run_subprocess_with_logging(with_retry("git -C {0} checkout {1}/{2}".format(io.escape_path(src_dir), remote, branch))):
         raise exceptions.SupplyError("Could not checkout [%s]. Do you have uncommitted changes?" % branch)
 
 
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -22,11 +22,6 @@ from esrally import exceptions
 from esrally.utils import io, process
 
 
-def with_retry(command):
-    # git has no native concept of retries
-    return f'/bin/bash -c "for i in {{1..5}}; do {command} && s=0 && break || s=1 && sleep 1; done; (exit $s)"'
-
-
 def probed(f):
     def probe(src, *args, **kwargs):
         # Probe for -C
@@ -72,13 +67,13 @@ def is_branch(src_dir, identifier):
 def clone(src, *, remote):
     io.ensure_dir(src)
     # Don't swallow subprocess output, user might need to enter credentials...
-    if process.run_subprocess_with_logging(with_retry("git clone %s %s" % (remote, io.escape_path(src)))):
+    if process.run_subprocess_with_logging("git clone %s %s" % (remote, io.escape_path(src))):
         raise exceptions.SupplyError("Could not clone from [%s] to [%s]" % (remote, src))
 
 
 @probed
 def fetch(src, *, remote):
-    if process.run_subprocess_with_logging(with_retry("git -C {0} fetch --prune --tags {1}".format(io.escape_path(src), remote))):
+    if process.run_subprocess_with_logging("git -C {0} fetch --prune --tags {1}".format(io.escape_path(src), remote)):
         raise exceptions.SupplyError("Could not fetch source tree from [%s]" % remote)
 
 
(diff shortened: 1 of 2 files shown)


esrally/utils/git.py
```diff
@@ -22,6 +22,11 @@
 from esrally.utils import io, process
 
 
+def with_retry(command):
+    # git has no native concept of retries
+    return f'/bin/bash -c "for i in {{1..5}}; do {command} && s=0 && break || s=1 && sleep 1; done; (exit $s)"'
```
b-deam, line 27: Given we now invoke real git operations in the test (including cloning, fetching, etc.), I figured it'd be a prerequisite to ensure that these commands were naively retryable.

b-deam:
@elasticmachine run rally/it-python310

b-deam:
CI failures are both:
```
14:03:27 =================================== FAILURES ===================================
14:03:27 ___________ test_authenticated_proxy_user_can_connect[in-memory-it] ____________
14:03:27 
14:03:27 cfg = 'in-memory-it'
14:03:27 http_proxy = HttpProxy(authenticated_url='http://testuser:testuser@127.0.0.1:3128/', anonymous_url='http://127.0.0.1:3128/')
14:03:27 
14:03:27     @it.rally_in_mem
14:03:27     def test_authenticated_proxy_user_can_connect(cfg, http_proxy):
14:03:27         env = dict(os.environ)
14:03:27         env["http_proxy"] = http_proxy.authenticated_url
14:03:27         env["https_proxy"] = http_proxy.authenticated_url
14:03:27         lines = process.run_subprocess_with_output(it.esrally_command_line_for(cfg, "list tracks"), env=env)
14:03:27         output = "\n".join(lines)
14:03:27         # rally should be able to connect, no warning
14:03:27 >       assert "[WARNING] Could not update tracks." not in output
14:03:27 E       AssertionError: assert '[WARNING] C...date tracks.' not in '\n____     ...------------'
14:03:27 E         '[WARNING] Could not update tracks.' is contained here:
14:03:27 E           
14:03:27 E           ____        ____
14:03:27 E           / __ \____ _/ / /_  __
14:03:27 E           / /_/ / __ `/ / / / / /
14:03:27 E           / _, _/ /_/ / / / /_/ /
14:03:27 E           /_/ |_|\__,_/_/_/\__, /...
14:03:27 E         
14:03:27 E         ...Full output truncated (35 lines hidden), use '-vv' to show
14:03:27 
14:03:27 it/proxy_test.py:98: AssertionError
```

I can't repro this on an ARM MacOS, but can on a Linux x86 machine. I can also reproduce this failure on `master`, so I don't think this is related to this PR.

b-deam:
lol
```
$ docker run --rm --name squid -d -v it/resources/squid//squidpasswords:/etc/squid/squidpasswords -v it/resources/squid//squid.conf:/etc/squid/squid.conf -p 3128:3128 datadog/squid
Unable to find image 'datadog/squid:latest' locally
docker: Error response from daemon: pull access denied for datadog/squid, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.

$ docker search squid
NAME                         DESCRIPTION                                     STARS     OFFICIAL   AUTOMATED
sameersbn/squid                                                              222                  [OK]
squidfunk/mkdocs-material    Documentation that simply works                 113                  [OK]
minimum2scp/squid            squid3 cache service container running on de…   66                   [OK]
ubuntu/squid                 Squid is a caching proxy for the Web. Long-t…   40
b4tman/squid                 Squid container based on Alpine Linux           17                   [OK]
squidex/squidex              Squidex Headless CMS                            13
lucacri/squid-ext-conf       Squid server on alpine, gathering configurat…   12
babim/squid                  Squid 4 on alpine linux. With or without auth   3                    [OK]
photoprism/mkdocs-material   Tracks stable major versions of squidfunk/mk…   2
cloudreach/squid-utm         Squid Proxy UTM                                 1
kasmweb/squid                A proxy intended to redirect users to a Kasm…   1
boynux/squid-exporter        Squid metrics exporter for Prometheus           1                    [OK]
squids/squids-exporter                                                       0
squids/mysql                                                                 0
squids/mycilium                                                              0
squids/mysqlleader                                                           0
squids/mycilium-amd64                                                        0
squidex/nginx-proxy                                                          0
squids/qfb-mysql-8.0                                                         0
squidex/caddy-proxy                                                          0
squidtoon99/apiserver-rs                                                     0
squids/qfb-backup-job                                                        0
squidex/squidex-identity     Squidex Identity Server                         0
squidcache/buildfarm         Images for the project's Jenkins build farm     0
squids/fluent-bit                                                            0
```

This image doesn't seem to exist anymore, but it was cached locally for me (which is why ITs worked locally).

tests/utils/git_test.py
```diff
@@ -140,41 +154,66 @@ def test_fetch_with_error(self):
...
+    def test_pull_ts(self):
+        git.pull_ts(TestGit.tmp_src_dir, "20160101T110000Z", remote="origin", branch="master")
+        assert "28474f4f097106ff3507be35958db0c3c8be0fc6" == git.head_revision(TestGit.tmp_src_dir)
 
-    def test_list_tags_with_tags_present(self):
-        expected_tags = ["2.6.0", "2.5.0", "2.4.0", "2.3.1", "2.3.0"]
-        tags = git.tags(TestGit.tmp_src_dir)
-        for tag in expected_tags:
-            assert tag in tags
+    @mock.patch("esrally.utils.process.run_subprocess_with_logging")
```
b-deam, line 181: I can't think of a way to properly test rebase or pull without doing things like pushing commits to a remote from _another_ cloned dir, which seems like it's getting a bit messy for unit test territory.

pquentin:
cla/check

b-deam:
@elasticmachine run rally/precommit

b-deam:

b-deam:
@pquentin - apologies for the mess, turns out the slight differences between MacOS Git and Linux git made the feedback loop quite long.

This is now ready for review.

esrally/utils/git.py
b-deam: Reverted in https://github.com/elastic/rally/pull/1602/commits/59b762e29b55d6262be9dd5b32b0b2d0d13b6296

tests/utils/git_test.py
```diff
@@ -17,63 +17,139 @@
...
+        # this is assuming that nobody stripped the git repo info in their Rally working copy
+        cls.original_src_dir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
+        cls.local_tmp_src_dir = io.escape_path("/tmp/rally-unit-test-local-dir")
+        cls.remote_tmp_src_dir = io.escape_path("/tmp/rally-unit-test-remote-dir")
+        cls.tmp_clone_dir = io.escape_path("/tmp/rally-unit-test-clone-dir")
+
+        # delete any pre-existing dirs (maybe last test setup cancelled early)
+        shutil.rmtree(TestGit.local_tmp_src_dir, ignore_errors=True)
+        shutil.rmtree(TestGit.remote_tmp_src_dir, ignore_errors=True)
+        shutil.rmtree(TestGit.tmp_clone_dir, ignore_errors=True)
```
pquentin, line 48: Can you please use `tmp_path` instead of `/tmp`? https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html

It makes sure to use the correct /tmp for every OS (macOS currently auto-deletes entries in /tmp after a few days, for example - not an issue here but can be surprising) and means you don't need to care about removing the directories, pytest removes older entries automatically. So you can remove all rmtree calls.

tests/utils/git_test.py
```diff
@@ -17,63 +17,139 @@
...
+        )
+
+        # fetch branches from remote
+        git.fetch(TestGit.local_tmp_src_dir, remote=TestGit.remote_repo)
+
+    @classmethod
+    def teardown_class(cls):
+        # delete copy of git dir
+        shutil.rmtree(TestGit.local_tmp_src_dir)
+        shutil.rmtree(TestGit.remote_tmp_src_dir)
```
pquentin, line 76: This won't be needed given if you use `tmp_path`.

tests/utils/git_test.py
```diff
@@ -87,176 +163,75 @@ def test_git_version_too_old(self, run_subprocess_with_logging, run_subprocess):
...
-    def test_fetch_successful(self, run_subprocess_with_logging):
-        run_subprocess_with_logging.return_value = 0
-        git.fetch("/src", remote="my-origin")
-        run_subprocess_with_logging.assert_called_with("git -C /src fetch --prune --tags my-origin")
+    def test_fetch_successful(self):
+        before_time = os.path.getmtime(os.path.join(TestGit.local_tmp_src_dir, ".git/FETCH_HEAD"))
+        git.fetch(TestGit.local_tmp_src_dir, remote=TestGit.remote_repo)
+        after_time = os.path.getmtime(os.path.join(TestGit.local_tmp_src_dir, ".git/FETCH_HEAD"))
+        fetch_head_modified = after_time > before_time
+        assert fetch_head_modified
```
pquentin, line 181: This seems brittle to me as it depends on the time resolution of the file system. I think it's good enough to notice that fetch does not fail.

tests/utils/git_test.py
```diff
@@ -87,176 +163,75 @@ def test_git_version_too_old(self, run_subprocess_with_logging, run_subprocess):
...
-        run_subprocess_with_logging.return_value = 0
-        run_subprocess.return_value = ""
-        assert git.tags("/src") == []
-        run_subprocess.assert_called_with("git -C /src tag")
+            git.checkout(TestGit.local_tmp_src_dir, branch=branch)
+        assert exc.value.args[0] == f"Could not checkout [{branch}]. Do you have uncommitted changes?"
+
+    def test_checkout_revision(self):
+        # Apple Git 'core.abbrev' defaults to return 7 char prefixes (09980cd)
+        # Linux defaults to return 8 char prefixes (09980cd5)
```
pquentin, line 200: I also get 7 char prefixes on my Linux laptop with git 2.38.1. I think it's enough to mention that minimum core.abbrev is 7 across all git versions

tests/utils/git_test.py
```diff
@@ -87,176 +163,75 @@ def test_git_version_too_old(self, run_subprocess_with_logging, run_subprocess):
...
+    def test_rebase(self, setup_teardown_rebase):
+        # fetch required first to get remote branch
+        git.fetch(TestGit.local_tmp_src_dir, remote=TestGit.remote_repo)
+        git.rebase(TestGit.local_tmp_src_dir, remote=TestGit.remote_repo, branch=TestGit.rebase_branch)
+        ops = " ".join(process.run_subprocess_with_output(f"git -C {TestGit.local_tmp_src_dir} reflog --since=10.seconds.ago"))
+        # reflog format differs from Apple Git to Linux
+        finished_pattern = re.compile(f"rebase (\\(finish\\)|finished).*returning to refs\\/heads\\/{TestGit.rebase_branch}")
+        started_pattern = re.compile(f"rebase:? (\\(start\\)*.|checkout).*{TestGit.remote_repo}\\/{TestGit.rebase_branch}")
+        assert re.search(started_pattern, ops)
+        assert re.search(finished_pattern, ops)
```
pquentin, line 228: This is brittle, not only because of the reflog format, but also because of the 10 seconds. We could match something that we do not want, and the machine could also hand for 10 seconds (especially in CI, say because a live migration happened or something).

What do you think about committing to the remote repo and then checking that the commit is here after the rebase?

tests/utils/git_test.py
```diff
@@ -17,63 +17,139 @@
...
+    @classmethod
+    def teardown_class(cls):
+        # delete copy of git dir
+        shutil.rmtree(TestGit.local_tmp_src_dir)
+        shutil.rmtree(TestGit.remote_tmp_src_dir)
+
+    @pytest.fixture
+    def teardown_clone(self):
+        yield
+        shutil.rmtree(TestGit.tmp_clone_dir)
```
pquentin, line 81: This seems quite implicit, why is this a fixture?

pquentin (changes requested):
Thank you! I think it's a big improvement.

tests/utils/git_test.py
b-deam: Thanks for the pointers! Addressed in https://github.com/elastic/rally/pull/1602/commits/529ea0a8dd8a9cdd47b33890eadbd05ea5cf13e4 and https://github.com/elastic/rally/pull/1602/commits/8f50ce76806a43d42ecf30fc300e7ff48cf724b8

tests/utils/git_test.py
b-deam: Addressed in https://github.com/elastic/rally/pull/1602/commits/529ea0a8dd8a9cdd47b33890eadbd05ea5cf13e4 and https://github.com/elastic/rally/pull/1602/commits/8f50ce76806a43d42ecf30fc300e7ff48cf724b8

tests/utils/git_test.py
b-deam: Addressed in https://github.com/elastic/rally/pull/1602/commits/ebbb7c1e08384a2f78c5692b903caf52ff5a6e14

tests/utils/git_test.py
b-deam: Sure. Addressed in https://github.com/elastic/rally/pull/1602/commits/c37f33409ec2f7668d01be524c4439f37480afdb

tests/utils/git_test.py
b-deam: Nice suggestion, and agreed RE: brittle. Addressed in https://github.com/elastic/rally/pull/1602/commits/8470aee7f82a920dfefc7bae0023761a028445f8

b-deam:

pquentin (approved):
Thanks, this looks good to me!

I'm not too happy at the amount of code in setup() that every test relies on, but I can only admit that the result is short and am not sure any alternative would be clearly superior.

Also would like to note that we're "only" loosing 3 seconds over this, not 20.

PR #1606, opened 2022-11-02
Fix broken log archival/upload where there CI failures

dliappis:
Currently, CI failures for it tests fail immediately and prohibit
log archival/upload which makes understanding failures an impossible
task.

The approach taken here is an approach successfully taken in a different
repository.

### DIFF
--- a/.ci/templates/it.yml
+++ b/.ci/templates/it.yml
@@ -12,7 +12,6 @@
       - shell: |
           #!/usr/local/bin/runbld
           archive_logs() {
-            # archive logs for ^^ google-cloud-storage uploader
             bash .ci/build.sh archive
           }
             
--- a/.ci/templates/it.yml
+++ b/.ci/templates/it.yml
@@ -11,6 +11,14 @@
     builders:
       - shell: |
           #!/usr/local/bin/runbld
-          set -o errexit
+          archive_logs() {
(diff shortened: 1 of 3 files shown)


dliappis:
By the way, I've noticed that IT failures don't cause an immediately failure, at least in a localized environments e.g.

image

I'll look into this more, but I wonder if anyone has noticed anything similar so far?

michaelbaamonde (approved):
LGTM.

michaelbaamonde:
> By the way, I've noticed that IT failures don't cause an immediately failure, at least in a localized environments e.g.

This is just how pytest works by default, AFAIK. You can have it abort on first failure ([see docs](https://docs.pytest.org/en/7.1.x/how-to/failures.html#stopping-after-the-first-or-n-failures)) but it'll run the full suite by default.

b-deam (approved):

dliappis:
I tested this using a local jenkins workflow and it seems to fix the problem:

```
09:42:56 [Checks API] No suitable checks publisher found.
09:42:56 [Google Cloud Storage Plugin] Found 1 files to upload from pattern: .rally/$BUILD_NUMBER.tar.bz2
09:43:53 [Google Cloud Storage Plugin] Uploading: .rally/3.tar.bz2
09:43:57 Finished: FAILURE
```

PR #1607, opened 2022-11-02
Cut down on the amount of INFO-level logging

michaelbaamonde:
This PR moves many INFO-level logging statements to DEBUG. These primarily concern the actor system, ES client instantiation, and relatively low-level load-driver operations.

On my machine, running `esrally race --track=geonames --test-mode` with `master` results in 2,295 lines of logs. This PR cuts it down to 650, with no (IMO) substantial loss of actionable information in the default case.

[This commit ](https://github.com/elastic/rally/commit/a42b9af0168d85b6de923f8ff3b96d1942b087dc)in particular makes it easier to follow what's going on during task execution. We'll now have a single line logged each time a worker finishes executing a task, including the time it took to complete the task. For example:

```
Worker[0] finished executing tasks ['delete-index'] in 0.107148 seconds
Worker[0] finished executing tasks ['create-index'] in 0.677319 seconds
Worker[0] finished executing tasks ['check-cluster-health'] in 0.011707 seconds
Worker[1] finished executing tasks ['index-append'] in 0.143370 seconds
```
And so on. Previously, this was scattered across multiple lines and difficult to parse.

All that said, I can see case-by-case arguments for some of these remaining at `INFO` as opposed to `DEBUG`, and I'm happy to adjust if there are places where I've been too ruthless. :) Git operations, for example, may be worth logging by default.

I'd suggest that reviewers give this a try locally with a few different tracks and see how it strikes them. The goal is to cut down on noise while preserving usefulness, and provide `DEBUG` as a fallback for trickier cases that require lower-level inspection. I've pinged the team broadly since this affects everyone to one degree or another.

### DIFF
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -67,19 +67,19 @@ def is_branch(src_dir, identifier):
 def clone(src, *, remote):
     io.ensure_dir(src)
     # Don't swallow subprocess output, user might need to enter credentials...
-    if process.run_subprocess_with_logging("git clone %s %s" % (remote, io.escape_path(src)), level=logging.DEBUG):
+    if process.run_subprocess_with_logging("git clone %s %s" % (remote, io.escape_path(src))):
         raise exceptions.SupplyError("Could not clone from [%s] to [%s]" % (remote, src))
 
 
 @probed
 def fetch(src, *, remote):
-    if process.run_subprocess_with_logging("git -C {0} fetch --prune --tags {1}".format(io.escape_path(src), remote), level=logging.DEBUG):
+    if process.run_subprocess_with_logging("git -C {0} fetch --prune --tags {1}".format(io.escape_path(src), remote)):
         raise exceptions.SupplyError("Could not fetch source tree from [%s]" % remote)
 
 
 @probed
 def checkout(src_dir, *, branch):
-    if process.run_subprocess_with_logging("git -C {0} checkout {1}".format(io.escape_path(src_dir), branch), level=logging.DEBUG):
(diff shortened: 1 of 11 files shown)


esrally/utils/git.py
```diff
@@ -67,19 +67,19 @@ def is_branch(src_dir, identifier):
 def clone(src, *, remote):
     io.ensure_dir(src)
     # Don't swallow subprocess output, user might need to enter credentials...
-    if process.run_subprocess_with_logging("git clone %s %s" % (remote, io.escape_path(src))):
+    if process.run_subprocess_with_logging("git clone %s %s" % (remote, io.escape_path(src)), level=logging.DEBUG):
```
b-deam, line 70: I think we should keep all of the git module logging at `INFO`. These commands are only invoked as a benchmark is starting, so their volume shouldn't be a concern, and they're generally useful in verifying that the correct actions took place (such as for building from source, teams, etc.).

Side note: we'll likely have some conflicts due to https://github.com/elastic/rally/pull/1602

michaelbaamonde: Yeah, this was one I was on the fence about. I'm fine with keeping at INFO, so I'll adjust. Thanks for the feedback!

inqueue (approved):
Thank you for this PR. I tested it against a new track I'm developing[^1] and so far so good. With timestamps removed and "zeroing" all Actor thread/process IDs, I see a significant improvement (though this analysis is not without flaws).

image

The number of logged lines with this PR dropped by over half:

```shell
grape:logs $ wc -l rally.log rally.log.1607            
     432 rally.log
     210 rally.log.1607
```

I'm good with what is removed, and good with this PR. LGTM!

[^1]: https://github.com/elastic/rally-tracks/pull/340

PR #1609, opened 2022-11-03
Remove TODO about remove install command

pquentin:
esrally stop --preserve-install works well, there is no need for a separate command.


dliappis (approved):
LGTM

PR #1610, opened 2022-11-03
Support specifying user tags as json

dliappis:
While the rest of the cli options (like `car-params`, `telemetry-params`) support json, in addition to command separate K:V string, this is currently not possible in Rally.

This commit adds additionaly support for user-specified tags as inlined json or referenced from a json file.

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -905,9 +905,9 @@ Rally usually installs and launches an Elasticsearch cluster internally and wipe
 ``user-tags``
 ~~~~~~~~~~~~~
 
-This is only relevant when you want to run :doc:`tournaments </tournament>`. You can use this flag to attach an arbitrary text to the meta-data of each metric record and also the corresponding race. This will help you to recognize a race when you run ``esrally list races`` as you don't need to remember the concrete timestamp on which a race has been run but can instead use your own descriptive names.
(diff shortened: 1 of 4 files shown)


docs/command_line_reference.rst
```diff
@@ -905,9 +905,9 @@ Rally usually installs and launches an Elasticsearch cluster internally and wipe
 ``user-tags``
 ~~~~~~~~~~~~~
 
-This is only relevant when you want to run :doc:`tournaments </tournament>`. You can use this flag to attach an arbitrary text to the meta-data of each metric record and also the corresponding race. This will help you to recognize a race when you run ``esrally list races`` as you don't need to remember the concrete timestamp on which a race has been run but can instead use your own descriptive names.
+You can use this flag to attach a arbitrary text to the meta-data of each metric record and also the corresponding race. This will help you to recognize a race when you run ``esrally list races`` as you don't need to remember the concrete timestamp on which a race has been run but can instead use your own descriptive names.
```
michaelbaamonde, line 908:
```suggestion
You can use this flag to attach arbitrary text to the meta-data of each metric record and also the corresponding race. This will help you to recognize a race when you run ``esrally list races`` as you don't need to remember the concrete timestamp on which a race has been run but can instead use your own descriptive names.
```

michaelbaamonde (approved):
LGTM pending CI. Very nice change!

dliappis:

PR #1616, opened 2022-11-14
Append build output instead of clobbering

b-deam:
Title.

[#es-perf-build Slack thread for context](https://elastic.slack.com/archives/C5AK2B6M8/p1668182102346229).


dliappis (approved):
LGTM

We may want to look at rotating the build log as a result of this PR on long running nightly servers.

PR #1619, opened 2022-11-15
Add shard data to search operation's detailed-results meta-data

b-deam:
With this commit we include the shard meta-data from Elasticsearch's 
search API response when specifying `detailed-results` for Rally's search
operation.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -1034,10 +1034,10 @@ class Query(Runner):
                 timed_out = props.get("timed_out", False)
                 took = props.get("took", 0)
 
-                shards_total = props.get("_shards.total")
-                shards_successful = props.get("_shards.successful")
-                shards_skipped = props.get("_shards.skipped")
-                shards_failed = props.get("_shards.failed")
+                shards_total = props.get("_shards.total", 0)
+                shards_successful = props.get("_shards.successful", 0)
+                shards_skipped = props.get("_shards.skipped", 0)
+                shards_failed = props.get("_shards.failed", 0)
 
                 return {
                     "weight": 1,
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -1015,11 +1015,29 @@ class Query(Runner):
             r = await self._raw_search(es, doc_type, index, body, request_params, headers=headers)
 
             if detailed_results:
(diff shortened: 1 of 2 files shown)


esrally/driver/runner.py
```diff
@@ -1015,12 +1015,30 @@ async def _request_body_query(es, params):
...
+                        "_shards.skipped",
+                        "_shards.failed",
+                    ],
+                )
                 hits_total = props.get("hits.total.value", props.get("hits.total", 0))
                 hits_relation = props.get("hits.total.relation", "eq")
                 timed_out = props.get("timed_out", False)
                 took = props.get("took", 0)
 
+                shards_total = props.get("_shards.total")
```
michaelbaamonde, line 1037: Should we have non-`None` defaults for these `props.get()` calls?

b-deam: Hmm, good point. I think if there's no `_shard` object in the response we should just omit it entirely, rather than populating it with a default value of `0` which could be misleading.

b-deam: Actually, I can't think of a scenario where `_shards` wouldn't be present. `0` is actually a good default.

b-deam: Addressed in https://github.com/elastic/rally/pull/1619/commits/9e9d96fffdce3af39221abc26f6d7d8c126a5769

michaelbaamonde (approved):
LGTM, thanks!

b-deam:

b-deam:
I think the rally-track-compat failure is unrelated, it's for `.rally/benchmarks/tracks/rally-tracks-compat/it/test_security.py`:
```
12:19:02 ----------------------------- Captured stdout call -----------------------------
12:19:02 
12:19:02     ____        ____
12:19:02    / __ \____ _/ / /_  __
12:19:02   / /_/ / __ `/ / / / / /
12:19:02  / _, _/ /_/ / / / /_/ /
12:19:02 /_/ |_|\__,_/_/_/\__, /
12:19:02                 /____/
12:19:02 
12:19:02 [INFO] Race id is [e931dacd-b0de-47fa-b156-799ebe31503d]
12:19:02 [ERROR] Cannot race. Error in load generator [0]
12:19:02 	No module named 'yaml'
```

Issue #1617, opened 2022-11-14
Add delay and duration parameters for the JFR telemetry device


dliappis:
The [JFR telemetry device](https://esrally.readthedocs.io/en/stable/telemetry.html#jfr) while very useful, currently, only captures flight recordings from the time that ES starts until it's been stopped. This makes it hard to grab more focused JFRs for specific tasks, e.g. the [scroll task](https://github.com/elastic/rally-tracks/blob/c86b3ad1e6fe831ff3cc61e0bca2b5134b16319d/http_logs/challenges/common/default-schedule.json#L145-L150) in `http_logs`.

The current workaround is to leverage [include|exclude tasks](https://esrally.readthedocs.io/en/stable/command_line_reference.html?highlight=include-tasks#include-tasks) to reduce the amount of operations but often it's not an option as indexing usually needs to take place anyway and also skipping preceding tasks tends to affect reproducability.

Until we get a way to coordinate tasks and telemetry device execution in Rally, a short-term improvement could be allowing to pass the `delay` and `duration` parameters to `jcmd`, as per the official [jcmd docs](https://docs.oracle.com/en/java/javase/19/docs/specs/man/jcmd.html) for `JFR.start`. Currently we only support one parameter (`recording-template`) for the `jfr` telemetry device, and adding those two should be a straightforward task.

debanjanc01:
Hey @dliappis, if this issue is up for grabs then I'd like to take it up.

dliappis:
@debanjanc01 sure go for it.

debanjanc01:
Thanks @dliappis ! Could you please assign this to me?

dliappis:
> Thanks @dliappis ! Could you please assign this to me?

Done

debanjanc01:
@dliappis I've raise the PR for this.

PR #1630, opened 2022-11-21
added delay and duration parameters to jfr telmetry device

debanjanc01:
* Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)? - YES
* Have you followed the [contributor guidelines](https://github.com/elastic/rally/blob/master/CONTRIBUTING.md)?  - YES
* Have you run `make check-all` successfully? - YES
* Did you choose a [descriptive title and description](https://chris.beams.io/posts/git-commit/) for your PR? - YES

---
This closes: https://github.com/elastic/rally/issues/1617

Two additional parameters have been added for JFR

1. `delay`
2. `duration`

Additionally, the code block for the existing parameter `recording-template` has been refactored to avoid code duplication.


dliappis:

esrally/telemetry.py
```diff
@@ -229,6 +229,8 @@ def instrument_java_opts(self):
 
     def java_opts(self, log_file):
         recording_template = self.telemetry_params.get("recording-template")
+        delay = self.telemetry_params.get("delay")
+        duration = self.telemetry_params.get("duration")
```
dliappis, line 233: I wonder if we should prefix these (and possible also the `recording-template`) with `jfr` i.e. `jfr-delay` and `jfr-duration`.
The rationale is that telemetry device parameters are not namespaced and the meaning of `delay` and `duration` is too wide to be dedicated only for the `jfr` telemetry device.

While at it, we could also rename `recording-template` but a) it'd be a breaking change for some and should also be documented in `migrate.rst` and b) `recording-template` is less generic and thus not so important to change right away.

WDYT?

esrally/telemetry.py
```diff
@@ -238,18 +240,19 @@ def java_opts(self, log_file):
...
         else:
             jfr_cmd += "-XX:StartFlightRecording=maxsize=0,maxage=0s,disk=true,dumponexit=true,filename={}".format(log_file)
-            if recording_template:
-                self.logger.info("jfr: Using recording template [%s].", recording_template)
-                jfr_cmd += ",settings={}".format(recording_template)
-            else:
-                self.logger.info("jfr: Using default recording template.")
+        if delay:
+            self.logger.info("jfr: Using delay [%s].", delay)
+            jfr_cmd += ",delay={}".format(delay)
```
dliappis, line 247:
```suggestion
            jfr_cmd += f",delay={delay}"
```

esrally/telemetry.py
```diff
@@ -238,18 +240,19 @@ def java_opts(self, log_file):
...
+        if duration:
+            self.logger.info("jfr: Using duration [%s].", duration)
+            jfr_cmd += ",duration={}".format(duration)
```
dliappis, line 250:
```suggestion
            jfr_cmd += f",duration={duration}"
```

esrally/telemetry.py
```diff
@@ -238,18 +240,19 @@ def java_opts(self, log_file):
...
+        if recording_template:
+            self.logger.info("jfr: Using recording template [%s].", recording_template)
+            jfr_cmd += ",settings={}".format(recording_template)
```
dliappis, line 253:
```suggestion
            jfr_cmd += f",settings={recording_template}"
```

dliappis (commented):
Thank you for the quick and meticulous PR!

I left a question/suggestion about how we expose the options as telemetry parameters (sorry the issue should have thought about it) and that preferably we should namespace them as we do for other telemetry params e.g. `node-stats-sample-internal` etc.

I haven't tested the PR myself, but triggered a CI run.

esrally/telemetry.py
debanjanc01: I agree that adding the `jfr` prefix would make it more consistent with the existing parameter naming convention.
If we are not adding it to the `recording-template` then wouldn't it break this consistency?

esrally/telemetry.py
debanjanc01: Had used the string.format as the existing code was using the same. 
Using f string would be a better alternative. Will be updating the same for all.

tests/telemetry_test.py
```diff
@@ -295,6 +295,142 @@ def test_sets_options_for_java_11_or_above_custom_recording_template(self):
...
+            log_root="/var/log",
+            java_major_version=random.randint(11, 999),
+        )
+        assert jfr.java_opts("/var/log/test-recording.jfr") == [
+            "-XX:+UnlockDiagnosticVMOptions",
+            "-XX:+DebugNonSafepoints",
+            "-XX:StartFlightRecording=maxsize=0,maxage=0s,disk=true,dumponexit=true,"
+            "filename=/var/log/test-recording.jfr,delay=10s,duration=20m,settings=profile",
+        ]
+
```
pquentin, line 433: Thanks for being that thorough in your tests. However, I think it's a bit too much. Each test should test for a specific thing, and each change should only break one test. Case in point, to change duration to jfr-duration and delay to jfr-delay, you now have to change twelve tests!

In practice, you don't have to test specifically for the Java version, it's done already. And you don't have to test for the "no parameters case", it's done already. You can reduce your twelve tests to one with all the settings enabled (params, delay and duration).

debanjanc01:
> Thank you for the quick and meticulous PR!
> 
> I left a question/suggestion about how we expose the options as telemetry parameters (sorry the issue should have thought about it) and that preferably we should namespace them as we do for other telemetry params e.g. `node-stats-sample-internal` etc.
> 
> I haven't tested the PR myself, but triggered a CI run.

I've let my replies. Shall I proceed with the changes? Additionally just wanted to know if there should be separate commits or a single squashed commit?

esrally/telemetry.py
```diff
@@ -238,18 +240,19 @@ def java_opts(self, log_file):
...
-            else:
-                self.logger.info("jfr: Using default recording template.")
         else:
             jfr_cmd += "-XX:StartFlightRecording=maxsize=0,maxage=0s,disk=true,dumponexit=true,filename={}".format(log_file)
-            if recording_template:
-                self.logger.info("jfr: Using recording template [%s].", recording_template)
-                jfr_cmd += ",settings={}".format(recording_template)
-            else:
-                self.logger.info("jfr: Using default recording template.")
+        if delay:
```
pquentin, line 245: When a `if` statement is directly below an existing if/elif/else statement, please add a newline to clarify that this is a different statement. This applies to duration and recording_template below.

pquentin (changes requested):
Looks great! Left two comments, one about code style and the other about testing strategy.

tests/telemetry_test.py
debanjanc01: Makes sense, will be updating this to just have a single test will all params.

esrally/telemetry.py
pquentin: I talked to @dliappis offline. It's OK to keep it like this for now. As PEP 8 says, a foolish consistency is the hobgoblin of little minds. :)

That said, in the future we'll want to switch. This requires a deprecation cycle, much like was done for relative-time (search for it in https://esrally.readthedocs.io/en/stable/migrate.html). Can you please create an issue about migrating to `jfr-recording-template` so that we don't forget to do it?

debanjanc01: Hey @pquentin, just to confirm the todos:

1. Add the prefix` jfr` to `delay` and `duration` making them `jfr-delay` and `jrf-duration` respectively.
2. Leave `recording-template` as is.
3. Create a separate issue for `jfr-recording-template`

dliappis:
> Shall I proceed with the changes? Additionally just wanted to know if there should be separate commits or a single squashed commit?

Feel free to proceed based on the feedback of @pquentin and mine. No particular preferences for the commits as anyway when it's time to merge we'll be squashing.

dliappis:

esrally/telemetry.py
debanjanc01: This has been updated in the latest commit.

esrally/telemetry.py
debanjanc01: I've pushed the changes in the latest commit. Additionally https://github.com/elastic/rally/issues/1631 has been raised for the `jfr-recording-template` migration.

docs/telemetry.rst
```diff
@@ -60,6 +60,8 @@ To enable ``jfr``, invoke Rally with ``esrally race --track=pmc --telemetry jfr`
 Supported telemetry parameters:
 
 * ``recording-template``: The name of a custom flight recording template. It is up to you to correctly install these recording templates on each target machine. If none is specified, the default recording template of Java flight recorder is used.
+* ``jfr-delay``: (Optional) Length of time to wait before starting to record (INTEGER followed by 's' for seconds 'm' for minutes or 'h' for hours, 0s). It is up to you to correctly pass in the value.
+* ``jfr-duration``: (Optional) Length of time to record. Note that 0s means forever (INTEGER followed by 's' for seconds 'm' for minutes or 'h' for hours, 0s). It is up to you to correctly pass in the value.
```
pquentin, line 64: Why do we we include `, 0s` at the end? I don't understand what this means.

pquentin (commented):
This looks great! We only need to test that it actually works. Have you tried?

Also left a small question.

docs/telemetry.rst
debanjanc01: 0s is default value for both the parameters. I've updated the documentation to reflect this properly.

dliappis:

debanjanc01:
> This looks great! We only need to test that it actually works. Have you tried?
> 
> Also left a small question.

Although I have checked that the final java cmd is as expected, I didn't get the opportunity to actually test this out.

pquentin (approved):
Thanks! This works.

I tried `esrally race --distribution-version=8.5.1 --test-mode --track=geonames --telemetry jfr --telemetry-params="jfr-delay:10s,jfr-duration:10s"` and did get a 10 seconds record starting after 10 seconds.

image

We don't usually merge on Fridays, and tomorrow is a [Shut it Down day](https://www.elastic.co/blog/culture-shut-it-down-days-help-elasticians-regroup-recharge) at Elastic, so we'll merge on Monday.

debanjanc01:
Thanks for the quick reviews and suggestions @dliappis @pquentin !

PR #1628, opened 2022-11-21
Use 8.x as a metrics store in it tests

pquentin:

PR #1632, opened 2022-11-23
Bump test es docker container to 7.17

b-deam:
As noted in https://github.com/elastic/rally/pull/1628, bumps to a version that supports MacOS ARM.


pquentin (approved):
Thanks! LGTM.

PR #1633, opened 2022-11-23
Remove internal chart generator

ebadyano:


ebadyano:

pquentin (changes requested):
The failure seems related:

> 22:59:41 esrally: error: argument subcommand: invalid choice: 'generate' (choose from 'race', 'list', 'delete', 'info', 'create-track', 'compare', 'build', 'download', 'install', 'start', 'stop')

ebadyano:
> The failure seems related:
> 
> > 22:59:41 esrally: error: argument subcommand: invalid choice: 'generate' (choose from 'race', 'list', 'delete', 'info', 'create-track', 'compare', 'build', 'download', 'install', 'start', 'stop')

missed removing the generate test.

it/generate_test.py
```diff
@@ -1,35 +0,0 @@
...
-
-import os
-
-import it
-
-
-@it.rally_in_mem
-def test_track_info_with_challenge(cfg, tmp_path):
-    cwd = os.path.dirname(__file__)
-    chart_spec_path = os.path.join(cwd, "resources", "sample-race-config.json")
```
pquentin, line 26: Can you please remove this file from the resources directory?

pquentin (approved):
Thanks! This looks great. Left two nits. We don't need another round of review here.

Can you please replace "avoid breaking the chart generator" with "avoid breaking the charts" in esrally/driver/driver.py?

it/generate_test.py
ebadyano: good catch thanks!

PR #1638, opened 2022-12-05
Don't use --short when retrieving revisions from git

### LINKED ISSUE #1635: Don't git clone elasticsearch on target machines when installing an artifact that's been built on a separate machine
## Bug description

When using the [build](https://esrally.readthedocs.io/en/latest/command_line_reference.html?highlight=build#build) subcommand -- introduced in #1576 -- to build a certain ES revision from source on machine A, then after transferring the artifact to the right location (`~/.rally/benchmarks/distributions/src`) on a different machine B, trying to install ES with the [install](https://esrally.readthedocs.io/en/latest/command_line_reference.html?highlight=build#install) subcommand results in an unnecessary `git clone` of the ES repo on machine B. As Rally won't actually build ES from source on machine B, this is both unnecessary and time consuming.

## Reproduction

(following repro is on the same machine, for simplicity)

1. Build ES

   ```shell
   $ esrally build --revision=82aeb478dbf83d164473b161801417fc1d59060e

   [INFO] Creating installable binary from source files
   [INFO] Creating installable binary from source files
   {
   "elasticsearch": "/home/dl/.rally/benchmarks/distributions/src/elasticsearch-82aeb478dbf-linux-x86_64.tar.gz"
   }
   ```

2. Wipe away the src directory (if you are not going to transfer the artifact to another machine)

    `~/.rally/benchmarks/src $ rm -rf elasticsearch`

3. Run the `install` command:

    ```shell
    $ esrally install --revision=82aeb478dbf83d164473b161801417fc1d59060e --runtime-jdk=bundled --network-host=127.0.0.1 --http-port=39200 --node-name=es01 --master-nodes=es01 --seed-hosts=127.0.0.1 --cluster-name=es01

    {
    "installation-id": "03a7c86d-268b-4347-9a24-c69580aaa0e2"
    }

    ---------------------------------
    [INFO] SUCCESS (took 200 seconds)
    ---------------------------------
    ```

    Observe that the source got re-pulled with git:

    ```shell
    $ cd ~/.rally/benchmarks/src/
    $ du -sh elasticsearch/
    1.4G	elasticsearch/
    ```

    (you can also observe git running during the `esrally install` with `ps` or [execsnoop](https://github.com/iovisor/bcc/blob/master/tools/execsnoop.py))

## Technical notes/tips

There is a comparison done in https://github.com/elastic/rally/blob/5e13334d63b70cbf58d358332511cb04dd134912/esrally/mechanic/supplier.py#L357-L358 however, using the above reproduction command we can see that `self.file_name` is:

```
> /home/dl/source/elastic/rally/esrally/mechanic/supplier.py(359)fetch()
    358         ipdb.set_trace()
--> 359         maybe_an_artifact = os.path.join(self.distributions_root, self.file_name)
    360         if os.path.exists(maybe_an_artifact):

ipdb> self.distributions_root
'/home/dl/.rally/benchmarks/distributions/src'
ipdb> self.file_name
'elasticsearch-82aeb478dbf83d164473b161801417fc1d59060e-linux-x86_64.tar.gz'
```

whereas the actual file on disk is `~/.rally/benchmarks/distributions/src/elasticsearch-82aeb478dbf-linux-x86_64.tar.gz` (i.e. it uses a shorter SHA).

dliappis:
Rally uses the git revision in several places, one of which when building Elasticsearch from source. Historically it used `--short`[^1] which results in shorter filenames, but they are unpredictable and more importantly cause an unnecessary git clone when using the `build` and `install` subcommands separately as described in #1635.

In this commit we switch to the full SHA length for all cases.

Closes #1635

[^1]: https://git-scm.com/docs/git-rev-parse#Documentation/git-rev-parse
      .txt---shortlength

### DIFF
--- a/esrally/utils/git.py
+++ b/esrally/utils/git.py
@@ -120,7 +120,7 @@ def checkout_revision(src_dir, *, revision):
 
 @probed
 def head_revision(src_dir):
-    return process.run_subprocess_with_output("git -C {0} rev-parse --short HEAD".format(io.escape_path(src_dir)))[0].strip()
+    return process.run_subprocess_with_output("git -C {0} rev-parse HEAD".format(io.escape_path(src_dir)))[0].strip()
 
 
 @probed


dliappis:
Reviewers: I spent some time thinking about what test could be meaningful for catching the referenced issue, but I can't think of a unit test that would be suitable; after all, the issue is that without this change the built artifact includes a revision that is unpredictable and possibly different from the revision passed with `--revision`. An IT test using a defined revision SHA IMHO would be problematic as the way that ES gets built changes over time plus it'd add unnecessary time spent building with gradle, unless we'd mock it out.

I also considered rather than removing `--short` SHA, to make the check [more lenient](https://github.com/elastic/rally/blob/5e13334d63b70cbf58d358332511cb04dd134912/esrally/mechanic/supplier.py#L357); this would mean using `git` and therefore either cloning the repo again or using GitHub network lookups to determine whether the value passed with `--revision` and whatever revision value is part of the filename are the same (any of which could be a shortened version).
    Cons: apart from flaky network connections, it would require knowledge of the repo URL itself which we don't store atm in the rally-teams [vanicall/config.ini](https://github.com/elastic/rally-teams/blob/master/cars/v1/vanilla/config.ini) (the repo URL, if available, would be used for a new method to calculate the shortened revision -- e.g. see [here](https://github.com/elastic/rally/blob/5e13334d63b70cbf58d358332511cb04dd134912/esrally/mechanic/supplier.py)


Finally, this PR means that we should encourage users to pass the entire SHA when using `--revision`. Maybe we could strive to convert the shortened SHA to the full length SHA in a future PR?

michaelbaamonde (approved):
I think this is fine as-is. I appreciate the thought you put into possible test scenarios, but I agree that they're impractical.

Re: converting a short SHA to a full one in the future, it doesn't really seem worth it to me.

dliappis:
> Re: converting a short SHA to a full one in the future, it doesn't really seem worth it to me.

The motivation behind this was purely for user friendliness since some tools and UIs might emit a shortened SHA, which if passed to Rally would cause again the same unnecessary cloning described in the issue https://github.com/elastic/rally/issues/1635 description.

PR #1643, opened 2023-01-03
Skip response parsing for raw requests

danielmitterdorfer:
With this commit we skip response parsing for raw requests. As users might leverage the `raw-request` operation type for any kind of request, it is possible that it produces arbitrarily large responses, which might introduce client-side bottlenecks. As we don't inspect the response body, we might as well just use the raw response internally and avoid an accidental bottleneck due to response parsing.


michaelbaamonde (approved):
The reasoning makes sense to me. Thanks @danielmitterdorfer!

danielmitterdorfer:
Thanks!

PR #1645, opened 2023-01-04
Fix to-date option for list subcommand

dliappis:
PR#1597 introduced new additional filters for list races including `--to-date`, but it doesn't work as expected (gets ignored) as the name of the key is different to what it should really be (and what we are testing for).
This commit fixes the bug making `--to-date` applicable.

It might make sense to refactor those hardcoded keys in a future PR, so that they are referenced from the same location in `rally.py` and `metrics.py`.


michaelbaamonde (approved):
Good catch. LGTM.

PR #1654, opened 2023-01-17
Add role metadata to node-stats telemetry device

b-deam:
It's useful to know what roles a node was configured for when analysing metrics in a remote metrics store. 

For example:
image

I've tested this with a local 8.5 metrics store:
```shell
esrally race --track=geopoint  --pipeline=benchmark-only --target-hosts=https://localhost:9222 --challenge=append-fast-with-conflicts --telemetry=node-stats  --client-options="timeout:180,verify_certs:false,basic_auth_user:'elastic',basic_auth_password:'changeme'" --kill-running-processes --test-mode
```

### DIFF
--- a/esrally/actor.py
+++ b/esrally/actor.py
@@ -259,7 +259,6 @@ def bootstrap_actor_system(try_join=False, prefer_local_only=False, local_ip=Non
         if coordinator_ip:
             # Make the coordinator node the convention leader
             capabilities["Convention Address.IPv4"] = "%s:1900" % coordinator_ip
-        print(system_base, capabilities)
         logger.info("Starting actor system with system base [%s] and capabilities [%s].", system_base, capabilities)
         return thespian.actors.ActorSystem(system_base, logDefs=log.load_configuration(), capabilities=capabilities)
     except thespian.actors.ActorSystemException:
--- a/esrally/actor.py
+++ b/esrally/actor.py
@@ -259,6 +259,7 @@ def bootstrap_actor_system(try_join=False, prefer_local_only=False, local_ip=Non
         if coordinator_ip:
             # Make the coordinator node the convention leader
             capabilities["Convention Address.IPv4"] = "%s:1900" % coordinator_ip
+        print(system_base, capabilities)
         logger.info("Starting actor system with system base [%s] and capabilities [%s].", system_base, capabilities)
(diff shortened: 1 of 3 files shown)


b-deam:

tests/telemetry_test.py
```diff
@@ -2146,7 +2146,8 @@ def test_stores_default_nodes_stats(self, metrics_store_put_doc):
         cfg = create_config()
         metrics_store = metrics.EsMetricsStore(cfg)
         node_name = [self.node_stats_response["nodes"][node]["name"] for node in self.node_stats_response["nodes"]][0]
-        metrics_store_meta_data = {"cluster": "remote", "node_name": node_name}
+        roles = [self.node_stats_response["nodes"][node]["roles"] for node in self.node_stats_response["nodes"]][0]
+        metrics_store_meta_data = {"cluster": "remote", "node_name": node_name, "roles": roles}
```
pquentin, line 2150: Would you mind moving this line after the call to recorder.record() to clarify that it does not affect the call but is just expected data? (This applies to the last three snippets too.)

pquentin (approved):
Thanks! LGTM. I added a small nit that is optional and does not require another round of review if you do implement it.

tests/telemetry_test.py
b-deam: No problem! Address in https://github.com/elastic/rally/pull/1654/commits/9c64b3ddc82ebc8c09bed05481290da2820009a3.

b-deam:
> AssertionError: It was not possible to build the docker image from Dockerfile-dev

@elasticmachine run rally/it-python310

esrally/actor.py
```diff
@@ -259,6 +259,7 @@ def bootstrap_actor_system(try_join=False, prefer_local_only=False, local_ip=Non
         if coordinator_ip:
             # Make the coordinator node the convention leader
             capabilities["Convention Address.IPv4"] = "%s:1900" % coordinator_ip
+        print(system_base, capabilities)
```
pquentin, line 262: leftover print

b-deam: 🤦

b-deam:
I raised https://github.com/elastic/rally/issues/1655 to track the CI failure, since it's not the first time we've seen it.

pquentin (approved):
Thanks for iterating! LGTM.

PR #1657, opened 2023-01-26
Cleanup params_test.py file

pquentin:
I tried to make the formatting consistent. I also compare full structures which works well in pytest because if a single item is wrong it shows a diff.


b-deam (approved):
LGTM! 

I really like this change, much more readable. Also, the diff errors are _much_ clearer.

PR #1658, opened 2023-01-26
Reduce the size of the readers list in create_readers

pquentin:
We have one BulkIndexParamSource per client, and if you look at how we fill the readers list, it's only using  `group` and `entry`. This won't have any effect in practice, the readers list will just contain less `None` entries.


pquentin:
Closing in favor of #1659 

I don't think the rally-tracks failure is related, for the record.

PR #1659, opened 2023-01-26
Simplify create_readers

pquentin:
This builds upon #1657. The second commit shows a bug in the current implementation of create_readers. It was hidden by the fact we multiply by num_clients which is often more than 1 in multi-corpora implementations.

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -963,32 +963,32 @@ def create_readers(
             offset, num_docs, num_lines = bounds(
                 docs.number_of_documents, start_client_index, end_client_index, num_clients, docs.includes_action_and_meta_data
             )
-            if num_docs > 0:
-                target = f"{docs.target_index}/{docs.target_type}" if docs.target_index else "/"
-                if docs.target_data_stream:
-                    target = docs.target_data_stream
-                logger.debug(
-                    "Task-relative clients at index [%d-%d] will bulk index [%d] docs starting from line offset [%d] for [%s] "
-                    "from corpus [%s].",
-                    start_client_index,
-                    end_client_index,
-                    num_docs,
-                    offset,
-                    target,
-                    corpus.name,
-                )
-                reader = create_reader(
-                    docs, offset, num_lines, num_docs, batch_size, bulk_size, id_conflicts, conflict_probability, on_conflict, recency
-                )
-                readers[-1].append(reader)
-                total_readers += 1
-            else:
+            if num_docs == 0:
                 logger.debug(
                     "Task-relative clients at index [%d-%d] skip [%s] (no documents to read).",
                     start_client_index,
                     end_client_index,
                     corpus.name,
                 )
+
+            target = f"{docs.target_index}/{docs.target_type}" if docs.target_index else "/"
+            if docs.target_data_stream:
+                target = docs.target_data_stream
+            logger.debug(
+                "Task-relative clients at index [%d-%d] will bulk index [%d] docs starting from line offset [%d] for [%s] "
+                "from corpus [%s].",
+                start_client_index,
+                end_client_index,
+                num_docs,
+                offset,
+                target,
+                corpus.name,
+            )
+            reader = create_reader(
+                docs, offset, num_lines, num_docs, batch_size, bulk_size, id_conflicts, conflict_probability, on_conflict, recency
+            )
+            readers[-1].append(reader)
+            total_readers += 1
 
     # Instead of reading all files from the first corpus, and then all files from the
     # second corpus, etc. we want to read the first file of all corpora, then the second
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -970,6 +970,7 @@ def create_readers(
                     end_client_index,
                     corpus.name,
                 )
+                continue
 
             target = f"{docs.target_index}/{docs.target_type}" if docs.target_index else "/"
             if docs.target_data_stream:
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -957,7 +957,7 @@ def create_readers(
     # stagger which corpus each client starts with for better parallelism
     reordered_corpora = corpora[start_client_index:] + corpora[:start_client_index]
     for corpus in reordered_corpora:
-        readers.append(collections.deque())
+        reader_queue = collections.dequeue()
         for docs in corpus.documents:
             offset, num_docs, num_lines = bounds(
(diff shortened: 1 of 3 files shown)


esrally/track/params.py
```diff
@@ -953,39 +953,53 @@ def create_readers(
...
-                    docs, offset, num_lines, num_docs, batch_size, bulk_size, id_conflicts, conflict_probability, on_conflict, recency
-                )
-            else:
+            if num_docs == 0:
                 logger.debug(
                     "Task-relative clients at index [%d-%d] skip [%s] (no documents to read).",
                     start_client_index,
                     end_client_index,
                     corpus.name,
                 )
```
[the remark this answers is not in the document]
pquentin: Good catch, thanks. Fixed in [`a9c0ce7` (#1659)](https://github.com/elastic/rally/pull/1659/commits/a9c0ce76e2d2b5945d9a5e0d79bccbe0ac5874b3) and added a test.

esrally/track/params.py
```diff
@@ -953,39 +953,54 @@ def create_readers(
...
+            readers[-1].append(reader)
+            total_readers += 1
+
+    # Instead of reading all files from the first corpus, and then all files from the
+    # second corpus, etc. we want to read the first file of all corpora, then the second
+    # file, etc.
+    staggered_readers = []
+    while total_readers > 0:
+        for reader_queue in readers:
+            if reader_queue:
```
danielmitterdorfer, line 1000: I *think* this is only needed because we eagerly add the dequeue to `readers`? If we'd only add it when there are actually documents that need to be read, we could get rid of this if.

esrally/track/params.py
```diff
@@ -953,39 +953,54 @@ def create_readers(
...
+                end_client_index,
+                num_docs,
+                offset,
+                target,
+                corpus.name,
+            )
+            reader = create_reader(
+                docs, offset, num_lines, num_docs, batch_size, bulk_size, id_conflicts, conflict_probability, on_conflict, recency
+            )
+            readers[-1].append(reader)
```
danielmitterdorfer, line 991: How about we assign the dequeue to a variable? This might be easier to read?

esrally/track/params.py
```diff
@@ -953,39 +953,54 @@ def create_readers(
...
                 logger.debug(
                     "Task-relative clients at index [%d-%d] skip [%s] (no documents to read).",
                     start_client_index,
                     end_client_index,
                     corpus.name,
                 )
-    return readers
+                continue
+
+            target = f"{docs.target_index}/{docs.target_type}" if docs.target_index else "/"
```
danielmitterdorfer, line 975: This and the following two lines add a bit of noise in a quite complicated function. Maybe we can help the reader by moving this into a separate function?

danielmitterdorfer (commented):
I only wanted to leave a few high-level comments: For better or worse, bulk-indexing is among the most complex parts of Rally so IMHO it pays off to be a bit more explicit than usual when doing changes. I'm thinking specifically of:

* Adding more comments on the algorithm in this function. What is it doing on a high-level?
* Simplifying the key functionality as much as possible, e.g. do we need the debug log statements?
* Simplifying control structures, e.g. can we reimplement this so we don't need a `continue` statement?
* Can we pick even more descriptive variable names (e.g. what differentiates `readers` from `staggered_readers`? The similar name implies e.g. similarity in data structure which is not the case)?

esrally/track/params.py
pquentin: No, actually this is needed because all queues don't necessarily have the same length. Say I have the following readers list of deques: [[doc1], [doc1, doc2]] and total_reads = 3. After one iteration I will have the following readers: [[], [doc2]] and total_reads = 1. I need to skip the empty deque.

I tried to explain this in [`8f05ed2` (#1659)](https://github.com/elastic/rally/pull/1659/commits/8f05ed285622ebc96940753e07ec418890ca16b0).

esrally/track/params.py
pquentin: Done in [`6870764` (#1659)](https://github.com/elastic/rally/pull/1659/commits/6870764ad9453f5ec8e54265bcc243c6ee4fafa2)

esrally/track/params.py
pquentin: So this was only needed for the debug logs, which gave me one more reason to remove them, which I have done in [`a26e18c` (#1659)](https://github.com/elastic/rally/pull/1659/commits/a26e18c84092601c1f116bbbc11b05a81a9585ce)

pquentin:
> * Adding more comments on the algorithm in this function. What is it doing on a high-level?

Done in [`8f05ed2` (#1659)](https://github.com/elastic/rally/pull/1659/commits/8f05ed285622ebc96940753e07ec418890ca16b0). This helped me find a bug which I fixed in [`dca5271` (#1659)](https://github.com/elastic/rally/pull/1659/commits/dca5271bfa7f31c51d1d2cf97044273136b0cac9).

> * Reducing noise as much as possible, e.g. do we need the debug log statements?

As mentioned above, this is done in [`a26e18c` (#1659)](https://github.com/elastic/rally/pull/1659/commits/a26e18c84092601c1f116bbbc11b05a81a9585ce).

> * Simplifying control structures, e.g. can we reimplement this so we don't need a `continue` statement?

Done in [`dcee647` (#1659)](https://github.com/elastic/rally/pull/1659/commits/dcee6470cd79f00b500a2ef6bfe599b378bcb919). This was much easier with the debug logs gone!
 
> * Can we pick even more descriptive variable names (e.g. what differentiates `readers` from `staggered_readers`? The similar name implies e.g. similarity in data structure which is not the case)?

Done in [`8f05ed2` (#1659)](https://github.com/elastic/rally/pull/1659/commits/8f05ed285622ebc96940753e07ec418890ca16b0). I also added types in [`6e35a40` (#1659)](https://github.com/elastic/rally/pull/1659/commits/6e35a40fcb85d89083d632cc4610c29c1f91c9f8) to clarify further the contents of the variables.

Can you please take another look?

pquentin:
> This LGTM, though I haven't done any local testing, and I'm temporarily without GH access. I can circle back in a few days if you'd like

@DJRickyB I appreciate the offer but you don't have to do this now, you know :)

pquentin:
@elasticmachine run rally/it-python38 please

pquentin:
Closing/reopening to get a RtD status.

pquentin:
@elasticmachine run rally/rally-tracks-compat please now that https://github.com/elastic/rally-tracks/pull/375 was merged.

PR #1662, opened 2023-02-01
Enable pyupgrade pre-commit check

pquentin:
[pyupgrade](https://github.com/asottile/pyupgrade) is a tool that upgrades code bases to newer versions of the Python syntax, which evolves with every minor version. The main thing it does in practice for Rally is to migrate % formats to f-strings, which are faster. However, pyupgrade by default upgrades some % formats to `str.format` calls when the expressions used are too complex and could harm readability. We don't want to do that as we have some highly optimized code and `str.format` is slower than % format:

```
∮ python -m timeit '"{} {} {}".format("foo", "bar", "baz")'
1000000 loops, best of 5: 231 nsec per loop
∮ python -m timeit '"%s %s %s" % ("foo", "bar", "baz")'
2000000 loops, best of 5: 148 nsec per loop
```

Thankfully by specifying `--keep-percent-format` we can keep % formats when they would have been migrated to `str.format`.

And the real reason I'm opening this pull request is that I needed one to test the Read the Docs integration. But it will be nice to stop having to think about f-strings in reviews.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,15 @@ repos:
     hooks:
       - id: isort
 
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
+    hooks:
+      - id: pyupgrade
+        args: [
+          "--py38-plus",
+          "--keep-percent-format"
+        ]
+
   - repo: local
     hooks:
       - id: pylint
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,13 @@
 repos:
+  - repo: https://github.com/asottile/pyupgrade
+    rev: v3.3.1
+    hooks:
+      - id: pyupgrade
+        args: [
+          "--py38-plus",
+          "--keep-percent-format"
+        ]
+
   - repo: https://github.com/psf/black
     rev: 22.3.0
     hooks:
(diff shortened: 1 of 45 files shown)


dliappis:
> ∮ python -m timeit '"{} {} {}".format("foo", "bar", "baz")'
> 1000000 loops, best of 5: 231 nsec per loop
> ∮ python -m timeit '"%s %s %s" % ("foo", "bar", "baz")'
> 2000000 loops, best of 5: 148 nsec per loop

The first example doesn't test the performance of f-string, bur rather the `string.format()` method which AFAIK is the slowest. This [link](https://www.reddit.com/r/Python/comments/pivojb/performance_comparison_of_string_formatting/?utm_source=share&utm_medium=web2x&context=3) has some benchmarks from all methods demonstrating that f-strings are the fastest.

pquentin:
Yes. The thing is pyupgrade by default will upgrade either to str.format or to f-strings, depending on the complexity of the variables declarations. Per https://github.com/asottile/pyupgrade#f-strings:

> note: pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability).

So without that option, sometimes % formats will become str.format:

```diff
-    print("* [#%s](%s)%s: %s" % (str(issue.number), issue.html_url, breaking_hint, issue.title))
+    print("* [#{}]({}){}: {}".format(str(issue.number), issue.html_url, breaking_hint, issue.title))
```

PR #1664, opened 2023-02-06
Fix precommit

pquentin:
pyupgrade noticed that with `from __future__ import annotations` I no longer needed to import Typing, List etc. so it made the change. But then pylint was not happy because I was not using an import.


dliappis (approved):
LGTM thanks for the quick turnaround

PR #1667, opened 2023-02-09
Fix git tests with sparse Rally git checkouts

pquentin:
The previous git tests relied on the Rally git checkout, but:

 * GitHub Actions does not checkout the whole history to make the git clone faster.
 * Using a dummy repo makes those tests 10x faster , with `pytest tests/utils/git_test.py` now taking 0.40s instead of 5.98s on my laptop.

I've also started using GitPython in tests which makes the setup/teardown easier to write and read in my opinion.

### DIFF
--- a/tests/utils/git_test.py
+++ b/tests/utils/git_test.py
@@ -28,6 +28,7 @@ from esrally.utils import git
 
 def commit(repo, *, date=None):
     file_name = os.path.join(repo.working_dir, "test")
+    # creates an empty file
     with open(file_name, "wb"):
         pass
     repo.index.add([file_name])
--- a/tests/utils/git_test.py
+++ b/tests/utils/git_test.py
@@ -49,14 +49,23 @@ def setup(request, tmp_path_factory):
     cls.tmp_clone_dir = str(tmp_path_factory.mktemp("rally-unit-test-clone-dir"))
 
     # create tmp git repos
-    cls.local_repo = Repo.init(cls.local_tmp_src_dir)
+    # Some recent git versions default to `main` but old versions don't accept
+    # --initial-branch. Until we migrate off Jenkins, let's default recent versions to
+    # master too.
+    try:
+        cls.local_repo = Repo.init(cls.local_tmp_src_dir, initial_branch="master")
+    except:
+        cls.local_repo = Repo.init(cls.local_tmp_src_dir)
     commit(cls.local_repo)
     cls.local_revision = cls.local_repo.heads["master"].commit.hexsha
     cls.local_repo.create_tag("local-tag-1", "HEAD")
     cls.local_repo.create_tag("local-tag-2", "HEAD")
     cls.local_repo.create_head(cls.local_branch, "HEAD")
 
-    cls.remote_repo = Repo.init(cls.remote_tmp_src_dir)
+    try:
+        cls.remote_repo = Repo.init(cls.remote_tmp_src_dir, initial_branch="master")
+    except Exception:
+        cls.remote_repo = Repo.init(cls.remote_tmp_src_dir)
     commit(cls.remote_repo, date="2016-01-01 00:00:00+0000")
     cls.old_revision = cls.remote_repo.heads["master"].commit.hexsha
     commit(cls.remote_repo)
--- a/tests/utils/git_test.py
+++ b/tests/utils/git_test.py
@@ -54,7 +54,7 @@ def setup(request, tmp_path_factory):
     # master too.
     try:
         cls.local_repo = Repo.init(cls.local_tmp_src_dir, initial_branch="master")
-    except:
+    except Exception:
         cls.local_repo = Repo.init(cls.local_tmp_src_dir)
     commit(cls.local_repo)
     cls.local_revision = cls.local_repo.heads["master"].commit.hexsha
--- a/tests/utils/git_test.py
+++ b/tests/utils/git_test.py
@@ -17,88 +17,76 @@
 
 import logging
 import os
-import shutil
 from unittest import mock
 
 import pytest
+from git import Repo
 
 from esrally import exceptions
-from esrally.utils import git, process
+from esrally.utils import git
+
+
+def commit(repo, *, date=None):
+    file_name = os.path.join(repo.working_dir, "test")
+    with open(file_name, "wb"):
+        pass
+    repo.index.add([file_name])
+    repo.index.commit("initial commit", commit_date=date)
 
 
 @pytest.fixture(scope="class", autouse=True)
 def setup(request, tmp_path_factory):
-    request.cls.remote_repo = "esrally"
-    request.cls.local_branch = "rally-unit-test-local-only-branch"
-    request.cls.remote_branch = "rally-unit-test-remote-only-branch"
-    request.cls.rebase_branch = "rally-unit-test-rebase-branch"
+    cls = request.cls
 
-    # this is assuming that nobody stripped the git repo info in their Rally working copy
-    request.cls.original_src_dir = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
-    request.cls.local_tmp_src_dir = str(tmp_path_factory.mktemp("rally-unit-test-local-dir"))
-    request.cls.remote_tmp_src_dir = str(tmp_path_factory.mktemp("rally-unit-test-remote-dir"))
-    request.cls.tmp_clone_dir = str(tmp_path_factory.mktemp("rally-unit-test-clone-dir"))
+    cls.local_remote = "remote_repo"
+    cls.local_branch = "rally-unit-test-local-only-branch"
+    cls.remote_branch = "rally-unit-test-remote-only-branch"
+    cls.rebase_branch = "rally-unit-test-rebase-branch"
+
+    cls.local_tmp_src_dir = str(tmp_path_factory.mktemp("rally-unit-test-local-dir"))
+    cls.remote_tmp_src_dir = str(tmp_path_factory.mktemp("rally-unit-test-remote-dir"))
+    cls.tmp_clone_dir = str(tmp_path_factory.mktemp("rally-unit-test-clone-dir"))
 
     # create tmp git repos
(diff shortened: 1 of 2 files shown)


pquentin:
Annoyingly Jenkins still uses Git 2.17 from 2019 which does not support `git init --initial-branch`. GitHub Actions [comes with Git 2.39.1 instead](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#tools), released a few days ago.

pquentin:
I fixed it by using "master" for now. According to `git init` output it's not safe to rely on this but it will do until the migration off Jenkins.

b-deam:
One high level comment before I review it is that the [GitPython project is in maintenance mode](https://github.com/gitpython-developers/GitPython#development-status):
> This project is in maintenance mode, which means that
>
> …there will be no feature development, unless these are contributed
> …there will be no bug fixes, unless they are relevant to the safety of users, or contributed
> …issues will be responded to with waiting times of up to a month

Given we're only using this in tests, I don't think it's a big deal, but I wanted to call it out.

tests/utils/git_test.py
```diff
@@ -17,88 +17,76 @@
...
+from git import Repo
 
 from esrally import exceptions
-from esrally.utils import git, process
+from esrally.utils import git
+
+
+def commit(repo, *, date=None):
+    file_name = os.path.join(repo.working_dir, "test")
+    with open(file_name, "wb"):
```
b-deam, line 31: Maybe just me, but I was a little confused at first.  
```suggestion
    file_name = os.path.join(repo.working_dir, "test")
    # creates an empty file
    with open(file_name, "wb"):
```

tests/utils/git_test.py
```diff
@@ -17,88 +17,76 @@
...
 
 @pytest.fixture(scope="class", autouse=True)
 def setup(request, tmp_path_factory):
-    request.cls.remote_repo = "esrally"
-    request.cls.local_branch = "rally-unit-test-local-only-branch"
-    request.cls.remote_branch = "rally-unit-test-remote-only-branch"
-    request.cls.rebase_branch = "rally-unit-test-rebase-branch"
+    cls = request.cls
+
+    cls.local_remote = "remote_repo"
```
b-deam, line 41: Nit, but I found the name confusing.
```suggestion
    cls.local_remote_name = "remote_repo"
```

b-deam (changes requested):
Thanks for this, I agree that the readability is significantly improved. 

Besides my comment on the dependency's maintenance status, this doesn't work in its current form on my Mac because my version of Git (`git version 2.37.1 (Apple Git-137.1)`) defaults to creating `main`.

I did get tests passing with something like this:
```diff
diff --git a/tests/utils/git_test.py b/tests/utils/git_test.py
index f161b5d36f887c66221a6b1a4f729749ddcb710a..0c2d93bde4c1fec0e5ec02cf9cd53836c5e97324 100644
--- a/tests/utils/git_test.py
+++ b/tests/utils/git_test.py
@@ -34,6 +34,14 @@ def commit(repo, *, date=None):
     repo.index.commit("initial commit", commit_date=date)
 
 
+def main_or_master(repo):
+    # default init branch can differ between Git versions
+    if "main" in repo.heads:
+        return "main"
+    else:
+        return "master"
+
+
 @pytest.fixture(scope="class", autouse=True)
 def setup(request, tmp_path_factory):
     cls = request.cls
@@ -50,16 +58,17 @@ def setup(request, tmp_path_factory):
     # create tmp git repos
     cls.local_repo = Repo.init(cls.local_tmp_src_dir)
     commit(cls.local_repo)
-    cls.local_revision = cls.local_repo.heads["master"].commit.hexsha
+    cls.main = main_or_master(cls.local_repo)
+    cls.local_revision = cls.local_repo.heads[cls.main].commit.hexsha
     cls.local_repo.create_tag("local-tag-1", "HEAD")
     cls.local_repo.create_tag("local-tag-2", "HEAD")
     cls.local_repo.create_head(cls.local_branch, "HEAD")
 
     cls.remote_repo = Repo.init(cls.remote_tmp_src_dir)
     commit(cls.remote_repo, date="2016-01-01 00:00:00+0000")
-    cls.old_revision = cls.remote_repo.heads["master"].commit.hexsha
+    cls.old_revision = cls.remote_repo.heads[cls.main].commit.hexsha
     commit(cls.remote_repo)
-    cls.remote_branch_hash = cls.remote_repo.heads["master"].commit.hexsha
+    cls.remote_branch_hash = cls.remote_repo.heads[cls.main].commit.hexsha
 
     cls.remote_repo.create_head(cls.remote_branch, "HEAD")
     cls.local_repo.create_remote(cls.local_remote, cls.remote_tmp_src_dir)
@@ -81,8 +90,8 @@ class TestGit:
         # undo rebase
         self.local_repo.head.reset("ORIG_HEAD", hard=True)
         # checkout starting branches
-        self.local_repo.head.reference = "master"
-        self.remote_repo.head.reference = "master"
+        self.local_repo.head.reference = self.main
+        self.remote_repo.head.reference = self.main
 
         # delete branches
         self.local_repo.delete_head(self.rebase_branch, force=True)
@@ -101,7 +110,7 @@ class TestGit:
         assert git.is_branch(self.local_tmp_src_dir, identifier=self.local_branch)
 
         # both remote, and local
-        assert git.is_branch(self.local_tmp_src_dir, identifier="master")
+        assert git.is_branch(self.local_tmp_src_dir, identifier=self.main)
 
     def test_is_not_branch_tags(self):
(.venv) bradleydeam in ~/perf/github.com/b-deam/rally on branch pr/1667 $ git diff > patch
(.venv) bradleydeam in ~/perf/github.com/b-deam/rally on branch pr/1667 $ cat patch 
diff --git a/tests/utils/git_test.py b/tests/utils/git_test.py
index f161b5d36f887c66221a6b1a4f729749ddcb710a..0c2d93bde4c1fec0e5ec02cf9cd53836c5e97324 100644
--- a/tests/utils/git_test.py
+++ b/tests/utils/git_test.py
@@ -34,6 +34,14 @@ def commit(repo, *, date=None):
     repo.index.commit("initial commit", commit_date=date)
 
 
+def main_or_master(repo):
+    # default init branch can differ between Git versions
+    if "main" in repo.heads:
+        return "main"
+    else:
+        return "master"
+
+
 @pytest.fixture(scope="class", autouse=True)
 def setup(request, tmp_path_factory):
     cls = request.cls
@@ -50,16 +58,17 @@ def setup(request, tmp_path_factory):
     # create tmp git repos
     cls.local_repo = Repo.init(cls.local_tmp_src_dir)
     commit(cls.local_repo)
-    cls.local_revision = cls.local_repo.heads["master"].commit.hexsha
+    cls.main = main_or_master(cls.local_repo)
+    cls.local_revision = cls.local_repo.heads[cls.main].commit.hexsha
     cls.local_repo.create_tag("local-tag-1", "HEAD")
     cls.local_repo.create_tag("local-tag-2", "HEAD")
     cls.local_repo.create_head(cls.local_branch, "HEAD")
 
     cls.remote_repo = Repo.init(cls.remote_tmp_src_dir)
     commit(cls.remote_repo, date="2016-01-01 00:00:00+0000")
-    cls.old_revision = cls.remote_repo.heads["master"].commit.hexsha
+    cls.old_revision = cls.remote_repo.heads[cls.main].commit.hexsha
     commit(cls.remote_repo)
-    cls.remote_branch_hash = cls.remote_repo.heads["master"].commit.hexsha
+    cls.remote_branch_hash = cls.remote_repo.heads[cls.main].commit.hexsha
 
     cls.remote_repo.create_head(cls.remote_branch, "HEAD")
     cls.local_repo.create_remote(cls.local_remote, cls.remote_tmp_src_dir)
@@ -81,8 +90,8 @@ class TestGit:
         # undo rebase
         self.local_repo.head.reset("ORIG_HEAD", hard=True)
         # checkout starting branches
-        self.local_repo.head.reference = "master"
-        self.remote_repo.head.reference = "master"
+        self.local_repo.head.reference = self.main
+        self.remote_repo.head.reference = self.main
 
         # delete branches
         self.local_repo.delete_head(self.rebase_branch, force=True)
@@ -101,7 +110,7 @@ class TestGit:
         assert git.is_branch(self.local_tmp_src_dir, identifier=self.local_branch)
 
         # both remote, and local
-        assert git.is_branch(self.local_tmp_src_dir, identifier="master")
+        assert git.is_branch(self.local_tmp_src_dir, identifier=self.main)
 
     def test_is_not_branch_tags(self):
         assert not git.is_branch(self.local_tmp_src_dir, identifier="2.6.0")
@@ -170,7 +179,7 @@ class TestGit:
 
     def test_head_revision(self):
         # minimum 'core.abbrev' is to return 7 char prefixes
-        git.checkout(self.local_tmp_src_dir, branch="master")
+        git.checkout(self.local_tmp_src_dir, branch=self.main)
         assert git.head_revision(self.local_tmp_src_dir).startswith(self.local_revision[:7])
 
     def test_pull_ts(self):

```

pquentin:
> One high level comment before I review it is that the [GitPython project is in maintenance mode](https://github.com/gitpython-developers/GitPython#development-status)

Yes, I've noticed that. But then I looked at the commit activity and noticed that this was still a very active project: https://github.com/gitpython-developers/GitPython/commits/main, despite existing for years. It's true that the original author has mostly moved on, but I think using this project is safe.

pquentin:
> Besides my comment on the dependency's maintenance status, this doesn't work in its current form on my Mac because my version of Git (`git version 2.37.1 (Apple Git-137.1)`) defaults to creating `main`.

Oh, that's surprising. GitHub Actions and my Fedora laptop bot have git 2.39.1 and the default is still master (but with a warning that it may change in the future). Anyway, I've added [`cf2684f` (#1667)](https://github.com/elastic/rally/pull/1667/commits/cf2684f57f769df0a0bd73a981889a780fbdfd5c) as a workaround until we migrate to GitHub Actions and can just use `main` by using `--initial-branch`.

pquentin:
Can you please take another look and check that you can run git tests on your Mac?

b-deam (approved):
Thanks for iterating, this works for me now.

PR #1671, opened 2023-02-16
Run lint and tests on GitHub Actions

pquentin:
This is part of our migration away from Jenkins. GitHub Actions is:

* is free for open source repos,
* is what everybody uses in the Python community (which helps external contributors)
* and comes with a big ecosystem of custom actions that’s helpful to get things done, like easy ways to setup Python or the cache

See https://github.com/elastic/rally-tracks/pull/379 for a similar pull request.

The result is https://github.com/pquentin/rally/actions/runs/4191392426. We can't see it in this repo until we merge this.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+          cache: pip
+          cache-dependency-path: pyproject.toml
+      - name: "Install dependencies"
+        run: python -m pip install .[develop]
+      - uses: pre-commit/action@v3.0.0
```
dliappis, line 20: As I am not super familiar with GitHub actions, I have one general question; seems that we are referencing various versioned "plugins" like `pre-commit/action@v3.0.0` etc.

I assume that they get regular updates, maybe even have security issues, breaking with newer Python versions etc.

What is the workflow for keeping up to date with them? Should we just expect that when there is a problem the PR check will fail, or is there something more automated/proactive that can be used?

pyproject.toml
```diff
@@ -96,6 +96,7 @@ develop = [
     "pytest-asyncio==0.19.0",
     "pytest-httpserver==1.0.5",
     "tox==3.25.0",
+    "nox==2022.11.21",
```
dliappis, line 99: If we are introducing `nox`, can we sunset `tox`?

dliappis (commented):
Functionally looks good. Left some questions.

.github/workflows/ci.yml
pquentin: There are two scenarios when it's useful to upgrade:

 * When there's a security issue (but we're not doing anything sensitive and there's no credentials to steal and the actions has read-only permissions)
 * When the version we use is deprecated (it shows up in the summary of the build, so we'll end up seeing it, https://github.com/elastic/rally-tracks/pull/346 is a nice example here)

We can use dependabot (which is now owned by GitHub) and configure it to open pull requests in case of security issues. Otherwise there's not much value in upgrading. Upgrading for every change is going to add a lot of noise.

I'll look into configuring dependabot.

pyproject.toml
pquentin: Not yet, we still run IT and rally-tracks-compat through tox. It will definitely be sunset in a few PRs.

.github/workflows/ci.yml
pquentin: I enabled Dependabot security updates in the settings section of this repository:

image

Let's see if the defaults work for us. If not, we can configure them using a dependabot.yml file: https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates

PR #1674, opened 2023-02-22
Allow reading metrics password from environment

### LINKED ISSUE #1670: Allow .ini definable in environment variable or pass as command line runtime value instead of persisting it in .ini file
**Describe the feature**:

Currently, metrics store password has to be defined in the `.ini` file. This is fine, but some users would prefer not to persist the password in the .ini file, so either allowing the password to be configured as an environment variable or allowing it to be passed as a runtime flag would be helpful in this use case scenario. 

The workaround for now is to move the .ini file in the `/dev/shm` directory and symlink it to avoid persisting the value in the disk.  

Thank you for the consideration!

pquentin:
Closes #1670


dliappis (approved):
LGTM thanks for adding this flexibility

PR #1675, opened 2023-02-28
Declare Buildkite pipelines

pquentin:
The .buildkite directory is still empty as I still need to iterate on the pipelines themselves. I would like to do that on a branch if possible as the pipeline is a bit tricky with Python tests running various Docker commands so I can't use a Dockerfile as I don't think Docker-in-Docker is supported.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,58 @@
+# Declare a Backstage Component that represents the Rally application.
+---
+# yaml-language-server: $schema=https://json.schemastore.org/catalog-info.json
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: rally
+  description:  Macrobenchmarking framework for Elasticsearch
+  annotations:
+    backstage.io/source-location: url:https://github.com/elastic/rally/
+    github.com/project-slug: elastic/rally
+    github.com/team-slug: elastic/es-perf
+    buildkite.com/project-slug: elastic/rally
+  tags:
+    - elasticsearch
+    - benchmark
+    - python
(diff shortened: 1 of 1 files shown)


nkammah:
This looks good and passes the [Backstage validator](https://backstage.elastic.dev/entity-validation).
The only thing I would defer to @jarpy is around declaring the dependency on marbi.

catalog-info.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - performance
+  links:
+    - title: Rally docs
+      url: https://esrally.readthedocs.io/en/stable/
+spec:
+  type: application
+  owner: group:es-perf
+  lifecycle: production
+  dependsOn:
+    - resource:marbi-buildkite-pipeline
```
nkammah, line 27: How does your service depends on marbi?

nkammah (approved):

catalog-info.yaml
jarpy: My fault because I sent @pquentin to the Marbi catalog info in the absence of real documentation :)

jarpy (approved):

jarpy:
Ooops!

Sorry for merging. I thought I was in a different repo. 😳

jarpy:
> I don't think Docker-in-Docker is supported.

It's supported if you opt to run on VMs instead of containers, but we encourage everyone to use containers if possible, and we appreciate the effort.

pquentin:
Thanks for the reviews and merge! Appreciated. Most of our tests don't rely on Docker, but for those that do we'll use VMs.

PR #1677, opened 2023-03-01
Define Buildkite IT pipeline file

pquentin:
I'm not using the default pipeline location (.buildkite/pipeline.yml) but https://github.com/elastic/rally/blob/buildkite/.buildkite/it/pipeline.yml in the buildkite branch that I'm using to test things.


dliappis (approved):
LGTM

PR #1678, opened 2023-03-01
Migrate away from Jenkins

pquentin:
- The rally-tracks-compat tests are run with GitHub Actions, where they appear to run fine, taking around 40 minutes to finish, as in Jenkins today
- The IT tests are run with Buildkite as they are more resource intensive and did not even finish after 2h30mn with GitHub Actions. They seem to take 15mn longer in Buildkite despite using the exact same n1-standard-4 GCP worker (1h10 instead of 55mn) and avoiding compilation
- As part of the move, I moved from tox to nox which is simpler and less verbose
- I added a test to make sure that our recommended installation using pyenv still worked since we don't use that anymore in CI.

Here's a Rally build in Buildkite: https://buildkite.com/elastic/rally/builds/16. It's not available to external contributors yet but I'm working on it.


pquentin:
Closing in favor of #1678

PR #1679, opened 2023-03-02
Migrate from Jenkins to Buildkite and GitHub Actions

pquentin:
I'm reopening #1678 from a branch since this pull request enables builds from forks but it needs to be merged to work.

* The rally-tracks-compat tests are run with GitHub Actions, where they appear to run fine, taking around 40 minutes to finish, as in Jenkins today. I've tried to authenticate to avoid being rate limited by GitHub when fetching assets for the elastic/endpoint track.
* The IT tests are run with Buildkite as they are more resource intensive and did not even finish after 2h30mn with GitHub Actions. They seem to take 15mn longer in Buildkite despite using the exact same n1-standard-4 GCP worker than Jenkins (1h10 instead of 55mn) and avoiding compilation.
* As part of the move, I moved from tox to nox which is simpler and less verbose
* I added a test to make sure that our recommended installation using pyenv still worked since we don't use that anymore in CI.
* For external contributors, we'll have to click "Approve" for GitHub Actions to run and "buildkite build this" to run Buildkite jobs.


.ci/jobs/defaults.yml
```diff
@@ -47,21 +47,3 @@
     publishers:
       - email:
           recipients: infra-root+build@elastic.co
-      - junit:
```
dliappis, line 50: For my understanding, is this file remaining due to https://github.com/elastic/rally/blob/master/.ci/jobs/release-docker.yml job?

dliappis (commented):
I just :heart: the amount of deleted files here!

Looks great. I left some minor comments.

michaelbaamonde (approved):
LGTM!

.ci/jobs/defaults.yml
pquentin: Yes, exactly. I'm not sure what exactly was needed or not so I left most of it.

.buildkite/it/run.sh
```diff
@@ -0,0 +1,63 @@
...
+  # restore settings to fail immediately on error
+  set -o errexit
+  set -o pipefail
+  return 0
+}
+echo "--- System dependencies"
+
+PYTHON_VERSION="$1"
+retry 5 sudo add-apt-repository --yes ppa:deadsnakes/ppa
+retry 5 sudo apt-get update && sudo apt-get install -y \
```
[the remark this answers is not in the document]
pquentin: Done in [`eacd70d` (#1679)](https://github.com/elastic/rally/pull/1679/commits/eacd70dcd97290663b4acb32c36c18a88c5ddd21)

Issue #1644, opened 2023-01-04
Single-node clusters use non-trivial discovery config when they shouldn't


danielmitterdorfer:
**Rally version** (get with `esrally --version`): `esrally 2.7.1.dev0 (git revision: 4e1335ee07e16a4af0de1a0b543a86595ca37803)` (current `master` branch)

**Description of the problem including expected versus actual behavior**:

When running a benchmark with Rally against a single node cluster, the Elasticsearch logs contain repeated warning messages like the following:

```
[2023-01-04T09:39:10,994][WARN ][o.e.c.c.Coordinator      ] [rally-node-0] This node is a fully-formed single-node cluster with cluster UUID [o1zdVpJiSMOVfeSz9nye_A], but it is configured as if to discover other nodes and form a multi-node cluster via the [discovery.seed_hosts=[127.0.0.1]] setting. Fully-formed clusters do not attempt to discover other nodes, and nodes with different cluster UUIDs cannot belong to the same cluster. The cluster UUID persists across restarts and can only be changed by deleting the contents of the node's data path(s). Remove the discovery configuration to suppress this message.
```

This indicates that the discovery configuration is not correct, see https://github.com/elastic/elasticsearch/issues/85222 for full details.

**Steps to reproduce**:
 1. `esrally race --track=geonames --challenge=append-no-conflicts-index-only --distribution-version=8.5.3`
 2. Inspect the corresponding Elasticsearch log

This can be fixed by *not* rendering the following line in rally-teams when there is only one node:

https://github.com/elastic/rally-teams/blob/74f96e3fab247e0c31d253408e4affd696a50eff/cars/v1/vanilla/templates/config/elasticsearch.yml#L72

However, as we pass the node ips as a pre-rendered string in the provisioner we *cannot* use  Jinja's [`length`](https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.length) filter to determine the number of items (would return the string length instead):

https://github.com/elastic/rally/blob/4c7141aac0179c38c08c134c25f7a23e6f07d2d9/esrally/mechanic/provisioner.py#L320

We could instead either provide the node ips as list *or* instead add an additional variable in the provisioner that contains the node count, which can then be used in the template in rally-teams. The latter is even backwards-compatible if we check for the existence of the template variable before evaluating it.

inqueue:
Thanks, @danielmitterdorfer. I will submit a PR proposal soon.

PR #1680, opened 2023-03-02
Revert "Add a variable in the provisioner for the seed node count"

pquentin:
Reverts elastic/rally#1647

Knowing the number of nodes does not have any useful purpose: https://github.com/elastic/rally/issues/1644#issuecomment-1386985098


dliappis (approved):
LGTM

PR #1681, opened 2023-03-02
Exclude closed pull requests from changelog

pquentin:
Unfortunately we state in CHANGELOG.md on PyPy and GitHub that we support Python 3.11 and provide binaries using PyOxidizer. I fixed all the other places (Discuss, GitHub release, internal announcements).


dliappis (approved):
LGTM thanks for catching this

PR #1669, opened 2023-02-10
Upgrade ES Client to 8.x

b-deam:
Note for reviewers: 
- I have tested this locally against 6.8.23, 7.13 - 7.17, and 8.6.0
- I have tested with both local and remote metrics stores (including Elastic Cloud)
- The use of the `body` parameter in client methods is pervasive across the code base, luckily for us the 8.6 release has an internal decorator that rewrites these params and logs a deprecation warning
- The previously global `request_timeout` param is now a transport option that must be set via the client's respective `options()` method (e.g. `es.options(request_timeout=..).search(..)`). Again, luckily for us the client rewrites this parameter and sets the corresponding transport option and logs a deprecation warning

A high level overview of my plan:
- [X]  Be able to `esrally race [...]` against:
  - [X] 8.6.x
  - [X] 7.17.x
  - [X] 6.8.x
- [X] Get all unit tests passing
- [x] Get all integration tests passing
  - [x] Rally tracks PR: https://github.com/elastic/rally-tracks/pull/380
  - [x] Event data PR: https://github.com/elastic/rally-eventdata-track/pull/115
- [x] Ensure support for all existing `--client-options`
- [X] Refactor exception handling [based on changes in 8.x](https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/release-notes.html#_changed)  
- [X] Address the various `TODO`s in the `es-client-upgrade` branch
- [x] Upgrade the `elasticsearch[async]` and `elastic-transport` libs to `8.6.1` and `8.4.0`, respectively
- [x] Address any deprecation warnings 
  - [x] Raised https://github.com/elastic/rally/issues/1672 to track the move away from the `body` parameter to properly typed/named parameters
  - [x] Raisedhttps://github.com/elastic/rally/issues/1673  to track the refactoring of `request-timeout` 
- [x] Fix SSL context propagation, and subsequent logging configuration (we currently emit too many warnings about insecure settings) 
- [x] Execute various benchmarks to establish whether we're introducing any client side bottlenecks with the upgrade
  - [x] `elastic/logs` is a given, because of its pervasive nature in a lot of our work
  - [x] Open to suggestions on other specific/targeted benchmarks, otherwise I'll fallback on the default track selection (`geonames`, `nyc_taxis` etc.)

Closes https://github.com/elastic/rally/issues/1350

PR #1683, opened 2023-03-06
Upgrade Elasticsearch client to 8.6.1

b-deam:
Reverting the revert, also includes a keyword argument fix for the missed `get_transform_stats` operation in Telemetry.

Relates: https://github.com/elastic/rally/pull/1669


esrally/driver/runner.py
```diff
@@ -2142,13 +2081,13 @@ class RestoreSnapshot(Runner):
...
-        await es.snapshot.restore(
-            repository=mandatory(params, "repository", repr(self)),
-            snapshot=mandatory(params, "snapshot", repr(self)),
-            wait_for_completion=params.get("wait-for-completion", False),
-            **api_kwargs,
-        )
+        repo = mandatory(params, "repository", repr(self))
+        snapshot = mandatory(params, "snapshot", repr(self))
+
+        await es.perform_request(method="POST", path=f"/_snapshot/{repo}/{snapshot}/_restore", **api_kwargs)
```
dliappis, line 2090: Should we leave a comment about why we aren't using `es.snapshot.restore` along with a TODO to fix in the future as soon as the client supports the missing parameter?

dliappis (commented):
Thanks for following up. Left a small ask so that we can follow up/clean up in the future.

esrally/driver/runner.py
b-deam: Yes! I forgot to add it and pushed so I could test some esbench use-cases.

pquentin (approved):
:fire:

PR #1684, opened 2023-03-06
Revert "Sync main branch (#1676)"

pquentin:
This reverts commit 7a5bb09fe67037b86b9739aa12d6742c6c837a1a. This job currently fails due to missing permissions as I gave GitHub Actions read-only permissions to avoid security issues. Since we don't need a main branch, the simple fix here is to stop running this job on pushes to master and delete the main branch for now.


dliappis (approved):
LGTM

PR #1686, opened 2023-03-24
Use simply expanded vars assignment in Makefile

danielmitterdorfer:
With `make` version 4.4, execution of make targets hangs indefinitely. Running it with `--debug=verbose` shows:

```
Makefile:24: not recursively expanding PY38 to export to shell function
Makefile:24: not recursively expanding PY38 to export to shell function
Makefile:25: not recursively expanding PY39 to export to shell function
Makefile:24: not recursively expanding PY38 to export to shell function
Makefile:25: not recursively expanding PY39 to export to shell function
Makefile:25: not recursively expanding PY39 to export to shell function
...
```

This continues eternally. With this commit we switch to [simply expanded variable
assignment](https://www.gnu.org/software/make/manual/html_node/Simple-Assignment.html) avoiding this unnecessary recursion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,26 +15,26 @@
 # specific language governing permissions and limitations
 # under the License.
 
-SHELL = /bin/bash
+SHELL := /bin/bash
 # We assume an active virtualenv for development
-PYENV_REGEX = .pyenv/shims
-PY_BIN = python3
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
There is a failing `rally-tracks-compat` check but to me it looks completely unrelated to my change.

PR #1687, opened 2023-03-24
Propagate request start/end lazily

danielmitterdorfer:
With this commit we only propagate request start and end time to the parent context if there actually is one. Previously we unconditionally propagated the request context which could lead to misleading errors when there was an open request context but no request was issued. This can happen when a mandatory parameter is missing and the runner was throwing exception. That exception would get masked by a `KeyError` because no request start was present yet and thus obfuscate the root cause.

### DIFF
--- a/esrally/client/context.py
+++ b/esrally/client/context.py
@@ -37,21 +37,19 @@ class RequestContextManager:
 
     @property
     def request_start(self):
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Awesome, this should really help with debugging errors!

Issue #1688, opened 2023-03-24
rally-tracks-compat CI check is failing


danielmitterdorfer:
On two of my PRs the `rally-tracks-compat` check is failing, see

* #1687 
* #1686 

I inspected e.g. https://github.com/elastic/rally/actions/runs/4510307701/jobs/7941093272?pr=1687 and it seems there is a problem with installing Elasticsearch via Rally:

```
md = 'esrally install --quiet --http-port=19200 --node=rally-node --master-nodes=rally-node --car=4gheap,trial-license,x-pack-ml --seed-hosts="127.0.0.1:19300" --revision=current'

    def run_command_with_output(cmd):
        try:
>           proc = subprocess.run(shlex.split(cmd), text=True, capture_output=True, check=True)

.nox/rally_tracks_compat/lib/python3.10/site-packages/pytest_rally/process.py:32: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = None, capture_output = True, timeout = None, check = True
popenargs = (['esrally', 'install', '--quiet', '--http-port=19200', '--node=rally-node', '--master-nodes=rally-node', ...],)
kwargs = {'stderr': -1, 'stdout': -1, 'text': True}
process = <Popen: returncode: 64 args: ['esrally', 'install', '--quiet', '--http-port=...>
Error: = '[ERROR] Cannot install. substring not found.\n', stderr = ''
retcode = 64
```

(note that `substring not found`)

PR #1690, opened 2023-03-24
Use correct params for create component template

danielmitterdorfer:
The Elasticsearch Python client does not accept a request body but rather requires parts of the body to be passed for the create component template API call. With this commit we adapt Rally to these changes.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -1454,8 +1454,8 @@ class CreateComponentTemplate(Runner):
     async def __call__(self, es, params):
         templates = mandatory(params, "templates", self)
         request_params = mandatory(params, "request-params", self)
-        for template, body in templates:
-            await es.cluster.put_component_template(name=template, body=body, params=request_params)
+        for name, body in templates:
+            await es.cluster.put_component_template(name=name, template=body["template"], params=request_params)
         return {
             "weight": len(templates),
             "unit": "ops",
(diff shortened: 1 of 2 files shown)


b-deam (approved):
Thanks for picking this up. LGTM.

b-deam:
Hmm, something's up with `esrally install --quiet --http-port=19200 --node=rally-node --master-nodes=rally-node --car=4gheap,trial-license,x-pack-ml --seed-hosts="127.0.0.1:19300" --revision=current` as part of `rally-tracks-compat`.

I can't repro this locally, so I suspect something is up with GH Actions since GitHub updated their SSH key on Friday. [The blog post mentions](https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/):
> GitHub Actions users may see failed workflow runs if they are using actions/checkout with the ssh-key option. We are updating the actions/checkout action in all our supported tags, including @v2, @v3, and @main. If you pin the action to a commit SHA and use the ssh-key option, you’ll need to update your workflow. You can read more about this process [in our official documentation for Actions security hardening](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#reusing-third-party-workflows).

As far as I can tell, we don't use the `ssh-key` option, and my GH Actions-fu is non existent. @pquentin can you help here?

b-deam:
I see now that CI failing is related to this https://github.com/elastic/rally/issues/1688

@pquentin agreed to take a look tomorrow.

Issue #1673, opened 2023-02-21
Start using `.options()` for transport parameters 


b-deam:
As part of [the ES client upgrade to 8.x](https://github.com/elastic/rally/pull/1669) the client now requires what were [previously request level parameters to be set at the client's respective transport level](https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/migration.html#migration-options).

For example, perhaps the most pervasive change here is the previously global, per-request setting `request-timeout`:
```python
# 7.x DEPRECATED USAGE (Don't do this!):
client.search(index="blogs", request_timeout=180)

# 8.0+ SUPPORTED USAGE:
client.options(request_timeout=180).search(index="blogs")
```

We need to refactor all of our runners (and [custom runners in rally-tracks](https://github.com/elastic/rally-tracks)) to set these parameters at the transport level.

b-deam:
Mechanically this was going to be a reasonable amount of work on its own, but it turns out that this new approach incurs a relatively high client side overhead.

After merging https://github.com/elastic/rally/pull/1705 I noticed that some nightly benchmarks were displaying higher than normal `latency` (`latency = processing_time + service_time`). Some reproductions and profiling later and it's clear that the `.options()` approach is _slower_. 

The absolute numbers aren't large, we're talking only a few hundred microseconds to a millisecond on each request, but some of our benchmarks are sensitive enough to notice this.
![image](https://user-images.githubusercontent.com/54515790/235043325-7374167b-d13c-4f63-853c-d5d4b5b720de.png)


Since merging  https://github.com/elastic/rally/pull/1705 `.options()` that causes us to reinstantiate a new client on _every_ request: 

```diff
name                                                                                                                                          ncall     tsub      ttot      tavg      
/home/esbench/.pyenv/versions/3.11.2/lib/python3.11/asyncio/base_events.py:1845 _UnixSelectorEventLoop._run_once                              3976      0.072129  5.481492  0.001379
/home/esbench/.pyenv/versions/3.11.2/lib/python3.11/asyncio/events.py:78 TimerHandle._run                                                     3977      0.011262  5.254178  0.001321
/home/esbench/rally/esrally/driver/driver.py:1843 AsyncExecutor.__call__                                                                      1         0.078444  5.195176  5.195176
/home/esbench/rally/esrally/driver/driver.py:1796 AsyncProfiler.__call__                                                                      1         0.001048  5.187659  5.187659
[...]

PR #1691, opened 2023-03-24
Ignore HTTP 404 consistently in runners

danielmitterdorfer:
Some runners that deleted resources have ignored HTTP 404 already but we did not apply this consistently. With this commit, all runners that delete resources ignore HTTP 404 to avoid spurious request errors.


b-deam (approved):
Thanks for this. LGTM.

Note to self that this'll all have to be refactored as part of https://github.com/elastic/rally/issues/1673 anyway

PR #1698, opened 2023-03-31
Update progress after GCS download finished

danielmitterdorfer:
With this commit we ensure that the progress indicator is updated once again after a download from a GCS bucket has finished.


danielmitterdorfer:
I noticed this because on a new track that I've created I always saw this after the track data have been downloaded and I was puzzled why it wouldn't say 100%.

```
[INFO] Downloading track data (2.2 GB total size)                                 [ 98.9%]
```

With this change:

```
[INFO] Downloading track data (2.2 GB total size)                                [100.0%]
```

dliappis (approved):
LGTM thanks for catching this

PR #1702, opened 2023-04-12
Include 'refresh' and 'flush' stats by default when using 'node-stats-include-indices

b-deam:
Title. These stats are super useful and we should include them by default.

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -805,6 +805,8 @@ class NodeStatsRecorder:
                 "indexing",
                 "search",
                 "merges",
+                "refresh",
+                "flush",
                 "query_cache",
                 "fielddata",
                 "segments",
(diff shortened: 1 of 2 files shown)


dliappis (commented):
> Title. These stats are super useful and we should include them by default.

Any thoughts about the performance impact? We used to have issues in the past, see the warning [here](https://esrally.readthedocs.io/en/stable/telemetry.html?highlight=warning%20refresh#node-stats) and while they don't look like big structures, I am always concerned about what happens under the hood in ES when asking for extra things in node-stats.

b-deam:
> > Title. These stats are super useful and we should include them by default.
> 
> Any thoughts about the performance impact? We used to have issues in the past, see the warning [here](https://esrally.readthedocs.io/en/stable/telemetry.html?highlight=warning%20refresh#node-stats) and while they don't look like big structures, I am always concerned about what happens under the hood in ES when asking for extra things in node-stats.

We already retrieve the full node stats output (no pre-filtering):
https://github.com/elastic/rally/blob/c90169e93288892661d739f63fad9cf883206dd0/esrally/telemetry.py#L942

So the only extra overhead is the additional 9 keys in the respective sample's dict.

dliappis (approved):
LGTM

PR #1705, opened 2023-04-18
Refactor some runners to set transport level options

b-deam:
With this commit we fix a bug that prevented any runner using the low
level `perform_request()` ES client method, and `_transport_request_params()`
helper method from passing in the `request-timeout` parameter.

Relates https://github.com/elastic/rally/issues/1673 
Relates https://github.com/elastic/rally-tracks/issues/393

Note that there's still follow up work to be done to refactor the remaining runners, specifically around overriding the client's `options()` method to handle the `distribution_version` param to support REST compatibility headers as discussed in https://github.com/elastic/rally/issues/1673, but that will only affect us when 9.x is released.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -204,16 +204,29 @@ class Runner:
         # filter Nones
         return dict(filter(lambda kv: kv[1] is not None, full_result.items()))
 
-    def _transport_request_params(self, params):
+    @staticmethod
+    def _transport_request_params(params):
+        """
+        Takes all of a runner's params and splits out request parameters, transport
+        level parameters, and headers into their own respective dicts.
+
+        :param params: A hash with all the respective runner's parameters.
+        :return: A tuple of the specific runner's params, request level parameters, transport level parameters, and headers, respectively.
+        """
+        transport_params = {}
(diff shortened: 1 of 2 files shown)


pquentin (approved):
Thanks! Looks good to me.

esrally/driver/runner.py
```diff
@@ -204,16 +204,29 @@ def _default_kw_params(self, params):
...
-        opaque_id = params.get("opaque-id")
-        if opaque_id is not None:
+
+        if request_timeout := params.pop("request-timeout", None):
+            transport_params["request_timeout"] = request_timeout
+
+        if (ignore_status := request_params.pop("ignore", None)) or (ignore_status := params.pop("ignore", None)):
+            transport_params["ignore_status"] = ignore_status
+
+        headers = params.pop("headers", None) or {}
```
dliappis, line 225: For my own curiosity: there is a possibility that the headers key exists but is not a dict, hence we need to do the `or {}` instead of just `params.pop("headers", {})`?

dliappis (approved):
LGTM from me too, thanks for the quick turnaround.

Issue #1598, opened 2022-10-20
Rally's Elasticsearch clients are sticky to a single IP when multiple resolve from a DNS lookup


b-deam:
## Description 
At a high level, this is what happens when Rally's runners connect to Elasticsearch for the first time:
```
                 +---------------------------+
                 |   Runner.__call__()       |
                 +---------------------------+
                                |
                                |
                                v
         +----------------------------------------------+
         |   RallyAsyncElasticsearch.perform_request()  |
         +----------------------------------------------+
                                |
                                |
                                v
         +--------------------------------------------+
         |   RallyAsyncTransport.perform_request()    |
         +--------------------------------------------+
                                |
                                |
                                v
         +-----------------------------------------+
         |   RallyAiohttpHttpNode.perform_request()|
         +-----------------------------------------+
                                |
                                |
                                v
         +---------------------------------------+
         |   ClientSession.request()             |
         +---------------------------------------+
                                |
                                |
                                v
         +---------------------------------------+
         |   TCPConnector.connect()              |
         +---------------------------------------+

```

Once a `ClientSession` is created, a given `ClientRequest` initiates a `Connection` via `TCPConnector.connect`. This `Connection` is persisted and re-used for the life of the session unless it is closed (e.g. keep alive expires, or explicitly closed), and the `TCPConnector` uses the `ClientRequest.connection_key` property to decide whether or not a `Connection` can be re-used, or if a new one should be established. 

The actual DNS resolution call is made via `TCPConnector.connect() > TCPConnector._create_connection() > TCPConnector._create_direct_connection() > TCPConnector._resolve_host()`. The thing is that `_resolve_host()` does **acutally** return all the IPs a given name resolves to, but the underlying logic [only every selects the _first succesful_ host from this list from which to establish a connection](https://github.com/aio-libs/aiohttp/blob/2b3db3a07cd1d40af46408fdcf3fbc12f7188a3a/aiohttp/connector.py#L1178-L1211). To explain why this is problematic, let's look at a real world example.

Using Elastic Cloud's `us-east-1` region on AWS as an example, we get back 3x ip addresses for our deployment. Each IP Address is an AWS LB that maps to a separate Elastic Cloud Proxy, each in a different Availability Zone. Requests to a given AZ's proxy are preferentially routed to Elasticsearch nodes within the same AZ to minimise cross-az traffic costs, which means that 'sticky' clients can end up causing imbalanced HTTP connections at the Elasticsearch cluster level: 
```shell
[...]

PR #1707, opened 2023-04-21
Evenly distribute client connections across hosts

b-deam:
With this commit we evenly distribute connections across
all IPs a given hostname resolves. 

Fixes https://github.com/elastic/rally/issues/1598 

Visualised:
|Bulk requests per AZ| HTTP Connections per Node|
| ---      | ---      |
|image| image |

### DIFF
--- a/tests/client/asynchronous_test.py
+++ b/tests/client/asynchronous_test.py
@@ -16,8 +16,13 @@
 # under the License.
 
 import json
+import math
+import random
+from unittest import mock
 
-from esrally.client.asynchronous import ResponseMatcher
+import pytest
+
+from esrally.client.asynchronous import RallyTCPConnector, ResponseMatcher
 
 
 class TestResponseMatcher:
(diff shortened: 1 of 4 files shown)


tests/client/asynchronous_test.py
```diff
@@ -49,3 +54,76 @@ def test_matches(self):
...
+    ):
+        """
+        TCPConnector._resolve_host() returns all the IPs a given name resolves to, but the underlying connection but the
+        logic in TCPConnector._create_direct_connection() only ever selects the first succesful host from this list.
+
+        So, we use the factory assigned client_id to deterministically return an IP and then place it at the beginning
+        of the list to evenly distributes connections across _all_ clients.
+
+        https://github.com/elastic/rally/issues/1598
+        """
```
pquentin, line 80: We already said that above, I don't think repeating it helps. I think I would just remove those lines?

tests/client/asynchronous_test.py
```diff
@@ -49,3 +54,76 @@ def test_matches(self):
     def assert_response_type(self, matcher, path, expected_response_type):
         response = json.loads(matcher.response(path))
         assert response["response-type"] == expected_response_type
+
+
+class TestResolveHost:
```
pquentin, line 59: You don't need a class here, the function itself is enough (just remove self)

pquentin (changes requested):
Thanks! This looks good.

tests/client/asynchronous_test.py
b-deam: 👍 addressed in https://github.com/elastic/rally/pull/1707/commits/fe762dfdc66ad80bec07fbcb3a0d4be7eec11a64

tests/client/asynchronous_test.py
b-deam: Addressed in https://github.com/elastic/rally/pull/1707/commits/fe762dfdc66ad80bec07fbcb3a0d4be7eec11a64

pquentin (approved):
Thanks for iterating! LGTM.

pquentin:
Something that just came to my mind: will we have a CI failure if aiohttp refactors their code to not call this method anymore?

PR #1710, opened 2023-05-01
Actually use client returned from '.options()'

b-deam:
@danielmitterdorfer kindly pointed out that we were actually using the new `.options()` API incorrectly by discarding the newly returned client with the respective options/headers set.

Relates https://github.com/elastic/rally/issues/1673


danielmitterdorfer (approved):
Thanks for addressing this! LGTM

b-deam:
@elasticmachine please run buildkite/rally

PR #1712, opened 2023-05-04
Downgrade Python 3.11 to work around aiohttp bug

pquentin:
Since https://github.com/python/cpython/pull/98540 which was included in Python 3.11.1, aiohttp can fail if two tasks are trying to close the connection concurrently.


danielmitterdorfer (approved):
LGTM. Thanks for the workaround!

PR #1715, opened 2023-05-11
Show Rally builds in Buildkite default view

pquentin:
Making this owned by the wider Elasticsearch team is nice in theory but in practice when I'm logged in to Buildkite I'd like to see all our builds, including Rally IT tests.


dliappis (approved):
LGTM (again based on description)

PR #1716, opened 2023-05-11
Prepare release Docker pipeline

pquentin:
Declaring the pipeline before actually working on it is useful as it allows iterating on a branch.


dliappis (approved):
Not super familiar with the pipeline language yet, so providing an LGTM based on PR description.

PR #1725, opened 2023-06-05
Make async context manager synchronous

pquentin:
It did not call any async functions, so this was only confusing analysis.


danielmitterdorfer (approved):
Nice simplification. LGTM

gbanasiak (approved):
LGTM :)

PR #1727, opened 2023-06-08
Allow configuring logging to use local time

pquentin:
I've thought about https://github.com/elastic/rally/pull/1699 from @vaibhavsolanki1193 quite a lot, and I think I'd rather *not* rename the formatter for now. Having used this pull request while developing Rally, I can confirm the automatic migration works, but it only works forward, not backwards, which makes bisecting Rally more difficult.

As a result, this pull request is #1699 without the formatter rename. The main author is still @vaibhavsolanki1193. If we do want to change the name eventually, those commits in my fork do it:

1. [Prepare migrating more than missing loggers](https://github.com/pquentin/rally/commit/e824cca945d34e23a174e5a3c9bd8566574cb4aa)
2. [Rename esrally.log formatter](https://github.com/pquentin/rally/commit/35181291e8811fb4c025372668859e7a9ed5189e)

Closes #1699


dliappis (approved):
Sounds like a logical and practical solution to me. Thanks for putting the thought.

LGTM

PR #1728, opened 2023-06-08
Revert "Upgrade Jinja2 to version 3.1.2"

pquentin:
Reverts elastic/rally#1711. @gbanasiak has noticed a performance regression with this upgrade that we would like to address. It currently affects the tail latencies of [one specific benchmark](https://elasticsearch-benchmarks.elastic.co/#tracks/percolator/nightly/default/90d) and appears to be due to the garbage collector pausing Rally. Since we're still investigating, I don't want to release Rally 2.8.0 with this change.

239218986-17cae921-ca2a-4b7b-b503-71487db7b018

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,8 +58,8 @@ dependencies = [
     "jsonschema==3.1.1",
     # License: BSD
     # transitive dependency Markupsafe: BSD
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM

Issue #1714, opened 2023-05-10
AttributeError: 'NoneType' object has no attribute '_abort' on aiohttp connection close


pquentin:
Two separate users of Rally stumbled upon the same aiohttp/asyncio exception.

```
  File "/.../rally/esrally/driver/driver.py", line 1783, in run
    await es.close()
  File "/.../.local/lib/python3.11/site-packages/elasticsearch/_async/client/__init__.py", line 557, in close
    await self.transport.close()
  File "/.../.local/lib/python3.11/site-packages/elastic_transport/_async_transport.py", line 386, in close
    await node.close()
  File "/.../.local/lib/python3.11/site-packages/elastic_transport/_node/_http_aiohttp.py", line 242, in close
    await self.session.close()
  File "/.../.local/lib/python3.11/site-packages/aiohttp/client.py", line 980, in close
    await self._connector.close()
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../.local/lib/python3.11/site-packages/aiohttp/connector.py", line 792, in close
    return super().close()
           ^^^^^^^^^^^^^^^
  File "/.../.local/lib/python3.11/site-packages/aiohttp/connector.py", line 409, in close
    self._close()
  File "/.../.local/lib/python3.11/site-packages/aiohttp/connector.py", line 439, in _close
    transport.abort()
  File "/.../.pyenv/versions/3.11.2/lib/python3.11/asyncio/sslproto.py", line 247, in abort
    self._ssl_protocol._abort()
    ^^^^^^^^^^^^^^^^^^^^^^^^^

AttributeError: 'NoneType' object has no attribute '_abort'
```

This error has been triggered by https://github.com/python/cpython/pull/98540 which fixed https://github.com/python/cpython/issues/98539. It is in Python 3.11.1 and not Python 3.11.0. Since downgrading to Python 3.11.0 in https://github.com/elastic/rally/pull/1712 we have not observed it.

@danielmitterdorfer analyzed the problem and concluded that the problem is that aiohttp attempts to close the transport several times and only checks for None in two places:

 1. https://github.com/aio-libs/aiohttp/blob/2be9318dc95a26940e9095673ebcbbf77dfd087b/aiohttp/connector.py#L357-L358
 2. https://github.com/aio-libs/aiohttp/blob/2be9318dc95a26940e9095673ebcbbf77dfd087b/aiohttp/connector.py#L411-L413

They should guard this with `if transport is not None and not transport.is_closing()` instead.

[...]

PR #1729, opened 2023-06-09
Bump Python to 3.11.4

pquentin:
Closes https://github.com/elastic/rally/issues/1714

We had to downgrade to 3.11.0 in https://github.com/elastic/rally/pull/1712 to work around an asyncio/aiohttp bug. Now that this bug was fixed in https://github.com/python/cpython/pull/104485 and released in Python 3.11.4, we can bump the Python 3.11 version.

@danielmitterdorfer @b-deam Is there an easy way to confirm this fixes the issue?


dliappis (approved):
LGTM

b-deam:
> Closes #1714
> 
> We had to downgrade to 3.11.0 in #1712 to work around an asyncio/aiohttp bug. Now that this bug was fixed in [python/cpython#104485](https://github.com/python/cpython/pull/104485) and released in Python 3.11.4, we can bump the Python 3.11 version.
> 
> @danielmitterdorfer @b-deam Is there an easy way to confirm this fixes the issue?

I was only able to reproduce on a very long running benchmark (took 10h+ to repro), so I've started one using this branch.

pquentin:
Awesome, thanks a lot for taking the time and energy!

PR #1732, opened 2023-06-12
Fix license URLs in create notice script

gbanasiak:
As in the title. Previous URLs were returning 404s.


pquentin (approved):
Thanks! LGTM.

PR #1733, opened 2023-06-12
Run all tests on release

pquentin:
@gbanasiak noticed while releasing Rally that we were only running simple unit tests. This pull request fixes the issue.


pquentin:
@elastic/es-perf Is anyone here using `make test` for development? I'm using `pytest` directly, so the change to `make test` does not affect me, but maybe it affects you?

gbanasiak (approved):
LGTM

Makefile
```diff
@@ -113,7 +114,7 @@ release-checks: check-venv
 	. $(VENV_ACTIVATE_FILE); ./release-checks.sh $(release_version) $(next_version)
 
 # usage: e.g. make release release_version=0.9.2 next_version=0.9.3
-release: check-venv release-checks clean docs it
+release: check-venv release-checks clean docs test it
```
dliappis, line 117: Is the intention to run the entirety of tests prior to `release`? If yes, should we include `lint` in this target too?

dliappis (commented):
No objections to this change. Makes sense.
Left one question.

Makefile
pquentin:
```suggestion
release: check-venv release-checks clean docs lint test it
```

dliappis (approved):
LGTM

pquentin:
buildkite build this please

PR #1738, opened 2023-06-20
Refactor version number checks

b-deam:
Serverless Elasticsearch doesn't return a `version.number` field from the [Info API](https://www.elastic.co/guide/en/elasticsearch/reference/current/info-api.html) (`/`) , which is actually something we rely on somewhat commonly throughout the codebase.

To fix this, we need a clearly defined way of determining whether or not Rally is talking to a Serverless Elasticsearch, this applies for both clients used to target the system under test, as well as any ancillary clients like those used by a remote metrics store.

In order to do so, we'll take two approaches:
- Create a `is_serverless | bool` property on both existing and new 'serverless' specific clients
- Refactor all version checks to default to selecting the 'minimum' version required for a specific conditional, and only use the `is_serverless` property if we need to do something specific for serverless

I think this is a balanced approach, given that serverless has no concept of 'versioning'.

Note:

~~The underlying client's [use of `.options()` complicates things](https://github.com/elastic/rally/issues/1673), because every call reinstantiates a new copy of itself, but without the ability for us to pass in any custom arguments (e.g. `distribution_version`). To work around this we must work out upfront whether or not we'd like the factory to create a serverless or non-serverless client, each of which have a static `is_serverless` attribute.~~ Fixed in https://github.com/elastic/rally/pull/1738/commits/abf43207e1ebde6899876dcc3ebadf5998e7eab1

<details>

<summary> Example tests </summary>

```
# http_logs with node-stats and data-stream-stats telemetry devices
esrally race --track http_logs --pipeline=benchmark-only --target-hosts https://serverless-es --client-options='{ "default": {"headers": {"X-Found-Cluster": "09895e10-efb8-42da-be30-888eb623d6cf.es"}, "basic_auth_user":"esbench", "basic_auth_password":"changeme", "verify_certs": false}}' --kill-running-processes --test-mode --track-params 'number_of_replicas: 1' --telemetry "node-stats, data-stream-stats"

# create-track
esrally create-track --track test --indices logs-221998 --target-hosts https://serverless-es --client-options '{ "default": {"headers": {"X-Found-Cluster": "09895e10-efb8-42da-be30-888eb623d6cf.es"}, "basic_auth_user":"elastic", "basic_auth_password":"changeme", "verify_certs": false}}'
```

</summary>

### DIFF
--- a/esrally/client/asynchronous.py
+++ b/esrally/client/asynchronous.py
@@ -285,20 +285,14 @@ class RallyIlmClient(IlmClient):
 class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
     def __init__(self, *args, **kwargs):
         distribution_version = kwargs.pop("distribution_version", None)
+        distribution_flavor = kwargs.pop("distribution_flavor", None)
         super().__init__(*args, **kwargs)
         # skip verification at this point; we've already verified this earlier with the synchronous client.
         # The async client is used in the hot code path and we use customized overrides (such as that we don't
         # parse response bodies in some cases for performance reasons, e.g. when using the bulk API).
         self._verified_elasticsearch = True
-        self._serverless = False
-
-        # this isn't always available because any call to self.options() doesn't pass any custom args
-        # to the constructor
-        # https://github.com/elastic/rally/issues/1673
-        if distribution_version:
-            self.distribution_version = distribution_version
-        else:
-            self.distribution_version = None
+        self.distribution_version = distribution_version
+        self.distribution_flavor = distribution_flavor
 
         # some ILM method signatures changed in 'elasticsearch-py' 8.x,
         # so we override method(s) here to provide BWC for any custom
@@ -307,7 +301,13 @@ class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
 
     @property
     def is_serverless(self):
-        return self._serverless
+        return versions.is_serverless(self.distribution_flavor)
+
+    def options(self, *args, **kwargs):
+        new_self = super().options(*args, **kwargs)
+        new_self.distribution_version = self.distribution_version
+        new_self.distribution_flavor = self.distribution_flavor
+        return new_self
 
     async def perform_request(
         self,
@@ -340,7 +340,7 @@ class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
         # Not applicable to serverless
         if not self.is_serverless:
             if self.distribution_version is not None and (
-                versions.Version.from_string(self.distribution_version) >= versions.Version.from_string("8.0.0") and not self.is_serverless
+                versions.Version.from_string(self.distribution_version) >= versions.Version.from_string("8.0.0")
             ):
                 _mimetype_header_to_compat("Accept", request_headers)
                 _mimetype_header_to_compat("Content-Type", request_headers)
@@ -412,17 +412,3 @@ class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
             response = ApiResponse(body=resp_body, meta=meta)  # type: ignore[assignment]
 
         return response
-
-
-class RallyAsyncElasticsearchServerless(RallyAsyncElasticsearch):
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        # we set this as an instance attribute because we can't afford to make any calls to external APIs to verify
-        # whether we're talking to a serverless cluster or not once we're executing the benchmark.
-        #
-        # the reason for this is because the client can reinstantiate itself (e.g. with a call to .options()) during
-        # the execution of the benchmark, which means external API calls add unnecessary latency, and it reinstantiates
-        # itself without the ability to pass any custom arguments (i.e. distribution_version) to the constructor
-        #
-        # see https://github.com/elastic/rally/issues/1673
-        self._serverless = True
--- a/esrally/client/asynchronous.py
+++ b/esrally/client/asynchronous.py
@@ -339,7 +339,7 @@ class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
         # see https://github.com/elastic/elasticsearch/issues/51816
         # Not applicable to serverless
         if not self.is_serverless:
-            if self.distribution_version is not None and (
+            if versions.is_version_identifier(self.distribution_version) and (
                 versions.Version.from_string(self.distribution_version) >= versions.Version.from_string("8.0.0")
             ):
                 _mimetype_header_to_compat("Accept", request_headers)
--- a/esrally/client/asynchronous.py
+++ b/esrally/client/asynchronous.py
@@ -290,8 +290,13 @@ class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
         # The async client is used in the hot code path and we use customized overrides (such as that we don't
         # parse response bodies in some cases for performance reasons, e.g. when using the bulk API).
         self._verified_elasticsearch = True
+        self._serverless = False
+
+        # this isn't always available because any call to self.options() doesn't pass any custom args
+        # to the constructor
+        # https://github.com/elastic/rally/issues/1673
         if distribution_version:
-            self.distribution_version = versions.Version.from_string(distribution_version)
+            self.distribution_version = distribution_version
         else:
             self.distribution_version = None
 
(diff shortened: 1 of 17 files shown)


pquentin:
buildkite build this please after https://github.com/elastic/rally/pull/1739

esrally/utils/versions.py
```diff
@@ -180,6 +184,8 @@ def best_match(available_alternatives, distribution_version):
         major, _, _, _ = components(distribution_version)
         if major > _latest_major(available_alternatives):
             return "master"
+    elif is_serverless(distribution_version):
+        return "master"
```
b-deam, line 188: We only every select the `master` branch of `rally-teams` when targeting a serverless cluster.

esrally/client/asynchronous.py
```diff
@@ -328,9 +337,13 @@ async def perform_request(
         # Converts all parts of a Accept/Content-Type headers
         # from application/X -> application/vnd.elasticsearch+X
         # see https://github.com/elastic/elasticsearch/issues/51816
-        if self.distribution_version is not None and self.distribution_version >= versions.Version.from_string("8.0.0"):
-            _mimetype_header_to_compat("Accept", request_headers)
-            _mimetype_header_to_compat("Content-Type", request_headers)
+        # Not applicable to serverless
+        if not self.is_serverless:
+            if self.distribution_version is not None and (
+                versions.Version.from_string(self.distribution_version) >= versions.Version.from_string("8.0.0") and not self.is_serverless
```
pquentin, line 343: You used `not self.is_serverless` twice here

esrally/client/asynchronous.py
```diff
@@ -290,8 +290,13 @@ def __init__(self, *args, **kwargs):
...
+        self._serverless = False
+
+        # this isn't always available because any call to self.options() doesn't pass any custom args
+        # to the constructor
+        # https://github.com/elastic/rally/issues/1673
         if distribution_version:
-            self.distribution_version = versions.Version.from_string(distribution_version)
+            self.distribution_version = distribution_version
         else:
             self.distribution_version = None
```
pquentin, line 301: Why don't we override the options() method ourselves like this instead?

```
def options(self, *, **kwargs):
   newself = super().options(**kwargs)
   newself.distribution_version = self.distribution_version
   newself._serverless = self._serverless
   return newself
```

It seems like it would solve both distribution_version being None (which is very annoying) and needing RallyAsyncElasticsearchServerless (which is not that bad but that I'd prefer to avoid if possible).

esrally/client/factory.py
```diff
@@ -31,7 +31,7 @@ class EsClientFactory:
     compatibility guarantees that are broader than the library's defaults.
     """
 
-    def __init__(self, hosts, client_options, distribution_version=None):
+    def __init__(self, hosts, client_options, distribution_version=None, distribution_flavor=None):
```
pquentin, line 34: Why do we accept `None` values here?

esrally/client/asynchronous.py
b-deam: Thanks, addressed in https://github.com/elastic/rally/pull/1738/commits/abf43207e1ebde6899876dcc3ebadf5998e7eab1

esrally/client/asynchronous.py
b-deam: Thank you! Implemented in https://github.com/elastic/rally/pull/1738/commits/abf43207e1ebde6899876dcc3ebadf5998e7eab1, testing with some proper benchmarks now.

esrally/client/factory.py
b-deam: The factory is also called before we know what the distribution version/flavor is, e.g:
https://github.com/elastic/rally/blob/abf43207e1ebde6899876dcc3ebadf5998e7eab1/esrally/client/factory.py#L326-L349

esrally/client/asynchronous.py
```diff
@@ -328,9 +337,13 @@ async def perform_request(
         # Converts all parts of a Accept/Content-Type headers
         # from application/X -> application/vnd.elasticsearch+X
         # see https://github.com/elastic/elasticsearch/issues/51816
-        if self.distribution_version is not None and self.distribution_version >= versions.Version.from_string("8.0.0"):
-            _mimetype_header_to_compat("Accept", request_headers)
-            _mimetype_header_to_compat("Content-Type", request_headers)
+        # Not applicable to serverless
+        if not self.is_serverless:
+            if self.distribution_version is not None and (
```
pquentin, line 342: Can we assume that `distribution_version` will be defined now?

esrally/client/factory.py
```diff
@@ -226,13 +220,26 @@ async def on_request_end(session, trace_config_ctx, params):
...
+
+        async def on_request_end(session, trace_config_ctx, params):
+            async_client.on_request_end()
+
+        trace_config = aiohttp.TraceConfig()
+        trace_config.on_request_start.append(on_request_start)
+        trace_config.on_request_end.append(on_request_end)
+        # ensure that we also stop the timer when a request "ends" with an exception (e.g. a timeout)
+        trace_config.on_request_exception.append(on_request_end)
+
```
pquentin, line 242: Would you mind reverting this change? (Move this back up and replace `async_client` with `RallyAsyncElasticsearch` in the async methods.)

esrally/client/synchronous.py
```diff
@@ -172,9 +180,12 @@ def perform_request(
         # Converts all parts of a Accept/Content-Type headers
         # from application/X -> application/vnd.elasticsearch+X
         # see https://github.com/elastic/elasticsearch/issues/51816
-        if self.distribution_version is not None and self.distribution_version >= versions.Version.from_string("8.0.0"):
-            _mimetype_header_to_compat("Accept", request_headers)
-            _mimetype_header_to_compat("Content-Type", request_headers)
+        if not self.is_serverless:
+            if self.distribution_version is not None and (
```
pquentin, line 184: Same as above

esrally/tracker/tracker.py
```diff
@@ -74,16 +74,19 @@ def create_track(cfg):
     track_name = cfg.opts("track", "track.name")
     indices = cfg.opts("generator", "indices")
     root_path = cfg.opts("generator", "output.path")
-    target_hosts = cfg.opts("client", "hosts")
-    client_options = cfg.opts("client", "options")
+    target_hosts = cfg.opts("client", "hosts").default
+    client_options = cfg.opts("client", "options").default
```
pquentin, line 78: What is this change doing?

esrally/tracker/tracker.py
```diff
@@ -74,16 +74,19 @@ def create_track(cfg):
...
+    client = factory.EsClientFactory(
+        hosts=target_hosts,
+        client_options=client_options,
+        distribution_version=distribution_version,
+        distribution_flavor=distribution_flavor,
     ).create()
 
-    info = client.info()
-    console.info(f"Connected to Elasticsearch cluster [{info['name']}] version [{info['version']['number']}].\n", logger=logger)
+    console.info(f"Connected to Elasticsearch cluster version [{distribution_version}]\n", logger=logger)
```
pquentin, line 89: Should we log the flavor too?

pquentin (commented):
Thanks for iterating! That looks much better.

esrally/client/asynchronous.py
b-deam: I think we can for the async client, just because we currently only create them in one place where we already know the distribution flavor/version. I mainly did this to be consistent between the two, but I think a better approach is to just check whether the value is a valid identifier as I've done in https://github.com/elastic/rally/pull/1738/commits/5f2db54f79ae9dc1b5c2bf3882319268a0945c65

esrally/client/synchronous.py
b-deam: Addressed in https://github.com/elastic/rally/commit/5f2db54f79ae9dc1b5c2bf3882319268a0945c65

esrally/client/factory.py
b-deam: Addressed in https://github.com/elastic/rally/pull/1738/commits/94d863418f013b6d1a5498fb476f93b8cfb31707

esrally/tracker/tracker.py
b-deam: Just a shorthand way of getting the same values with:
```python
target_hosts.all_hosts[opts.TargetHosts.DEFAULT]
client_options.all_client_options[opts.TargetHosts.DEFAULT]
```

esrally/tracker/tracker.py
b-deam: I thought about that, but when you're targeting stateful elasticsearch you just get `default` back, which is a little confusing, so I just left it out entirely. Happy to include it if you think it's important though.

pquentin: Yes I think it's important to have Rally confirm what sort of cluster it thinks it's benchmarking.

b-deam: Sure, addressed in https://github.com/elastic/rally/pull/1738/commits/27f1f1268e3c087eb41cdfa07e2f32aba3640e8b. I've kicked off a long running test benchmark too. Will likely merge early next week.

dliappis (approved):
LGTM thank you for this massive change!

PR #1743, opened 2023-06-29
Bulk action refresh param

inqueue:
This commit adds support for the bulk `?refresh` query parameter to address [feedback](https://github.com/elastic/rally-tracks/pull/425#issuecomment-1612358005) for a new workload for serverless. Possible values for `refresh` are `"true"`, `"false"`, and `"wait_for"`, consistent with https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-refresh.

@pquentin @b-deam you have been tagged as reviewers. If you are fine with the approach, then I will add docs and take this PR out of draft form.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -501,6 +501,8 @@ class BulkIndex(Runner):
             bulk_params["timeout"] = params["timeout"]
         if "pipeline" in params:
             bulk_params["pipeline"] = params["pipeline"]
+        if "refresh" in params:
+            bulk_params["refresh"] = params["refresh"]
 
         with_action_metadata = mandatory(params, "action-metadata-present", self)
         bulk_size = mandatory(params, "bulk-size", self)
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -634,6 +634,7 @@ class BulkIndexParamSource(ParamSource):
             raise exceptions.InvalidSyntax("'batch-size' must be numeric")
 
         self.ingest_percentage = self.float_param(params, name="ingest-percentage", default_value=100, min_value=0, max_value=100)
+        self.refresh = params.get("refresh")
         self.param_source = PartitionBulkIndexParamSource(
             self.corpora,
             self.batch_size,
@@ -644,6 +645,7 @@ class BulkIndexParamSource(ParamSource):
             self.on_conflict,
             self.recency,
             self.pipeline,
+            self.refresh,
             self._params,
         )
 
@@ -705,6 +707,7 @@ class PartitionBulkIndexParamSource:
         on_conflict,
         recency,
         pipeline=None,
+        refresh=None,
         original_params=None,
     ):
         """
@@ -732,6 +735,7 @@ class PartitionBulkIndexParamSource:
         self.on_conflict = on_conflict
         self.recency = recency
         self.pipeline = pipeline
+        self.refresh = refresh
         self.original_params = original_params
         # this is only intended for unit-testing
         self.create_reader = original_params.pop("__create_reader", create_default_reader)
--- a/tests/driver/runner_test.py
+++ b/tests/driver/runner_test.py
@@ -1243,6 +1243,82 @@ class TestBulkIndexRunner:
 
         es.bulk.assert_awaited_with(body=bulk_params["body"], params={})
 
+    @mock.patch("elasticsearch.Elasticsearch")
+    @pytest.mark.asyncio
+    async def test_bulk_index_success_with_refresh_true(self, es):
+        bulk_response = {
+            "errors": False,
+            "took": 8,
+            "items": [{"create": {"_index": "test", "result": "created", "status": 201, "forced_refresh": True}}],
+        }
+        es.bulk = mock.AsyncMock(return_value=bulk_response)
+
+        bulk = runner.BulkIndex()
+
+        bulk_params = {
+            "body": _build_bulk_body("index_line"),
+            "index": "test",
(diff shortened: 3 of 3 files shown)


esrally/track/params.py
```diff
@@ -732,6 +735,7 @@ def __init__(
         self.on_conflict = on_conflict
         self.recency = recency
         self.pipeline = pipeline
+        self.refresh = refresh
```
b-deam, line 738: Noting that this initial review is just for the approach, but don't forget to add this param to the docstring for `PartitionBulkIndexParamSource`

b-deam:
Approach looks fine to me, but curious how you intend to handle more complex configurations where want to change the `refresh` param dynamically? Will we require users to implement a custom param source?

inqueue:
> Approach looks fine to me, but curious how you intend to handle more complex configurations where want to change the `refresh` param dynamically? Will we require users to implement a custom param source?

I am not too concerned with changing the `refresh `param` dynamically. I do not have a solution for doing this in a deterministic way. For now, it is good enough to issue a manual refresh to the target in a parallel operation. Right now, I am more concerned with the use of the `refresh` parameter to avoid situations where search-idle optimizations kick in for smaller corpora where the use case of this parameter is most applicable.

Does it make more sense to change the possible values to strings only? Like:

* `sync` - Equivalent to `?refresh=wait_for`
* `async` - Equivalent to `?refesh=true`
* `default` - Equivalent to `?refresh=false`

This would leave the door open for a future parameter value like `sometimes` where `?refresh=true` (or `?refresh=wait_for`) for a subset of requests as you have suggested elsewhere.

inqueue:
> Does it make more sense to change the possible values to strings only? Like:
> 
> * `sync` - Equivalent to `?refresh=wait_for`
> * `async` - Equivalent to `?refesh=true`
> * `default` - Equivalent to `?refresh=false`

The latest commit uses this model. It can be tested against serverless with the track in https://github.com/elastic/rally-tracks/pull/425.

```
esrally race --track=serverless_k8s --pipeline=benchmark-only --client-options='{"default": {"use_ssl": true, "verify_certs": false, "basic_auth_user":"${ESUSER}","basic_auth_password":"${ESPASSWORD}"}}' --telemetry=node-stats,shard-stats,data-stream-stats --telemetry-params="node-stats:node-stats-include-indices" --target-hosts=${TARGET} --track-params="bulk_size:100,ingest_percentage:10,bulk_refresh:async" --kill-running-processes --test-mode
```

esrally/driver/runner.py
```diff
@@ -501,6 +504,18 @@ async def __call__(self, es, params):
             bulk_params["timeout"] = params["timeout"]
         if "pipeline" in params:
             bulk_params["pipeline"] = params["pipeline"]
+        if "refresh" in params:
+            refresh_params = {"sync": "wait_for", "async": "true", "default": "false"}
```
b-deam, line 508: I think we just use a 1:1 mapping of the values to the ES API, any 'dynamic' behaviour can use a parameter source to set those values.

esrally/driver/runner.py
```diff
@@ -501,6 +504,18 @@ async def __call__(self, es, params):
...
+            bulk_params["refresh"] = refresh_params.get(params["refresh"])
+
+            if bulk_params["refresh"] is None:
+                self.logger.warning(
+                    "Using default bulk refresh parameter value '%s' in place of unrecognized specfified parameter value '%s'. "
+                    "Use one of %s.",
+                    refresh_params["default"],
+                    params["refresh"],
+                    ", ".join(refresh_params.keys()),
+                )
```
b-deam, line 518: I don't think this needs to be a warning, maybe just an info message.

b-deam (approved):
Left some comments, but it otherwise LGTM.

esrally/driver/runner.py
inqueue: > Edit: I'm not actually sure what we're checking here? If the param isn't set shouldn't we just default to default?

This check is for invalid values, like the test [here](https://github.com/elastic/rally/pull/1743/files#diff-026a25af4075b17db2736adf14aa844c0aaeb2bc995b4b4ae94c432b0380b51fR1365).

esrally/driver/runner.py
```diff
@@ -505,16 +505,14 @@ async def __call__(self, es, params):
...
-            if bulk_params["refresh"] is None:
-                self.logger.warning(
-                    "Using default bulk refresh parameter value '%s' in place of unrecognized specfified parameter value '%s'. "
-                    "Use one of %s.",
-                    refresh_params["default"],
+            valid_refresh_values = ("wait_for", "true", "false")
+            if params["refresh"] in valid_refresh_values:
+                bulk_params["refresh"] = params["refresh"]
+            else:
+                self.logger.info(
```
b-deam, line 512: Sorry for my last comment, I see now what you're checking for. I think both `INFO` and `WARNING` aren't enough this case, and we should probably just error out because of an invalid value - otherwise we risk returning misleading results and the corresponding log message is buried deep in the file somewhere. 

Maybe we can error out similar to how it's done here?:
https://github.com/elastic/rally/blob/b1a822c19734cb350961eb15d34649ac24cda9e2/esrally/driver/runner.py#L2609-L2612

b-deam: We could probably also just lean on the client itself to return an error, if you pass something in that's not valid does it return anything meaningful?

b-deam (approved):
Left one minor comment around unsupported values, but no need for another review cycle. LGTM.

PR #1748, opened 2023-07-12
Document actor system

pquentin:
You can see this live at https://github.com/pquentin/rally/blob/document-actor-system/docs/architecture/actor_system.md.

### DIFF
--- a/docs/architecture/actor_system.md
+++ b/docs/architecture/actor_system.md
@@ -37,7 +37,7 @@ The sequence diagram below starts with `BenchmarkActor` defined in
 
  * Setup the Elasticsearch cluster if the pipeline requires it (not covered
    here)
- * Prepare the benchmark which involves the `TrackPreparator` and
+ * Prepare the benchmark which involves the `TrackPreparationActor` and
    `TaskExecutionActor` actors.
  * Start the benchmark, which will involve the `Worker` actor that will
    delegate the actual work to `AsyncIoAdapter`, which will run an asyncio loop
@@ -64,23 +64,23 @@ sequenceDiagram
     DriverActor ->> Driver: __init__
     DriverActor ->> Driver: prepare_benchmark
     Driver ->> DriverActor: prepare_track
-    DriverActor ->> TrackPreparator: __init__
-    DriverActor -->> TrackPreparator: Bootstrap
-    TrackPreparator -->> DriverActor: ReadyForWork
-    DriverActor -->> TrackPreparator: PrepareTrack
-    TrackPreparator ->> TaskExecutionActor: __init__
-    TrackPreparator -->> TaskExecutionActor: StartTaskLoop
+    DriverActor ->> TrackPreparationActor: __init__
+    DriverActor -->> TrackPreparationActor: Bootstrap
+    TrackPreparationActor -->> DriverActor: ReadyForWork
+    DriverActor -->> TrackPreparationActor: PrepareTrack
+    TrackPreparationActor ->> TaskExecutionActor: __init__
+    TrackPreparationActor -->> TaskExecutionActor: StartTaskLoop
     loop
-        TaskExecutionActor -->> TrackPreparator: ReadyForWork
-        TrackPreparator -->> TaskExecutionActor: DoTask
+        TaskExecutionActor -->> TrackPreparationActor: ReadyForWork
+        TrackPreparationActor -->> TaskExecutionActor: DoTask
         loop
             TaskExecutionActor -->> TaskExecutionActor: WakeupMessage
         end
     end
-    TaskExecutionActor -->> TrackPreparator: WorkerIdle
-    TrackPreparator -->> DriverActor: TrackPrepared
-    DriverActor -->> TrackPreparator: ActorExitRequest
-    TrackPreparator -->> TaskExecutionActor: ActorExitRequest
+    TaskExecutionActor -->> TrackPreparationActor: WorkerIdle
+    TrackPreparationActor -->> DriverActor: TrackPrepared
+    DriverActor -->> TrackPreparationActor: ActorExitRequest
+    TrackPreparationActor -->> TaskExecutionActor: ActorExitRequest
     Driver -->> BenchmarkActor: PreparationComplete
 ```
 
--- a/docs/architecture/actor_system.md
+++ b/docs/architecture/actor_system.md
@@ -3,7 +3,7 @@
 At its heart, Rally is a distributed system. It has been designed that way to
 allow using multiple load drivers in the same benchmark, to ensure that Rally
 is never a bottleneck. In the vast majority of cases, using a powerful load
-driver is enough, but benchmarks large Elasticsearch clusters containing tens
+driver is enough, but benchmarking large Elasticsearch clusters containing tens
 or hundreds of nodes can require more load drivers.
 
 ## Thespian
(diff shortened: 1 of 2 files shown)


docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
+# Rally Actor System
+
+At its heart, Rally is a distributed system. It has been designed that way to
+allow using multiple load drivers in the same benchmark, to ensure that Rally
+is never a bottleneck. In the vast majority of cases, using a powerful load
+driver is enough, but benchmarks large Elasticsearch clusters containing tens
```
inqueue, line 6:
```suggestion
driver is enough, but benchmakring large Elasticsearch clusters containing tens
```

docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
...
+
+The sequence diagram below starts with `BenchmarkActor` defined in
+`racecontrol.py` which does the high-level scheduling:
+
+ * Setup the Elasticsearch cluster if the pipeline requires it (not covered
+   here)
+ * Prepare the benchmark which involves the `TrackPreparator` and
+   `TaskExecutionActor` actors.
+ * Start the benchmark, which will involve the `Worker` actor that will
+   delegate the actual work to `AsyncIoAdapter`, that will run an asyncio loop
```
inqueue, line 43:
```suggestion
   delegate the actual work to `AsyncIoAdapter`, which will run an asyncio loop
```

docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
...
+`Driver` contains most of the logic, it is not an actor, so it relies on
+`DriverActor` to send and receive messages. This was done for two reasons:
+
+ * `Driver` can be unit-tested without bringing the actor system
+ * An earlier attempt at removing the actor system failed.
+
+### PrepareBenchmark
+
+Here's the sequence diagram for benchmark preparation. Dotted lines
+means messages sent with the actor systems, while plain lines are
```
inqueue, line 57:
```suggestion
mean messages sent with the actor systems, while plain lines are
```

docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
...
+documented](https://thespianpy.com/doc/using.html) and battle tested.
+Additionally, the maintainer has always been responsive and helpful.
+
+## Sequence diagrams
+
+Rally has a number of actors that all inherit from `actor.RallyActor`. This
+document focuses on the actors needed to *prepare* and *run* a benchmark, with
+the following limitations:
+
+ * The mechanic actors that can setup an Elasticsearch cluster are not covered
```
inqueue, line 30:
```suggestion
 * The mechanic actors that can set up an Elasticsearch cluster are not covered
```

inqueue (changes requested):
Thank you for this. I left some minor suggestions.

docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
...
+is never a bottleneck. In the vast majority of cases, using a powerful load
+driver is enough, but benchmarks large Elasticsearch clusters containing tens
+or hundreds of nodes can require more load drivers.
+
+## Thespian
+
+Actors are managed by [Thespian](https://thespianpy.com/doc/) which provide us
+with the following features:
+
+ * Works with Linux and macOS (and Windows, but Rally does not need that)
```
b-deam, line 14:
```suggestion
 * Works with Linux and macOS (and Windows, though Rally does not support Windows)
```

gbanasiak (commented):
Many thanks! This helps in building the Rally mental map.

I went through `PrepareBenchmark` so far. A few remarks, a wish list for later mostly:
* For consistency, `TrackPreparator` should be renamed to `TrackPreparationActor` in the charts as that's the name of the class I think.
* It would be great to include 10k feet description of _what_ each component is doing, e.g. 1) `TrackPreparationActor` loads a track and track _plugins_ (custom extensions such as parameter sources and runners), gets a list of track _processors_ (`TrackProcessor` interface) from `TrackProcessorRegistry` and stores them in a queue. Then it sends each processor for execution to `TaskExecutionActor`. 2) `TackExecutionActor` executes each processor in a background task and verifies its status periodically. Once task completes, it reports its readiness with `WorkComplete`. Not sure what level of details makes sense here - too much is just a repetition of code, too little forces us to go through the flow again if we revisit this after enough time.
* It would be nice to explain how individual components grow as the scale of the system increases, e.g. `TrackPreparationActor` set grows with the number of load driver hosts, while `TaskExecutionActor` set can grow with the number of cores declared per load driver.

docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
+# Rally Actor System
+
+At its heart, Rally is a distributed system. It has been designed that way to
+allow using multiple load drivers in the same benchmark, to ensure that Rally
+is never a bottleneck. In the vast majority of cases, using a powerful load
```
dliappis, line 5: Another reason (the first one, IIRC) is that distributing Rally also provides an easy way to handle multiple target hosts and ensure that they all build/install/start the required ES version via the `race` subcommand (this is for example the mode used on the Hetzner nightly benchmarks).
Later on we got the dedicated/decoupled `esrally build/install/start/stop` subcommands (which we use e.g. in esbench) and these are decoupled from the actor system.

pquentin: The current formulation isn't correct then, however my understanding is that this is achievable without a full-fledged actor system, while we could easily end up re-implementing Thespian badly to support multiple load drivers.

pquentin:
> I went through `PrepareBenchmark` so far. A few remarks, a wish list for later mostly:

I fixed `TrackPreparationActor`, would you like contributing your other fixes to this branch or in another pull request?

docs/architecture/actor_system.md
dliappis: > however my understanding is that this is achievable without a full-fledged actor system, while we could easily end up re-implementing Thespian badly to support multiple load drivers.

I mostly agree. I don't think the risk of supporting >1 load drivers without the actor system would necessarily result in re-implementing Thespian, but as always, the devil is in the details.

docs/architecture/actor_system.md
```diff
@@ -0,0 +1,117 @@
...
+   delegate the actual work to `AsyncIoAdapter`, that will run an asyncio loop
+   in a thread, which is how `AsyncExecutor` runs many Elasticsearch async
+   clients.
+
+You'll notice that `DriverActor` and `Driver` are tightly coupled. While
+`Driver` contains most of the logic, it is not an actor, so it relies on
+`DriverActor` to send and receive messages. This was done for two reasons:
+
+ * `Driver` can be unit-tested without bringing the actor system
+ * An earlier attempt at removing the actor system failed.
```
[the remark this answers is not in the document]
gbanasiak: I've removed this bullet in https://github.com/elastic/rally/pull/1748/commits/ac1aec8d28bbfb0623dd1e579a241f281344eafb.

docs/architecture/actor_system.md
gbanasiak: I've changed the wording slightly in https://github.com/elastic/rally/pull/1748/commits/ac1aec8d28bbfb0623dd1e579a241f281344eafb.

gbanasiak:
@dliappis @inqueue I've adjusted description based on the previous discussion and also updated diagrams for `StartBenchmark` section while I was reviewing the code myself. I will continue with extra additions in this doc in other PRs, if needed. PTAL.

dliappis (approved):
This LGTM -- so great to see these internal aspects of Rally documented.

One thing I wanted to ask is how come this isn't living under the [official docs](https://esrally.readthedocs.io/en/latest/) e.g. in the Reference Documentation or Additional Information sections? (is it because [Mermaid](https://mermaid.js.org/) isn't available?)

gbanasiak:
> One thing I wanted to ask is how come this isn't living under the [official docs](https://esrally.readthedocs.io/en/latest/) e.g. in the Reference Documentation or Additional Information sections? (is it because [Mermaid](https://mermaid.js.org/) isn't available?)

It should be possible to use Mermaid with Sphinx but it would require additional package, see https://github.com/mgaitan/sphinxcontrib-mermaid. That's work in progress. If/when this matures we can include this in official documentation.

PR #1749, opened 2023-07-12
Cleanup actor system code

pquentin:
See the the commits for the details.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -395,7 +395,7 @@ class TaskExecutionActor(actor.RallyActor):
         self.pool = concurrent.futures.ThreadPoolExecutor(max_workers=1)
         self.executor_future = None
         self.wakeup_interval = 5
-        self.parent = None
+        self.task_preparation_actor = None
         self.logger = logging.getLogger(__name__)
         self.track_name = None
         self.cfg = None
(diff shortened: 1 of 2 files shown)


esrally/driver/driver.py
```diff
@@ -402,13 +402,13 @@ def __init__(self):
 
     @actor.no_retry("task executor")  # pylint: disable=no-value-for-parameter
     def receiveMsg_StartTaskLoop(self, msg, sender):
-        self.parent = sender
+        self.task_preparation_actor = sender
```
pquentin, line 405: `self.parent` is still used in various other methods (including in `__init__`)

esrally/driver/driver.py
```diff
@@ -468,15 +468,15 @@ def __init__(self):
...
 
     @actor.no_retry("track preparator")  # pylint: disable=no-value-for-parameter
-    def receiveMsg_Bootstrap(self, msg, sender):
+    def receiveMsg_Bootstrap(self, msg, driver_actor):
         # load node-specific config to have correct paths available
         self.cfg = load_local_config(msg.config)
         # this instance of load_track occurs once per host, so install dependencies if necessary
         load_track(self.cfg, install_dependencies=False)
-        self.send(sender, ReadyForWork())
+        self.send(driver_actor, ReadyForWork())
```
pquentin, line 479: What are your thoughts about setting `self.driver_actor = sender` here instead of `receiveMsg_PrepareTrack` since it's received first?

esrally/racecontrol.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, name, description, target, stable=True):
         :param name: A short name of the pipeline. This name will be used to reference it from the command line.
         :param description: A human-readable description what the pipeline does.
         :param target: A function that implements this pipeline
-        :param stable True iff the pipeline is considered production quality.
+        :param stable True if the pipeline is considered production quality.
```
pquentin, line 62: For what it's worth, this is probably not a typo, "iff" is a mathematical abbrevation meaning "if and only if". But it does not add anything here, so "if" is fine.

pquentin (commented):
Thanks!

esrally/driver/driver.py
gareth-ellis: I'm not sure what happened there - I did go through and change there, I did the same with children and then found that was used in the RallyActor class, so reverted that - perhaps I managed to revert some of the parent changes too...will fix

esrally/driver/driver.py
gareth-ellis: That would make more sense, I hadn't thought about that

pquentin:
Thanks! LGTM. You will have to approve yourself as I can't approve my own pull request.

gareth-ellis (approved):

PR #1754, opened 2023-08-01
Upgraded Elasticsearch python client version to 8.9.0

maxhniebergall:
This change is required to incorporate an API change in the 8.9 version of the elasticsearch python client, which is needed for a new rally track I am creating

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ dependencies = [
     #   urllib3: MIT
     #   aiohttp: Apache 2.0
 
-    "elasticsearch[async]==8.6.1",
+    "elasticsearch[async]==8.9.0",
     "elastic-transport==8.4.0",
     "urllib3==1.26.9",
     "docker==6.0.0",


b-deam:
Hi @maxhniebergall - thanks for the PR. 

Unfortunately the python client version can be/is a sensitive area in the codebase, as there's a chance that the upgrade introduces difficult to debug performance regressions, even upgrading a seemingly unrelated dependency recently caused us to spend a week debugging outlier query latencies in our nightly benchmarks (see https://github.com/elastic/rally/pull/1728). 

This means that we need to spend quite some time running benchmarks and testing to ensure there's no subtle regressions introduced via the client upgrade, which takes quite some time. 

> This change is required to incorporate an API change in the 8.9 version of the elasticsearch python client, which is needed for a new rally track I am creating

Can you discuss this API change a bit more? The client has an 'escape hatch' in `client.perform_request()` that allows you to craft your own request? Here's an example:
https://github.com/elastic/rally/blob/2479411283a36de362ef5e08725fe2454717f951/esrally/driver/runner.py#L1891-L1895

maxhniebergall:
@b-deam, thanks for letting me know about the performance implications. I will try to use the perform request method instead.

PR #1755, opened 2023-08-02
Add blob-store-stats telemetry device and tests

b-deam:
This commit adds a new `blob-store-stats` telemetry device that based on the sample rate collects per-cluster, and per-node blob store statistics as they are reported by the respective API.

You can test this yourself with:
```
$ esrally race [...] --telemetry='blob-store-stats'
```

And then checking in the configured remote metrics store:
image

Notes:
- This is intentionally unlisted from `esrally list telemetry` given the elevated privileges required to collect these metrics
  - Running without the correct privleges will show:
   ```
  elasticsearch.NotFoundError: NotFoundError(404, 'Request for uri [/_internal/blob_store/stats] with method [GET] exists but is not available when running in serverless mode', 'Request for uri [/_internal/blob_store/stats] with method [GET] exists but is not available when running in serverless mode')
  ```
- The API doesn't return node names, only IDs. Refactoring the existing metadata collection or adding a separate call to the Info API to allow us to do lookups seems like overkill here

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2386,7 +2386,7 @@ class BlobStoreStats(TelemetryDevice):
         self.sample_interval = telemetry_params.get("blob-store-stats-sample-interval", 1)
         if self.sample_interval <= 0:
             raise exceptions.SystemSetupError(
-                f"The telemetry parameter 'blob-store-stats-sample-interval' must be greater than zero " f"but was {self.sample_interval}."
+                f"The telemetry parameter 'blob-store-stats-sample-interval' must be greater than zero but was {self.sample_interval}."
             )
         self.specified_cluster_names = self.clients.keys()
         self.metrics_store = metrics_store
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2359,3 +2359,143 @@ class DiskUsageStats(TelemetryDevice):
                 knn_vectors = field_info.get("knn_vectors_in_bytes", 0)
                 if knn_vectors > 0:
                     self.metrics_store.put_value_cluster_level("disk_usage_knn_vectors", knn_vectors, meta_data=meta, unit="byte")
+
+
+class BlobStoreStats(TelemetryDevice):
+    internal = False
+    command = "blob-store-stats"
+    human_name = "Blob Store Stats"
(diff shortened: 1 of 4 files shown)


inqueue (changes requested):
My tests with `--telemetry=blob-store-stats` don't show Rally recording results to the metrics store just yet.

b-deam:
> My tests with `--telemetry=blob-store-stats` don't show Rally recording results to the metrics store just yet.

Hmm, mine is working.. can you try filtering for `name: blob-store-stats` on the `rally-metrics*` index pattern?

`esrally race --pipeline=benchmark-only --telemetry='node-stats, recovery-stats, segment-stats, shard-stats, "blob-store-stats]`:
image

esrally/telemetry.py
```diff
@@ -2359,3 +2359,143 @@ def handle_telemetry_usage(self, response):
...
+        :param metrics_store: The configured metrics store we write to.
+        """
+        super().__init__()
+
+        self.telemetry_params = telemetry_params
+        self.clients = clients
+        self.sample_interval = telemetry_params.get("blob-store-stats-sample-interval", 1)
+        if self.sample_interval <= 0:
+            raise exceptions.SystemSetupError(
+                f"The telemetry parameter 'blob-store-stats-sample-interval' must be greater than zero " f"but was {self.sample_interval}."
```
gbanasiak, line 2389: Nit: why the split in 2 string blocks?

gbanasiak (approved):
LGTM. I've tested it with both serverless and non-serverless and got expected results.

esrally/telemetry.py
dliappis: I have seen `make lint|format` (i.e. [black](https://github.com/elastic/rally/blob/a7387ae7ddc53098d640130900a6f5cf2aa5c85e/.pre-commit-config.yaml#L11-L14) automatically do that.

pquentin: Right it's unlikely to be intentional but would be nice to fix just before merging (does not need a new review)

b-deam: Just keeping you all on your toes.

pquentin (approved):

PR #1756, opened 2023-08-03
Retrieve serverless build hash from nodes info API

gbanasiak:
This PR adds retrieval of build hash for serverless clusters from nodes info API (`_nodes?filter_path=**.build_hash`).

The PR introduces 2 configuration settings under `driver` section:
* `serverless.mode` - equals `true` if Rally targets serverless cluster
* `serverless.operator` - equals `true` if Elasticsearch user has operator privileges

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -706,8 +706,8 @@ class Driver:
         else:
             self.wait_for_rest_api(es_clients)
             self.target.cluster_details = self.retrieve_cluster_info(es_clients)
-            serverless_mode = self.config.opts("driver", "serverless.mode", default_value=False)
-            serverless_operator = self.config.opts("driver", "serverless.operator", default_value=False)
+            serverless_mode = self.config.opts("driver", "serverless.mode", default_value=False, mandatory=False)
+            serverless_operator = self.config.opts("driver", "serverless.operator", default_value=False, mandatory=False)
             if serverless_mode and serverless_operator:
                 build_hash = self.retrieve_build_hash_from_nodes_info(es_clients)
                 self.logger.info("Retrieved actual build hash [%s] from serverless cluster.", build_hash)
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -47,7 +47,7 @@ from esrally import (
 from esrally.client import delete_api_keys
 from esrally.driver import runner, scheduler
 from esrally.track import TrackProcessorRegistry, load_track, load_track_plugins
-from esrally.utils import console, convert, net, versions
+from esrally.utils import console, convert, net
 
 
 ##################################
@@ -710,7 +710,7 @@ class Driver:
             serverless_operator = self.config.opts("driver", "serverless.operator", default_value=False)
             if serverless_mode and serverless_operator:
                 build_hash = self.retrieve_build_hash_from_nodes_info(es_clients)
-                self.logger.info(f"Retrieved actual build hash [{build_hash}] from serverless cluster.")
+                self.logger.info("Retrieved actual build hash [%s] from serverless cluster.", build_hash)
                 self.target.cluster_details["version"]["build_hash"] = build_hash
 
         # Avoid issuing any requests to the target cluster when static responses are enabled. The results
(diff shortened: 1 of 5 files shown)


esrally/driver/driver.py
```diff
@@ -661,6 +661,16 @@ def retrieve_cluster_info(self, es):
             self.logger.exception("Could not retrieve cluster info on benchmark start")
             return None
 
+    def retrieve_build_hash_from_nodes_info(self, es):
+        try:
+            nodes_info = es["default"].nodes.info(filter_path="**.build_hash")
+            nodes = list(nodes_info["nodes"].keys())
```
inqueue, line 667: What do you think about getting the build hash this way instead of populating a list?

```python
In [52]: next(iter(nodes_info["nodes"].items()))[1]["build_hash"]
Out[52]: '4890e5b6b19f9a20c5f62318cf600be156fbc5e3'
```

inqueue (approved):
I left one suggestion. Otherwise, another review round is not needed. LGTM

esrally/driver/driver.py
b-deam: I find that much harder to read than the current implementation.

ebadyano:
Thank you for the change! Does it make sense to add a test in `driver_test.py`?

gbanasiak:
@ebadyano 

> Does it make sense to add a test in `driver_test.py`?

Sure, added in https://github.com/elastic/rally/pull/1756/commits/b9a2ab8752a458a2cd29003756a24bfec0a0bd54. Please take a look.

esrally/driver/driver.py
```diff
@@ -696,6 +707,12 @@ def prepare_benchmark(self, t):
         else:
             self.wait_for_rest_api(es_clients)
             self.target.cluster_details = self.retrieve_cluster_info(es_clients)
+            serverless_mode = self.config.opts("driver", "serverless.mode", default_value=False, mandatory=False)
+            serverless_operator = self.config.opts("driver", "serverless.operator", default_value=False, mandatory=False)
+            if serverless_mode and serverless_operator:
+                build_hash = self.retrieve_build_hash_from_nodes_info(es_clients)
+                self.logger.info("Retrieved actual build hash [%s] from serverless cluster.", build_hash)
+                self.target.cluster_details["version"]["build_hash"] = build_hash
```
dliappis, line 715: What about the stats in the metric store?

There is some code in https://github.com/elastic/rally/blob/c3b04f4d97de7c379bd78905d83df8bc71c7d229/esrally/telemetry.py#L1778-L1781 and in my tests it results in

image

(correctly)

but 

image

whereas I see in the logs `Retrieved actual build hash [ad5c80e1b49f10b42d438253d7cc0b9753c156b9] from serverless cluster.`

Additionally, shouldn't it also be collected by the [node stats telemetry device](https://github.com/elastic/rally/blob/c3b04f4d97de7c379bd78905d83df8bc71c7d229/esrally/telemetry.py#L779-L842)?

dliappis (commented):
Thank you. This seems to work but I have some questions around populating the retrieved field in the remote metric store.

esrally/driver/driver.py
gbanasiak: Telemetry device modifications are scoped separately in https://elasticco.atlassian.net/browse/ES-6459. After discussion with @dliappis we thought the necessary changes to override build hash in `ClusterEnvironmentInfo` are small, so I went ahead and added [da34a7d](https://github.com/elastic/rally/pull/1756/commits/da34a7d843237e1c387e081b3949f37e31afe9fb). PTAL.

esrally/telemetry.py
```diff
@@ -1774,8 +1775,10 @@ def on_benchmark_start(self):
             self.logger.exception("Could not retrieve cluster version info")
             return
         distribution_flavor = client_info["version"].get("build_flavor", "oss")
-        # build hash will only be available on serverless if the client has operator privs
+        # serverless returns dummy build hash which gets overridden when running with operator privileges
```
dliappis, line 1778: I'd add a `# TODO` here that we'll likely refactor `Telemetry()` to accept the config object and allow handling the `revision` based on `driver.serverless` settings

dliappis (approved):
Thanks for iterating. LGTM.

esrally/telemetry.py
gbanasiak: Addressed in https://github.com/elastic/rally/pull/1756/commits/bbd8cabe478798b71b6623526f4a22f356af8eb5.

PR #1759, opened 2023-08-09
Ensure serverless.mode/operator are always set

pquentin:
And invert one condition to avoid using `not`.

The idea is to avoid having to encode the default value (with the annoying `mandatory=False`) in every place where we will read those configuration values.


gbanasiak (approved):
LGTM

PR #1761, opened 2023-08-15
Update blob store stats telemetry device

b-deam:
The response format of the blob store stats API has changed, this PR updates the telemetry device to handle this, as well as flatten the metric fields. 

Technically a breaking change, but it's been done within releases, and this is an internal device anyway.

Relates https://github.com/elastic/rally/pull/1755


inqueue (approved):
I am seeing blob stats data in a run with this PR branch. LGTM

PR #1762, opened 2023-08-16
Sort AUTHORS file using en_US.UTF-8

pquentin:
This achieves two things: making the output stable and avoiding ASCIIbetical order where lower case authors are put last.


inqueue (approved):
LGTM, much better

gareth-ellis (approved):
LGTM

Issue #1752, opened 2023-07-25
Track execution can hang with multiple load generator machines


b-deam:
**Rally version** (get with `esrally --version`): `esrally 2.8.1.dev0 (git revision: e39834d1b7cb3062f36976c647fde27d6ee31a0e)`

**Invoked command**:

```
esrally race --pipeline=benchmark-only --track-repository="default" --track="elastic/logs" --challenge="logging-indexing" --exclude-tasks='["compression-stats"]' --telemetry='["node-stats", "recovery-stats", "segment-stats", "shard-stats"]' --on-error="continue" --target-hosts=~/.esbench/environments/fe92f286-5dd0-40ce-bb30-ece43f4d78b1/simple-experiment-0-target-hosts.json --client-options=~/.esbench/environments/fe92f286-5dd0-40ce-bb30-ece43f4d78b1/simple-experiment-0-client-options.json --track-params=~/.esbench/environments/fe92f286-5dd0-40ce-bb30-ece43f4d78b1/simple-experiment-0-track-params.json --telemetry-params=~/.esbench/environments/fe92f286-5dd0-40ce-bb30-ece43f4d78b1/simple-experiment-0-telemetry-params.json --user-tags=~/.esbench/environments/fe92f286-5dd0-40ce-bb30-ece43f4d78b1/simple-experiment-0-user-tags.json --race-id=7f23da13-b98b-4617-b3fc-10a362767545 --load-driver-hosts=10.10.63.127,10.10.65.177
```

**OS version**:
`Linux rally-0 5.4.0-1083-aws #90~18.04.1-Ubuntu SMP Fri Aug 5 08:15:25 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux`

**Description of the problem including expected versus actual behavior**:

When running the `elastic/logs` track with multiple load drivers, the execution hangs after this message:
```
[INFO] Racing on track [elastic/logs], challenge [logging-indexing] and car ['external'] with version [8.10.0-SNAPSHOT].
```

**Steps to reproduce**:
 1.  Setup multiple loaddrivers
 2. Execute the elastic/logs track
 3. Wait for it to hang after data generation

**Provide logs (if relevant)**:

This issue looks similar to https://github.com/elastic/rally/issues/1206, if we look at the coordinator actor logs we see:
```
[INFO] Preparing file offset table for [/home/esbench/.rally/benchmarks/data/elastic/logs/redis-app-slowlogs/document-0.json] ... [OK]
[...]

PR #1763, opened 2023-08-17
Add 'Bootstrap' message handler to 'Worker' actor

b-deam:
This commit adds a 'new' `Bootstrap` message handler to the `Worker` actor, which is responsible for ensuring that the new actor has the correct configuration and track modules loaded into its system path. This restores the ability to run tracks across multiple load driver machines.

The reason tracks worked across `Worker` actors on coordinator machine (where the race was invoked) is because the default behaviour for new actors created with `multiprocTCPBase` is that they fork (in the Unix sense) from their parent process. When a process is forked, it inherits the memory and other properties from the parent, including the system path. 

When a new `Worker` was created on a remote load driver machine the existing parent was missing paths, meaning there were missing modules for the load path.

Notes:
- This was actually supposed to have been 'fixed' in https://github.com/elastic/rally/pull/1315, but there was a [misnamed handler called `receiveMsg_RallyConfig` that was unused](https://github.com/elastic/rally/pull/1315/files#diff-7c569d2e27cf4d285afa9e91cf4e6919e9f50c8fda95de56e1a3bcf4b3554aedR1122-R1125), meaning it was unlikely this ever worked  
- This dead code was coincidentally removed only days ago in https://github.com/elastic/rally/pull/1749

Fixes https://github.com/elastic/rally/issues/1752
Relates https://github.com/elastic/rally/issues/1206

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -24,6 +24,7 @@ import logging
 import math
 import multiprocessing
 import queue
+import sys
 import threading
 import time
 from dataclasses import dataclass
@@ -305,7 +306,9 @@ class DriverActor(actor.RallyActor):
             self.wakeupAfter(datetime.timedelta(seconds=DriverActor.WAKEUP_INTERVAL_SECONDS))
 
     def create_client(self, host, cfg):
-        return self.createActor(Worker, targetActorRequirements=self._requirements(host))
+        worker = self.createActor(Worker, targetActorRequirements=self._requirements(host))
+        self.send(worker, Bootstrap(cfg))
+        return worker
 
     def start_worker(self, driver, worker_id, cfg, track, allocations, client_contexts=None):
         self.send(driver, StartWorker(worker_id, cfg, track, allocations, client_contexts))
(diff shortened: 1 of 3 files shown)


esrally/driver/driver.py
```diff
@@ -1210,6 +1213,14 @@ def __init__(self):
         self.wakeup_interval = Worker.WAKEUP_INTERVAL_SECONDS
         self.sample_queue_size = None
 
+    @actor.no_retry("worker")  # pylint: disable=no-value-for-parameter
+    def receiveMsg_Bootstrap(self, msg, sender):
+        self.driver_actor = sender
+        # load node-specific config to have correct paths available
+        self.cfg = load_local_config(msg.config)
```
pquentin, line 1220: Do we really need to attach those to `self`? If we do we should set them to `None` in `__init__` but it looks they're only used locally.

Additionally, I think having both `cfg` and `msg.config` is confusing. What do you think about `node_config` instead of `cfg`?

pquentin (approved):
Wow, that's a nice catch! I left a small nit about naming.

esrally/driver/driver.py
b-deam: Good points! I refactored these attributes in https://github.com/elastic/rally/pull/1763/commits/34566f1fc0d8ccc237af64b4a78f3ab452a1fa8b as I noticed we were needlessly setting the config twice otherwise.

b-deam:
buildkite test this please

b-deam:
Hmm, there's some race condition here related to multiple workers on the same host all loading the track (i.e. running git operations) in parallel:
```
2023-08-21 04:42:59,804 ActorAddr-(T|:61276)/PID:44024 esrally.utils.process INFO fatal: Unable to create '/Users/bradleydeam/.rally/benchmarks/tracks/rally-tracks-compat/.git/index.lock': File exists.
```

I'll work out a fix.

b-deam:
Confirmed working as of latest commit:
image

PR #1764, opened 2023-08-17
Rename master -> driver_actor

gareth-ellis:
Spotted another less clear variable name, this fixes that.


pquentin (approved):
Thanks, LGTM!

PR #1765, opened 2023-08-21
Upload Rally to PyPI using Trusted Publishers

pquentin:


b-deam (approved):
I tested the release script, which LGTM. But I'm unfamiliar with GH Actions, so my review there is worth taking with a very large grain of salt.

pquentin:
Grain of salt taken, thanks!

image

PR #1768, opened 2023-08-23
Detect serverless operator status automatically

pquentin:
This can be tested like https://github.com/elastic/rally/pull/1760, but you no longer need to explicitly set the operator status in rally.ini, instead it should be detected automatically.

### DIFF
--- a/esrally/client/factory.py
+++ b/esrally/client/factory.py
@@ -336,8 +336,9 @@ def cluster_distribution_version(hosts, client_options, client_factory=EsClientF
     """
     # no way for us to know whether we're talking to a serverless elasticsearch or not, so we default to the sync client
     es = client_factory(hosts, client_options).create()
-    # unconditionally wait for the REST layer - if it's not up by then, we'll intentionally raise the original error
-    wait_for_rest_layer(es)
+
+    # wait_for_rest_layer  calls the Cluster Health API, which is not available for unprivileged users on Serverless
+    # As a result, we need to call the info API first to know if we can call wait_for_rest_layer().
     version = es.info()["version"]
 
     version_build_flavor = version.get("build_flavor", "oss")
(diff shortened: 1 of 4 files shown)


esrally/client/factory.py
```diff
@@ -346,7 +347,16 @@ def cluster_distribution_version(hosts, client_options, client_factory=EsClientF
     # version number does not exist for serverless
     version_number = version.get("number", version_build_flavor)
 
-    return version_build_flavor, version_number, version_build_hash
+    serverless_operator = False
+    if version_build_flavor == "serverless":
```
b-deam, line 351: Can we use the `is_serverless` helper function for this?
https://github.com/elastic/rally/blob/b0b5e05a40552bc0a6a0d2f2191b95c56327c23d/esrally/utils/versions.py#L36-L37

esrally/client/factory.py
```diff
@@ -346,7 +347,16 @@ def cluster_distribution_version(hosts, client_options, client_factory=EsClientF
...
+        authentication_info = es.perform_request(method="GET", path="/_security/_authenticate")
+        serverless_operator = authentication_info.body.get("operator", False)
+
+    if not version_build_flavor == "serverless" or serverless_operator is True:
```
b-deam, line 355: same as https://github.com/elastic/rally/pull/1768/files#r1303824651

esrally/client/factory.py
```diff
@@ -336,8 +336,9 @@ def cluster_distribution_version(hosts, client_options, client_factory=EsClientF
     """
     # no way for us to know whether we're talking to a serverless elasticsearch or not, so we default to the sync client
     es = client_factory(hosts, client_options).create()
-    # unconditionally wait for the REST layer - if it's not up by then, we'll intentionally raise the original error
-    wait_for_rest_layer(es)
+
+    # wait_for_rest_layer  calls the Cluster Health API, which is not available for unprivileged users on Serverless
```
b-deam, line 340: Can we please update the docstring to include the new `serverless_operator` return value?

b-deam (approved):
I get the sense that it might be better to split out the retrieval of the versions, operator status, and REST API health check into discrete functions, but I don't think it's worth blocking this PR.

gareth-ellis (approved):
LGTM

b-deam:
> > I get the sense that it might be better to split out the retrieval of the versions, operator status, and REST API health check into discrete functions, but I don't think it's worth blocking this PR.
> 
> Well this is definitely quite ugly. I want to use a single Elasticsearch client here but did not want to affect the stateful uses of this function too much. This is the best compromise I found. However, if we think the logic of this function is fine, then I'm not convinced separating it into other functions would have helped that much. The logic is complicated but 20 lines is not that much in my opinion, even though Uncle Bob would disagree.

I agree, no need for any changes. :shipit:

PR #1770, opened 2023-08-25
Exclude telemetry devices based on serverless status

pquentin:
Can be tested using:

```
esrally race --track-path=$HOME/src/rally-tracks/geonames --test-mode --target-hosts=... --pipeline=benchmark-only --client-options=serverless.json --on-error=abort --telemetry=ccr-stats
```

This will run the race and print:

```
[INFO] Excluding telemetry device [ccr-stats] as it is unavailable on serverless.                                                         
```

To test without operator status, you should either include https://github.com/elastic/rally/pull/1768 or set `serverless.operator` to `false` in the rally.ini `[driver]` section.

TODO:

- [x] Tests
- [x] Figure out why the message is printed twice

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -93,6 +93,7 @@ class Telemetry:
         for device in self.devices:
             if self._enabled(device):
                 if self.serverless_mode and not self._available_on_serverless(device):
+                    # Only inform about exclusion if the user explicitly asked for this device
                     if getattr(device, "command", None) in self.enabled_devices:
                         console.info(f"Excluding telemetry device [{device.command}] as it is unavailable on serverless.")
                     continue
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -92,9 +92,12 @@ class Telemetry:
     def on_benchmark_start(self):
         for device in self.devices:
             if self._enabled(device):
+                if self.serverless_mode and not self._available_on_serverless(device):
+                    if device.command in self.enabled_devices:
+                        console.info(f"Excluding telemetry device [{device.command}] as it is unavailable on serverless.")
+                    continue
+
                 device.on_benchmark_start()
-            if self._excluded_on_serverless(device):
-                console.info(f"Excluding telemetry device [{device.command}] as it is unavailable on serverless.")
 
     def on_benchmark_stop(self):
         for device in self.devices:
@@ -108,14 +111,6 @@ class Telemetry:
 
     def _enabled(self, device):
         return device.internal or device.command in self.enabled_devices
-
-    def _excluded_on_serverless(self, device):
-        return (
-            self.serverless_mode
-            and not device.internal
-            and device.command in self.enabled_devices
-            and not self._available_on_serverless(device)
-        )
 
     def _available_on_serverless(self, device):
         if self.serverless_operator:
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -25,7 +25,7 @@ import tabulate
 
 from esrally import exceptions, metrics, time
 from esrally.metrics import MetaInfoScope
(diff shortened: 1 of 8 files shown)


pquentin:
To be clear, here is the current output:

```
[INFO] Race id is [de8d87c7-26d2-4caa-a8b6-5747281babb1]                                                                                                       
[INFO] Treating parallel task in challenge [indexing-querying] as public.                                                                                      
[INFO] Excluding [put-settings], [check-cluster-health], [force-merge], [wait-until-merges-finish] as challenge [indexing-querying] is run on serverless.      
[INFO] Racing on track [pmc], challenge [indexing-querying] and car ['external'] with version [8.10.0].                                                        
                                                                                                                
[INFO] Excluding telemetry device [ccr-stats] as it is unavailable on serverless.       
Running delete-index                                                           [100% done]
Running create-index                                                           [100% done]
Running initial-index-append                                                   [100% done]
Running refresh-after-index                                                    [100% done]
Running refresh-after-force-merge                                              [100% done]
Running phrase,articles_monthly_agg_uncached,default,term,articles_mont...     [100% done]
[INFO] Excluding telemetry device [ccr-stats] as it is unavailable on serverless.       
```

The empty line between "Racing" and "Excluding" is annoying, but I want to remove the second "Excluding" message between the  "Running" messages and the summary report.

esrally/telemetry.py
```diff
@@ -91,6 +93,8 @@ def on_benchmark_start(self):
         for device in self.devices:
             if self._enabled(device):
                 device.on_benchmark_start()
+            if self._excluded_on_serverless(device):
```
b-deam, line 96: Don't we want this before calling `on_benchmark_start()`, and to also `continue`?

pquentin: Oh, of course, thanks. I was so focused on the message that I forgot the functionality. :sweat_smile: 

I'll leave it like that for now to make sure my tests catch it.

pquentin:
OK, it should be correct now! A nice win for test-driven development. Please take another look.

esrally/telemetry.py
```diff
@@ -92,9 +92,12 @@ def detach_from_node(self, node, running):
     def on_benchmark_start(self):
         for device in self.devices:
             if self._enabled(device):
+                if self.serverless_mode and not self._available_on_serverless(device):
+                    if device.command in self.enabled_devices:
```
b-deam: I find this logic confusing. Isn't this check redundant given that `self._enabled` checks this already?

esrally/telemetry.py
```diff
@@ -90,6 +92,11 @@ def detach_from_node(self, node, running):
     def on_benchmark_start(self):
         for device in self.devices:
             if self._enabled(device):
+                if self.serverless_mode and not self._available_on_serverless(device):
+                    if getattr(device, "command", None) in self.enabled_devices:
```
b-deam, line 96: I find this logic confusing. Isn't this check redundant given that `self._enabled` checks this already?

pquentin: This isn't redundant because `self._enabled` also checks if the command is internal. But I don't want to print a message for those internal commands. If we were not printing a message, we would not need this condition at all: notice that the `continue` is outside it. I've added [`57b2d5d` (#1770)](https://github.com/elastic/rally/pull/1770/commits/57b2d5d406321aef999fc520eb80b10bb37ee45e) to clarify this. What do you think?

b-deam: Ah, ofc. Thanks for clarifying. The comment helps!

b-deam (approved):
LGTM! :shipit:

esrally/telemetry.py
pquentin: Thanks! I would not say "of course", it took me way too many attempts to get to that version. :sweat_smile:

PR #1775, opened 2023-09-07
Test Rally with operator on Serverless during CI

pquentin:


.buildkite/it/run_serverless.sh
```diff
@@ -0,0 +1,66 @@
...
+export TERM=dumb
+export LC_ALL=en_US.UTF-8
+export TZ=Etc/UTC
+export DEBIAN_FRONTEND=noninteractive
+# https://askubuntu.com/questions/1367139/apt-get-upgrade-auto-restart-services
+sudo mkdir -p /etc/needrestart
+echo "\$nrconf{restart} = 'a';" | sudo tee -a /etc/needrestart/needrestart.conf > /dev/null
+
+# based on https://gist.github.com/sj26/88e1c6584397bb7c13bd11108a579746?permalink_comment_id=4155247#gistcomment-4155247
+function retry {
```
gareth-ellis, line 14: It could be good to add this to a seperate file and source, since the same function is used in run.sh

gareth-ellis (approved):
LGTM - Thanks!

it_serverless/serverless_test.py
```diff
@@ -0,0 +1,137 @@
...
+    credentials = serverless_api(
+        "POST",
+        f"/api/v1/serverless/projects/elasticsearch/{serverless_project['id']}{GET_CREDENTIALS_ENDPOINT}",
+    )
+
+    es_endpoint = serverless_project["endpoints"]["elasticsearch"]
+    es_hostname = urllib.parse.urlparse(es_endpoint).hostname
+    rally_target_host = f"{es_hostname}:443"
+
+    print("Waiting DNS propagation")
```
gbanasiak, line 70:
```suggestion
    print("Waiting for DNS propagation")
```

it_serverless/serverless_test.py
```diff
@@ -0,0 +1,137 @@
...
+            print(json.dumps(info.body, indent=2))
+
+            authenticate = es.perform_request(method="GET", path="/_security/_authenticate")
+            print("GET /_security/_authenticate")
+            print(json.dumps(authenticate.body, indent=2))
+
+            break
+        except Exception as e:
+            print(f"GET / Failed with {type(e)}")
+            time.sleep(10)
```
gbanasiak, line 99: There's a potential of infinite wait here I think. Can we limit it time-wise?

gbanasiak (commented):
Thank you. Should there also be a test with Python 3.8 by analogy to non-serverless pipeline? Please also check small remarks inline.

it_serverless/serverless_test.py
pquentin: Thanks, done in [`80be169` (#1775)](https://github.com/elastic/rally/pull/1775/commits/80be169d84598f3d88a56e9c3b28406fefb11f51)

pquentin:
> Thank you. Should there also be a test with Python 3.8 by analogy to non-serverless pipeline?

I would rather not do this, because this is specifically about testing serverless, and it costs resources (we spin up a project on Elastic Cloud). This is also in line with the idea that Rally should eventually support a single Python version.

gbanasiak (approved):
Thank you, LGTM.

PR #1779, opened 2023-09-13
Simplify ClusterEnvironmentInfo telemetry device

gbanasiak:
`ClusterEnvironmentInfo` telemetry device unnecessarily calls nodes stats API to populate metadata fields in a metric store which duplicates `ExternalEnvironmentInfo`. As `ClusterEnvironmentInfo` is not tagged with `serverless.Status.Internal` status, the additional bonus of the simplification is unblocking Rally when ran with public serverless privileges.

I've tested the change comparing metric store content (`rally-metrics-*`) with/without simplification. On each run the amount of documents with the following fields was equal:
* serverless cluster with operator privileges - `meta.jvm_vendor` field,
* stateful ESS cluster with superuser privileges - `meta.jvm_vendor` and `meta.attribute_region` fields.

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -1707,7 +1707,7 @@ class DiskIo(InternalTelemetryDevice):
 
 
 def store_node_attribute_metadata(metrics_store, nodes_info):
-    # push up all node level attributes to cluster level if the values are identical for all nodes
+    # push up all node level attributes to cluster level iff the values are identical for all nodes
     pseudo_cluster_attributes = {}
     for node in nodes_info:
         if "attributes" in node:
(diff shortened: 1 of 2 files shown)


esrally/telemetry.py
```diff
@@ -1707,7 +1707,7 @@ def store_system_metrics(self, node, metrics_store):
 
 
 def store_node_attribute_metadata(metrics_store, nodes_info):
-    # push up all node level attributes to cluster level iff the values are identical for all nodes
+    # push up all node level attributes to cluster level if the values are identical for all nodes
```
pquentin, line 1710: iff is mathematical notation for "if and only if" that @danielmitterdorfer liked using in Rally but I agree it does not add value here

pquentin (approved):
Thanks! LGTM.

esrally/telemetry.py
gbanasiak: Thank you for this useful context. I decided to revert this comment change for consistency in https://github.com/elastic/rally/pull/1779/commits/81348f48eaafa2d70d69d81d251c84b6ca6b9977.

pquentin (approved):

PR #1780, opened 2023-09-14
Bump version to 2.9.1

pquentin:
@Leaf-Lin has hit https://github.com/elastic/rally/pull/1772 and the other commits can help with serverless testing (even though support is not complete yet).


gareth-ellis (approved):
LGTM

PR #1782, opened 2023-09-22
Add public user to serverless test

pquentin:
Rally does not currently support serverless public users, which is why we use `pytest.xfail` (which was simpler to use than `pytest.mark.xfail` due to the parametrization). Usefully, it does print the error, which is currently:

```
elasticsearch.NotFoundError: NotFoundError(404, 'Request for uri [/_cluster/health?wait_for_nodes=%3E%3D1] with method [GET] exists but is not available when running in serverless mode', 'Request for uri [/_cluster/health?wait_for_nodes=%3E%3D1] with method [GET] exists but is not available when running in serverless mode')
```

I'd suggest reviewing the two commits separately as the first reorganizes the code.

### DIFF
--- a/.buildkite/pull-requests.json
+++ b/.buildkite/pull-requests.json
@@ -4,6 +4,11 @@
       "enabled": true,
       "pipeline_slug": "rally-it",
       "allow_org_users": true
+    },
+    {
+      "enabled": true,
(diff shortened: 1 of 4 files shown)


gbanasiak (approved):
Thank you. LGTM.

PR #1783, opened 2023-09-22
Run Rally IT Serverless pipeline on pull requests

pquentin:
Noticed in https://github.com/elastic/rally/pull/1782 that the pipeline does not run on pull requests from forks.


gareth-ellis (approved):
LGTM

PR #1786, opened 2023-09-27
Add support for zstd-compression

### LINKED ISSUE #1781: Add support for zstd-compressed corpora
Rally supports various compression formats such as gz or bzip. It does not support the [zstd format](https://github.com/facebook/zstd) which is perfoming significantly better in disk usage and decompression speed in my experiments. I've compressed 183GB corpus with `pbzip2` and `pzstd`, both with the maximum compression level that is supported by the respective tool.

|Format|Size on disk [GB]|Size on disk [GB]|Relative size [%]|
|------|-----------------|-----------------|-----------------|
|bzip  |18613471805      |18               |100              |
|zstd  |11215205385      |11               |60               |

Also decompression speed is vastly superior (times measured with `time`, table contains the output of `real`, i.e. wall clock time):

|Format|Time to decompress [s]|Relative time [%]|
|------|----------------------|-----------------|
|bzip  |388                   |100              |
|zstd  |144                   |36               |

Therefore I propose to add support for zstd compression to Rally similar to bzip support: The fast option would require `pzstd` to be on `PATH` and a fallback can be based on the [Python zstd implementation](https://pypi.org/project/zstd/).

For reference:

* Compress data: `pzstd -19 corpus.json -o corpus.json.zstd` (`19` denotes the maximum compression level)
* Decompress data: `pzstd -d corpus.json.zstd -o corpus.json`

danielmitterdorfer:
With this commit we add support for zstd compressed corpora. Compared to bzip, the zstd format produces compressed files that are roughly 40% smaller and took around a third of the time to decompress in our tests.

Closes #1781


pquentin (approved):
I haven't tested it but I assume you have and the code looks good to me.

I also checked that you're using the fastest python-zstandard API and the only one that supports reading across ZSTD frames. (This was an issue in urllib3 which had to use the standard library abstraction which is slower and where we need to pass `read_across_frames=True` explicitly!)

danielmitterdorfer:
Thanks! Yes, I've tested both cases (with the library and the native binary). Also, the library I've picked seems to be the most mature of them.

PR #1785, opened 2023-09-26
Fix serverless public user in trivial tracks

gbanasiak:

PR #1789, opened 2023-10-06
Add serverless-aware runners

gbanasiak:
This PR passes configuration context to `Query`, `DeleteIndex` and `DeleteComposableTemplate` runners to make them serverless-aware and change their behaviour if Rally targets serverless cluster.

Changes:
* `Query`, `DeleteIndex` and `DeleteComposableTemplate` stop getting and setting cluster settings through `set_destructive_requires_name()` as cluster settings are not available to serverless users. In case of `DeleteIndex` this replaces the earlier crude workaround through `contextlib.suppress` introduced in https://github.com/elastic/rally/pull/1785.
* The `DeleteComposableTemplate` runner starts to ignore `delete_matching_indices` template option. Wildcard deletes are not available to serverless users (!). `DeleteIndexTemplate` runner is not fixed as legacy templates are not available in serverless.
* **Request cache becomes turned on by default for serverless non-operator use in the `Query` runner** (`search`, `paginated-search`, `composite-agg` and `scroll-search` operations). In non-serverless use we typically rely on `"index.requests.cache.enable": false` index setting for this purpose but it is not available to serverless users. As part of this change I've removed request params in `/_search/scroll` which is present in certain `Query` scenarios, as `/_search/scroll` does not accept `request_cache` parameter and it is the _only_ parameter used.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -47,10 +47,10 @@ def register_default_runners(config=None):
     register_runner(track.OperationType.ForceMerge, ForceMerge(), async_runner=True)
     register_runner(track.OperationType.IndexStats, Retry(IndicesStats()), async_runner=True)
     register_runner(track.OperationType.NodeStats, NodeStats(), async_runner=True)
-    register_runner(track.OperationType.Search, Query(config), async_runner=True)
-    register_runner(track.OperationType.PaginatedSearch, Query(config), async_runner=True)
-    register_runner(track.OperationType.CompositeAgg, Query(config), async_runner=True)
-    register_runner(track.OperationType.ScrollSearch, Query(config), async_runner=True)
+    register_runner(track.OperationType.Search, Query(config=config), async_runner=True)
+    register_runner(track.OperationType.PaginatedSearch, Query(config=config), async_runner=True)
+    register_runner(track.OperationType.CompositeAgg, Query(config=config), async_runner=True)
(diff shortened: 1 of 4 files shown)


esrally/driver/runner.py
```diff
@@ -230,6 +229,20 @@ def _transport_request_params(params):
...
+    Mixin to create configuration-aware runners
+    """
+
+    def __init__(self, config, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.serverless_mode = False
+        self.serverless_operator = False
+        if config:
+            self.serverless_mode = config.opts("driver", "serverless.mode", mandatory=False, default_value=False)
+            self.serverless_operator = config.opts("driver", "serverless.operator", mandatory=False, default_value=False)
```
b-deam, line 243: I haven't used multiple inheritance much before as I've always been told to avoid it at all costs. I don't know how we all feel about this, but presumably it wouldn't hurt to have all runners be 'configuration' aware, in which case we could just add this logic to `Runner` itself.. I think, I'm probably overlooking something obvious.

gbanasiak: I had `config` in `Runner` initially but had not found the right way to express it's optional character, and then got inspired by `Delegator` mixin. I went back to `Runner` in https://github.com/elastic/rally/pull/1789/commits/670de571987c35866d00737cbe647cacb889cb0b. PTAL.

b-deam (approved):

PR #1791, opened 2023-10-15
Add ESQL operator

dnhatn:
This pull request introduces the ES|QL operator. This operator, for now, supports only two main parameters: `query` and `filter`.


esrally/track/track.py
```diff
@@ -701,44 +701,45 @@ class OperationType(Enum):
     CompositeAgg = (18, AdminStatus.No, serverless.Status.Public)
     WaitForCurrentSnapshotsCreate = (19, AdminStatus.No, serverless.Status.Internal)
     Downsample = (20, AdminStatus.No, serverless.Status.Internal)
+    Esql = (21, AdminStatus.No, serverless.Status.Internal)
```
b-deam, line 704: This should be `serverless.Status.Public` if we expect users _without_ operator privileges to be able to use it, which I think we do.

dnhatn: Thanks, Brad! I chose `serverless.Status.Internal` because we haven't exposed this API in the serverless yet. However, I think it should be okay to use `serverless.Status.Public` here. I have adjusted this in https://github.com/elastic/rally/pull/1791/commits/a053ecce9e85cbbca795b5bac7e2925461840cf8.

b-deam: Oh, right. All this enum does is change the behaviour of Rally based on whether it is being executed against a serverless cluster with or without operator privileges. 

If you set 'Internal', then this operation is skipped in any track you run without operator privileges. If you set 'Public', then Rally will execute the operation regardless of the user's privileges. If the API isn't exposed yet, it means that this operation will fail.

dnhatn: Ok, I've reverted this back to `Internal`. We will change this to Public when we expose ESQL in serverless.

b-deam: Sorry for the back and forth here, but changing the enum could/may require us to cut a new Rally release, which I don't think is a huge problem, but it may be less work to just merge this as `Public` and accept that the operations will fail until we expose the API.

@gbanasiak - you're probably the person with the most experience around cutting releases and serverless changes, WDYT?

gbanasiak: Today `_query` endpoint and thus the new `esql` operation have `serverless.Status.Blocked` because the endpoint is available neither to serverless operator nor to a normal user. I would prefer to start with `Blocked` and come back and switch it to `Public` once API becomes exposed in serverless. We should be ready to cut a new Rally release quickly _just_ for this.

dnhatn: @gbanasiak That makes sense. I pushed https://github.com/elastic/rally/pull/1791/commits/fc7d959e10784dcae2c42f8a04b04ff4c9d9432b to use `Blocked`. Thank you!

esrally/driver/runner.py
```diff
@@ -2829,6 +2830,23 @@ def __repr__(self, *args, **kwargs):
         return "field-caps"
 
 
+class Esql(Runner):
+    async def __call__(self, es, params):
```
b-deam, line 2834: There are a couple of high level properties [that operations should expose](https://esrally.readthedocs.io/en/latest/track.html?highlight=request_timeout#operations), namely:
- `request-timeout` 
- `headers`
- `opaque-id`

You can add support for these by using the `Runner._transport_request_params` helper method:
https://github.com/elastic/rally/blob/fc7d959e10784dcae2c42f8a04b04ff4c9d9432b/esrally/driver/runner.py#L210-L231

An example of usage:
https://github.com/elastic/rally/blob/fc7d959e10784dcae2c42f8a04b04ff4c9d9432b/esrally/driver/runner.py#L1971

b-deam (approved):
LGTM. I left one comment about using a helper method around configuring the different request, transport, and header parameters but no need for another review.

dnhatn:
@b-deam @gbanasiak Thanks for reviews.

dnhatn:
@elasticmachine run rally/it-python310 please

docs/track.rst
```diff
@@ -3132,6 +3132,40 @@ Meta-data
...
+~~~~~~~~~~~~~
+
+With the operation type ``esql`` you can execute `ES|QL query <https://www.elastic.co/guide/en/elasticsearch/reference/master/esql.html>`_.
+
+Properties
+""""""""""
+
+* ``query`` (mandatory): An ES|QL query starts with a source command followed processing commands.
+* ``filter`` (optional): A query filter defined in `Elasticsearch query DSL <https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl.html>`_.
+* ``body`` (optional): The query body.
```
craigtaverner, line 3145: The description here should indicate that this is used for anything additions to put into the body, since the query will already be placed there. For example, we use this for `pragma`.

Issue #1354, opened 2021-10-06
test-mode support for source files with fewer than 1k documents


michaelbaamonde:
It may be useful for `test-mode` to relax its assumption that source files within corpora contain 1k documents. I say this for three reasons:

1.  New users trying out Rally for the first time may--as I did :smile: --run `esrally
     create-track` on a very small index in an already-running local Elasticsearch cluster. The resulting track will fail when run in `test-mode` due to the `document-count` being less than 1,000. The error message is rather clear, for what it's worth.
2.  While the docs do [clearly state](https://esrally.readthedocs.io/en/stable/adding_tracks.html?highlight=test-mode#adding-support-for-test-mode) that `test-mode` assumes that data files contain 1k documents, the error message that one receives if their track does not comply with this assumption is confusing. See [below](#misleading-error-message-example).
3.  There may be legitimate use cases for actually benchmarking workloads with corpora containing fewer than 1,000 documents. In order to use `test-mode` on tracks that define such workloads, users would have to jump some through hoops to set things up in a way that will satisfy its assumptions.

Relaxing this assumption would be simple to implement (I have a POC on a branch) if there's agreement that this is worth doing.

If there are good reasons to preserve this assumption, however, I'd advocate for a refactoring that would surface a clearer error in the scenario described in (2) above and illustrated below.

## Misleading error message

Here is a minimal track and source file that will produce the misleading error message. I tested this with `esrally 2.3.1.dev0 (git revision: f2d8a11)`.

<details>
  <summary>track.json</summary>
        
``` json
{
  "version": 2,
  "description": "Minimal track for illustrating --test-mode issue",
  "indices": [
    {
      "name": "test"
    }
  ],
  "corpora": [
    {
      "name": "test",
      "documents": [
        {
          "source-file": "documents.json",
          "document-count": 1,
          "uncompressed-bytes": 15
        }
      ]
    }
  ],
  "schedule": [
    {
      "operation": {
        "operation-type": "bulk",
        "bulk-size": 1,
        "target-index": "test"
      }
    }
  ]
}

```  

</details>

<details>
  <summary>documents.json</summary>
        
``` json
{"foo": "bar"}
```

</details>

Example Rally invocation:

``` bash
esrally race --distribution-version=7.15.0 --track-path=~/debug-test-mode --test-mode
```

Output (truncated):

``` example
[INFO] Race id is [70ab67cc-d623-4c6c-aa97-cf107dc85286]
[INFO] Preparing for race ...

[...]

PR #1792, opened 2023-10-16
Simplify test mode with small document sets

gbanasiak:
https://github.com/elastic/rally-tracks/pull/469 adds very small indices for enrichment purposes. This scenario is not supported by test mode which requires `-1k` files with 1000 documents. This leads to an awkward workaround where original corpus gets dummy content just to reach 1000 documents in `-1k` files.

This PR addresses this limitation by skipping the `-1k` requirement for document sets with 1000 and less documents. In such case, original document set file is taken when running in test mode.

### DIFF
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -299,12 +299,14 @@ Congratulations, you have created your first track! You can test it with ``esral
 Adding support for test mode
 ----------------------------
 
-You can check your track very quickly for syntax errors when you invoke Rally with ``--test-mode``. Rally postprocesses its internal track representation as follows:
+You can check your track very quickly for syntax errors when you invoke Rally with ``--test-mode``.
+
+In test mode Rally postprocesses its internal track representation as follows:
 
 * Iteration-based tasks run at most one warmup iteration and one measurement iteration.
 * Time-period-based tasks run at most for 10 seconds without warmup.
 
-Rally also post-processes all data file names in the internal track representation:
+In test mode Rally also post-processes all data file names:
 
 * If ``documents.json`` has 1000 documents or fewer, Rally uses it (no modifications).
 * If ``documents.json`` has more than 1000 documents, Rally assumes an additional ``documents-1k.json`` file is present and uses it. You need to prepare these additional files manually. Pick 1000 documents for every data file in your track and store them in a file with the ``-1k`` suffix. On Linux you can do it as follows: ``head -n 1000 documents.json > documents-1k.json``.
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -304,9 +304,10 @@ You can check your track very quickly for syntax errors when you invoke Rally wi
 * Iteration-based tasks run at most one warmup iteration and one measurement iteration.
 * Time-period-based tasks run at most for 10 seconds without warmup.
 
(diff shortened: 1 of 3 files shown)


inqueue (changes requested):

pquentin:
Is this closing <https://github.com/elastic/rally/issues/1354>?

gbanasiak:
> Is this closing https://github.com/elastic/rally/issues/1354?

Yes although addressing it wasn't the initial intention here - I wasn't aware this issue exists. Looking at the discussion there the novelty now is the arrival of a case where we have both big/regular corpus ("main" index) and very small corpus (enrichment indices) where test mode is still useful. This scenario wasn't thought of back in 2021, and I think it tips the scale towards the change of test mode behaviour.

docs/adding_tracks.rst
```diff
@@ -304,7 +304,9 @@ You can check your track very quickly for syntax errors when you invoke Rally wi
 * Iteration-based tasks run at most one warmup iteration and one measurement iteration.
 * Time-period-based tasks run at most for 10 seconds without warmup.
 
-Rally also postprocesses all data file names. Instead of ``documents.json``, Rally expects ``documents-1k.json`` and assumes the file contains 1.000 documents. You need to prepare these data files though. Pick 1.000 documents for every data file in your track and store them in a file with the suffix ``-1k``. We choose the first 1.000 with ``head -n 1000 documents.json > documents-1k.json``.
+Rally also post-processes all data file names. Instead of ``documents.json``, Rally expects ``documents-1k.json`` and assumes the file contains 1.000 documents. You need to prepare these data files though. Pick 1.000 documents for every data file in your track and store them in a file with the suffix ``-1k``. We choose the first 1.000 with ``head -n 1000 documents.json > documents-1k.json``.
+
+If your file has 1000 or fewer documents, you can skip creating the ``-1k`` variant. Rally will use the original file in such case.
```
gareth-ellis, line 309: > in such case

This doesn't sound right - I would probably say either in such a case (but this doesnt sound great either), in such cases or even just something like "Rally will use the original file instead"

Reading the code, it is in fact even that even if a -1k file exists, it will be ignored if there are less than 1000 documents, perhaps this should also be highlighted.

gareth-ellis (commented):
I think the idea of this change of behaviour is good - I have one comment about the wording in the readme

tests/track/loader_test.py
```diff
@@ -1327,11 +1338,22 @@ def test_post_processes_track_spec(self):
             ],
             "corpora": [
                 {
-                    "name": "unittest",
+                    "name": "unittest-more-than-1k-documents",
```
inqueue, line 1341: Can we rename this test?

```suggestion
                    "name": "unittest-1k-documents",
```

inqueue (approved):
I left one minor adjustment suggestion. Otherwise, LGTM!

tests/track/loader_test.py
gbanasiak: The idea of this test is to check how test mode processor changes the track spec. On entry there's more than 1k documents in this corpus, but after test mode is applied it gets reduced to 1k. What do you think about [e2b2ca0](https://github.com/elastic/rally/pull/1792/commits/e2b2ca084e871f6046285d0c30c5700d5e1098e4) wording?

docs/adding_tracks.rst
gbanasiak: WDYT about https://github.com/elastic/rally/pull/1792/commits/6675a6efbb17dfbcb8c6f358ac69ca24902dd4d6 ?

gareth-ellis: LGTM

gareth-ellis (approved):

PR #1795, opened 2023-11-02
Add props to prevent typos in config properties

### LINKED ISSUE #1646: DRY/Refactor hard coded config file keys
The motivation of this issue is to prevent bugs like #1645.

The Config object uses various keys for storing configuration properties. Currently many of those get populated during argument parsing (e.g. https://github.com/elastic/rally/blob/4c7141aac0179c38c08c134c25f7a23e6f07d2d9/esrally/rally.py#L1024-L1029), and the same keys are referenced, again as strings, in other code locations e.g. in [metrics.py](https://github.com/elastic/rally/blob/4c7141aac0179c38c08c134c25f7a23e6f07d2d9/esrally/metrics.py#L1497-L1544).

This approach is brittle and error prone. We should DRY things up and instead reference the config options via meaningful variable names in some module (e.g. [config.py](https://github.com/elastic/rally/blob/master/esrally/config.py)) such that a typo becomes a real error.

sakurai-youhei:
This PR will eliminate numerous strings around config by typo-proof enums.

In short,
`cfg.opts("benchmarks", "local.dataset.cache")`
will be rewritten to
`cfg.opts(*props.BENCKMARKS.LOCAL.DATASET.CACHE)`

The following points are aimed in the implementation:

1. Not require any changes in section and key names.
2. Support auto-completion by IDEs.
3. Keep the existing code width the same as possible.
4. Prepare the way to describe config properties in the code.
5. Enable to express both dot-notated sections and keys.
6. Make unpreferable unpacking fail on properties in the middle of dot notation.
  -> i.e. With `props.SECTION.DOT.NOTATED.KEY`, `func(*props.SECTION.DOT.NOTATED)` raises `TypeError`.

In return, the following restrictions are introduced:

- No way to declare the below kind of config keys.
  ```ini
  [section]
  spam = many
  spam.eggs = a few
  ```
  -> i.e. `func(*props.SECTION.SPAM)` raises `TypeError`; you can still write `func("section", "spam")`, though.
- Can't also express `camelCase`, `PascalCase`, `MACRO_CASE`, `kebab-case`, `COBOL-CASE`, and `bracket[notation]`.
- (possibly more)

Closes #1646

**TODOs**:

- [ ] Write unit tests.
- [ ] Write script to check typos between before and after strings elimination.
- [ ] Replace the rest of config `(section, key)` pairs.

### DIFF
--- a/esrally/props.py
+++ b/esrally/props.py
@@ -38,7 +38,7 @@ class PropEnum(str, Enum, metaclass=PropEnumType):
         if isinstance(value, PropEnumType):
             return value  # Nested Enum should stay original
         elif isinstance(value, str):
-            # This prefixing lets A.B.C.D values not just "d" but "a.b.c.d"
+            # This prefixing lets A.B.C.D value not just "d" but "a.b.c.d"
             value = f"{cls.__qualname__.lower()}.{value}"
             member = str.__new__(cls, value)
             member._value_ = value
--- a/esrally/props.py
+++ b/esrally/props.py
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from enum import auto, EnumType, member, StrEnum
+from enum import EnumType, StrEnum, auto, member
 
 
 class PropEnumType(EnumType):
@@ -38,7 +38,7 @@ class PropEnum(StrEnum, metaclass=PropEnumType):
 class Property(PropEnum):
     def __sections(self):
         node = globals()
-        for name in type(self).__qualname__.split('.'):
+        for name in type(self).__qualname__.split("."):
             node = node[name]
             if issubclass(node, Section):
                 yield name
@@ -54,7 +54,12 @@ class Property(PropEnum):
         return self.value.replace(f"{self.__section}.", "", 1)
 
     def __iter__(self):
-        return iter((self.__section, self.__key, ))
+        return iter(
+            (
+                self.__section,
+                self.__key,
+            )
+        )
 
     def __repr__(self):
         return f"<{type(self).__qualname__}.{self.name}: {self.value}>"
--- a/esrally/props.py
+++ b/esrally/props.py
@@ -1,0 +1,93 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+from enum import auto, EnumType, member, StrEnum
+
+
+class PropEnumType(EnumType):
+    __iter__ = None
+
+
+class PropEnum(StrEnum, metaclass=PropEnumType):
+    def __new__(cls, value):
+        if isinstance(value, PropEnumType):
+            return value
+        elif isinstance(value, str):
+            value = f"{cls.__qualname__.lower()}.{value}"
+            member = str.__new__(cls, value)
+            member._value_ = value
+            return member
+        else:
+            raise TypeError(f"{value!r} is not a string or enum")
+
+
+class Property(PropEnum):
+    def __sections(self):
+        node = globals()
+        for name in type(self).__qualname__.split('.'):
+            node = node[name]
+            if issubclass(node, Section):
+                yield name
+            else:
+                break
+
+    @property
+    def __section(self):
+        return '.'.join(map(str.lower, self.__sections()))
+
+    @property
+    def __key(self):
+        return self.value.replace(f"{self.__section}.", "", 1)
+
+    def __iter__(self):
+        return iter((self.__section, self.__key, ))
+
+    def __repr__(self):
+        return f"<{type(self).__qualname__}.{self.name}: {self.value}>"
+
+
+class Section(Property):
+    pass
+
+
+class Key(Property):
+    pass
+
+
+def desc(*_):
+    return auto()
+
+
+class SYSTEM(Section):
+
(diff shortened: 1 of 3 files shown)


esrally/props.py
```diff
@@ -0,0 +1,113 @@
...
+class SYSTEM(Section):
+    @member
+    class ADMIN(Key):
+        TRACK = desc()
+
+    @member
+    class LIST(Key):
+        FROM_DATE = desc()
+        MAX_RESULTS = desc()
+        TO_DATE = desc()
```
gbanasiak, line 105: I'm wondering if the following descriptions would be useful and whether we should add them for every member? These can be derived from `rally.py` argparse help. It's a bit of a duplication but would make it easier to decipher the intention behind each configuration key.

```suggestion
        # list records only from this date
        FROM_DATE = desc()
        # the limit of the number of records to return
        MAX_RESULTS = desc()
        # list records only to this date
        TO_DATE = desc()
```

gbanasiak (commented):
That's DRY indeed and looks fine :) I left small comments below. Can you confirm we could convert to underscores (e.g. `list.config.option` -> `list_config_option`) just by changes in `PropEnum` and `Property` classes in the future?

esrally/props.py
```diff
@@ -0,0 +1,113 @@
...
+    from enum import EnumType
+except ImportError:
+    from enum import EnumMeta as EnumType
+
+try:
+    from enum import member
+except ImportError:
+
+    def member(enum):
+        return enum
```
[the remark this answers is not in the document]
sakurai-youhei: Done. Thank you.

esrally/props.py
sakurai-youhei: I prepared `desc()` to write each property description as not comment but str for future use. But, as I'm not particular about this way, if you think it's against common sense, let's adopt the comment style.

esrally/props.py
```diff
@@ -0,0 +1,113 @@
...
+    def __new__(cls, value):
+        if isinstance(value, PropEnumType):
+            return value
+        elif isinstance(value, str):
+            value = f"{cls.__qualname__.lower()}.{value}"
+            member = str.__new__(cls, value)
+            member._value_ = value
+            return member
+        else:
+            raise TypeError(f"{value!r} is not a string or enum")
```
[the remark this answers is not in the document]
sakurai-youhei: I wrote down my best understanding in the comments. Hope they clarify things a little. But note that they may have errors because the enum is too hard to understand. I'm not confident until I test the code by unit test cases later.

sakurai-youhei:
@gbanasiak Thank you for your comment.

However, I became a bit suspicious that this hacky code is sustainable, even though I opened this PR. Also, if you plan to convert dot notations (`list.config.option`) to underscores (`list_config_option`), there might be a more straightforward way. Another thought is about another PR https://github.com/elastic/rally/pull/1798, which is probably more acceptable for everyone.

What do you think? Can we archive (close without merging and leave) this PR until this enum constants gets really crucial?

esrally/props.py
```diff
@@ -0,0 +1,138 @@
...
+            """
+            the limit of the number of records to return
+            """
+        )
+
+        TO_DATE = desc(
+            """
+            list records only to this date
+            """
+        )
```
gbanasiak, line 130: I like it. I would only collapse each `desc()` to a single line as there could be many of those.

esrally/props.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+class PropEnum(str, Enum, metaclass=PropEnumType):
+    """StrEnum-like Enum but supports multi-level nesting"""
+
+    def __new__(cls, value):
+        if isinstance(value, PropEnumType):
+            return value  # Nested Enum should stay original
+        elif isinstance(value, str):
+            # This prefixing lets A.B.C.D value not just "d" but "a.b.c.d"
```
gbanasiak, line 41:
```suggestion
            # String member A.B.C.D value is prefixed to obtain "a.b.c.d" instead of "d"
```

gbanasiak:
@sakurai-youhei Thank you for iterating.

I think https://github.com/elastic/rally/pull/1798 is a step in the right direction. It will help working on https://github.com/elastic/rally/pull/1795 or similar, and tighten existing use of configuration, but by itself will not fully address https://github.com/elastic/rally/issues/1646 due to the following:
* any combination of section and key is still allowed
* verification is done only if `cfg` instance is properly annotated with `Optional[config.Config]` or similar.

This wasn't stated explicitly but I've assumed we've embarked on the following journey:
* introduce literals and selective `mypy` configuration to provide a full list of sections and keys to work on next
* introduce `props` (gradually)
* once all literals replaced with `props`, switch `Config` methods to accept `props` instead of section/key strings.
Perhaps this las

I've already got accustomed to https://github.com/elastic/rally/pull/1795 logic and don't find it hacky, but if you see a more straightforward solution we can pursue this instead. We already looked at 2 proposals earlier in https://github.com/elastic/rally/issues/1646, and both were more repetitive in nature than this PR.

We can obviously also close this draft if you don't want to invest more time in it. Enum constants are not a top priority. It is a nice to have. Also, it is understood community contributions are not "guaranteed". It's your free time after all.

The switch from dot notation to underscores is an option for the future, but I don't think we want to introduce such breaking change now. I was just confirming this proposal is flexible enough to accommodate this path, I think it is.

sakurai-youhei:
@gbanasiak Thank you for your comment.

[1]
> once all literals replaced with `props`, switch `Config` methods to accept `props` only instead of section/key strings.

The above didn't come up in my mind, but that's a good idea indeed.

[2]
> The switch from dot notation to underscores is an option for the future, but I don't think we want to introduce such breaking change now. I was just confirming this proposal is flexible enough to accommodate this path, I think it is.

Thinking about it over a long period, the answer is YES. The props can be reworked when the time will have come. I thought the switch would happen in the near future. But if it's not, I don't have anything other than this PR to achieve the DRY as of now.

[3]
> I've already got accustomed to #1795 logic and don't find it hacky

That's good to know. :) OK, I will resume working on this PR once #1798 gets done.

esrally/props.py
sakurai-youhei: Done. Thanks!

sakurai-youhei:
I'm closing this PR to clean up my PR list. Please feel free to reopen PR and/or ping me anytime.

P.S. Although @gbanasiak has welcomed this approach once, I've been questioning this PR by myself because of the hackiness. But I believe introducing mypy through https://github.com/elastic/rally/pull/1798 was not a wrong decision. I appreciate the time and effort given to my PRs. Thank you very much.

PR #1799, opened 2023-11-06
Checkout tag on Docker release

gbanasiak:
Docker release workflow does not checkout the right tag before the Docker build.
This PR fixes this using approach taken from [eland](https://github.com/elastic/eland/blob/6cecb454e35890bd487f275e94f7c477c7acecd3/.buildkite/release-docker/run.sh#L18-L23).


gareth-ellis (approved):
LGTM

PR #1800, opened 2023-11-07
Bump version to 2.10.0

gbanasiak:


gbanasiak:
> Has the generation of Authors changed? It seems that capitals are being sorted differently now

https://github.com/elastic/rally/pull/1762 didn't quite work for me on my Mac. I'll fix that.

Issue #1802, opened 2023-11-14
Compare does not work with tracks with "metric" index names


gareth-ellis:
In for example the elastic/logs track we produce metric names with the index in.

E.g ``` .ds-logs-endpoint.events.file-default-2023.11.02-000001 process.name.text norms```

This means that if we want to compare two runs that were run on different days, we cannot as they produce different 

Doing so produces:
[ERROR] Cannot compare. '.ds-logs-endpoint.events.file-default-2023.11.01-000001'.

We should skip uncomparable metrics without erroring out.

PR #1482, opened 2022-05-06
Allow compare to continue for missing disk usage stats

inqueue:
This commit fixes an issue introduced by the [one with the beaker](https://github.com/elastic/rally/pull/1428#issuecomment-1027321314) when running `esrally compare` where disk usage stats are not present for prior benchmarks.

```diff
$ esrally compare --baseline=1ad6f592-14c7-4ffd-8c93-4eef6555eb9d --contender=bfdfe50a-d62f-45df-ac9a-6a1a5295d45f

    ____        ____
   / __ \____ _/ / /_  __
  / /_/ / __ `/ / / / / /
 / _, _/ /_/ / / / /_/ /
/_/ |_|\__,_/_/_/\__, /
                /____/


Comparing baseline
  Race ID: 1ad6f592-14c7-4ffd-8c93-4eef6555eb9d
-  Race timestamp: 2021-08-09 20:00:56
  Challenge: logging-querying
  Car: external
  User tags: env-id=4d577a92-4cf4-4da2-b1dd-948efb4a6d15

with contender
  Race ID: bfdfe50a-d62f-45df-ac9a-6a1a5295d45f
  Race timestamp: 2022-05-04 19:09:41
  Challenge: logging-querying
  Car: external
  User tags: env-id=18de9c61-7975-4aa9-800c-d2a984299268, setup=esbench

------------------------------------------------------
    _______             __   _____
   / ____(_)___  ____ _/ /  / ___/_________  ________
  / /_  / / __ \/ __ `/ /   \__ \/ ___/ __ \/ ___/ _ \
 / __/ / / / / / /_/ / /   ___/ / /__/ /_/ / /  /  __/
/_/   /_/_/ /_/\__,_/_/   /____/\___/\____/_/   \___/
------------------------------------------------------

[ERROR] Cannot compare. 'NoneType' object is not iterable.
```

### rally.log

```sh
2022-05-06 21:26:32,719 -not-actor-/PID:70929 esrally.rally ERROR A fatal error occurred while running subcommand [compare].
Traceback (most recent call last):
  File "/Users/jbryan/dev/src/rally/esrally/rally.py", line 910, in dispatch_sub_command
    reporter.compare(cfg, args.baseline, args.contender)
  File "/Users/jbryan/dev/src/rally/esrally/reporter.py", line 49, in compare
    ComparisonReporter(cfg).report(race_store.find_by_race_id(baseline_id), race_store.find_by_race_id(contender_id))
[...]

PR #1803, opened 2023-11-14
Skip metrics that are missing in either contender or baseline

gareth-ellis:
Fixes: #1802 

This fixes compare failing if a metric is missing from one of baseline or contender. It seems this functionality was partly thought about already via e.g #1482 though this adds further protection.

### DIFF
--- a/esrally/reporter.py
+++ b/esrally/reporter.py
@@ -666,7 +666,6 @@ class ComparisonReporter:
         totals.sort()
 
         lines = []
-        skip = 0
         for index, _total, field in totals:
             for stat in disk_usage_fields(baseline_stats):
                 if index in collated_baseline and index in collated_contender:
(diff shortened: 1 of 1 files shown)


esrally/reporter.py
```diff
@@ -684,6 +689,8 @@ def _report_disk_usage_stats_per_field(self, baseline_stats, contender_stats):
                         formatter=partial(convert.bytes_to_unit, unit),
                     )
                 )
+        if total > 0:
+            print_internal(f"Skipped {skip} statistics as missing from either baseline or contender")
```
gbanasiak, line 693: The intention was to check `skip > 0` condition here I think?

gbanasiak (commented):
Thank you for fixing this.

This prints twice because `_metrics_table()` is called twice in [here](https://github.com/elastic/rally/blob/9d301718029f53357cf43b914ece1d00193598b7/esrally/reporter.py#L397-L398).

Based on https://github.com/elastic/rally/pull/1482#issuecomment-1125381714 I think we should simplify and avoid a warning. WDYT?

gareth-ellis:
Sure - i've removed the logic around the warning. I have pushed to my branch, however the PR hasnt updated yet...

gbanasiak (approved):

gbanasiak:
Thank you for iterating, LGTM.

PR #1808, opened 2023-11-28
Remove 'aiohttp' pin

b-deam:
[3.9.1 was released this week](https://pypi.org/project/aiohttp/), which includes the fix for https://github.com/aio-libs/aiohttp/issues/7864, so removing the pin as installing will fetch the latest version.

Verified the fix with the reproduction script in https://github.com/elastic/rally/pull/1806


gbanasiak (approved):
LGTM

Issue #1810, opened 2023-12-01
Use the _index_template API for metrics templates


inqueue:
Today, Rally uses the legacy [`put_template`](https://elasticsearch-py.readthedocs.io/en/v8.11.0/api.html#elasticsearch.client.IndicesClient.put_template) API for creating index templates on the metrics store. Rally should instead use [`put_index_template`](https://elasticsearch-py.readthedocs.io/en/v8.11.0/api.html#elasticsearch.client.IndicesClient.put_index_template) (composable templates) when creating metrics index templates so it is using the most current API and to advance it one step toward sending metrics to a serverless project.

```diff
2023-12-01 23:20:04,259 -not-actor-/PID:1624 esrally.rally ERROR Cannot run subcommand [race].
Traceback (most recent call last):
  File "/home/esbench/rally/esrally/rally.py", line 1183, in dispatch_sub_command
    race(cfg, args.kill_running_processes)
  File "/home/esbench/rally/esrally/rally.py", line 931, in race
    with_actor_system(racecontrol.run, cfg)
  File "/home/esbench/rally/esrally/rally.py", line 961, in with_actor_system
    runnable(cfg)
  File "/home/esbench/rally/esrally/racecontrol.py", line 399, in run
    raise e
  File "/home/esbench/rally/esrally/racecontrol.py", line 396, in run
    pipeline(cfg)
  File "/home/esbench/rally/esrally/racecontrol.py", line 72, in __call__
    self.target(cfg)
  File "/home/esbench/rally/esrally/racecontrol.py", line 335, in benchmark_only
    return race(cfg, external=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/rally/esrally/racecontrol.py", line 293, in race
    raise exceptions.RallyError(result.message, result.cause)
esrally.exceptions.RallyError: Traceback (most recent call last):
  File "/home/esbench/rally/esrally/metrics.py", line 110, in guarded
    return target(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/.local/lib/python3.11/site-packages/elasticsearch/_sync/client/utils.py", line 414, in wrapped
    return api(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/.local/lib/python3.11/site-packages/elasticsearch/_sync/client/indices.py", line 2891, in put_template
    return self.perform_request(  # type: ignore[return-value]
[...]

PR #1811, opened 2023-12-02
Use composable index templates for metric indices

inqueue:
In this commit, Rally is being switched from using legacy index templates to newer composable templates for metrics indices. [Legacy index templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates-v1.html) have been deprecated since ES 7.8 in favor of composable templates, and are not supported in ES Serverless. Rally metrics index templates have just one modification to make them composable templates. No other modifications have been made.

* Elasticsearch favors composable templates over legacy index templates when both have matching index patterns. After this change, new metrics indices will use the new index templates.

Closes https://github.com/elastic/rally/issues/1810.

### DIFF
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -7,7 +7,7 @@ Migrating to Rally 2.10.1
 The metrics store now uses composable templates 
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-* Existing legacy index templates will are not removed, however, composable templates take precedence (applies to races, results, and metrics).
+* Existing legacy index templates are not removed, however, composable templates take precedence (applies to races, results, and metrics).
 * The new templates will be applied when new races, results, and metrics indices are created.
 * The index template for annotations will only be applied if the ``rally-annotations`` index is not present.
 
--- a/docs/migrate.rst
+++ b/docs/migrate.rst
@@ -1,6 +1,16 @@
 Migration Guide
 ===============
 
+Migrating to Rally 2.10.1
+-------------------------
+
+The metrics store now uses composable templates 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Existing legacy index templates will are not removed, however, composable templates take precedence (applies to races, results, and metrics).
+* The new templates will be applied when new races, results, and metrics indices are created.
+* The index template for annotations will only be applied if the ``rally-annotations`` index is not present.
+
 Migrating to Rally 2.10.0
 -------------------------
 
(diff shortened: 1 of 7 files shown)


esrally/metrics.py
```diff
@@ -50,13 +50,13 @@ def __init__(self, client, cluster_version=None):
...
-        return self.guarded(self._client.indices.put_template, name=name, **tmpl)
+        return self.guarded(self._client.indices.put_index_template, name=name, **tmpl)
 
     def template_exists(self, name):
-        return self.guarded(self._client.indices.exists_template, name=name)
+        return self.guarded(self._client.indices.exists_index_template, name=name)
 
     def delete_template(self, name):
-        self.guarded(self._client.indices.delete_template, name=name)
+        self.guarded(self._client.indices.delete_index_template, name=name)
```
gbanasiak, line 59: Nit: I don't think we're using template removal anywhere. Perhaps we could get rid of it?

gbanasiak (approved):
LGTM.

What do you think about adding a migration guide entry explaining the consequences of this change to _existing_ ES metric stores:
* legacy index templates are not removed automatically but composable index templates take precedence (applies to races, results and metrics),
* index template for annotations only gets created when `rally-annotations` index is missing, so full migration from legacy to composable would require cumbersome renaming of `rally-annotations` index, followed by Rally run (`esrally add annotation ...`), followed by reindexing of old data to new `rally-annotations` index.

inqueue:
> LGTM.
> 
> What do you think about adding a migration guide entry explaining the consequences of this change to _existing_ ES metric stores:
> 
> * legacy index templates are not removed automatically but composable index templates take precedence (applies to races, results and metrics),
> * index template for annotations only gets created when `rally-annotations` index is missing, so full migration from legacy to composable would require cumbersome renaming of `rally-annotations` index, followed by Rally run (`esrally add annotation ...`), followed by reindexing of old data to new `rally-annotations` index.

Hi @gbanasiak, the requested changes have been applied. Would you mind taking another look?

docs/migrate.rst
```diff
@@ -1,6 +1,16 @@
 Migration Guide
 ===============
 
+Migrating to Rally 2.10.1
+-------------------------
+
+The metrics store now uses composable templates 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* Existing legacy index templates will are not removed, however, composable templates take precedence (applies to races, results, and metrics).
```
gbanasiak, line 10:
```suggestion
* Existing legacy index templates are not removed, however, composable templates take precedence (applies to races, results, and metrics).
```

gbanasiak (approved):
Thank you for iterating. LGTM.

PR #1816, opened 2024-01-04
Add FS metrics to node-stats telemetry device

b-deam:
Adds overall (total) FS metrics to the default list of stats collected by the `node-stats` telemetry device. 

Note that we ignore the per-device stats here, as that'd require some remodelling of the `node-stats` document structure to support having per-device documents.


inqueue (approved):
I've been meaning to do this, THANK YOU. LGTM

PR #1823, opened 2024-01-15
Measure time until last response chunk

### LINKED ISSUE #1822: Service time measurement stops at first response byte
**Rally version** (get with `esrally --version`): `esrally 2.10.0.dev0 (git revision: df49d4d2c45d3dbcd2e6791524b0c5812430105e)`

**Invoked command**: N/A

**Configuration file (located in `~/.rally/rally.ini`))**: N/A

**JVM version**: N/A

**OS version**: N/A

Rally uses [aiohttp tracing callbacks](https://docs.aiohttp.org/en/stable/tracing_reference.html#) to measure service time, specifically:

* `on_request_start` to start a timer
* `on_request_end` to stop the timer

The [state diagram in the aiohttp docs](https://docs.aiohttp.org/en/stable/tracing_reference.html#) suggests that `on_request_end` is the callback that is called when the response has been completely received, however, looking at the implementation in aiohttp, this is not correct. Instead, this callback is invoked when the first byte (or initial chunk) of the response is received, see ([initial await](https://github.com/aio-libs/aiohttp/blob/63bea9d21498ccda73472fabc060d67cd83e2c7a/aiohttp/client.py#L548) and [callback invocation](https://github.com/aio-libs/aiohttp/blob/63bea9d21498ccda73472fabc060d67cd83e2c7a/aiohttp/client.py#L655-L658)) but not when the response body has been read. This happens in higher-level code (see [aio implementation of elastic-transport](https://github.com/elastic/elastic-transport-python/blob/ad30469c5be0ea009411c2dccd5866d54477a8e4/elastic_transport/_node/_http_aiohttp.py#L191) that reads the body).

Consequently, Rally stops the timer when the first byte of the response is received, instead of the entire response. As any code that uses the Elasticsearch client cannot make any progress before the body has been received, service time measurement is too "optimistic" at the moment. In many cases the difference will be small but if the response is large this can make a difference of several hundred milliseconds.

**Steps to reproduce**:
 1. Run a Rally track that produces large responses (e.g. Universal Profiling, only available internally at the moment)
 2. Observe the time difference between using the current callback and the fixed version (see below)

**Fix**

The following snippet fixes the issue:

```diff
diff --git a/esrally/client/factory.py b/esrally/client/factory.py
index 4a4c0179..daaaeeee 100644
--- a/esrally/client/factory.py
+++ b/esrally/client/factory.py
@@ -225,7 +225,7 @@ class EsClientFactory:
 
         trace_config = aiohttp.TraceConfig()
         trace_config.on_request_start.append(on_request_start)
-        trace_config.on_request_end.append(on_request_end)
+        trace_config.on_response_chunk_received.append(on_request_end)
         # ensure that we also stop the timer when a request "ends" with an exception (e.g. a timeout)
         trace_config.on_request_exception.append(on_request_end)
```

The fix works because Rally [continuously updates the end time on each callback invocation](https://github.com/elastic/rally/blob/21ae64a8f8e1776838e086f7694e2649530c6abc/esrally/client/context.py#L95) so in the end, the timestamp corresponds to the timestamp when the *last* chunk has been received.

danielmitterdorfer:
Rally uses timers internally to measure when a request started and when it has completed. However, it has stopped measuring as soon as the first byte of the response has been received instead of waiting for the full response. This can skew service time measurement if the response is large and Elasticsearch streams it.

With this commit we wait until the last chunk of the response has been received. This results in a more realistic service time metric in these cases.

Closes #1822

### DIFF
--- a/esrally/client/asynchronous.py
+++ b/esrally/client/asynchronous.py
@@ -90,6 +90,17 @@ class StaticRequest(aiohttp.ClientRequest):
         return self.response
 
 
+# we use EmptyStreamReader here because it overrides all methods with
+# no-op implementations that we need.
+class StaticStreamReader(aiohttp.streams.EmptyStreamReader):
+    def __init__(self, body):
+        super().__init__()
+        self.body = body
+
+    async def read(self, n: int = -1) -> bytes:
+        return self.body.encode("utf-8")
+
+
 class StaticResponse(aiohttp.ClientResponse):
     def __init__(
         self,
@@ -122,11 +133,9 @@ class StaticResponse(aiohttp.ClientResponse):
         self._protocol = connection.protocol
         self._connection = connection
         self._headers = CIMultiDictProxy(CIMultiDict())
+        self.content = StaticStreamReader(self.static_body)
         self.status = 200
         return self
-
-    async def read(self):
-        return self.static_body.encode("utf-8")
 
 
 class ResponseMatcher:
--- a/esrally/client/factory.py
+++ b/esrally/client/factory.py
@@ -225,7 +225,16 @@ class EsClientFactory:
 
         trace_config = aiohttp.TraceConfig()
         trace_config.on_request_start.append(on_request_start)
-        trace_config.on_request_end.append(on_request_end)
+        # It is tempting to register this callback on `TraceConfig.on_request_end()`. However, aiohttp will call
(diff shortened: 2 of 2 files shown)


dpifke-elastic (approved):
LGTM.

PR #1824, opened 2024-01-22
Fix telemetry docs 

gareth-ellis:
Fix broken link in telemetry docs

I spotted the link back to cluster management was broken, so fixed it


gbanasiak (approved):
LGTM

PR #1825, opened 2024-01-22
Make Jit telelmetry compatible with JDK 9+

gareth-ellis:
I found that the JIT telemetry device doesn't work with JDK > 16 - in JDK 9 the -XX:+TraceClassLoading option was replaced with -Xlog:class+load=info , JDK 9-15 it was changed for the user if they supplied the older one, in jdk16 + the option is rejected and the JVM fails to start.


ebadyano (approved):
LGTM! Thank you!

PR #1827, opened 2024-01-30
Track dataset size

danielmitterdorfer:
With this commit we introduce a new metric "Datset size" that tracks not only local disk usage but also disk usage of partially mounted snapshots. This allows to track disk usage also in Serverless.


inqueue (approved):
LGTM, thanks!

PR #1828, opened 2024-01-30
Fixes #1595: Add method to process.py to return both return code and logs

### LINKED ISSUE #1595: Return both exit code, and output in process module
In https://github.com/elastic/rally/pull/1591 it became clear that it'd be nice to modify an existing function in our [`process` module](https://github.com/elastic/rally/blob/master/esrally/utils/process.py) to return both the stdout/stderr of the subprocess as well as the exit code.

favilo:
As requested in https://github.com/elastic/rally/issues/1595

[x] Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
[x] Have you followed the [contributor guidelines](https://github.com/elastic/rally/blob/master/CONTRIBUTING.md)?
[x] Have you run `make check-all` successfully?
[x] Did you choose a [descriptive title and description](https://chris.beams.io/posts/git-commit/) for your PR?

### DIFF
--- a/esrally/utils/process.py
+++ b/esrally/utils/process.py
@@ -24,10 +24,13 @@ from typing import Callable, Dict, List
 
 import psutil
 
+LogLevel = int
+FileId = int
+
 
 def run_subprocess(command_line: str) -> int:
     """
-    Runs the provided command line in a subprocess. All output will be returned in the `CompletedProcess.stdout` field.
+    Runs the provided command line in a subprocess.
 
     :param command_line: The command line of the subprocess to launch.
     :return: The process' return code
@@ -65,10 +68,6 @@ def exit_status_as_bool(runnable: Callable[[], int], quiet: bool = False) -> boo
         if not quiet:
             logging.getLogger(__name__).exception("Could not execute command.")
         return False
-
-
-LogLevel = int
-FileId = int
 
 
 def run_subprocess_with_logging(
--- a/esrally/utils/process.py
+++ b/esrally/utils/process.py
@@ -25,14 +25,14 @@ from typing import Callable, Dict, List
 import psutil
 
 
-def run_subprocess(command_line: str) -> subprocess.CompletedProcess:
+def run_subprocess(command_line: str) -> int:
     """
     Runs the provided command line in a subprocess. All output will be returned in the `CompletedProcess.stdout` field.
 
     :param command_line: The command line of the subprocess to launch.
-    :return: The `CompletedProcess` object for the subprocess. `.returncode` contains the process' return code
+    :return: The process' return code
     """
-    return subprocess.run(command_line, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True, check=False)
+    return subprocess.call(command_line, shell=True)
 
 
 def run_subprocess_with_output(command_line: str, env: Dict[str, str] = None) -> List[str]:
@@ -78,7 +78,7 @@ def run_subprocess_with_logging(
     stdin: FileId = None,
     env: Dict[str, str] = None,
     detach: bool = False,
-) -> subprocess.CompletedProcess:
+) -> int:
     """
     Runs the provided command line in a subprocess. All output will be captured by a logger.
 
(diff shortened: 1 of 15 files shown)


gbanasiak (commented):
LGTM, but https://github.com/elastic/rally/issues/1595 lacks explicit clarification of what's exactly the win. I think we can find it [here](https://github.com/elastic/rally/pull/1591#discussion_r986874797).

WDYT about using the modified `run_subprocess()` to avoid the double call in [is_branch()](https://github.com/elastic/rally/blob/5e6b1bc869f0406024bd796b825511dc13a6d133/esrally/utils/git.py#L52) as part of this PR. On quick scan that's probably the only improvement in the `git` module we can make as other methods use different commands in a sequence?

favilo:
> LGTM, but #1595 lacks explicit clarification of what's exactly the win. I think we can find it [here](https://github.com/elastic/rally/pull/1591#discussion_r986874797).
> 
> WDYT about using the modified `run_subprocess()` to avoid the double call in [is_branch()](https://github.com/elastic/rally/blob/5e6b1bc869f0406024bd796b825511dc13a6d133/esrally/utils/git.py#L52) as part of this PR? On quick scan that's probably the only improvement in the `git` module we can make as other methods use different commands in a sequence?

What do you think about the latest change?

gbanasiak:
> What do you think about the latest change?

Thank you for iterating. https://github.com/elastic/rally/pull/1828/commits/21caafa2f0ffd405ff93fc26d5fed706fd624b57 eliminates duplication but made me realize we're missing the logging now which I think should be kept for troubleshooting purposes. We could add snowflake logging at `git` module level, or add logging option to `run_subprocess()`. Or, rewrite `run_subprocess_with_logging` to be based on `subprocess.run` and return `CompletedProcess`, but this would extend the scope of the PR. Which option do you find the most elegant?

favilo:
> > What do you think about the latest change?
> 
> Thank you for iterating. [21caafa](https://github.com/elastic/rally/commit/21caafa2f0ffd405ff93fc26d5fed706fd624b57) eliminates duplication but made me realize we're missing the logging now which I think should be kept for troubleshooting purposes. We could add snowflake logging at `git` module level, or add logging option to `run_subprocess()`. Or, rewrite `run_subprocess_with_logging()` to be based on `subprocess.run()` and return `CompletedProcess`, but this would extend the scope of the PR. Which option do you find the most elegant?

Adding logging option to `run_subprocess()` feels like the least effort solution here. I wouldn't mind having `run_subprocess_with_logging()` return a Popen object which would also have `.returncode` and `.stdout`, but that has a _lot_ more uses, and since we are already logging, do we need to keep track of the output separately?

I could certainly be convinced that the latter one is the correct approach, but maybe not in this small PR?

gbanasiak:
IT tests are failing as it turns out `run_subprocess_with_logging()` is also used in https://github.com/elastic/rally-teams, e.g. [here](https://github.com/elastic/rally-teams/blob/9ea8b51e2dec6aca3181ea7394f66c242da8b46f/cars/v1/x_pack/base/config.py#L60C5-L66C61). I wasn't aware of that. This changes the picture considerably. I think we should go for a new method in the `process` module for backward compatibility. The problem is https://github.com/elastic/rally-teams code can be interpreted by a range of Rally versions.

esrally/utils/process.py
```diff
@@ -20,15 +20,22 @@
...
 import time
+from typing import Callable, Dict, List
 
 import psutil
 
 
-def run_subprocess(command_line):
+def run_subprocess(command_line: str) -> int:
+    """
+    Runs the provided command line in a subprocess. All output will be returned in the `CompletedProcess.stdout` field.
```
gbanasiak, line 30:
```suggestion
    Runs the provided command line in a subprocess.
```

esrally/utils/process.py
```diff
@@ -60,7 +67,18 @@ def exit_status_as_bool(runnable, quiet=False):
         return False
 
 
-def run_subprocess_with_logging(command_line, header=None, level=logging.INFO, stdin=None, env=None, detach=False):
+LogLevel = int
+FileId = int
```
gbanasiak, line 71: Nit: Should type definitions be moved to the top of the file?

gbanasiak (approved):
LGTM. Please change the title of the PR as it's no longer relevant. I've left 2 minor comments.

favilo:
Huh, I see the IT tests failed, but looking at the logs, I don't actually see any failures noted.

PR #1830, opened 2024-02-20
Add looped mode to bulk operation

gbanasiak:
Adds `looped` mode to bulk operation. That is useful in benchmarks where we want Rally to ingest for a specific amount of time even if corpus is not big enough to sustain indexing at a specific rate for this time.

This mode is meant to be used together with `time-period` or `iterations` at the task level. Otherwise Rally will never complete the task.

Example use:
```
{
  "name": "some-bulk-task",
  "operation": {
    "operation-type": "bulk",
    "bulk-size": 5000,
    "looped": true
  },
  "time-period": 300,
  "target-throughput": 2
}
```


gbanasiak:
Re-running IT test failure of which seems unrelated.

esrally/track/params.py
```diff
@@ -763,7 +768,12 @@ def params(self):
         # self.internal_params always reads all files. This is necessary to ensure we terminate early in case
         # the user has specified ingest percentage.
         if self.current_bulk == self.total_bulks:
-            raise StopIteration()
+            # start from the beginning in looped mode, otherwise stop the run
+            if self.looped:
+                self.current_bulk = 0
+                self._init_internal_params()
```
inqueue, line 774: With looped mode, we will have the possibility to add an override for bulk and batch sizes in steps if it is something we can use in the future.

inqueue (approved):
LGTM

esrally/track/params.py
gbanasiak: The idea is looped operation should be limited at the task level, and you can have as many tasks with different batch sizes, etc. as you want. Looped mode basically means "keep on generating bulk requests until some external criteria stop the task".

PR #1831, opened 2024-02-20
Add client ID to metrics

gbanasiak:
When analyzing Rally metrics with runs with multiple clients it is impossible to filter for a specific client ID. This PR adds new `meta.client_id` field to every metric record which can be attributed to a specific client, i.e. is not aggregated like throughput.


favilo (approved):
LGTM

inqueue (approved):
LGTM

Issue #1834, opened 2024-02-28
Large meta.error-description field fails Elasticsearch metric store ingest


gbanasiak:
**Rally version** (get with `esrally --version`):

```
esrally 2.10.0.dev0 (git revision: a2c09a751d7e2797fde531f90aea43ac1375c987)
```

**Description of the problem including expected versus actual behavior**:

In certain scenarios, Rally can produce large `meta.error-description` field in `rally-metrics-*` documents which cannot be indexed by Elasticsearch and fails a race. The `meta.error-description` field is mapped as `keyword` which has a term byte-length limit of 32766 bytes imposed by Lucene.

The characteristic symptom is the following error:
```
Document contains at least one immense term in field="meta.error-description" (whose UTF8 encoding is longer than the max length 32766), all of which were skipped.
```

**Provide logs (if relevant)**:

`2024-02-27 20:37:56,153 ActorAddr-(T|:37759)/PID:484 esrally.driver.runner WARNING Bulk request failed: [HTTP status: 409, message: [-ChI7I0BiZHehfRA873C]: version conflict, document already exists (current version [1]) | HTTP status: 409, message: [-ChI7I0BiZHehfRA877C]: version conflict, document already exists (current version [1]) | HTTP status: 409, message: [-ChI7I0BiZHehfRA87_C]: version conflict, document already exists (current version [1]) | HTTP status: 409, message: [-ChI7I0BiZHehfRA87nB]: version conflict, document already exists (current version [1]) | [..]`

[...]

PR #1835, opened 2024-02-28
Avoid immense term error in metrics metadata fields

gbanasiak:
To avoid immense term error in metrics metadata fields:
* `meta.error-description` field mapping changes from `keyword` to `wildcard` which fits this type of content better while still allowing the search,
* remaining dynamically mapped text fields get `"ignore_above": 32766` parameter (see [doc](https://www.elastic.co/guide/en/elasticsearch/reference/8.12/ignore-above.html)),
* `meta.error-description` field generated in `BulkIndex` runner gets limited to 5 unique errors only for better log readability - currently, in the worst case the field is concatenated from errors from _all_ bulk documents.

Addresses https://github.com/elastic/rally/issues/1834.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -659,13 +659,28 @@ class BulkIndex(Runner):
             error_details.add((data["status"], None))
 
     def error_description(self, error_details):
+        """
+        Generates error description with an arbitrary limit of 5 errors.
+
+        :param error_details: accumulated error details
+        :return: error description
+        """
         error_descriptions = []
-        for status, reason in error_details:
-            if reason:
-                error_descriptions.append(f"HTTP status: {status}, message: {reason}")
+        is_truncated = False
+        for count, error_detail in enumerate(error_details):
+            status, reason = error_detail
+            if count < 5:
+                if reason:
+                    error_descriptions.append(f"HTTP status: {status}, message: {reason}")
+                else:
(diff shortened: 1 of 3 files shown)


esrally/driver/runner.py
```diff
@@ -659,13 +659,28 @@ def extract_error_details(self, error_details, data):
...
+            status, reason = error_detail
+            if count < 5:
+                if reason:
+                    error_descriptions.append(f"HTTP status: {status}, message: {reason}")
+                else:
+                    error_descriptions.append(f"HTTP status: {status}")
             else:
-                error_descriptions.append(f"HTTP status: {status}")
-        return " | ".join(sorted(error_descriptions))
+                is_truncated = True
```
gareth-ellis, line 678: What do you think about summarising the remaining errors? E.g Truncated 20 X 409, 30 X 503 etc? That way we can easily see if the errors we got back were at least all of the same status code, or whether there was other statuses hidden further along?

gbanasiak: Great idea, let me add this :)

gbanasiak: Added in https://github.com/elastic/rally/pull/1835/commits/6ce175adbd18fe48dc461ed8fc7356156d5dac3f.

inqueue (approved):
LGTM

gareth-ellis (approved):

PR #1836, opened 2024-03-01
Update create-track in rally

gareth-ellis:
Add option to increase batch size when creating a track to speed up download of data. Update track layout to adhere to best practices

Creating tracks from large corpus can take quite a bit of time, so I have added the option to increase the batch size of scan, so if a user is running with a stable enough network connection and enough hardware resources on the rally instance, they can increase the batch size to speed up download.

I have also updated the track layout to better match how we tend to layout our tracks

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,11 +41,11 @@ jobs:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
         os:
-          - macos-latest
+          - macos-13
           - ubuntu-latest
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


gareth-ellis:
Yes, that makes sense. I'll update those too

ebadyano:
tested with regular index and data streams. I think once we have updated docs it's good to go!

One interesting thing for data-streams scenario: we create a regular index in the new track, it's the same as the original behaviour and out of scope for this pr, but I wonder if for data-streams it makes sense if we create a track that index docs into data streams?..

inqueue (approved):
++ @ebadyano's comments. The rest LGTM.

PR #1838, opened 2024-03-21
[it tests] Upload logs as buildkite artifact even on failure

favilo:

### DIFF
--- a/.buildkite/it/run.sh
+++ b/.buildkite/it/run.sh
@@ -4,6 +4,11 @@ set -eo pipefail
 
 source .buildkite/retry.sh
 
+function upload_logs {
+    echo "--- Upload artifacts"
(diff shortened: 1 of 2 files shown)


.buildkite/it/run.sh
```diff
@@ -4,6 +4,11 @@ set -eo pipefail
 
 source .buildkite/retry.sh
 
+function upload_logs {
+    echo "--- Upload artifacts"
+    buildkite-agent artifact upload "${RALLY_HOME}/.rally/logs/*.log"
+}
+
```
favilo, line 11: Should this go somewhere common, like `.buildkite/retry.sh`? It relies on `$RALLY_HOME` being set, so it _could_ be fragile, is why I didn't just go ahead and do it.

gareth-ellis (approved):
LGTM

gbanasiak (approved):

PR #1839, opened 2024-03-22
Add support for Python 3.12

favilo:


pyproject.toml
```diff
@@ -88,20 +89,20 @@ s3 = [
...
     "pytest==7.1.2",
     "pytest-benchmark==3.4.1",
     "pytest-asyncio==0.19.0",
     "pytest-httpserver==1.0.5",
     "tox==3.25.0",
-    "nox==2022.11.21",
+    "nox==2024.3.2",
     "sphinx==5.1.1",
     "furo==2022.06.21",
     "github3.py==3.2.0",
```
ebadyano, line 108: Looks like pylint 3.0.4 supports 3.12?

gbanasiak:
@favilo I think we should follow-up by removing 3.8 support. Its EOL is October 2024.

PR #1841, opened 2024-04-12
Send version with esql

nik9000:
Elasticsearch version 8.14 will require a `version` parameter. This defaults it to the first released version. That version is supported in 8.13.3, but not before that.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -2892,6 +2892,7 @@ class Esql(Runner):
         query = mandatory(params, "query", self)
         body = params.get("body", {})
         body["query"] = query
+        body["version"] = params.get("version", "2024.04.01.🚀")
         query_filter = params.get("filter")
         if query_filter:
             body["filter"] = query_filter
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -2892,7 +2892,7 @@ class Esql(Runner):
         query = mandatory(params, "query", self)
         body = params.get("body", {})
         body["query"] = query
-        body["version"] = params.get("version", "2024.04.01.🚀")
+        body["version"] = params.get("version", "2024.04.01")
         query_filter = params.get("filter")
         if query_filter:
             body["filter"] = query_filter
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ Once your changes and tests are ready to submit for review:
 
     Ensure that all tests pass by running `make check-all`. This runs sequentially lint checks, unit tests and integration tests. These can be executed in isolation using `make lint`, `make test` and `make it` respectively, in case you need to iterate over a subset of tests.
 
-    Note: Integration tests are much slower than unit tests.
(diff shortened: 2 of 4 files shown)


esrally/driver/runner.py
```diff
@@ -2892,6 +2892,7 @@ async def __call__(self, es, params):
         query = mandatory(params, "query", self)
         body = params.get("body", {})
         body["query"] = query
+        body["version"] = params.get("version", "2024.04.01.🚀")
```
alex-spies, line 2895: :rocket:

alex-spies (approved):
LGTM

pquentin:
Do we want to support versions of Elasticsearch that don't know about this parameter?

nik9000:
> Do we want to support versions of Elasticsearch that don't know about this parameter? Or at least wait for 8.13.3 to be released?

I haven't checked, https://github.com/elastic/elasticsearch/pull/107438 should make rally work even without this PR and this one can wait until we upgrade.

As for versions of ES.... I dunno! ESQL's not GA in 8.13 so I'd be ok dropping any executions against it. But waiting to 8.13.next seems fine, especially if we have that ES PR.

esrally/driver/runner.py
gbanasiak: Just curious, what's the purpose of UTF emoji? To have an option to release 2 versions in the same day? Or have an easy-to-remember mnemonic?

gbanasiak:
> https://github.com/elastic/elasticsearch/pull/107438 should make rally work even without this PR

Rally uses Python client 8.6.1, so looking at [this](https://github.com/nik9000/elasticsearch/blob/1fa03e084becc29466525e09cb4e04c9a2572135/x-pack/plugin/esql/src/main/java/org/elasticsearch/xpack/esql/action/RestEsqlQueryAction.java#L99-L101) it won't? I think we need to merge this before `version` is required on the ES side.

Please also add documentation of `version` parameter under https://github.com/elastic/rally/blob/master/docs/track.rst#esql.

esrally/driver/runner.py
gbanasiak: @pquentin was kind enough and provided context behind the emoji. As this is just for fun and optional, I would actually prefer to skip it to avoid questions such as mine in the future from Rally code users.

nik9000: Dropped.

alex-spies:
> Please also add documentation of `version` parameter under https://github.com/elastic/rally/blob/master/docs/track.rst#esql.

Added to docs, thanks for the pointer!

nik9000:
run CI / unit macOS 3.8 (pull_request)

gbanasiak (approved):
LGTM

gbanasiak:
@nik9000 We started seeing benchmark failures with ESQL operations after https://github.com/elastic/elasticsearch/pull/107433 got merged, so it's time to merge this one.

nik9000:
Merged!

nik9000:
sorry for the delay

PR #1846, opened 2024-05-17
Update create track

gareth-ellis:
I seem to mess up my old branch - this is the same as PR #1836 
Add option to increase batch size when creating a track to speed up download of data. Update track layout to adhere to best practices

Creating tracks from large corpus can take quite a bit of time, so I have added the option to increase the batch size of scan, so if a user is running with a stable enough network connection and enough hardware resources on the rally instance, they can increase the batch size to speed up download.

I have also updated the track layout to better match how we tend to layout our tracks

There was feedback to add more docs, so that's done too

### DIFF
--- a/docs/adding_tracks.rst
+++ b/docs/adding_tracks.rst
@@ -27,11 +27,15 @@ The track generator will create a folder with the track's name in the specified 
 
     > find tracks/acme
     tracks/acme
+    tracks/acme/challenges
+    tracks/acme/challenges/default.json
     tracks/acme/companies-documents.json
     tracks/acme/companies-documents.json.bz2
     tracks/acme/companies-documents-1k.json
     tracks/acme/companies-documents-1k.json.bz2
     tracks/acme/companies.json
+    tracks/acme/operations
+    tracks/acme/operations/default.json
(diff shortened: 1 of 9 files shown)


ebadyano:
@gareth-ellis Thank you for updating the docs. LGTM.

One additional thing, since you are alreday updating the docs do you mind adding an option for data streams? During last review when I tested the change with datastreams I didn't realize I needed to specify `--data-streams` to actually make track create to include them in the track. See https://github.com/elastic/rally/pull/1531/files
I will test it shortly to confirm it works

gareth-ellis:
The data-stream option is already in the docs - https://esrally.readthedocs.io/en/stable/command_line_reference.html#data-streams - or were you meaning something different?

ebadyano:
> The data-stream option is already in the docs - https://esrally.readthedocs.io/en/stable/command_line_reference.html#data-streams - or were you meaning something different?

You are right, I somehow missed it.

gareth-ellis:
I'd like to add the option to setup writing back to the datastream too - that isn't currently in place, but I thought that could be in a seperate PR

ebadyano (approved):
LGTM

PR #1848, opened 2024-05-21
Pin `requests` library until we fix `docker-py`

favilo:


pyproject.toml
```diff
@@ -50,7 +50,7 @@ dependencies = [
     "urllib3==1.26.18",
     "docker==6.0.0",
     # avoid specific requests version to fix bug in docker-py
-    "requests > 2.26.0, != 2.32.0",
+    "requests==2.31.0",
```
inqueue, line 53: There is a 2.31.1. Can we set it to `<2.32.0`?

inqueue (changes requested):

pyproject.toml
favilo: Done

inqueue (approved):
LGTM

PR #1849, opened 2024-05-23
Remove esql version

alex-spies:
ES|QL in 8.14 will not require a `version` parameter in requests to `_query`, after all.

Remove the `version` parameter from the ES|QL runner, essentially reverting https://github.com/elastic/rally/pull/1841.

This should be merged only after https://github.com/elastic/elasticsearch/pull/108919 is merged to avoid breakage.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,7 +60,7 @@ Once your changes and tests are ready to submit for review:
 
     Ensure that all tests pass by running `make check-all`. This runs sequentially lint checks, unit tests and integration tests. These can be executed in isolation using `make lint`, `make test` and `make it` respectively, in case you need to iterate over a subset of tests.
 
-    Note: Integration tests are much slower than unit tests.
+    Note: Integration tests are much slower than unit tests and require `docker-compose`.
 
 3. Sign the Contributor License Agreement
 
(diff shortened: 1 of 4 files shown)


gbanasiak:
The last released Rally version 2.10.0 does not send the `version` and we are about to release version 2.11.0. I think we should refrain from releasing 2.11.0 until we remove `version`. Does that sound reasonable to you @alex-spies ?

The IT tests are failing which is expected. Once https://github.com/elastic/elasticsearch/pull/108919 is merged they should pass.

gbanasiak:
Also, if we ever have to investigate ES|QL regression in the range of commits where `version` is required, we will have to use specific Rally commit ID (from before this PR), which is not a major problem.

alex-spies:
Thanks for your comments @gbanasiak , very good points.

> The last released Rally version 2.10.0 does not send the version and we are about to release version 2.11.0. I think we should refrain from releasing 2.11.0 until we remove version. Does that sound reasonable to you @alex-spies ?

That sounds very reasonable! I expect that ES 8.15 will want to disallow `version` again, so it'd be better if rally didn't send this; if it's not too much trouble, waiting until https://github.com/elastic/elasticsearch/pull/108919#pullrequestreview-2073142605 is merged and backported would be _great_ IMHO.

> Also, if we ever have to investigate ES|QL regression in the range of ES commits where version is required, we will have to use specific Rally commit ID (from before this PR), which is not a major problem.

That is great to hear, thank you!

gbanasiak (approved):
LGTM assuming tests start passing once https://github.com/elastic/elasticsearch/pull/108919 is in

gbanasiak:
https://github.com/elastic/elasticsearch/pull/108919 is merged, so I've initiated CI re-runs.

gbanasiak:
We will have to refrain from merging until ES change propagates to all serverless environments.

pquentin:
buildkite test this please

PR #1851, opened 2024-05-29
Assume no operator if security not enabled in serverless

gbanasiak:
When testing locally serverless might be running without security enabled. In this case, `GET _security/_authenticate` fails. This PR falls back to non-operator privileges in such case.

### DIFF
--- a/esrally/client/factory.py
+++ b/esrally/client/factory.py
@@ -372,11 +372,9 @@ def cluster_distribution_version(hosts, client_options, client_factory=EsClientF
         # pylint: disable=import-outside-toplevel
         from elasticsearch.exceptions import ApiError
 
-        try:
+        with contextlib.suppress(ApiError):
             authentication_info = es.perform_request(method="GET", path="/_security/_authenticate")
             serverless_operator = authentication_info.body.get("operator", False)
-        except ApiError:
-            serverless_operator = False
 
     if not versions.is_serverless(version_build_flavor) or serverless_operator is True:
         # if available, unconditionally wait for the REST layer - if it's not up, we'll intentionally raise the original error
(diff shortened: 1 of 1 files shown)


esrally/client/factory.py
```diff
@@ -367,8 +367,16 @@ def cluster_distribution_version(hosts, client_options, client_factory=EsClientF
...
+
+        # determine operator status if security enabled, otherwise assume operator privileges
+        # pylint: disable=import-outside-toplevel
+        from elasticsearch.exceptions import ApiError
+
+        try:
+            authentication_info = es.perform_request(method="GET", path="/_security/_authenticate")
+            serverless_operator = authentication_info.body.get("operator", False)
+        except ApiError:
+            serverless_operator = False
```
gbanasiak, line 379: I can also use `pass` here. Which is more readable?

gareth-ellis: I'd say like you have it is fine, and probably clearer

gareth-ellis (approved):
LGTM

esrally/client/factory.py
pquentin, line 379: I think the explicit assignment is more readable than `pass`. Another option is using `contextlib.suppress`:

```suggestion
        with contextlib.suppress(ApiError):
            authentication_info = es.perform_request(method="GET", path="/_security/_authenticate")
            serverless_operator = authentication_info.body.get("operator", False)
```

pquentin: However, the comment above says "otherwise assume operator privileges", so should this be assigned to True?

gbanasiak: Good catch. That was my initial assumption which failed in testing. Code is correct, but comment requires a change. I like the brevity of `contextlib`, so let me use that as well :)

pquentin (approved):
Thanks! LGTM.

PR #1852, opened 2024-05-30
Update pypi publish to 1.8.14

gareth-ellis:
During release of 2.11.0 I discovered our version of PyPi publish is no longer compatible with the latest metadata version, this updates the action version, which bundles a later pkginfo


pquentin (approved):
Thanks! LGTM.

gareth-ellis:
I'm going to skip waiting for the full pipelines to complete, as I have not changed anything that those use, and I need to rerun once merged anyway

PR #1853, opened 2024-05-30
Update _version.py

gareth-ellis:
Now 2.11.0 is released we need to update the _version file


ebadyano:
I thought in the past we would update the dev to be the next version, did we stop doing that?

gareth-ellis:
The release instructions say to set to released.dev0 - since we don't know yet if the next will be a fix, minor or major

ebadyano (approved):
lgtm

PR #1854, opened 2024-05-30
[Metrics store] Add support for custom templates

inqueue:
This change adds the ability to customize indices in the metrics store. The goal is to support ILM policies and eventually shift to using data streams.

* Support the addition of component templates having the `@custom` suffix.
* Add documentation.

### DIFF
--- a/docs/metrics.rst
+++ b/docs/metrics.rst
@@ -6,6 +6,34 @@ Metrics Records
 
 At the end of a race, Rally stores all metrics records in its metrics store. Metrics can be kept in memory or written to a dedicated Elasticsearch cluster (not the cluster where Rally ran its benchmarks). This can be configured in the `[reporting] section <https://esrally.readthedocs.io/en/stable/configuration.html#reporting>`_.
 
+Rally writes data to the following indices in the Elasticsearch metrics store:
+
+* ``rally-races-*``: Contains the metadata of a benchmark. This is a subset of the metrics records but only contains the metadata of a benchmark.
+* ``rally-metrics-*``: Contains all metrics records.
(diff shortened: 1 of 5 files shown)


gbanasiak:
```
TypeError: put_index_template() got an unexpected keyword argument 'ignore_missing_component_templates'
```

We would need to modify assumption on metric store ES version to support this approach. We're still maintaining compatibility with 6.8 for target cluster which I think is something we could drop, but 7.17.x must stay for now. Unfortunately `ignore_missing_component_templates` was added in 8.7. Do we want to bifurcate and have different requirements for metric store and target cluster?

inqueue:
> ```
> TypeError: put_index_template() got an unexpected keyword argument 'ignore_missing_component_templates'
> ```
> 
> We would need to modify assumption on metric store ES version to support this approach. We're still maintaining compatibility with 6.8 for target cluster which I think is something we could drop, but 7.17.x must stay for now. Unfortunately `ignore_missing_component_templates` was added in 8.7. Do we want to bifurcate and have different requirements for metric store and target cluster?

Thanks, Grzegorz. I hadn't realized `ignore_missing_component_templates` was so new. The parameter is not even supported in elasticsearch-py until 8.14 via https://github.com/elastic/elasticsearch-py/pull/2525 (Rally uses 8.6.1 currently). 

Rally would need to load an empty `*@custom` component template to support this, if one did was not already in the metrics cluster:

```
{
  "template": {}
}
```

The ultimate goal was to slowly move toward using data streams, or at least provide the option for it, to allow for better management of metrics retention, and pushing metrics to serverless. We are not there yet on any of these fronts, and therefore, closing this PR.

PR #1859, opened 2024-06-21
Modify rally to allow multiple cars with complex configuration

favilo:
We'd made some decisions in the past to only allow a single car with a python based configuration.

I'm not certain of the reasoning behind that, but I assume it was because this simplified the evaluation of cars.

This changes that to allow multiple cars to run with complex python based configuration. In particular, it allows us to run APM Tracing and X-Pack-Security simultaneously, which is something I need for benchmarking reasons.


esrally/mechanic/team.py
```diff
@@ -67,11 +69,11 @@ def __init__(self, root_path, entry_point):
...
             if BootstrapHookHandler(Component(root_path=p, entry_point=Car.entry_point)).can_load():
-                if not root_path:
-                    root_path = p
+                if p not in root_paths:
+                    root_paths.append(p)
                 # multiple cars are based on the same hook
-                elif root_path != p:
-                    raise exceptions.SystemSetupError(f"Invalid car: {name}. Multiple bootstrap hooks are forbidden.")
+                # elif root_paths != p:
+                #     raise exceptions.SystemSetupError(f"Invalid car: {name}. Multiple bootstrap hooks are forbidden.")
```
AI-IshanBhatt, line 76: Looks like a leftover comment.

tests/mechanic/team_test.py
```diff
@@ -48,17 +48,17 @@ def test_load_known_car(self):
         car = team.load_car(self.team_dir, ["default"], car_params={"data_paths": ["/mnt/disk0", "/mnt/disk1"]})
         assert car.name == "default"
         assert car.config_paths == [os.path.join(current_dir, "data", "cars", "v1", "vanilla", "templates")]
-        assert car.root_path is None
+        assert car.root_path == []
```
gbanasiak, line 51: Nit: When looking at this change it struck me, `root_path` could be renamed to `root_paths` in `Car` class.

gbanasiak (commented):
I've tested with:

```
esrally install --distribution-version=8.14.0 --node-name="rally-node-0" --network-host="127.0.0.1" --http-port=39200 --master-nodes="rally-node-0" --seed-hosts="127.0.0.1:39300" --team-path=../rally-teams --car="defaults,x-pack-security,apm-tracing" --car-params=/tmp/car_params.json
```

and confirmed both X-Pack and APM Tracing (https://github.com/elastic/rally-teams/pull/85) configuration was included.

The change LGTM overall. Thank you.

Apart from the enhancement I like the fact that type hints are introduced here and there. However, current mypy configuration in Rally is _extremely_ narrow after https://github.com/elastic/rally/pull/1798 (prior to that PR there were _no_ mypy checks at all), so in https://github.com/elastic/rally/pull/1859/commits/13f1272b4e7d713a73c3fc27c9b79aa877b4b0cd I've added override section where we could add modules that we are revisiting to include hints. For starters, I've included the following modules, which is a _subset_ of modules touched in the PR:
```
[[tool.mypy.overrides]]
module = [
    "esrally.mechanic.team",
    "esrally.utils.modules",
]
[..]
```
Feel free to add remaining modules changed in the PR (`make lint` to trigger mypy check).

Majority of https://github.com/elastic/rally/pull/1859/commits/13f1272b4e7d713a73c3fc27c9b79aa877b4b0cd is about fixing `tests` after mypy scope was fixed to include the entirety of `esrally`, `it` and `tests`, but I also had to fix some details in the introduced hints.

mypy was introduced through pre-commit and now I'm no longer happy with this approach. pre-commit creates a dedicated virtual environment so we need to specify additional dependencies separately and explicitly (`--install-types` is not recommended as per https://github.com/pre-commit/mirrors-mypy). I would prefer to have a single source of dev optional dependencies and include them both in `pyproject.toml` and mypy configuration. We can streamline this later.

esrally/client/asynchronous.py
```diff
@@ -223,7 +223,7 @@ def __init__(self, config):
         self._loop = None
         self.client_id = None
         self.trace_configs = None
-        self.enable_cleanup_closed = None
+        self.enable_cleanup_closed = False
```
gbanasiak, line 226: @favilo This is the only material change in https://github.com/elastic/rally/pull/1859/commits/13f1272b4e7d713a73c3fc27c9b79aa877b4b0cd I think. I've followed base classes and confirmed `enable_cleanup_closed` defaults to `False`, please check.

esrally/utils/modules.py
```diff
@@ -34,78 +36,89 @@ class ComponentLoader:
...
-            m = importlib.import_module(p)
-            if p == root_module_name:
+        for name, p in self._modules(module_dirs, component_name, root_path):
+            self.logger.debug("Loading module [%s]: %s", name, p)
+            spec = importlib.util.spec_from_file_location(name, p)
+            if spec is None or spec.loader is None:
+                raise exceptions.SystemSetupError(f"Could not load module [{name}]")
+            m = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(m)
+            if name == root_module_name:
```
gbanasiak, line 74: The change from `importlib.import_module()` to `importlib.util.spec_from_file_location() + importlib.util.module_from_spec()` is to differentiate between 2 modules with the same name but different absolute paths, e.g. `rally-teams/cars/v1/x_pack/base/config.py` and `rally-teams/cars/v1/apm_tracing/config.py`? WDYT about adding a comment with some context?

esrally/client/asynchronous.py
favilo: Yeah, that is why I chose `False`, because it's the default in the base class.

esrally/client/asynchronous.py
```diff
@@ -77,7 +77,7 @@ async def send(self, conn: "Connection") -> "ClientResponse":
         self.response = self.response_class(
             self.method,
             self.original_url,
-            writer=self._writer,
+            writer=self._writer,  # type: ignore[arg-type]  # TODO remove this ignore when introducing type hints
```
favilo, line 80: We could also just an an `assert self._writer is not None` above this line, and it won't complain. 

Oh, wait... It might not be necessary depending on the `response_class` we've got, huh?

tests/mechanic/team_test.py
favilo: So, I attempted this just now again. And the issue is we equate `Car` and `Plugin` in several places, just referring to `component.root_path` directly. 

So, instead of forcibly rewriting those to conditionally use the correct field, I will just keep it as is for now.

.pre-commit-config.yaml
```diff
@@ -22,7 +22,13 @@ repos:
     rev: v1.6.1
     hooks:
       - id: mypy
+        additional_dependencies: [
+          "elasticsearch[async]==8.6.1",
+          "elastic-transport==8.4.1",
+          "types-tabulate",
```
favilo, line 28: Do we want to add a version here to match the one we're using?

tests/mechanic/team_test.py
gbanasiak: Ok, thanks for checking.

esrally/client/asynchronous.py
gbanasiak: I haven't analyzed this deeply and just went for a safe type ignore, but thinking about this now wouldn't such assertion fail at runtime? (I haven't tested)

My understanding is `StaticRequest` class extends `aiohttp.ClientRequest` to support static API responses feature of Rally which allows us to test Rally itself providing curated responses from non-existing ES (ES delays and performance eliminated from the equation). The extension is done by overwriting `send()` method. However, `_writer` property is updated in the _original_ `send()` method. Otherwise it defaults to `None` because `_writer` property is taken from `__writer` class variable which is set to `None` initially.

.pre-commit-config.yaml
gbanasiak: You're right, https://github.com/elastic/rally/pull/1859/commits/30849e6fe0450e166331d0769e12aac4967039a4.

gbanasiak:
@favilo Do you think we should add `esrally.utils.io` and `esrally.utils.process` to mypy overrides too? My thinking is, if we start adding type hints somewhere, let's mypy do a proper check there, instead of a narrowed-down one. This will raise more mypy errors though, and the scope of PR will increase. Alternatively, we can retract type hint changes from those 2 modules, and add them in a separate PR.

gbanasiak (approved):
I am happy with current form. What is left are non-critical comments.

favilo:
> @favilo Do you think we should add `esrally.utils.io` and `esrally.utils.process` to mypy overrides too? My thinking is, if we start adding type hints somewhere, let's mypy do a proper check there, instead of a narrowed-down one. This will raise more mypy errors though, and the scope of PR will increase. Alternatively, we can retract type hint changes from those 2 modules, and add them in a separate PR.

I will add those modules in, I like the idea of decreasing debt continuously. And I still need this branch around for my config to load properly

favilo:
@gbanasiak I went a little crazy with those two files... Turned on a stricter mode of `mypy` than I absolutely needed. But this way, we can be sure we are future proof

esrally/client/asynchronous.py
favilo: Ah, that is frustrating. Eh, we can get back to it later, then

pyproject.toml
```diff
@@ -10,13 +10,13 @@ path = "esrally/_version.py"
 name = "esrally"
 dynamic = ["version"]
 authors = [
-  {name="Daniel Mitterdorfer", email="daniel.mitterdorfer@gmail.com"},
+    { name = "Daniel Mitterdorfer", email = "daniel.mitterdorfer@gmail.com" },
```
favilo, line 13: Sorry, ran a toml formatter on this file.

pyproject.toml
```diff
@@ -143,12 +146,26 @@ junit_logging = "all"
...
+# opportunistically introduced whenever we revisit a group of modules.
+#
+# The intention of the following largely reduced global config scope is
+# verification of argument types in config.Config methods while introducing
+# configuration properties (props). The intention of "disable_error_code" option
+# is to keep "arg-type" error code, while disabling other error codes.
+# Ref: https://github.com/elastic/rally/pull/1798
 [tool.mypy]
-python_version = 3.8
+python_executable="$VIRTUAL_ENV/bin/python"
```
favilo, line 159: @gbanasiak, can you test `make lint` with this line? I'm hoping this works everywhere.

It seems to work properly with both my local setup and the CI so far.

gbanasiak (approved):
The changes in `esrally.utils.io` and `esrally.utils.process` LGTM.

gbanasiak:
I'm testing simplified mypy config in https://github.com/elastic/rally/pull/1859/commits/2058474dc796f3e81f7bc4379041e82a57e1c414.

gbanasiak:
https://github.com/elastic/rally/pull/1859/commits/56ef7e5224af543f5f8b2aaa06324c56926b3d23 changed file object type in `FileOffsetTable` from generic IO to byte-based IO, but files are opened in text-mode in this IO wrapper. The `split()` is called against `str` not `bytes` at runtime here: https://github.com/elastic/rally/blob/56ef7e5224af543f5f8b2aaa06324c56926b3d23/esrally/utils/io.py#L554

```
[..]
  File "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1720474499404871070/elastic/rally-it/.nox/it-3-8/lib/python3.8/site-packages/esrally/utils/io.py", line 656, in skip_lines
    offset, remaining_lines = file_offset_table.find_closest_offset(number_of_lines_to_skip)

  File "/opt/buildkite-agent/builds/bk-agent-prod-gcp-1720474499404871070/elastic/rally-it/.nox/it-3-8/lib/python3.8/site-packages/esrally/utils/io.py", line 558, in find_closest_offset
    line_number, offset_in_bytes = (int(i) for i in line.strip().split(b";"))

TypeError: must be str or None, not bytes
```

Issue #1862, opened 2024-06-27
Elasticsearch cluster authentication credentials can leak into rally.log


inqueue:
Elasticsearch authentication credentials can leak into rally.log when using the `--kill-running-processes` CLI argument and there is a running process to kill. Rally should not leak basic authentication credentials or API keys to rally.log.

#### Problem code line
* https://github.com/elastic/rally/blob/18e4a2ca8f8be9e1e0d9354f63e95aec9bbfde5f/esrally/utils/process.py#L184

#### Example log

```
2024-06-27 09:54:12,866 -not-actor-/PID:1698183 esrally.utils.process INFO Killing lingering process with PID [1696855] and command line [['/home/user/.conda/envs/rally/bin/python', '/home/user/.conda/envs/rally/bin/esrally', 'race', '--track-path=.', '--pipeline=benchmark-only', '--target-hosts=https://10.13.31.10:9200', "--client-options=basic_auth_user:'elastic',basic_auth_password:'TheLeakedPassword',use_ssl:true,verify_certs:false,timeout:60", '--kill-running-processes']].
```

PR #1863, opened 2024-06-27
Avoid logging ES credentials from running Rally processes

inqueue:
Replace the basic authentication password and API key with `'*****'` before creating the log entry when killing a running process.

Support both `--client-options` CLI formats:

- [x] name:value pairs
- [x] inline JSON

Closes https://github.com/elastic/rally/issues/1862

### DIFF
--- a/esrally/utils/process.py
+++ b/esrally/utils/process.py
@@ -181,7 +181,7 @@ def find_all_other_rally_processes() -> List[psutil.Process]:
 
 def redact_cmdline(cmdline: list) -> List[str]:
     """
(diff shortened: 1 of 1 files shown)


gbanasiak (approved):
LGTM, thank you. I've fixed incorrect method call after the latest refactor in https://github.com/elastic/rally/pull/1863/commits/d5a6508fdc9aae38c821e32dae738a41354e8e94, see https://github.com/elastic/rally/pull/1863/commits/f6a3a893d974ba51c4018c556602b2d570b47fcf.

PR #1865, opened 2024-07-16
Add an option to filter by challange in list command

AI-IshanBhatt:
`esrally list races` command does not have an option to filter races based on the **challenge**.

One would have to rely on --benchmark-name to filter the races and if user did not provide the tag while creation the race it is very hard to find out the race of the interest.

This commit addresses that.


favilo (approved):
Seems straightforward enough

PR #1868, opened 2024-08-02
Bulk http status & 429 retries

gareth-ellis:
Future direction:
- need to avoid messy nested fields 
- service_time should be a sum of all retries
- follow approach of other products, such as logstash


gbanasiak:
Context: https://elastic.slack.com/archives/C03P67VQAP2/p1721939793707489

gareth-ellis:
My approach so far seems to work - Please take a look

esrally/driver/runner.py
```diff
@@ -527,19 +531,28 @@ async def __call__(self, es, params):
...
+            else self.simple_stats(bulk_size, unit, response, api_kwargs, emit_lines_to_retry=retries_on_429 > 0)
+        )
+        for _ in range(retries_on_429):
             retry_stats.append(stats)
             if len(lines_to_retry) == 0:
                 break
+            self.logger.warning("Retrying %d documents that previously resulted in a 429.", len(lines_to_retry) / 2)
             api_kwargs["body"] = lines_to_retry
-            bulk_size = len(lines_to_retry) / 2
+            bulk_size = len(lines_to_retry) / 2  # at this point the data always contains action metadata.
```
fressi-elastic: Can you explain why bulk size is the half of lines to retry?

esrally/driver/runner.py
```diff
@@ -526,20 +527,75 @@ async def __call__(self, es, params):
...
+                break
+        if len(retry_stats) == 1:
+            stats = retry_stats[0]
+        else:
+            for stats in retry_stats:
+                total_success += stats["success-count"]
+                total_error += stats["error-count"]
+                total_time += stats["took"]
+                sum_bulk_request_size_bytes += stats["bulk-request-size-bytes"]
+                sum_total_document_size_bytes += stats["total-document-size-bytes"]
```
gbanasiak: Isn't there a risk of `KeyError` in case of simple stats here? There are no `bulk-request-size-bytes` and `total-document-size-bytes` keys in stats dictionary returned by `simple_stats()` ?

esrally/driver/runner.py
```diff
@@ -526,20 +527,75 @@ async def __call__(self, es, params):
...
+            if detailed_results
+            else self.simple_stats(bulk_size, unit, response, api_kwargs, emit_lines_to_retry=retries_on_429 > 0)
+        )
+        retry_stats.append(stats)
+        for _ in range(retries_on_429):
+            if len(lines_to_retry) == 0:
+                break
+            self.logger.warning("Retrying %d documents that previously resulted in a 429.", len(lines_to_retry) / 2)
+            api_kwargs["body"] = lines_to_retry
+            bulk_size = len(lines_to_retry) / 2  # at this point the data always contains action metadata.
```
gbanasiak: > at this point the data always contains action metadata

This is a slight off-topic:

I've spent some time digging once I saw this comment. I think bulk runner always receives action-and-metadata lines in the `body` param today (see [here](https://github.com/elastic/rally/blob/a9c3ab99ee4029c592ee7b85912fa1fa19fad225/esrally/track/params.py#L1062)). If corpus does not include them they are generated in earlier processing stages. I don't quite understand this from the above code:

```
        if with_action_metadata:
            api_kwargs.pop("index", None)
            # only half of the lines are documents
            response = await es.bulk(params=bulk_params, **api_kwargs)
        else:
            response = await es.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
```

The `only half of the lines are documents` comment suggests the `else` clause is different, but it isn't. There's nothing in `bulk()` method of ES client that would magically add action-and-metadata lines. Also `doc_type` is ignored I think, it's a leftover from old ES versions.

I think we could simplify / remove this.

esrally/driver/runner.py
```diff
@@ -526,20 +527,75 @@ async def __call__(self, es, params):
...
+                }
+            else:
+                stats = {
+                    "success-count": total_success,
+                    "error-count": total_error,
+                    "retry-count": retry_count,
+                    "took": total_time,
+                    "success": len(lines_to_retry) == 0,
+                    "retried": retry_count > 0,
+                }
```
gbanasiak: This exposes the details of stats at `_call__()` level which were previously hidden either in `simple_stats()` or `detailed_stats()`. Can we avoid this? We could have a method that iterates through response documents, and:
* calls another method (passed as an argument) that increases stats counters for each document,
* builds a retry list (optionally).

gbanasiak (commented):
This is an interesting coupling problem with `super().detailed_stats(...)` used in elastic/logs. Tracks version control is only for ES, not for Rally, so changing custom `detailed_stats()` say like below will fail when calling parent's method in old Rally.

```
    def detailed_stats(self, params, response, emit_lines_to_retry=False):
        stats = super().detailed_stats(params, response, emit_lines_to_retry) <--- HERE
        return {**stats, **params["param-source-stats"]}
```

It would be best if we minimized the surface between tracks code and rally code. Our [docs](https://esrally.readthedocs.io/en/stable/advanced.html#creating-your-own-operations-with-custom-runners) only mention top-level runner call with `es` and `params` arguments. Due to this I would be happy to declare that what this particular runner is doing as not supported. I would even go as far as clarify this in documentation.

My vote would be to change elastic/logs custom runner to avoid calling `detailed_stats()` completely and backport this to _all_ branches where modified Rally might potentially be used, so I'd say all 8.x up until now.

Something like this perhaps?

```
class RawBulkIndex(BulkIndex):
    async def __call__(self, es, params):
        meta_data = await super().__call__(es, params)
        if params.get("detailed-results", False):
            meta_data.update(params["param-source-stats"])
        return meta_data
```

The alternative would be to _keep_ `detailed_results()` as is, and determine documents to retry separately but that means iterating through a response twice which would impact processing time.

I'm curious about @fressi-elastic thoughts on that one as well.

I'm adding further comments below.

---

I think we also planned on adding exponential back-off. Is this still a plan, or not in this PR?

esrally/driver/runner.py
gareth-ellis: Ah yes - you can see which variant i've been testing with ;) fixed in 04879dbb744934119bde082d1486230778c38b89

esrally/driver/runner.py
gareth-ellis: Yea - it seems at one point we calculated the number of documents here, but that was removed and the comment not removed. I'll remove the comment for now, I think i would need to test a little more whether we have code paths that use the with_action_metadata or not

esrally/driver/runner.py
```diff
@@ -515,31 +515,86 @@ async def __call__(self, es, params):
...
         if with_action_metadata:
             api_kwargs.pop("index", None)
-            # only half of the lines are documents
             response = await es.bulk(params=bulk_params, **api_kwargs)
         else:
             response = await es.bulk(doc_type=params.get("type"), params=bulk_params, **api_kwargs)
-
-        stats = self.detailed_stats(params, response) if detailed_results else self.simple_stats(bulk_size, unit, response)
+        retry_stats = []
+        total_success = total_error = total_time = sum_bulk_request_size_bytes = sum_total_document_size_bytes = 0
```
fressi-elastic: Please think about regrouping variables (and part of the behavior) in some data class. There are too many variables here and it is getting hard to keep track of all the variables this function has. In alternative please also think about splitting the method in sub-methods.

esrally/driver/runner.py
```diff
@@ -515,31 +515,86 @@ async def __call__(self, es, params):
...
+        stats, lines_to_retry = (
```
fressi-elastic: I generally prefer the simplicity of a normal if block. I see no advantage on putting the if in the middle instead of the beginning of the block.

fressi-elastic: Could we have `lines_to_retry` as a new attribute of stats? Which type `stats` is? Without type hints this is very hard to follow for me.

esrally/driver/runner.py
```diff
@@ -515,31 +515,86 @@ async def __call__(self, es, params):
...
+            self.detailed_stats(params, response, emit_lines_to_retry=retries_on_429 > 0)
+            if detailed_results
+            else self.simple_stats(bulk_size, unit, response, api_kwargs, emit_lines_to_retry=retries_on_429 > 0)
+        )
+        retry_stats.append(stats)
+        for _ in range(retries_on_429):
+            if len(lines_to_retry) == 0:
```
fressi-elastic: if not lines_to_retry

esrally/driver/runner.py
```diff
@@ -591,7 +640,11 @@ def _utf8len(line):
                 shards_histogram[sk]["item-count"] += 1
             if data["status"] > 299 or ("_shards" in data and data["_shards"]["failed"] > 0):
                 bulk_error_count += 1
+                max_doc_status = max(max_doc_status, data["status"])
```
fressi-elastic, line 687: Why are you keeping the maximum value of status codes? Which is the purpose?

esrally/driver/runner.py
```diff
@@ -611,24 +666,42 @@ def _utf8len(line):
         if "ingest_took" in response:
             stats["ingest_took"] = response["ingest_took"]
 
-        return stats
+        return stats, lines_to_retry
+
+    def get_bulk_lines(self, params):
```
fressi-elastic, line 713: The `return` statements here make all `elfi` replaceable with `if` and the final `else` rendundant.

esrally/driver/runner.py
```diff
@@ -611,24 +666,42 @@ def _utf8len(line):
...
+            return params["body"].split(b"\n")
+        elif isinstance(params["body"], str):
+            return params["body"].split("\n")
+        elif isinstance(params["body"], list):
+            return params["body"]
+        else:
+            raise exceptions.DataError("bulk body is not of type bytes, string, or list")
 
-    def simple_stats(self, bulk_size, unit, response):
+    def simple_stats(self, bulk_size, unit, response, params, emit_lines_to_retry=False):
```
fressi-elastic: You are adding the same logic to both sible_stats and detailed_stats. You could make instead a 3rd function that add missing data after calling one of the two. Current code would be much simpler and easy to understand:

```Python
if detailed_results:
  stats = self.detailed_stats(...)
else:
  stats = self.simple_stats(...)
if retries_on_429 > 0:
  lines_to_retry = self.lines_to_retry(stats)
```

fressi-elastic (changes requested):
I spot some opportunities to simplify your code and remove code duplication.

I also don't think changing the return type of `simple_stats` and `detailed_stats` is save and necessary having tracks extending this class and potentially already calling this methods. It would be simpler adding a third method to extract retry_lines and fill additional data to stats dictionary.

esrally/driver/runner.py
gareth-ellis: Because that's how bulk requests works - there is an action line, and a data line, when we talk about 2500 documents, the request will contain 5000 lines

esrally/driver/runner.py
gareth-ellis: I've change this slightly 1895973

esrally/driver/runner.py
```diff
@@ -457,6 +458,80 @@ def escape(v):
         return str(v)
 
 
+@dataclass
+class BulkStats:
+    """Metrics collected from a single bulk indexing request."""
+
+    success_count: Optional[int]
+    error_count: int
```
gbanasiak, line 466: Why `success_count` is optional? Can we default both `success_count` and `error_count` to 0?

esrally/driver/runner.py
```diff
@@ -457,6 +458,80 @@ def escape(v):
...
+        self.took = (self.took or 0) + (other.took or 0)
+        self.bulk_request_size_bytes += other.bulk_request_size_bytes
+        self.total_document_size_bytes += other.total_document_size_bytes
+        if self.retry_count == 0:
+            self.retry_count = 2
+        else:
+            self.retry_count += 1
+        self.error_429_indices = other.error_429_indices
+        self.request_status = other.request_status
+        self.max_doc_status = other.max_doc_status
```
gbanasiak, line 502: The fact that some things get increased while others replaced when accumulating might not be obvious to the first-time reader. I think I would add a comment on top of `accumulate()` to explain that.

esrally/driver/runner.py
```diff
@@ -515,30 +590,54 @@ async def __call__(self, es, params):
...
-        stats = self.detailed_stats(params, response) if detailed_results else self.simple_stats(bulk_size, unit, response)
+        stats = self._parse_stats(params, bulk_size, unit, response, api_kwargs, detailed_results)
+
+        for _ in range(retries_on_429):
+            if not stats.error_429_indices:
+                break
+            lines_to_retry = self._build_retry_body(api_kwargs, stats.error_429_indices)
+            self.logger.warning("Retrying %d documents that previously resulted in a 429.", len(lines_to_retry) / 2)
+            api_kwargs["body"] = lines_to_retry
+            bulk_size = len(lines_to_retry) / 2
```
gbanasiak, line 612: I would replace `len(lines_to_retry) / 2` with `len(stats.error_429_indices)`.

esrally/driver/runner.py
```diff
@@ -515,30 +590,54 @@ async def __call__(self, es, params):
...
+        meta_data.update(stats.as_dict())
         return meta_data
 
+    def _parse_stats(self, params, bulk_size, unit, response, api_kwargs, detailed_results):
+        if detailed_results:
+            return self.detailed_stats(params, response)
+        return self.simple_stats(bulk_size, unit, response, api_kwargs)
+
+    def _build_retry_body(self, api_kwargs, error_429_indices):
+        bulk_lines = self.get_bulk_lines(api_kwargs)
```
gbanasiak: `bulk_lines` remains the same throughout entire lifetime of the runner's call. Can we avoid recalculating it on every retry?

esrally/driver/runner.py
```diff
@@ -515,30 +590,54 @@ async def __call__(self, es, params):
...
+                break
+            lines_to_retry = self._build_retry_body(api_kwargs, stats.error_429_indices)
+            self.logger.warning("Retrying %d documents that previously resulted in a 429.", len(lines_to_retry) / 2)
+            api_kwargs["body"] = lines_to_retry
+            bulk_size = len(lines_to_retry) / 2
+            response = await es.bulk(params=bulk_params, **api_kwargs)
+            retry_result = self._parse_stats(params, bulk_size, unit, response, api_kwargs, detailed_results)
+            stats.accumulate(retry_result)
+            if response.meta.status not in (200, 201, 429):
+                self.logger.warning("%s after bulk request retry. Payload: %s", response.meta.status, lines_to_retry)
```
gbanasiak: There is a potential for very long output here. I would only inform about selected stats at `warning` level.

esrally/driver/runner.py
gbanasiak: It gives us an indication of what was the worst treatment any of the potentially thousands of documents received as part of processing a bulk request. Status codes are sorted by severity in a sense.

gbanasiak (commented):
This looks good to me. I've added some non-blocking comments after going through the code. Can we have a unit test showing at least a single retry in action together with stats accumulation?

It would be great to add exponential back-off, but this can be done in a follow-up.

esrally/driver/runner.py
gareth-ellis: yes. I can switch that to debug - i was using it for diagnosing another issue,

tests/driver/runner_test.py
```diff
@@ -1517,6 +1595,189 @@ async def test_bulk_index_success_with_refresh_invalid(self, es):
...
+            ),
+        ]
+        es.bulk = mock.AsyncMock(side_effect=bulk_responses)
+        bulk = runner.BulkIndex()
+        result = await bulk(es, dict(params))
+        assert result["index"] == "test"
+        assert result["unit"] == "docs"
+        assert result["weight"] == 3
+        assert result["retried"] is True
+        assert result["retry-count"] == 3
```
gbanasiak: Shouldn't `retry-count` be equal 2? 1 regular call + 2 retries...

gareth-ellis: Fixed - i had initially set this as a count of requests, so after the first retry we set to 2 rather than 1. Fixed that now

gbanasiak (approved):
LGTM

fressi-elastic (approved):

PR #1869, opened 2024-08-03
Elastic client now uses url_prefix from host dictionary.

RichTeaMan:
I have several Elastic clusters hosted in a single Kubernetes cluster. For policy reasons, I can only expose port 443. Instead, I use different URL paths to get the right combination of cluster and elastic/kibana. As a result, Rally was failing as `--target-hosts` would end up stripping most of the URL.

This change checks if a URL prefix has been configured and passes that through in the ES client factory.

I have ran `make check-all`, the unit tests all work but there are some integration test failures. However, I also get the same failures on master so I'm unsure what to do with them.

I have signed the contributor agreement.


gareth-ellis:
Thanks for the contribution - Yes, the integration tests require some API keys etc, so tend to be run via the CI. I unblocked the CI for your changes, so lets let them run there and see.

gareth-ellis:
Buildkite test this please

gbanasiak:
@elasticmachine update branch

gbanasiak:
Buildkite test this please

gbanasiak:
Buildkite test this please

gbanasiak (approved):
We somehow forgot about this PR, sorry. The change looks good. I've added additional tests, and expanded documentation a bit. Thank you for your contribution.

PR #1871, opened 2024-08-31
[ES-8436] feat: Change Dockerfile base images to Wolfi

favilo:
Running the docker image with a mounted `/rally/.rally` directory that already contains the configuration file didn't work. So I had to change how we parse the logging config and rally.ini files.

Now we dealy the parsing of the environment variables to read time, to make running from docker and outside of docker compatible with each other.

This will require changes to the buildkite scripts to allow pushing a regular docker image for use with nightlies, and potentially rebuilding older images when vulnerabilities are found.

### DIFF
--- a/docker/Dockerfiles/Dockerfile-dev
+++ b/docker/Dockerfiles/Dockerfile-dev
@@ -1,8 +1,10 @@
-FROM cgr.dev/chainguard/wolfi-base
+FROM docker.elastic.co/wolfi/python:3.11.7-dev
 ARG RALLY_VERSION
 ARG RALLY_LICENSE
 
 ENV RALLY_RUNNING_IN_DOCKER=True
+
+USER root
 
 RUN apk update 
 RUN apk add curl git gcc pigz bash
@@ -37,15 +39,6 @@ RUN mkdir -p /rally/.rally && \
     chown -R 1001:0 /rally/
 
 USER 1001
-
-# Install pyenv
-RUN git clone --depth=1 https://github.com/pyenv/pyenv.git /rally/.pyenv
-ENV PYENV_ROOT="/rally/.pyenv"
-ENV PATH=/rally/.local/bin:$PATH
-ENV PATH="$PYENV_ROOT/shims:$PYENV_ROOT/bin:$PATH"
-
-RUN pyenv install 3.11.7
-RUN pyenv global 3.11.7
 
 RUN mkdir -p /rally/esrally
 COPY pyproject.toml /rally/
--- a/docker/Dockerfiles/Dockerfile-dev
+++ b/docker/Dockerfiles/Dockerfile-dev
@@ -1,19 +1,12 @@
-FROM docker.elastic.co/wolfi/python:3.11.7-dev
-ARG RALLY_VERSION
-ARG RALLY_LICENSE
-
-ENV RALLY_RUNNING_IN_DOCKER=True
+FROM docker.elastic.co/wolfi/python:3.11.7-dev AS builder
 
 USER root
 
 RUN apk update 
-RUN apk add curl git gcc pigz bash zstd
+RUN apk add curl git gcc pigz bash zstd bzip2 gzip
 
 # pbzip2 doesn't have a package for wolfi, so we build it from source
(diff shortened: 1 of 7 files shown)


docker/Dockerfiles/Dockerfile-dev
```diff
@@ -1,54 +1,33 @@
...
-    apt-get -y upgrade && \
-    rm -rf /var/lib/apt/lists/*
+USER root
 
-RUN mkdir -p /rally/esrally
-COPY pyproject.toml /rally/
-COPY README.md /rally/
-COPY esrally/ /rally/esrally/
+RUN apk update 
+RUN apk add curl git gcc pigz bash
```
gareth-ellis, line 10: Could we add zstd too? Rally supports that and some of ours tracks use zstd compression for artifacts

favilo: That seems completely reasonable

esrally/log.py
```diff
@@ -95,16 +95,29 @@ def install_default_log_config():
         source_path = io.normalize_path(os.path.join(os.path.dirname(__file__), "resources", "logging.json"))
         with open(log_config, "w", encoding="UTF-8") as target:
             with open(source_path, encoding="UTF-8") as src:
-                # Ensure we have a trailing path separator as after LOG_PATH there will only be the file name
-                log_path = os.path.join(paths.logs(), "")
-                # the logging path might contain backslashes that we need to escape
-                log_path = io.escape_path(log_path)
```
gbanasiak, line 101: We're dropping backlash escaping which was meant to help with Windows judging by https://github.com/elastic/rally/pull/829. But we never officially supported Windows so I'm fine with it.

docker/Dockerfiles/Dockerfile-dev
```diff
@@ -1,51 +1,31 @@
-################################################################################
-# Build stage 0 `builder`:
-# Install Rally from source inside a virtualenv
-################################################################################
+FROM docker.elastic.co/wolfi/python:3.11.7-dev AS builder
```
gbanasiak, line 1: This is probably a good opportunity to define the rule going forward. Will we keep this version aligned with the version of the nightlies (3.11.7), or perhaps we should bump it up all the way to the newest supported version (3.12.x)?

gbanasiak (commented):
This looks good. I like the increased portability the delayed resolution of `${CONFIG_DIR}` and `${LOG_PATH}` variables offer but I think this should get proper description in [here](https://esrally.readthedocs.io/en/2.11.0/configuration.html#rally-configuration) and [here](https://esrally.readthedocs.io/en/2.11.0/configuration.html#logging). It could also be useful to mention it in [Docker section](https://esrally.readthedocs.io/en/2.11.0/docker.html), perhaps [here](https://esrally.readthedocs.io/en/2.11.0/docker.html#configuration)?

esrally/log.py
favilo: Oh, how did I not see this? I am happy to add this logic back to the code.

docs/docker.rst
```diff
@@ -70,6 +70,17 @@ To customize Rally you can create your own ``rally.ini`` and bind mount it using
 
     docker run -v /home/<myuser>/custom_rally.ini:/rally/.rally/rally.ini elastic/rally ...
 
+As a quality of life improvement, the ``rally.ini`` and ``logging.json`` files can use ``${CONFIG_DIR}`` in ``rally.ini`` and ``${LOG_PATH}`` in ``logging.json`` in order to
+make the files more portable. For example::
+
+* In ``rally.ini``, you can set ``root.dir = ${CONFIG_DIR}/benchmarks`` instead of hard-coding the path as ``/rally/.rally/benchmarks``
+* In ``logging.json``, you can set ``"filename": "${LOG_PATH}/rally.log"`` instead of hard-coding the path as ``"filename": "/rally/.rally/logs/rally.log"``
```
gbanasiak, line 77: This bulleted list does not look good in the review. Double colon perhaps?

gbanasiak (approved):
LGTM. One cosmetic docs change needed.

PR #1872, opened 2024-09-02
Fix revision help text

gareth-ellis:
I found whilst trying to run rally and build elasticsearch from a different branch that our help text is wrong. This fixes that

The correct syntax is branch@version, rather than `@branch@version`


gbanasiak (approved):
LGTM

PR #1876, opened 2024-09-13
Cast batch size as int

gareth-ellis:
Spotted this whilst making some other changes to create-track.


gareth-ellis:
@elasticmachine update branch

inqueue (approved):
LGTM

PR #1877, opened 2024-09-20
Docs: fix level of (warmup-)interations

maxjakob:
iterations, warmup-iterations and target-throughput are defined on the [schedule level](https://esrally.readthedocs.io/en/stable/track.html#schedule).


dpifke-elastic (approved):
Good catch, thanks.

Issue #187, opened 2016-11-22
Add support for index templates


danielmitterdorfer:
To simplify handling of tracks in certain cases, we should provide an ability to use [index templates](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html). The suggested syntax is:

```json
"templates": [
    {
                    "name": "my-default-index-template",
                    "index-pattern": "my-index-*",
                    "template": "default-template.json"
      }
]
```

This element needs to be placed as a sibling to `indices`. Either `indices` or `templates` are required. It is also possible to use `indices` together with `templates`. Templates are created before indices and in the order they have been defined.

Upon start of a benchmark, Rally will not delete and create individual indices but instead do:

* `DELETE /my-index-*`: Note that this is done regardless of the `auto-managed` setting of the indices (see #189)
* `DELETE /_template/my-default-index-template`
* `PUT /_template/my-default-index-template`

PR #1873, opened 2024-09-06
feat: Buildkite pipeline for building dev docker images [ES-9318]

favilo:

PR #1878, opened 2024-09-24
Build multi-arch Docker image

pquentin:
This pull request builds on top of #187. It adds multi-arch builds as done in Eland nearly one year ago: https://github.com/elastic/eland/pull/615. The main difference is that in Rally we test the image before pushing it. However, this workflow is not supported for multi-arch images: https://github.com/docker/buildx/issues/59. As a workaround, we build a linux/amd64 image first, test it, and then build and push the actual multi-arch image with linux/amd64 and linux/arm64. While it is named linux/arm64, it's supported natively on Apple silicon hardware, such as M1/M2/M3 MacBooks.

To review this, the only relevant commit is https://github.com/elastic/rally/commit/515ce21ed178fbcb04120af33ce55e6110359ea8. I believe this should only be merged *after* #1871, because it's a different concern, and may need to be reverted independently.

### DIFF
--- a/docker/Dockerfiles/Dockerfile-dev
+++ b/docker/Dockerfiles/Dockerfile-dev
@@ -1,8 +1,10 @@
-FROM cgr.dev/chainguard/wolfi-base
+FROM docker.elastic.co/wolfi/python:3.11.7-dev
 ARG RALLY_VERSION
 ARG RALLY_LICENSE
 
 ENV RALLY_RUNNING_IN_DOCKER=True
+
+USER root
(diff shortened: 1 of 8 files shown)


favilo:
Hah, I've also been working on https://github.com/elastic/rally/pull/1873

favilo:
Hah, I've also been working on https://github.com/elastic/rally/pull/1873

pquentin:
Excellent! Closing in favor of #1873.

PR #1882, opened 2024-10-17
fixes #1881 - composite operation failing when track_total_hits is false

### LINKED ISSUE #1881: operation-type:composite fails when track_total_hits is false
We are using a composite operation to benchmark searchable snapshots using async search like below:

```json
{
  "operation": {
    "operation-type": "composite",
    "name": "date_histogram_by_day-90d-hotfrozen",
    "iterations": 10,
    "requests": [
      {
        "stream": [
          {
            "name": "async-search",
            "operation-type": "submit-async-search",
            "index": "hotfrozenlogs",
            "request-params": {
              "track_total_hits": "false"
            },
            "body": {
              "size": 0,
              "query": {
                "bool": {
                  "filter": [
                    {
                      "range": {
                        "@timestamp": {
                          "gte": "2024-01-01T00:00:00",
                          "lt": "2024-04-01T00:00:00"
                        }
                      }
                    },
                    {
                      "term": {
                        "log.file.path": {
                          "value": "/var/log/messages/salttouch"
                        }
                      }
                    }
                  ]
                }
              },
              "aggs": {
                "time": {
                  "date_histogram": {
                    "field": "@timestamp",
                    "calendar_interval": "1d"
                  }
                }
              }
            }
          }
        ]
      },
      {
        "operation-type": "get-async-search",
        "retrieve-results-for": [
          "async-search"
        ]
      },
      {
        "operation-type": "delete-async-search",
        "delete-results-for": [
          "async-search"
        ]
      }
    ]
  }
}
```
And its failing with:

`esrally.exceptions.RallyError: Cannot run task [date_histogram_by_day-90d-hotfrozen]: Cannot execute [user-defined context-manager enabled runner for [composite]]. Provided parameters are: ['operation-type', 'name', 'iterations', 'requests', 'include-in-reporting']. Error: ['total'].`

The log shows GetAsyncSearch is trying to read the `total` in: https://github.com/elastic/rally/blob/543b4dcf6d602db2e9ef37ddf134a7833b9f530f/esrally/driver/runner.py#L2530

```
File "/usr/local/lib/python3.8/site-packages/esrally/driver/runner.py", line 2530, in __call__
  "hits": response["response"]["hits"]["total"]["value"],
KeyError: 'total'
```

ugosan:
This PR addresses an issue (#1881) where a composite operation using async search fails when track_total_hits is set to false in the request. The failure occurs because the response does not include the hits.total field, causing a KeyError during the retrieval of async search results.

Changes:
- Updated submit-async-search runner to handle track_total_hits when set to false.
- Added checks to avoid accessing hits.total when track_total_hits is false.
- Modified the composite context to store the value of track_total_hits and use it during result processing.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -2501,7 +2501,17 @@ class SubmitAsyncSearch(Runner):
         # id may be None if the operation has already returned
         search_id = response.get("id")
         CompositeContext.put(op_name, search_id)
+        
+        #track_total_hits is true by default
+        CompositeContext.put("track_total_hits", True)
+        
+        if "track_total_hits" in params.get("body").keys():
+            CompositeContext.put("track_total_hits", bool(eval(params.get("body").get("track_total_hits").capitalize())))
+            
+        if "track_total_hits" in request_params.keys():
+            CompositeContext.put("track_total_hits", bool(eval(request_params.get("track_total_hits").capitalize())))
 
+             
     def __repr__(self, *args, **kwargs):
         return "submit-async-search"
 
@@ -2527,11 +2537,14 @@ class GetAsyncSearch(Runner):
             success = success and not is_running
             if not is_running:
                 stats[search] = {
-                    "hits": response["response"]["hits"]["total"]["value"],
-                    "hits_relation": response["response"]["hits"]["total"]["relation"],
                     "timed_out": response["response"]["timed_out"],
                     "took": response["response"]["took"],
                 }
+                
+                if CompositeContext.get("track_total_hits"):
+                    stats[search]["hits"] = response["response"]["hits"]["total"]["value"]
+                    stats[search]["hits_relation"] = response["response"]["hits"]["total"]["relation"]
+            
 
         return {
             # only count completed searches - there is one key per search id in `stats`


esrally/driver/runner.py
```diff
@@ -2495,10 +2495,13 @@ def __repr__(self, *args, **kwargs):
 class SubmitAsyncSearch(Runner):
     async def __call__(self, es, params):
         request_params = params.get("request-params", {})
+
+        # enforce wait_for_completion_timeout = 0 so there is always an id associated with the async search
+        request_params["wait_for_completion_timeout"] = 0
```
gareth-ellis, line 2500: How about checking if wait_for_completion_timeout is set in the request, then if not set it to 0, otherwise allow the current value? That would make it possible for users to override, if they would like. We should still keep a ticket open to see if we can work round it failing if the request completes synchronously

esrally/driver/runner.py
```diff
@@ -2495,10 +2495,13 @@ def __repr__(self, *args, **kwargs):
...
         response = await es.async_search.submit(body=mandatory(params, "body", self), index=params.get("index"), params=request_params)
 
         op_name = mandatory(params, "name", self)
-        # id may be None if the operation has already returned
+
```
gareth-ellis, line 2504:
```suggestion
```

esrally/driver/runner.py
```diff
@@ -2521,18 +2524,21 @@ async def __call__(self, es, params):
         searches = mandatory(params, "retrieve-results-for", self)
         request_params = params.get("request-params", {})
         stats = {}
+
```
gareth-ellis, line 2527:
```suggestion
```

esrally/driver/runner.py
ugosan: agreed, let me add that

gareth-ellis (approved):
LGTM, thank you!

PR #1884, opened 2024-10-22
Add Jinja filters for creating dynamic date ranges based on now()

inqueue:
In serverless, data is considered hot when document timestamps are within the boost window and cold when timestamps are outside the boost window. Tracks using date ranges for corpus generation are limited to static dates, presenting a problem for nightly benchmarks where all ingested data is indexed cold.

This commit adds two custom Jinja filters to the track loader to create dynamic date ranges. These simple functions are most useful in the serverless context for any track utilizing date ranges for corpus generation. Usage:

#### `get_start_date(format)`

Pair with the `now` global template variable to produce a start date:

```
"start-date": {{ now | get_start_date('%Y-%m-%d') | tojson }}
```

Result:

```json
"start-date": "2024-10-22"
```

#### `get_end_date(duration_days, format)`

```
"end-date": {{ now | get_end_date(3, '%Y-%m-%d') | tojson }}
```

Result:

```json
"end-date": "2024-10-25"
```

### Documentation

Documentation will be added with a subsequent commit to this PR. The [documentation](https://esrally.readthedocs.io/en/latest/advanced.html#extensions) on the `now` global filter does not mention the return value is a float as it is the result of `time.time()`.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -6,7 +6,7 @@ Advanced Topics
 Templating Track Files with Jinja2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Rally uses `Jinja2 <http://jinja.pocoo.org/docs/dev/>`_ as a template language so you can use Jinja2 expressions in track files.
+Rally uses `Jinja2 <https://jinja.palletsprojects.com/en/latest/>`_ as a template language so you can use Jinja2 expressions in track files.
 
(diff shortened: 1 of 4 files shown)


martijnvg (commented):
This looks great 👍 . I think we can use this in several tracks that use data generation. I would say we also want this for stateful benchmark runs, since it just makes a benchmark more realistic.

favilo (approved):
Looks good

PR #1885, opened 2024-10-23
feat: Add ability to run esrallyd inside docker [ES-9146]

favilo:
This waits for child processes to exit before exiting itself. This is necessary, because `docker -d` will just terminate the container once the main process ends. 

Also adding ability to log json files.


pyproject.toml
```diff
@@ -84,6 +84,8 @@ dependencies = [
     "zstandard==0.21.0",
     # License: Python Software Foundation License
     "typing-extensions==4.12.2",
+    # License:  BSD-2-Clause license 
+    "python-json-logger==2.0.7",
```
favilo, line 88: This isn't used directly, but it allows us to specify JSON logs in our `logging.json` configuration. Useful if you are using docker, and can't add dependencies in your virtual environment.

See change in [configuration.rst] for example of how to do this.

esrally/utils/process.py
```diff
@@ -229,3 +229,35 @@ def rally_process(p: psutil.Process) -> bool:
...
+) -> bool:
+    """
+    Waits for all child processes to terminate.
+
+    :param timeout: The maximum time to wait for child processes to terminate (default: None).
+    :param callback: A callback to call as each child process terminates.
+        The callback will be passed the PID and the return code of the child process.
+    :param list_callback: A callback to tell caller about the child processes that are being waited for.
+
+    :return: True if no child processes found, False otherwise.
```
dpifke-elastic, line 254: This seems backwards to the actual behavior.

esrally/utils/process.py
```diff
@@ -229,3 +229,35 @@ def rally_process(p: psutil.Process) -> bool:
         )
 
     kill_all(rally_process)
+
+
+def current_pid() -> int:
```
dpifke-elastic, line 234: Why do we need to wrap os.getpid?

dpifke-elastic (approved):
LGTM, just one (maybe dumb) question and a doc update.

esrally/utils/process.py
favilo: Heh, this was left over from when I was using `psutil` for the purpose, but realized this was a much faster call.

PR #1890, opened 2024-11-13
ci: Fix lint errors due to mypy with python 3.12

favilo:
Now that we're using python 3.12 for the `lint` github action, we need to fix the errors that it is reporting.

[Lint errors in question](https://github.com/elastic/rally/actions/runs/11823806034/job/32943896624)

* Have you signed the [contributor license agreement](https://www.elastic.co/contributor-agreement)?
* Have you followed the [contributor guidelines](https://github.com/elastic/rally/blob/master/CONTRIBUTING.md)?
* Have you run `make check-all` successfully?
* Did you choose a [descriptive title and description](https://chris.beams.io/posts/git-commit/) for your PR?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
     # License: MIT
     "urllib3==1.26.19",
     # License: Apache 2.0
-    "aiohttp==3.11.0",
+    "aiohttp==3.10.8",
     "docker==6.0.0",
     # avoid specific requests version to fix bug in docker-py
     "requests<2.32.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,13 +41,12 @@ classifiers = [
 ################################################################################################
 dependencies = [
     # License: Apache 2.0
-    # transitive dependencies:
-    #   urllib3: MIT
-    #   aiohttp: Apache 2.0
-
     "elasticsearch[async]==8.6.1",
     "elastic-transport==8.4.1",
+    # License: MIT
     "urllib3==1.26.19",
+    # License: Apache 2.0
+    "aiohttp==3.11.0",
     "docker==6.0.0",
     # avoid specific requests version to fix bug in docker-py
     "requests<2.32.0",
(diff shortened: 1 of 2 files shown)


gbanasiak:
This is caused by `aiohttp` update, which is a transitive/unpinned dependency in Rally.

After local venv recreation:
```
-aiohttp==3.10.8
+aiohttp==3.11.0
```

It's either/or it seems. Either we pin `aiohttp` and previous version is OK, or we let `aiohttp` move and then need to remove mypy ignore. WDYT @favilo ?

favilo:
How about we pin to the current version. Then we can remove the ignore and have an updated dependency, without being forced to upgrade from an unpinned one.

pyproject.toml
```diff
@@ -41,13 +41,12 @@ classifiers = [
 ################################################################################################
 dependencies = [
     # License: Apache 2.0
-    # transitive dependencies:
-    #   urllib3: MIT
-    #   aiohttp: Apache 2.0
-
     "elasticsearch[async]==8.6.1",
     "elastic-transport==8.4.1",
+    # License: MIT
```
favilo, line 46: I've also moved the license comment to here.

favilo:
Ah, so currently Python 3.8 isn't supported. So I'll pin it to 3.10.8. But I'm going to take this opportunity to just remove python 3.8 support in the near future. I've already got it working with 3.12 just fine.

That will allow us to do a lot more of the improvements we were interested in doing, too.

gbanasiak (approved):

pyproject.toml
```diff
@@ -41,13 +41,12 @@ classifiers = [
...
     "urllib3==1.26.19",
+    # License: Apache 2.0
+    "aiohttp==3.10.8",
```
gbanasiak, line 49: Nit: The latest 3.10.x is 3.10.11, actually.

gbanasiak: I think we were inspired by the state of local env too much, so https://github.com/elastic/rally/pull/1890/commits/eddc70cf3c9ea6f50d1d61243d2b81b8f3961548.

PR #1891, opened 2024-11-13
fix: Use named parameter for `_to_datetime` to pass default_tz

favilo:
I noticed in the PST timezone, `tests/track/loader_test.py::TestTemplateRender::test_render_simple_template` was failing because the default timezone wasn't getting set correctly.


inqueue (approved):
Nice find!

favilo:
Sigh, can't merge until #1890 can fix the lints

PR #1897, opened 2024-12-13
Update architecture diagrams

gbanasiak:
Differentiates between local class instance creation (`__init__()`) within the same process and Thespian actor creation which creates a new process (`startActor()`). Also, parentheses added to methods names for easier differentiation between Thespian messages and method calls.


favilo (approved):
Looks pretty good

PR #1898, opened 2024-12-13
feat: Add schedule for dev-docker builds daily. [ES-9318]

favilo:
Also fix the configuration to skip the input when the build is triggered by a schedule or webhook.

### DIFF
--- a/.buildkite/dev-docker/pipeline.yml
+++ b/.buildkite/dev-docker/pipeline.yml
@@ -2,7 +2,7 @@ steps:
   - input: "Build parameters"
     # If the build is triggered by merging a PR to master, we want to build from the master branch.
     # If the build is triggered by a schedule, we want to build from the current branch
-    if: (build.source != "webhook" && build.branch == "master") || build.source != "schedule" 
+    if: '!(build.source == "webhook" && build.branch == "master") || build.source != "schedule"'
     fields:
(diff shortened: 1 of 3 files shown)


gbanasiak:
@elasticmachine update

gbanasiak:
@elasticmachine update branch

favilo:
Thanks @gbanasiak, I've actually fixed the condition to be correct. I've used this elsewhere and it works this way correctly. And I had a weird merge, so I've just rebased ontop of the most recent `master` branch, and force-pushed

.buildkite/dev-docker/pipeline.yml
```diff
@@ -1,6 +1,8 @@
 steps:
   - input: "Build parameters"
-    if: build.source != "schedule"
+    # If the build is triggered by merging a PR to master, we want to build from the master branch.
+    # If the build is triggered by a schedule, we want to build from the current branch
+    if: '!((build.source == "webhook" && build.branch == "master") || build.source == "schedule")'
```
gbanasiak, line 5: Nit: The second comment is misleading to me as the current branch in case of scheduled build is... master. To me the point is we don't want to ask for inputs (and thus stop the pipeline) in case of scheduled builds and builds triggered by a PR merge to master.

gbanasiak (approved):
LGTM

PR #1904, opened 2025-01-21
Reduce log volume forwarded through Thespian framework

gbanasiak:
This reduces the amount of Thespian messages containing forwarded logs contributing to https://github.com/elastic/rally/issues/1448.

I don't understand `actorSystemCapabilityCheck()` behavior, as `DEBUG` logs should not be forwarded. I've raised this with Thespian maintainer, but have not received additional explanations so far. As we cannot get lower than `DEBUG` I'm simply removing those logs.

The `WARNING` log about SSL verification being off being forwarded through Thespian is expected.

See https://github.com/thespianpy/Thespian/issues/82#issuecomment-2551853924 for additional context, and potential follow-up in the future.


gareth-ellis (approved):
LGTM

PR #1908, opened 2025-01-29
Improve error messages regarding track syntax errors

gareth-ellis:
I've encountered a few cryptic error messages whilst tweeking a track.

Issues this solves:
- If we fail to transform the templates completely due to a syntax error, we write that the full output is listed XXXX - it isnt, we never get round to writing it, so I am catching the TemplateSyntaxError and writing another error message. It would be nice if we can say _what_ the error is, or at least where, but we can't yet. If I find a way I will add that to the PR / a new PR
- If you don't include "operation", you get an Attribute error, but no information where in the track. Now we include a more helpful message that should help users see where the issue may be.

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -1122,8 +1122,8 @@ class TrackFileReader:
                 track_spec = json.loads(rendered)
             except jinja2.exceptions.TemplateSyntaxError as te:
                 self.logger.exception("Could not load [%s] due to Jinja Syntax Exception.", track_spec_file)
-                msg = f"Could not load '{track_spec_file}' due to Jinja Syntax Exception."
-                msg += f"The track file ({tmp.name}) likely hasn't been writen"
+                msg = f"Could not load '{track_spec_file}' due to Jinja Syntax Exception. "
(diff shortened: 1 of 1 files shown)


esrally/track/loader.py
```diff
@@ -1120,6 +1120,11 @@ def read(self, track_name, track_spec_file, mapping_dir):
                     f.write(rendered)
                 self.logger.info("Final rendered track for '%s' has been written to '%s'.", track_spec_file, tmp.name)
                 track_spec = json.loads(rendered)
+            except jinja2.exceptions.TemplateSyntaxError as te:
+                self.logger.exception("Could not load [%s] due to Jinja Syntax Exception.", track_spec_file)
+                msg = f"Could not load '{track_spec_file}' due to Jinja Syntax Exception."
```
dpifke-elastic, line 1125: Nit: no space between this sentence and the next.  (And do we want a period at the end of the second sentence?)

dpifke-elastic (approved):
LGTM.

esrally/track/loader.py
gareth-ellis: resolved in dbd778e81716790b280011c6cf0721f324822a6b

PR #1909, opened 2025-01-31
Add check if headers is not None before treating it like a dict

gareth-ellis:
Martijn came across an issue with ccr-stats, where we are updating headers, but we don't have any request headers set - so this bit of code blows up. In other cases (e.g blob-store-stats) we typically run against serverless, which we specifically don't try and update the headers for.

### DIFF
--- a/esrally/client/common.py
+++ b/esrally/client/common.py
@@ -20,7 +20,7 @@ _COMPAT_MIMETYPE_SUB = _COMPAT_MIMETYPE_TEMPLATE % (r"\g<1>",)
 def _mimetype_header_to_compat(header, request_headers):
     # Converts all parts of a Accept/Content-Type headers
     # from application/X -> application/vnd.elasticsearch+X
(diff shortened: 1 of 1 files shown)


martijnvg (approved):
LGTM - I can confirm that this fix addresses error with ccr-stats telemetry device.

esrally/client/common.py
```diff
@@ -20,7 +20,7 @@ def _client_major_version_to_str(version: tuple) -> str:
 def _mimetype_header_to_compat(header, request_headers):
     # Converts all parts of a Accept/Content-Type headers
     # from application/X -> application/vnd.elasticsearch+X
-    mimetype = request_headers.get(header, None)
+    mimetype = request_headers.get(header, None) if request_headers else None
```
fressi-elastic, line 23: LGTM. By the way I think the following would be more clear to read:

```Python
if request_headers:
  mimetype  = request_headers.get(header, None)
  if mimetype:
     ...
```

PR #1911, opened 2025-02-03
Replace deprecate use of resources.read_text() function.

fressi-elastic:
This fixes the issue by replacing the deprecated function with the code suggested by Python documentation.


favilo (approved):
Looks good to me

fressi-elastic:
I just added the unit test.

PR #1913, opened 2025-02-07
Add ability to provide a list as input for disk-usage-stats

gareth-ellis:
We currently expect a comma separated list of indices, however its possible to provide a list, which then ends up crashing rally at the end of a race. Since the parameter is in plural, it makes sense to at least support providing a list. This checks that if the parameter is a list, and if so convert to comma separated


gbanasiak (commented):
Sorry, haven't seen this earlier.

There's another comma-separated string in telemetry: `node-stats-include-indices-metrics` in `node-stats`. It has a different behavior, exception is thrown when parameter is not a string. This happens on start, so that's less frustrating for a user. WDYT about aligning both to either support both formats (string and list) or throw a similar exception?

gbanasiak (approved):
LGTM

PR #1653, opened 2023-01-16
Upgrade thespian to 4.0.0 to fix hang on macOS

pquentin:
Closes #1575

See https://github.com/thespianpy/Thespian/issues/70 for more context.

PR #1916, opened 2025-02-21
Revert "Upgrade thespian to 4.0.0 to fix hang on macOS (#1653)"

gbanasiak:
This reverts commit fa5fbe0bb9993b60e33eaa36c2e73da8a935cdab.

Thespian 4.0.0 is causing issues on systems with multiple interfaces / IP addresses. During convention registration, actors might be presenting unexpected IP address which leads to communication failures, if one IP path is filtered.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,7 +62,7 @@ dependencies = [
     "Jinja2==3.1.5",
     "markupsafe==2.0.1",
     # License: MIT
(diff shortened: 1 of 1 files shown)


gareth-ellis (approved):

gbanasiak:
Reason for the revert: https://github.com/thespianpy/Thespian/issues/86

Issue #1918, opened 2025-03-03
Flaky internal `mypy` error appearing randomly when executing make lint.


fressi-elastic:
It looks like that executing multiple process of mypy in parallel is hitting some kind concurrency issue when executing (for example typing `make ling`). The issue can be avoided by asking `pre-commit` to process the files in sequence (alias avoid tool parallel execution for mypy hook).

PR #1919, opened 2025-03-03
Suppress a flaky internal mypy error appearing randomly when executing `make lint`.

fressi-elastic:
Fixes Rally issue #1918.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,8 @@ repos:
           "--config",
           "pyproject.toml"
         ]
+        # Suppress a flaky internal mypy error appearing randomly when executing make lint.
+        require_serial: true
 
   - repo: local
     hooks:


gbanasiak:
@fressi-elastic I've removed `\@sakurai-youhei`, no need to engage him. Better to use @elastic/es-perf unless someone has specific context. The scope of the change is bigger than the title, is that expected?

fressi-elastic:
> @fressi-elastic I've removed `\@sakurai-youhei`, no need to engage him. Better to use @elastic/es-perf unless someone has specific context. The scope of the change is bigger than the title, is that expected?

No the scope is the one in the title. I include other changes by mistake. I am sorry for it. I am going to remove unrelated changes from this one.

gbanasiak (approved):
I assume this makes `mypy` check slower under `pre-commit`, but correctness is more important than speed. We can revisit if this becomes too slow for us for some reason.

favilo (approved):

fressi-elastic:
The speed difference are irrelevants with my M4 14 cores MacBook. I also guess that the reason I see so many errors is because I was working with so many fast cores. With less and slower cores probably race conditions are harder to spot. Anyway considering the correctness and reliability of tests is much more important than speed I agree that we should wait the bug is fixed on `mypy` side before restoring this configuration as it was before.

Issue #1923, opened 2025-03-03
Create a tool to help pretty dumping JSON like objects to logs.


fressi-elastic:
It would be nice to have a tool helping visualizing changes in index templates.

The tool is intended to be used for printing JSON like objects to log files.

It should also integrate a functionality to compare two object dumps and highlight differences.

PR #1924, opened 2025-03-03
Introduce tools for pretty printing JSON like objects and their differences.

fressi-elastic:
This closes Rally issue #1923


gbanasiak (approved):
LGTM

PR #1929, opened 2025-03-19
Upgrade google-auth to 2.38.0

dpifke-elastic:


pyproject.toml
```diff
@@ -76,7 +76,7 @@ dependencies = [
     #   google-crc32c: Apache 2.0
     "google-resumable-media[requests]==1.1.0",
     # License: Apache 2.0
-    "google-auth==1.22.1",
+    "google-auth==2.17.3",
```
gareth-ellis, line 79: Any reason we dont go straight to 2.38.0 (released 2025) - 2.17.3 is already almost 2 years old, and from what I can see there arent any breaking changes since 2.0.0

dpifke-elastic: I matched the version to what we're using internally (1.22 was blocking Rally and esbench from being imported together).  I'll bump the version here to 2.38.0 and we can upgrade esbench as well.

gareth-ellis (approved):

PR #1930, opened 2025-03-19
Update offline-install.sh

gareth-ellis:
We don't upload sha256 of the artifacts - this adds the generation. Internal release documents to be updated shortly


dpifke-elastic (approved):
LGTM.

Issue #1931, opened 2025-03-21
Bug - error when creating track on Serverless existing Serach project


aavidan:
I have a Serverless ES project with existing index.
When running the `create-track` command:
```
./rally --skip-update \
create-track --track=uiam-first-track --target-hosts="https://atara-elasticsearch-project-fe4ad5.es.us-east-1.aws.dev.elastic.cloud" \
--client-options="use_ssl:true,verify_certs:false,basic_auth_user:'admin',basic_auth_password:'removed'" \
--indices="atara-uiam-index" --output-path=~/tracks
```

It fails with 
```
esrally/tracker/index.py", line 77, in extract_index_mapping_and_settings
    index_settings = filter_ephemeral_index_settings(details["settings"]["index"])
KeyError: 'index'
```
The reason is that the index in Serverless was created with an empty `settings`.

Fixing the line to be the following
```
            index_settings = filter_ephemeral_index_settings(details["settings"].get("index", {}))
```

seems to solve the issue.

This is the first time I work with this project so I would like to get feedback if you think I should create a PR.
Thank you.

gareth-ellis:
Hello,

Yes that would work - i guess its an area that hasnt been tested as such, since especially in stateful most templates would contain index settings, whilst in serverless some of those settings arent available.

I'd suggest 
`            index_settings = filter_ephemeral_index_settings(details.get("settings", {}).get("index", {}))
`
that way if there's ever a case where settings is also missing, we dont fall over.

Please feel free to open a PR

PR #1932, opened 2025-03-24
issue #1931: `create-track` command supports serverless.

aavidan:
Support Serverless ES project with existing index when running the `create-track` command. Serverless might not have entries like settings and index.

 https://github.com/elastic/rally/issues/1931


gareth-ellis (approved):
LGTM - Thanks!

PR #1935, opened 2025-03-28
Upgrade gh-action-pypi-publish

dpifke-elastic:
Publishing esrally 2.12.0 to PyPI failed because our tooling wrote metadata version 2.4, and it appears the prior version of this action only understands up to version 2.3.


gareth-ellis (approved):
LGTM

PR #1937, opened 2025-04-08
Fix snapshot restore

gbanasiak:
This fixes the following problems in snapshot restore runner:
* `wait-for-completion` is ignored and not passed as request parameter
* client-side timeout is not a request parameter, but transport-level setting of Python client


luigidellaquila (approved):
I don't have the full picture, but for what I can see it LGTM

inqueue (approved):
LGTM

PR #1942, opened 2025-06-06
Add tool to handle file parts.

fressi-elastic:
This is the first module of a new package intended for accelerating tracks files by using multipart downloading.

Add a module used to handle set operations between portion of files.

The idea is having a base data struct called Range representing a continuous portion of a file (with a start and end position).
This struct is stored in groups or non continuous ranges so that we can represents any non continuous portion of a file. In this way we can identify portions of the file so that it is easier to distribute file transfers to multiple worker threads. It also facilitate a way to represent a transient status of a partially downloaded file so we can later resume the transfer from the last know status.

The abstract base class RangeSet has 3 implementations:
  - `class EmptyRange` for representing an empty portion of a file.
  - `class Range` for representing a continuous portion of a file.
  - `class RangeThree` for representing N disjoin portion of a file (implemented as a binary tree).

### DIFF
--- a/esrally/storage/_range.py
+++ b/esrally/storage/_range.py
@@ -1,0 +1,415 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+from __future__ import annotations
+
+import sys
+from abc import abstractmethod
+from collections.abc import Hashable, Iterable, Iterator, Sequence, Set
+from itertools import chain, islice
+from typing import Any, overload
+
+MAX_LENGTH = sys.maxsize
+
+
+class RangeSet(Sequence["Range"], Set["Range"], Hashable):
+    """RangeSet is the abstract base class of all implementations of sets of ranges.
+
+    A range set is an immutable sequence of disjoint ranges sorted by its start value. It implements some mixin methods
+    so that the implementation of a range sets is going to be lighter. It implements either the behaviour of a set
+    and of a sequence.
+    """
+
+    @property
+    @abstractmethod
+    def start(self) -> int:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def end(self) -> int:
+        raise NotImplementedError
+
+    @property
+    @property
+    @abstractmethod
+    def size(self) -> int:
+        raise NotImplementedError
+
+    @abstractmethod
+    def split(self, max_size: int = MAX_LENGTH) -> tuple[Range | EmptyRange, RangeSet]:
+        raise NotImplementedError
+
+    def __str__(self) -> str:
+        return ",".join(str(r) for r in self)
+
+    def __repr__(self) -> str:
+        return f"rangeset('{str(self)}')"
+
+    @abstractmethod
+    def __or__(self, other: Iterable[Range]) -> RangeSet:  # type: ignore
+        raise NotImplementedError
+        # if not isinstance(other, Iterable):
+        #     raise TypeError(f"{other} is not iterable")
+        # return _union(self, other)
+
+    @abstractmethod
+    def __and__(self, other: Iterable[Range]) -> RangeSet:
+        raise NotImplementedError
+
+    @abstractmethod
+    def __sub__(self, other: Iterable[Range]) -> RangeSet:
+        raise NotImplementedError
+
+    @overload
+    def __getitem__(self, key: int) -> Range:
+        pass
+
+    @overload
+    def __getitem__(self, key: slice) -> RangeSet:
+        pass
+
+    def __getitem__(self, key: int | slice) -> RangeSet:
+        if isinstance(key, int):
+            if key < 0:
+                raise IndexError(f"index key can't be negative: {key} < 0")
+            try:
+                return next(islice(self, key, key + 1))
+            except StopIteration:
+                raise IndexError(f"index key is too big: {key} >= {len(self)}") from None
+        if isinstance(key, slice):
+            if key.step not in [None, 1]:
+                raise ValueError(f"invalid slice step value: {key.step} is not 1 | None")
+            start = 0 if key.start is None else 0
+            end = 0 if key.stop is None else key.stop
+            step = 1 if key.step is None else key.step
+            return _rangeset(islice(self, start, end, step))
+        elif isinstance(key, RangeSet):
+            return self & key
+        raise TypeError(f"invalid key type: {key} is not int | slice | RangeSet")
+
+    @abstractmethod
+    def __eq__(self, other: Any) -> bool:
+        raise NotImplementedError
+
+    @abstractmethod
+    def __hash__(self) -> int:
+        raise NotImplementedError
+
+
+class EmptyRange(RangeSet):
+
+    def __contains__(self, item: Any) -> bool:
+        return False
+
+    def __len__(self) -> int:
+        return 0
+
+    def __iter__(self) -> Iterator[Range]:
+        return iter(tuple())
+
+    def __bool__(self) -> bool:
+        return False
+
+    def __or__(self, other: Iterable[Range]) -> RangeSet:  # type: ignore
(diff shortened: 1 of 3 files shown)


esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+    @abstractmethod
+    def start(self) -> int:
+        raise NotImplementedError
+
+    @property
+    @abstractmethod
+    def end(self) -> int:
+        raise NotImplementedError
+
+    @property
```
favilo, line 47: `@property` is listed twice

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+        """It returns subtract the other range sets from this one."""
+        return _rangeset(self._remove(others))
+
+    @abstractmethod
+    def _remove(self, others: Iterable[Range]) -> Iterable[Range]:
+        raise NotImplementedError
+
+    @overload
+    def __getitem__(self, i: int) -> Range:
+        """It returns the renge at the ith position."""
```
favilo: Nit: spelling

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+    def __sub__(self, others: Iterable[Range]) -> RangeSet:
+        """It returns subtract the other range sets from this one."""
+        return _rangeset(self._remove(others))
+
+    @abstractmethod
+    def _remove(self, others: Iterable[Range]) -> Iterable[Range]:
+        raise NotImplementedError
+
+    @overload
+    def __getitem__(self, i: int) -> Range:
```
favilo, line 91: Should this be `RangeSet` or `Range`? I suppose `Range` is a subclass, but what does `mypy` think of this?

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+        """It returns the renge at the ith position."""
+
+    @overload
+    def __getitem__(self, i: slice) -> RangeSet:
+        """It returns a set of ranges selected using a slice."""
+
+    def __getitem__(self, i: int | slice) -> RangeSet:
```
favilo, line 98: Hmm, `slice` is a fairly unusual type to use here. I like it, but I think it needs some usage examples. 

```python
range = Range(...)
assert range[0:-1] is not None
```

Or some such

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+        return 0
+
+    def __eq__(self, other: Any) -> bool:
+        return isinstance(other, EmptyRange)
+
+    def __hash__(self) -> int:
+        return 0
+
+
+NO_RANGE = EmptyRange()
```
favilo: Maybe annotate with `Final[EmptyRange]`?

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+
+    @property
+    def end(self) -> int:
+        return self._end
+
+    @property
+    def size(self):
+        return self._end - self._start
+
+    def _intersect(self, others: Iterable[Range]) -> Iterator[Range]:
```
favilo, line 192: What if we return a `RangeTree` here instead of an Iterator? I guess either works, I just was wondering what use a `RangeTree` actually would be otherwise.

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+            return self.start == other.start and self.end == other.end
+        return False
+
+    def __hash__(self) -> int:
+        return hash((self.start, self.end))
+
+
+class RangeTree(RangeSet):
+
+    def __init__(self, left: RangeSet, right: RangeSet):
```
favilo, line 262: Do we want to enforce some sort of ordering invariant? Make it a binary tree or something, and ensure that they don't overlap? Otherwise, it should just be a `Range`, though `rangeset()`  already does that. 

I just want to ensure either people can't use these constructors directly the wrong way, or we enforce our invariants, if we are okay with that use case.

favilo (commented):
Left some comments. I think it looks pretty good, and I like the test cases, I just think we should enforce some invariants.

esrally/storage/_range.py
fressi-elastic: Because RangeSet is a sorted sequence of items of type Range, this should always return one single Range or raise an exception. Range is a sequence of one single Range, returning itself on this method. I know this could look confusing.

esrally/storage/_range.py
fressi-elastic: I had to implement slices because it is being expected when using RangeSet with unpacking expressions like the following:

```python
first, *others = self
```

Above code is equivalent to:
```
first, others = self[0], self[1:]
```

To conform with the expected interface in case of an `int` key, then the result has to be a single Range, while in case of a slice it has to be a sequence of Range (I.E. a RangeSet).

esrally/storage/_range.py
fressi-elastic: Implementing algorithms would be simpler and lighter using a generator, even if later debugging and reading the code would be harder. In case you think this adds some complexity in exchange for some minor unproved advantage in efficiency then I would remove this method and implement the __and__ operators in-place. What do you prefer?

esrally/storage/_range.py
fressi-elastic: Please tell me the change I am going to commit after this looks like what do you have in mind.

favilo:
> > invariants
> 
> @favilo Please be more specific: what would you improve? what do you mean with "invariants"?
> 
> Thank you

Specifically, down in what I assume you want to be the main entry point to this module, you have trees getting constructed in a balanced binary tree layout. But since the classes are not protected in any way, the constructors can be used counter to that. The invariant in this case would be that the trees passed in are in sorted order, non-overlapping, and ideally balanced. 

I'm not sure how important the balanced part of that is, but the other code you have depends on the ordered and non-overlapping nature to be correct. Like the start and end properties of the tree.

esrally/storage/_range.py
favilo: Eh, it wasn't a strong opinion of mine, I'm fine with it as it is. I just wanted to know your reasoning for the iterator

esrally/storage/_range.py
favilo: After reading that, I still think we should add some enforcement. 

I am of the opinion that bad states should be impossible to get into. Either enforced by the type system and not exporting the base classes, or by asserting that some condition is true at runtime.

So the options here, would be only having `rangeset()`and maybe `RangeSet` in the module's `__all__`, or by ensuring that we are keeping our invariant true.

fressi-elastic: I agree. I will forbid bad state to be possible.

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+        assert isinstance(left, (Range, RangeTree)), "left must be an instance of Range | RangeTree"
+        assert isinstance(right, (Range, RangeTree)), "right must be an instance of Range | RangeTree"
+        assert left.start < left.end < right.start < right.end, "left and right ranges sets aren't sorted and disjoint"
```
favilo, line 265: Oh man, I like this. I was expecting a more complex condition, but this is just a one line, simple check. 

Thanks for adding this

favilo (approved):
Awesome, thanks for this. It'll be a great improvement

gareth-ellis (approved):

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+    try:
+        left = next(it)
+    except StopIteration:
+        return
+
+    for right in it:
+        if right.start <= left.end and left.start <= right.end:
+            # It merges touching ranges to a single one.
+            left = Range(min(left.start, right.start), max(right.end, left.end))
+            continue
```
gbanasiak, line 361: Isn't `left.start <= right.end` always met due to earlier sort? Also couldn't `min(left.start, right.start)` be simplified to `left.start` ?

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+        return self._end - self._start
+
+    def _intersect(self, others: Iterable[Range]) -> Iterator[Range]:
+        # pylint: disable=protected-access
+        for r in _combine(others):
+            if r.end <= self.start:
+                continue
+            if r.start >= self.end:
+                break
+            yield Range(max(r._start, self._start), min(r._end, self._end))
```
gbanasiak, line 199: Nit: Why mixed use of public properties and private attributes, e.g. `r.start` vs. `r._start`?

esrally/storage/_range.py
```diff
@@ -0,0 +1,430 @@
...
+    def __init__(self, msg: str, ranges: RangeSet = NO_RANGE):
+        super().__init__(msg)
+        self.range = ranges
+
+
+class EmptyRangeError(RangeError):
+    pass
+
+
+def _combine(ranges: Iterable[Range]) -> Iterable[Range]:
```
gbanasiak, line 340: Shouldn't generator functions returned type be rather annotated with `Iterator`, instead of `Iterable`? They return an iterator object, and we should return the most specific type possible?

gbanasiak (commented):
Leaving some questions for learning purposes.

esrally/storage/_range.py
fressi-elastic: Please have a look at this comment: https://github.com/elastic/rally/pull/1945#discussion_r2149557426

PR #1943, opened 2025-06-06
Add some threading synchronization tools.

fressi-elastic:
This tools are intended to be used by a new package intended for accelerating tracks files downloads by using multipart downloading.

It provides following threading synchronization tools:

- `class ContinuousTimer` implements a thread that every `interval` seconds runs a function.
- `class TimedEvent` implements similar behavior of `threading.Event` but with the additional feature of recording the time the event is set.
- `class WaitGroup` implements on top of `TimedEvent` with the addition of a mutable counter that sets the event and notify waiter threads when the counter returns to 0. It is intended to be used in a similar way as the GoLang WaitGroup type with the purpose of tracking the count of some limited resources (like workers threads, number of connections) and waiting for a group of worker threads to finish their job.

### DIFF
--- a/esrally/utils/threads.py
+++ b/esrally/utils/threads.py
@@ -158,11 +158,12 @@ class WaitGroup(TimedEvent):
     """
 
     MAX_COUNT = sys.maxsize
+    _count = 0
 
     def __init__(self, count: int = 0, max_count: int = MAX_COUNT):
         super().__init__()
-        self._count = count
-        self._max_count = max(1, max_count)
+        self.max_count = max_count
+        self.add(count)
 
     def clear(self):
         self._count = 0
@@ -178,9 +179,15 @@ class WaitGroup(TimedEvent):
 
     @max_count.setter
     def max_count(self, value: int) -> None:
-        self._max_count = max(1, value)
+        if value < 1:
+            raise ValueError("max_count must be greater than or equal to 1")
+        self._max_count = value
 
     def add(self, value: int) -> bool:
+        """It increments `count` of `value` times. `value` can be negative to reduce the `count`.
+        :raises: ValueError if `count` + `value` is negative.
+        :raises: WaitGroupLimitError in case the sum `count` + `value` is greater than `max_count`.
+        """
         with self._cond:
             new_value = self._count + value
             if new_value < 0:
@@ -191,4 +198,5 @@ class WaitGroup(TimedEvent):
         return new_value == 0 and self.set()
 
     def done(self) -> bool:
+        """It subtracts 1 from `count`."""
         return self.add(-1)
--- a/esrally/utils/threads.py
+++ b/esrally/utils/threads.py
@@ -135,6 +135,9 @@ class TimedEvent:
 class WaitGroupLimitError(Exception):
     """Raised when a wait group reach max workers limit."""
 
+    def __init__(self, msg: str, max_count: int):
+        self.max_count = max_count
+
 
 class WaitGroup(TimedEvent):
     """It implements a go-lang style wait group on top of a timed event.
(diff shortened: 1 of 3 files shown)


esrally/utils/threads.py
```diff
@@ -0,0 +1,189 @@
...
+from __future__ import annotations
+
+import sys
+import threading
+import time
+from collections.abc import Callable
+
+
+class ContinuousTimer(threading.Thread):
+    """It calls a function every a specified number of seconds:
```
favilo: Nit: every specified number of seconds

esrally/utils/threads.py
```diff
@@ -0,0 +1,189 @@
...
+    def max_count(self) -> int:
+        return self._max_count
+
+    @max_count.setter
+    def max_count(self, value: int) -> None:
+        self._max_count = max(1, value)
+
+    def add(self, value: int) -> bool:
+        with self._cond:
+            new_value = max(0, self._count + value)
```
favilo: It might be better to raise an exception when we drop below 0. We don't want more code to call `done()` than we've allocated capacity for, do we?

esrally/utils/threads.py
```diff
@@ -0,0 +1,189 @@
...
+class TimedEvent:
+    """It re-implement threading.Event objects with a couple of additional functionalities.
+
+    TimedEvent manages a flag that can be set to true with the set() method and reset
+    to false with the clear() method. The wait() method blocks until the flag is
+    true. The flag is initially false.
+
+    On top of the threading.Event functionalities:
+        - the set() method returns True in case the flag was False at the time it has been called;
+        - its time() method returns the value of monotonic.time() at the time set has been called first.
```
favilo: Why `monotonic()` and not `monotonic_ns()`? NIT: Also it's `time.monotonic()`

tests/utils/threads_test.py
```diff
@@ -0,0 +1,164 @@
...
+    timer.cancel()
+
+    assert timer.wait(timeout=5.0), "finished not set"
+
+    assert duration >= interval * count, f"finished early: {duration}"
+    assert interval <= times[0], f"started early: {times[0]}"
+    assert duration >= times[-1], f"finished early: {times[-1]}"
+
+
+def test_timed_event():
```
favilo: Couldn't this be done with `pytest.mark.parameterize()`?

favilo (commented):
I've left some questions

tests/utils/threads_test.py
fressi-elastic: Sure!

favilo (approved):
Looks good. Thanks

fressi-elastic:
Tomorrow evening I will merge this, if there any more comments.

esrally/utils/threads.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def __init__(self, msg: str, max_count: int):
+        super().__init__(msg)
+        self.max_count = max_count
+
+
+class WaitGroup(TimedEvent):
+    """It implements a go-lang style wait group on top of a timed event.
+
+    After N-times the done method is called, the event will be set. This can be used for waiting for waiting for
```
gbanasiak, line 146: Nit: `waiting for` twice

tests/utils/threads_test.py
```diff
@@ -0,0 +1,161 @@
...
+    assert wg.max_count == max_count
+    assert not wg
+    assert not notified.is_set()
+
+    try:
+        wg.add(1)
+    except WaitGroupLimitError:
+        pass
+    else:
+        pytest.fail("didn't raised exception")
```
gbanasiak, line 140: Nit: `didn't raise`

gbanasiak (commented):
Coming late to the party, found 2 small nits.

esrally/utils/threads.py
fressi-elastic: Resolved in PR #1946

tests/utils/threads_test.py
fressi-elastic: Resolved in PR #1946

PR #1945, opened 2025-06-09
Add HTTP adapter for the storage client.

fressi-elastic:
With the purpose of using multipart downloading to speed up tracks downloading, this introduces an HTTP client build on top of the `requests` library. The adapter is being registered for the "http" and "https" protocols. Other adapter types could be introduced later to use other libraries and protocols.

### Please note this PR is built on top of PR #1942, please review it first. ###


fressi-elastic:
@favillo @gareth-ellis Please have a look at this. Parent PR have been pulled from the master branch.

fressi-elastic:
This is ready for review again.

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,147 @@
...
+        self._classes: dict[str, type[Adapter]] = {}
+        self._adapters: dict[type[Adapter], Adapter] = {}
+        self._lock = threading.Lock()
+        self._cfg = cfg
+
+    @classmethod
+    def from_config(cls, cfg: Config) -> AdapterRegistry:
+        registry = cls(cfg)
+        adapters_specs = (
+            cfg.opts(section="storage", key="storage.adapters", default_value=ADAPTER_CLASS_NAMES, mandatory=False)
```
gareth-ellis, line 123: If we're expanding the config options, can we update the docs too?

docs/configuration.rst
```diff
@@ -116,6 +116,38 @@ Define a secure connection to an Elastic Cloud cluster::
     datastore.user = rally
     datastore.password = the-password-to-your-cluster
 
+storage
+~~~~~~~
+
+This section defines how client is configured to transfer big files (I.E. track files) from
+remote servers. Available options are:
+
+* ``storage.adapters`` is a command separated list of storage adapters implementations specified using the following
```
gbanasiak:
```suggestion
* ``storage.adapters`` is a comma-separated list of storage adapter implementations specified using the following
```

docs/configuration.rst
```diff
@@ -116,6 +116,38 @@ Define a secure connection to an Elastic Cloud cluster::
...
+  format:
+
+  ``<python module name>:<adapter class name>``
+
+  Here is an example of valid value for http(s) adapter::
```
gbanasiak, line 130: I think a section called `HTTPAdapter` should start here specifying which options are valid for this specific adapter implementation. Different adapters might have different options.

docs/configuration.rst
```diff
@@ -116,6 +116,38 @@ Define a secure connection to an Elastic Cloud cluster::
...
+
+* ``storage.http.chunk_size`` is used to specify the size of the buffer is being used for transferring chunk of files.
+
+* ``storage.http.max_retries`` is used to configure the maximum number of retries for making HTTP adapter requests.
+  it accept a numeric value to simply specify total number of retries. Examples::
+
+    [storage]
+    storage.http.max_retries = 3
+
+  For a more complex uses it accepts a dictionary of parameters (defined in yaml/json format) to be passed to the
```
gbanasiak, line 147: Is there a strong reason to introduce YAML parsing here? It's a first in Rally. Wouldn't JSON be enough, e.g. `json.loads()` ?

esrally/storage/_http.py
```diff
@@ -0,0 +1,157 @@
...
+        if max_retries:
+            adapter = requests.adapters.HTTPAdapter(max_retries=max_retries)
+            self.mount("http://", adapter)
+            self.mount("https://", adapter)
+
+
+class HTTPAdapter(Adapter):
+    """It implements the adapter interface for http(s) protocols using the requests library."""
+
+    __adapter_prefixes__ = ("http://", "https://")
```
gbanasiak: Nit: The `://` part will be repeated every time I think. Why not drop it? Also, it has a name in URL structure, it's called a "scheme". WDYT about renaming this to `__adapter_url_schemes__` or similar? Consequently, `prefixes` in method arguments could be replaced with `url_schemes`, or similar.

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,147 @@
...
+    def create(cls, url: str, content_length: int | None = None, accept_ranges: bool | None = None, ranges: RangeSet = NO_RANGE) -> Head:
+        if accept_ranges is None:
+            accept_ranges = bool(ranges)
+        return cls(url=url, accept_ranges=accept_ranges, content_length=content_length, ranges=ranges)
+
+
+class Adapter(ABC):
+    """Base class for storage class client implementation"""
+
+    __adapter_prefixes__: tuple[str, ...] = tuple()
```
gbanasiak: A comment explaining what a prefix is would be nice. It's not obvious on the first pass.

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def register_class(self, cls: type[Adapter], prefixes: Iterable[str]) -> type[Adapter]:
+        with self._lock:
+            keys_to_move: set[str] = set()
+            for p in prefixes:
+                self._classes[p] = cls
+                # Adapter types are sorted in descending order by prefix length.
+                keys_to_move.update(k for k in self._classes if len(k) < len(p))
+            for k in keys_to_move:
+                self._classes[k] = self._classes.pop(k)
```
gbanasiak: Why can't we simply use 1:1 mapping between URL scheme and the adapter?

Is it realistic we will ever use longest-prefix match? If the answer is yes, can we have a comment for the last line that it moves the key to the end of the dictionary for easier read? Also wouldn't hurt to have a unit test that shows the sorting and longest-prefix match in action.

esrally/storage/_http.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @classmethod
+    def from_config(cls, cfg: Config) -> Session:
+        max_retries: urllib3.Retry | int | None = 0
+        max_retries_text = cfg.opts("storage", "storage.http.max_retries", MAX_RETRIES, mandatory=False)
+        if max_retries_text:
+            try:
+                max_retries = int(max_retries_text)
+            except ValueError:
+                max_retries = urllib3.Retry(**yaml.safe_load(max_retries_text))
```
gbanasiak: As per earlier comment, could we use `json.loads()` here instead?

esrally/storage/_http.py
```diff
@@ -0,0 +1,157 @@
...
+    if len(ranges) > 1:
+        raise NotImplementedError(f"unsupported multi range requests: ranges are {ranges}")
+    headers["range"] = f"bytes={ranges[0]}"
+    return headers
+
+
+def content_length_from_headers(headers: CaseInsensitiveDict) -> int | None:
+    value = headers.get("content-length", "*").strip()
+    if value == "*":
+        return None
```
gbanasiak: Nit: Why `*` instead of more natural `None`?

esrally/storage/_http.py
```diff
@@ -0,0 +1,157 @@
...
+    try:
+        return int(value)
+    except ValueError:
+        raise ValueError(f"invalid content-length value: {value}") from None
+
+
+def accept_ranges_from_headers(headers: CaseInsensitiveDict) -> bool | None:
+    got = headers.get("accept-ranges", "").strip()
+    if not got:
+        return None
```
gbanasiak, line 129: Nit: Why `""` instead of default `None`?

esrally/storage/_range.py
```diff
@@ -404,12 +404,12 @@ def rangeset(ranges: str | Iterable[Range]) -> RangeSet:
...
-    if not isinstance(ranges, str):
-        return _rangeset(_combine(ranges))
-    return _rangeset(_combine(_parse(ranges)))
+    if isinstance(ranges, str):
+        ranges = list(_parse(ranges))
+    return _rangeset(_combine(ranges))
 
 
-def _parse(text: str) -> Iterator[Range]:
+def _parse(text: str) -> Iterable[Range]:
```
gbanasiak, line 412: I think `Iterator` is actually correct as this is a generator function, and generator functions return iterators. Also see https://github.com/elastic/rally/pull/1942#discussion_r2140466660.

pyproject.toml
```diff
@@ -85,6 +85,10 @@ dependencies = [
     "python-json-logger==2.0.7",
     # License: Apache 2.0
     "ecs-logging==2.2.0",
+    # License: MIT
+    "pyyaml==6.0.2",
+    # License: MIT
+    "types-PyYAML==6.0.2"
```
gbanasiak: If YAML parsing is needed, `types-PyYAML` should be moved to `[project.optional-dependencies]` to `develop` list.

gbanasiak (commented):
This looks good. I left minor comments.

esrally/storage/_http.py
fressi-elastic: This value is being used as a prefix, ideally allowing to match a larger part than just che scheme. I copied the same approach from `requests` library (https://requests.readthedocs.io/en/latest/user/advanced/#transport-adapters). Prefixes are matched started from the longest to the shortest. This allows for example to add special adapters for some specific URL, leaving the not-matching URLs to be caught by a more generic implementation (like the HTTP(S) one). Think about we want to write a special adapter for one specific end point, while being https, it needs some ad hoc logic on top of the regular HTTP.

esrally/storage/_adapter.py
fressi-elastic: DONE.

esrally/storage/_http.py
fressi-elastic: None don't implements `.strip` method. I will use "" instead. Because this value comes from a remote servers, I would like to be flexible to the type of values could reach our client (where it is not too complex to do).

esrally/storage/_http.py
fressi-elastic: Because "" implements `.strip()` method.

esrally/storage/_range.py
fressi-elastic: `Iterator` implements one method more than `Iterable`. Both are correct here. I don't have strong preferences, just I am used to accept `Iterable` as methods parameters for generic uses because it is implemented either by collections and generators, while `Iterator` it is only by generators. Therefore for simplicity and usability I return the bare minimum interface I actually use, to avoid relying on some more things in the code when later consuming it. Anyway if you prefer I will change it. I makes no difference to me in this specific case.

pyproject.toml
fressi-elastic: YAML removed here.

fressi-elastic:
@gbanasiak Can you give another look at this before I can merge it? I tried to address all your comments. I hope I sattisfied all your concerns.

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,155 @@
...
+    def create(cls, url: str, content_length: int | None = None, accept_ranges: bool | None = None, ranges: RangeSet = NO_RANGE) -> Head:
+        if accept_ranges is None:
+            accept_ranges = bool(ranges)
+        return cls(url=url, accept_ranges=accept_ranges, content_length=content_length, ranges=ranges)
+
+
+class Adapter(ABC):
+    """Base class for storage class client implementation"""
+
+    # Collection of URL prefixes used to be associate an adapter implementation to a remote file URL.
```
gbanasiak:
```suggestion
    # Collection of URL prefixes used to associate an adapter implementation to a remote file URL.
```

esrally/storage/_adapter.py
gbanasiak: As we're keeping the prefixes, can you please comment before line 139 to indicate the intention is to move `k` at the end of the dictionary? This looked like a no-op to me initially.

Also, can we have a unit test that confirms overlapping prefixes work as intended? Now unit test covers `http://` + `https://` scenario. We should also have something like `test` and `test123` scenario, where longest-prefix match actually matters.

esrally/storage/_range.py
gbanasiak: I only want to agree on a rule we could all follow. Me being the first person that doesn't follow it today :) We've been discussing this with @favilo on other occasions and I think the rule that makes sense is: accept the most _generic_ types possible as method arguments, and return the most _specific_ type possible. WDYT? In this particular case the change is from more specific to less specific type.

gbanasiak (approved):
LGTM. I suggested one more unit test.

esrally/storage/_http.py
gbanasiak: I think prefixes beyond URL scheme are only hypothetically useful in Rally but OK.

esrally/storage/_range.py
fressi-elastic: In the specific case of iterators I would always use `Iterable` unless you really need to use the `next()` method and only because we should not expect this method to be required at all in 99% of the cases. The only case you actually need to use the `next()` method you can anyway translate any `Iterable` in an `Iterator` by use the iter function. I actually never see using `Iterator` in any interface, at least the only real reason is because it sounds the same as Java and other languages. WDYT?

gareth-ellis (approved):

esrally/storage/_range.py
gbanasiak: That's fine. It looks like there's no simple rule then ;-)

esrally/storage/_adapter.py
fressi-elastic: I agree with you. This could be done better. I am rewriting the code so it is more obvious. Please have a look at it later as soon I have updated the code.

fressi-elastic:
@gbanasiak can you please give it one last check?

esrally/storage/_adapter.py
```diff
@@ -130,26 +135,20 @@ def from_config(cls, cfg: Config) -> AdapterRegistry:
...
             for p in prefixes:
-                self._classes[p] = cls
-                # Adapter types are sorted in descending order by prefix length.
-                keys_to_move.update(k for k in self._classes if len(k) < len(p))
-            for k in keys_to_move:
-                self._classes[k] = self._classes.pop(k)
+                self._classes.append(AdapterClassEntry(p, cls))
+            # The list of adapter classes is kept sorted from the longest prefix to the shorter to ensure that matching
+            # a shorter URL prefix will never hide matching a longer one.
+            self._classes.sort(key=lambda e: len(e.url_prefix), reverse=True)
```
gbanasiak, line 142: Yes, much easier to read, thank you.

PR #1949, opened 2025-06-13
Create storage client able to redirecting downloads to mirror servers

fressi-elastic:
This continue works on top of PR #1945


docs/configuration.rst
```diff
@@ -116,6 +116,42 @@ Define a secure connection to an Elastic Cloud cluster::
     datastore.user = rally
     datastore.password = the-password-to-your-cluster
 
+storage
+~~~~~~~
+
+This section defines how client is configured to transfer big files (I.E. track files) from
+remote servers. Available options are:
+
+* ``storage.adapters`` is a comma-separated list of storage adapter implementations specified using the following
```
favilo: Might want to have a section of `Available Adapters` here. I know it's only `HTTPAdapter` now, but when this does grow, it'd be nice to have.

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Collection of URL prefixes used to associate an adapter implementation to a remote file URL.
+    # This value will be overridden by `Adapter` subclasses to be consumed by `AdapterRegistry` class.
+    # Example:
+    #   ```
+    #   class HTTPAdapter(Adapter):
+    #       # The value will serve to associate any URL with "https" scheme to `HTTPAdapter` subclass.
+    #       __adapter_URL_prefixes__ = ("http://",)
+    #   ```
+    __adapter_URL_prefixes__: tuple[str, ...] = tuple()
```
favilo: Why tuple and not list? Just curious

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @classmethod
+    def from_config(cls, cfg: Config) -> Adapter:
```
favilo, line 77: I think I'd rather have this `raise NotImplementedError` since we shouldn't be using this? and make it `@abstractmethod`?

esrally/storage/_adapter.py
```diff
@@ -0,0 +1,154 @@
...
+            registry.register_class(obj, obj.__adapter_URL_prefixes__)
+        return registry
+
+    def register_class(self, cls: type[Adapter], prefixes: Iterable[str]) -> type[Adapter]:
+        with self._lock:
+            for p in prefixes:
+                self._classes.append(AdapterClassEntry(p, cls))
+            # The list of adapter classes is kept sorted from the longest prefix to the shorter to ensure that matching
+            # a shorter URL prefix will never hide matching a longer one.
+            self._classes.sort(key=lambda e: len(e.url_prefix), reverse=True)
```
favilo, line 150: This is probably fine performance-wise, but it feels like it might be a good place to use a `Trie` data structure. However, that would require adding a new dependency, so I'm not sure the work to do it is worth the minimal performance gain we'd see. Since this is very likely going to remain fairly small.

esrally/storage/_client.py
```diff
@@ -0,0 +1,220 @@
...
+    Writable,
+)
+from esrally.storage._mirror import MirrorList
+from esrally.storage._range import NO_RANGE, RangeSet
+from esrally.utils.threads import WaitGroup, WaitGroupLimitError
+
+LOG = logging.getLogger(__name__)
+
+MIRRORS_FILES = "~/.rally/storage/mirrors.yml"
+MAX_CONNECTIONS = 4
```
favilo: Nit: Call this `DEFAULT_MAX_CONNECTIONS` so it's clear that this can be overridden by the config

esrally/storage/_client.py
```diff
@@ -0,0 +1,220 @@
...
+            self._stats[_server_key(url)].append(ServerStats(url, start_time, end_time))
+            # The cached value could be an exception, or a head. In this way it will not retry previously failed
+            # urls until the TTL expires.
+            self._cached_heads[url] = value, start_time
+
+        return _head_or_raise(value)
+
+    def resolve(self, url: str, content_length: int | None = None, accept_ranges: bool = False, ttl: float = 60.0) -> Iterator[Head]:
+        """It looks up mirror list for given URL and yield mirror heads.
+        :param url:
```
favilo: Nit: Why not fill this out with the obvious?

esrally/storage/_client.py
```diff
@@ -0,0 +1,220 @@
...
+        if not mirror_urls:
+            yield self.head(url, ttl=ttl)
+            return
+
+        # It shuffles mirror URLs in the hope two threads will try different mirrors for the same URL.
+        self._random.shuffle(mirror_urls)
+
+        # It uses measured latencies and the number of connections to affect the order of the mirrors.
+        weights = {u: self._average_latency(u) * self._random.uniform(1.0, self._server_connections(url).count) for u in mirror_urls}
+        # LOG.debug("weights: %s", weights)
```
favilo: Hmm, these is a low enough log level, why not leave them in? Or just change them to `LOG.trace`?

esrally/storage/_client.py
```diff
@@ -0,0 +1,220 @@
...
+                    stats.popleft()
+        return stats
+
+    def _average_latency(self, url: str, ttl: float | None = 60.0) -> float:
+        stats = self._server_stats_from_url(url, ttl)
+        if not stats:
+            return 0.0
+        return sum(s.latency for s in stats) / len(stats)
+
+    def monitor(self):
```
favilo, line 204: Should there be a `with self._lock:` here? I'm worried about a race condition, but I suppose it'd just be old data if we do end up calling another thread during this. And of course until the GIL is gone, we won't really have any issues, anyway.

esrally/storage/_range.py
```diff
@@ -404,12 +404,12 @@ def rangeset(ranges: str | Iterable[Range]) -> RangeSet:
     :param ranges: it could represent a string to be parsed or a sequence of ranges.
     :return: A RangeSet obtained from the given text.
     """
-    if not isinstance(ranges, str):
-        return _rangeset(_combine(ranges))
-    return _rangeset(_combine(_parse(ranges)))
+    if isinstance(ranges, str):
+        ranges = list(_parse(ranges))
```
favilo: Curious why we need to alocate a list here instead of just storing the `Iterable` that is returned

favilo (commented):
This looks pretty good, but I had some comments.

esrally/storage/_adapter.py
fressi-elastic: Tuple is a constant, while list is mutable. This is the only reason. The actual only requirement is to be Itarable because the order is not important, so I am going to change it.

esrally/storage/_adapter.py
fressi-elastic: We are using this for now in tests. Normally adapter types are going to re-implement it only if they need to parse some specific configuration. Otherwise the regular `Adapter` implementation is fine. The `cls` parameter is going to be any Adapter subclass, so this method servers any subclass that are ignoring the configuration options.

esrally/storage/_adapter.py
fressi-elastic: Indeed. At the beginning we should have very little number of adapters in most of cases, and the HTTPAdapter should be valid for most of the cases (hopefully). Speed will never be a problem here. I would expect other esrally users could be implementing their own adapters for their specific cases, but we don't to care about it as the interface with adapters is very very simple and narrow. The list of adapters classes could anyway be restricted/expanded by editing the configuration file, so any change to esrally is necessary only for adapters of public interest.

esrally/storage/_client.py
fressi-elastic: Is this a standard? Because I omitted `DEFAULT_` prefix every where as almost all constants in storage should be having it for coherence. Please confirm me which is your preference in this context.

esrally/storage/_client.py
fressi-elastic: I removed them because it was producing too much output. I guess I should reintroduce them at least for debugging the first times we are going in production. I'll put more.

esrally/storage/_range.py
fressi-elastic: No string reason, I guess some branch merge left this from a previous implementation before merging to master. Going to revert this.

esrally/storage/_adapter.py
fressi-elastic: I replaced this with a comma-separated string list. This is harder to confuse. `Iterable[str]` is easy to be confused with `str` as the second implements the interface and finally the linter will not help catching the problem.

esrally/storage/_adapter.py
fressi-elastic: I update the method doc-string

esrally/storage/_range.py
fressi-elastic: I fixed. Thank you for spotting it out. It was from a parent PR this was build on top.

esrally/storage/_client.py
favilo: Nah, it's fine. This was just a slight issue

tests/storage/client_test.py
```diff
@@ -43,7 +43,7 @@
 
 class HTTPSAdapter(Adapter):
 
-    __adapter_URL_prefixes__ = ("https:",)
+    __adapter_URL_prefixes__ = "https://"
```
favilo, line 46: Heh, I think technically this was correct before the double slash. But don't change it.

Originally, the double slash meant that the host was "authoratative" or something like that, so it was just incidental, but now I believe it's guaranteed to exist, so this is better.

favilo (approved):

docs/configuration.rst
```diff
@@ -133,6 +133,44 @@ remote servers. Available options are:
     storage.adapters = esrally.storage._http:HTTPAdapter
 
 
+  At this point in time `esrally.storage._http:HTTPAdapter` is the only existing `Adapter` implementations intended
+  for public use and that is already the default one. So it is required to edit this option for special customizations.
+
+* ``storage.max_connections`` represents the maximum number of client connections to be made against the same server or
+  bucket. The default value is 8.
+
+* ``storage.mirror_files`` is used to provide a json file that specify the mapping for mirrors URLs resolution.
```
gbanasiak, line 142: @fressi-elastic Can we add a comment it's comma-separated list?

esrally/storage/_client.py
```diff
@@ -0,0 +1,232 @@
...
+            urls = self._mirrors.resolve(url)
+        except ValueError:
+            urls = [url]
+        else:
+            if len(urls) > 1:
+                # It shuffles mirror URLs in the hope two threads will try different mirrors for the same URL.
+                self._random.shuffle(urls)
+
+                # It uses measured latencies and the number of connections to affect the order of the mirrors.
+                weights = {u: self._average_latency(u) * self._random.uniform(1.0, self._server_connections(url).count) for u in urls}
```
gbanasiak, line 127: @fressi-elastic Why do we need random value here?

gbanasiak (commented):
Left some after-party questions.

PR #1952, opened 2025-06-20
Create module to implement multipart downloading using threads and previously submitted client. 

fressi-elastic:
This introduces a new class `Transfer` supporting multi-part downloading. The class is internally thread safe. It makes uses of a thread pool executor for spawning more threads. The module also support resuming downloads from an interrupted state. It uses the source URL and `content_length` to check if the file has been changed on the server side before resuming the download.

*NOTE*: part of this has been taken out to the following this one is made on top of PRs:
- [x] PR #1955


esrally/storage/_transfer.py
```diff
@@ -0,0 +1,464 @@
...
+        self._path = path
+        self._started = threads.TimedEvent()
+        self._workers = threads.WaitGroup(max_count=max_connections)
+        self._finished = threads.TimedEvent()
+        self._multipart_size = multipart_size
+        self._head = head
+        self._todo = todo
+        self._fds: list[FileDescriptor] = []
+        self._executor = executor
+        self._errors = list[Exception]()
```
favilo: I didn't realize this was proper syntax. Could you use this for `_fds`, too? It's clever

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,464 @@
...
+    return f"{value:.2f}TB"
+
+
+def pretty_time(value: int | float) -> str:
+    value = float(value)
+    if value < 60.0:
+        return f"{value:.0f}s"
+    value /= 60.0
+    if value < 60.0:
+        return f"{value:.2f}m"
```
favilo: Hmm, I wonder if we'd prefer to have a `10m20s` format, instead of `10.33m` format

esrally/utils/threads.py
```diff
@@ -89,7 +89,7 @@ def set(self) -> bool:
             if self._flag:
                 return False
             self._flag = True
-            self._time = time.monotonic_ns()
+            self._time = time.monotonic()
```
favilo: Just curious about why going back to float instead of int ns?

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,464 @@
...
+                    ranges = fd.ranges
+                self._client.get(self._head.url, fd, ranges)
+        except StreamClosedError as ex:
+            LOG.info("task cancelled: %s: %s", self.url, ex)
+            cancelled = True
+        except ServiceUnavailableError as ex:
+            with self._lock:
+                count = self._workers.count
+                max_count = max(1, count - 1)
+                self._workers.max_count = max_count
```
favilo, line 312: How will we trim down or stop the running threads to enforce this? Or do we expect all of them above the lookout would receive the error?

esrally/utils/threads.py
fressi-elastic: Because precision is not that important here, and because using 1s as a unit makes every computation much simpler that using 1ns. Think about computing time duration to print statistics... or computing how much to sleep... if you are used to time.time() like me (that returns a float) then monotonic would look more natural to use. In this specific case we are computing difference of times and the consuming them as secodns, therefore having the values already in seconds is less error prone and simpler.

esrally/storage/_transfer.py
fressi-elastic: Threads are being allocated by a thread pool executor. By submitting tasks to the thread pool, workers threads pull tasks from the shared executor queue and execute them one by one. To limit the number connections in the extreme case it reaches the limit of the total allowed connections on the client and server side for all the mirror servers, the client raises an exception that is captured here. The transfer object handle this case by reducing the maximum number of workers (and by consequence connections). After this code is executed the task ends eventually without re-scheduling itself for another execution. The max_count here is used to limit the number of tasks that can be scheduled and executed at the same time (`self._workers` object works as a semaphore raising exception when the limit is reached). The weakness of this mechanism is that there is no place where this `max_count` number is  increased after the object is created. The eventuality this happen for now is not proved to happening with AWS or rally tracks services, therefore for now it is OK like that. We will see if a better mechanism (may be with a TTL timeout) has to be implemented after we see this working in nightly executions.

esrally/storage/__init__.py
```diff
@@ -1,16 +0,0 @@
-# Licensed to Elasticsearch B.V. under one or more contributor
-# license agreements. See the NOTICE file distributed with
-# this work for additional information regarding copyright
```
gbanasiak: we want to keep the license header

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,430 @@
...
+class TransferStatus(enum.Enum):
+    INITIALIZED = 0
+    QUEUED = 1
+    DOING = 2
+    DONE = 3
+    CANCELLED = 4
+    FAILED = 5
+
+
+MULTIPART_SIZE = 8 * 1024 * 1024
```
gbanasiak, line 53: I'm curious, is there specific reason for using 8 MiB as a chunk of work?

esrally/storage/_transfer.py
gbanasiak: > I didn't realize this was proper syntax.

Same. Can we change to standard notation? I assume that's equivalent to:

```
self._errors: list[Exception] = []
```

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,430 @@
...
+    """OutputStream provides an output stream wrapper able to prevent exceeding writing given range."""
+
+    def __init__(self, path: str, ranges: RangeSet = NO_RANGE):
+        if os.path.isfile(path):
+            fd = open(path, "r+b")
+        else:
+            os.makedirs(os.path.dirname(path), exist_ok=True)
+            fd = open(path, "w+b")  # pylint: disable=consider-using-with
+        if ranges and ranges.start > 0:
+            fd.seek(ranges.start)
```
gbanasiak, line 561: Ok, so we will be seeking potentially non-linearly to non-existent locations in a file which actually works. And we are using separate file descriptor in each thread, so this should be thread safe. I feel this needs some high-level description somewhere in the comment.

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,430 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+
+class FileWriter(FileDescriptor):
+    """OutputStream provides an output stream wrapper able to prevent exceeding writing given range."""
+
+    def __init__(self, path: str, ranges: RangeSet = NO_RANGE):
+        if os.path.isfile(path):
+            fd = open(path, "r+b")
```
gbanasiak, line 556: Is there a guarantee 2 threads will not truncate the file twice? I think this is not done under lock now?

esrally/storage/_transfer.py
gbanasiak: My understanding is we let running threads complete their 8 MiB chunks or fail for some reason. The `max_count` throttling affects how many threads will run in parallel further. Is this correct?

esrally/utils/pretty.py
```diff
@@ -101,3 +102,39 @@ def _flat(o: Any) -> abc.Generator[tuple[str, str], None, None]:
                     yield str(k1), v2
     else:
         yield "", json.dumps(o)
+
+
+def seconds(s: float | int) -> str:
```
gbanasiak: There's some overlap here with `utils.convert`, also for `size()`. Do you see an opportunity to avoid this repeat? Just making sure you're aware of that module.

tests/storage/transfer_test.py
```diff
@@ -0,0 +1,181 @@
...
+    mismach_status_url=TransferCase(resume_status={"done": "128-255", "url": MISMATCH_URL, "content_length": 1024}, want_done="0-1023"),
+    # It tests mismatching content_length in the status file produces re-starting transfer from the beginning
+    mismach_status_content_length=TransferCase(
+        resume_status={"done": "128-255", "url": MISMATCH_URL, "content_length": 212}, want_done="0-1023"
+    ),
+)
+def test_transfer(case: TransferCase, executor: DummyExecutor, tmpdir: os.PathLike) -> None:
+    client = DummyClient.from_config(Config())
+    head = client.head(case.url)
+    path = os.path.join(str(tmpdir), os.path.basename(head.url))
```
gbanasiak: Nit: `os.path.basename()` against URL is hacky but works. Adding `urllib.parse` would make more sense.

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,430 @@
...
+            raise ValueError(f"url field not found in status file: {status_filename}")
+        if url != self.url:
+            raise ValueError(f"mismatching url in status file: '{status_filename}', got '{url}', want '{self.url}'")
+
+        content_length = document.get("content_length")
+        if content_length is None:
+            raise ValueError(f"content_length field not found in status file: {status_filename}")
+
+        if self.content_length != content_length:
+            raise ValueError(f"mismatching content length: got: {content_length}, want {self.content_length}")
```
gbanasiak: Lack of length difference does not guarantee the same content. Similar assumption is made in `_client.py` when resolving mirrors. Perhaps we could add `Last-Modified` header into the mix, i.e. mirror's `Last-Modified` must be the same or newer than the `Last-Modified` from the source to be considered?

But we're using compressed files, so decompression should fail some CRC check if we happen to download chunks from 2 different files. It's our last resort.

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,430 @@
...
+                    LOG.debug("transfer done: %s", self.url)
+                    return
+                if self._todo:
+                    # It eventually spawns another thread for multipart transfer
+                    self.start()
+                assert isinstance(fd, FileWriter)
+                ranges: RangeSet = NO_RANGE
+                if self._head.accept_ranges:
+                    ranges = fd.ranges
+                self._client.get(self._head.url, fd, ranges)
```
gbanasiak: This will subsequently call `_client.resolve()` but I don't see where content length is passed, so effectively we're not filtering the list of mirrors based on content length?

tests/storage/transfer_test.py
```diff
@@ -0,0 +1,181 @@
...
+            resume=case.resume,
+        )
+    except ValueError as exc:
+        assert case.want_error is not None
+        assert isinstance(exc, case.want_error)
+        return
+
+    assert case.want_error is None
+    transfer.start()
+    executor.execute_tasks()
```
gbanasiak: My understanding is this emulates completion of a single task because we submit further tasks when executing the first one, and only execute once. Have you considered adding a unit test with actual thread pool?

gbanasiak (commented):
This looks great. I left a bunch of questions to understand the code better.

esrally/storage/_transfer.py
fressi-elastic: Yes every task will download of the TODO chunks of files then release its thread to do the next task in the queue. Before taking downloading and after downloading it eventually schedule another task for execution according to the state of the transfer before and after downloading. This make the number of scheduled tasks to increase and therefore the number of threads requested for its execution. It ideally quickly scale up exponentially to the double of the number of threads at each pass. But at some point it quickly reach the max_count limit and stop scheduling new tasks until the count of goes down. We keep the max_count >= 1 so that there is always at least one thread trying to download the file.

esrally/storage/__init__.py
fressi-elastic: Restored.

esrally/storage/_transfer.py
fressi-elastic: I copied this value from the S3 client library provided by Amazon. No idea which value would be better for us. Smaller values will make the client sending more requests for the same file adding some additional charge to the infrastructure. Bigger values could speed up things, but reduce the opportunity of dynamically distribute the effort between workers.

esrally/storage/_transfer.py
fressi-elastic: True, this should be clarified better in the code.

esrally/storage/_transfer.py
favilo: Ooh, good point about decompression. We should probably add a check for CRC failure and redownload in that case.

I also really like the `Last-Modified` header. What about also using `Content-Digest`? I'm not sure if it's supported everywhere, but we could pass the `Want-Content-Digest` header in the request, and fallback to `Last-Modified` if it's missing. However, that would differ based on `Content-Range`, so it's only good for comparing the header of the full contents, not the individual pieces.

There is also `Repr-Digest`, which might be good for range requests, though I'm still reading about the differences with that and `Content-Digest`.

fressi-elastic: It is clear that there are many ways to improve this. Can we do it as a separate feature after recollecting some ideas?

esrally/storage/_transfer.py
fressi-elastic: It was missing. I implemented it today.

esrally/storage/_transfer.py
fressi-elastic: Correct. I just fixed it by putting this under the lock in `Trasfer` class. Good catch!

tests/storage/transfer_test.py
fressi-elastic: True!

tests/storage/transfer_test.py
fressi-elastic: I on purpose execute all submitted tasks until the moment  executor.execute_tasks is called to check the effects of a single task execution. This allow me to check every intermediate step individually case by case and verify pre and post conditions. Up to now I tested this class with a real thread pool only in a script that does connect to real servers. Shall I add an unit test that emulates the transfer of whole a file split by parts using multiple parallel threads?

esrally/storage/_transfer.py
fressi-elastic: I was making some test and I saw that our origin server seams to ignore my Want-Repr-Digest request. But I found something interesting in the headers.

```
X-Goog-Hash: crc32c=kijlTQ==, md5=j0hWDprkR82dy1JTH0VBXA==
```

We could match these values with similar ones asked from mirror servers.

esrally/storage/_transfer.py
gbanasiak: Just making sure this was seen, so unresolving this.

esrally/storage/_transfer.py
gbanasiak: > Can we do it as a separate feature after recollecting some ideas?

Absolutely. We are just sharing ideas. I'm good as long as we manually test an edge case of 2 compressed archives with the _same_ length but _different_ content, multiplexed from 2 different mirror locations. I'd like to see Rally _fail_ in this case while decompressing before running a race. We can do it as part of review of the final PR in this saga. Once we observe how this works in production we can later make adjustments.

tests/storage/transfer_test.py
gbanasiak: > Shall I add an unit test that emulates the transfer of whole a file split by parts using multiple parallel threads?

That was my thought, unless it is more logical to do it in following PR? In the existing test I would make a comment explaining intended state is after a _single_ task execution. I was puzzled by case content initially.

esrally/storage/_transfer.py
fressi-elastic: I added a first draft implementation of using `crc32c` checksum to compare server side heads and the status file. It is  still missing a client side verification to validate downloaded data.

gbanasiak: Do we know the state of adoption of `X-Goog-Hash` header in AWS/Azure/GCP?

Edit: If it's not universally adopted let's not complicate the code now.

fressi-elastic:
I taken out some changes to the new PR #1955 this one is now based on.

esrally/storage/_transfer.py
fressi-elastic: Done

esrally/utils/threads.py
fressi-elastic: I reverted this one to nano seconds. I am now using convert module for pretty printing time durations, so having this as nano seconds is no a big deal anymore.

esrally/utils/pretty.py
fressi-elastic: I improved `convert` module and then based this one on that one.

esrally/storage/_transfer.py
fressi-elastic: X-Goog-Hash is something that comes from GCP (our actual source of true). We can store the value coming from there in the other storage services and then read it in the way specified by the actual implementation (AWS or Azure). The hash value where is supported allow to quickly check any transfer is succeeded just it have been done (either from server or client side). At the current state on the client side it only check the value from the servers matches (where is available, IE only GCP). The other server types and the client side have to be supported yet. I propose to do it at the moment we are going to make the uploads of the files as it is where the value is going to be passed to those other service types.

tests/storage/transfer_test.py
fressi-elastic: Done.

esrally/storage/_transfer.py
fressi-elastic: @gbanasiak @favilo Did I answered your questions?

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,582 @@
...
+
+class Transfer:
+    """Transfers class implements multipart file transfers by submitting tasks to an Executor.
+
+    By using a thread pool executor like concurrent.futures.ThreadPoolExecutor to delegate its tasks, it implements
+    parallel file transfers.
+
+    It splits the file in parts of up to multipart_size bytes, and allocate a task for downloading each part.
+    Each task is being executed by a thread provided by the Executor. Each thread after executing the task will
+    eventually execute the next task. Every task open local file for writing. Then it seeks to the beginning of the
```
gbanasiak:
```suggestion
    eventually execute the next task. Every task opens local file for writing. Then it seeks to the beginning of the
```

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,582 @@
...
+    file part before writing transferred data from there. On the network side it ask the server to send only the
```
gbanasiak:
```suggestion
    file part before writing transferred data from there. On the network side it asks the server to send only the
```

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,582 @@
...
+    part of the remote file that has to be written to the local file.
+
+    By reducing the size of the multipart_size it reduces the scope of each task, increasing the number of tasks and
+    parts of file to transfer. This produces some additional work because in increases the number of IO operations and
+    of request to the remote service. The advantage of that is making the task allocation between multiple threads more
```
gbanasiak:
```suggestion
    parts of file to transfer. This produces some additional work because it increases the number of IO operations and
    of requests to the remote service. The advantage of that is making the task allocation between multiple threads more
```

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,582 @@
...
+            LOG.info("transfer cancelled: %s: %s", self.url, ex)
+            cancelled = True
+        except ServiceUnavailableError as ex:
+            # The client raised this error because the maximum number of client connections for each remote server
+            # has been already established.
+            with self._lock:
+                count = self._workers.count
+                max_count = max(1, count - 1)
+                self._workers.max_count = max_count
+            # In case of the count of concurrent tasks is greater than 1 then if forbid to reschedule it.
```
gbanasiak:
```suggestion
            # In case of the count of concurrent tasks is greater than 1 then it forbids to reschedule it.
```

What is the reasoning behind `cancelled = count > 1`? We reduce the number of concurrent workers in line 312, so even if `self.start()` is called in `finally` block it would not do anything because of `WaitGroupLimitError` exception.

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,582 @@
...
+        finally:
+            # It decreases the number of scheduled tasks, allowing a new tasks to be submit.
+            self._workers.done()
+            self.save_status()
+            if self._errors:
+                if self._finished.set():
+                    # The first task entered here will write the error(s) to the log fine.
+                    LOG.error("transfer failed: %s:\n - %s", self.url, "\n - ".join(str(e) for e in self._errors))
+            elif cancelled:
+                # It will submit any task for execution.
```
gbanasiak, line 333: `self.start()` is not called, so will it?

esrally/storage/_transfer.py
```diff
@@ -0,0 +1,582 @@
...
+                fd = FileWriter(self.path, todo)
+                # It registers the file descriptor so that it can be closed from another thread.
+                self._fds.append(fd)
+            with fd:
+                try:
+                    yield fd
+                finally:
+                    with self._lock:
+                        try:
+                            # It de-registers the file descriptor so that we can't be closed from another thread.
```
gbanasiak:
```suggestion
                            # It de-registers the file descriptor so that it can't be closed from another thread.
```

tests/storage/transfer_test.py
```diff
@@ -0,0 +1,248 @@
...
+        want_init_todo="0-49",
+        document_length=50,
+        want_init_document_length=50,
+        want_final_written="0-49",
+        want_final_document_length=50,
+        want_final_error=RuntimeError,
+    ),
+    # It tests multipart working when multipart_size < content_length.
+    no_document_length=TransferCase(want_init_document_length=None, want_init_todo="0-", want_final_done="0-1023", document_length=None),
+    # It tests when max_connections < 0.
```
gbanasiak:
```suggestion
    # It tests when max_connections == 0.
```

gbanasiak (approved):
LGTM. Many thanks for multiple iterations. I've left non-blocking comments.

esrally/storage/_transfer.py
fressi-elastic: Start will not be called on purpose in case cancelled is true. Another thread will eventually call it. Cancelled is being called when the the transfer is being shut down (streams closed error) or when the server refuses to handle so many connections, so we intentionally leave threads to die (with the exception of the last one) without rescheduling the task for the next executions, actually reducing the number of threads assigned to the transfers one by one after task completion.

tests/storage/transfer_test.py
fressi-elastic: Done.

gbanasiak:
I've re-tried serverless IT test, it failed for unrelated reasons.

PR #1956, opened 2025-07-07
Pin aiosignal

gbanasiak:
Temporary solution for https://github.com/elastic/rally/actions/runs/16075552010/job/45369828493.


favilo (commented):
Good idea. Might be time to switch over to uv for that nice locking behavior

favilo (approved):

PR #1957, opened 2025-07-07
Introducing skip-telemetry parameter

NickDris:
This PR introduces a new optional flag, `--skip-telemetry`, to the `esrally stop` command. When specified, Rally will skip running telemetry devices during the stop phase, preventing the collection of potentially invalid or irrelevant metrics.

### Motivation

- Avoids collecting and reporting telemetry from failed or incomplete benchmarks, which can pollute metrics dashboards and downstream analyses with unwanted irrelevant data.
- Provides users with more control over when telemetry is collected

### Usage
To skip telemetry collection during the stop phase, use:
`esrally stop --skip-telemetry`

### DIFF
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -84,16 +84,16 @@ class DockerLauncher:
         logging.error(msg)
         raise exceptions.LaunchError(msg)
 
-    def stop(self, nodes, metrics_store):
+    def stop(self, nodes, metrics_store, skip_telemetry=False):
         self.logger.info("Shutting down [%d] nodes running in Docker on this host.", len(nodes))
         for node in nodes:
             self.logger.info("Stopping node [%s].", node.node_name)
-            if metrics_store:
+            if metrics_store and not skip_telemetry:
                 telemetry.add_metadata_for_node(metrics_store, node.node_name, node.host_name)
             node.telemetry.detach_from_node(node, running=True)
             process.run_subprocess_with_logging(self._docker_compose(node.binary_path, "down"))
             node.telemetry.detach_from_node(node, running=False)
-            if metrics_store:
+            if metrics_store and not skip_telemetry:
                 node.telemetry.store_system_metrics(node, metrics_store)
 
 
@@ -223,12 +223,12 @@ class ProcessLauncher:
 
         return wait_for_pidfile(pid_path)
 
-    def stop(self, nodes, metrics_store):
+    def stop(self, nodes, metrics_store, skip_telemetry=False):
         self.logger.info("Shutting down [%d] nodes on this host.", len(nodes))
         stopped_nodes = []
         for node in nodes:
             node_name = node.node_name
-            if metrics_store:
+            if metrics_store and not skip_telemetry:
                 telemetry.add_metadata_for_node(metrics_store, node_name, node.host_name)
             try:
                 es = psutil.Process(pid=node.pid)
@@ -258,6 +258,6 @@ class ProcessLauncher:
 
                 node.telemetry.detach_from_node(node, running=False)
             # store system metrics in any case (telemetry devices may derive system metrics while the node is running)
-            if metrics_store:
+            if metrics_store and not skip_telemetry:
                 node.telemetry.store_system_metrics(node, metrics_store)
         return stopped_nodes
--- a/esrally/mechanic/launcher.py
+++ b/esrally/mechanic/launcher.py
@@ -84,15 +84,16 @@ class DockerLauncher:
         logging.error(msg)
         raise exceptions.LaunchError(msg)
 
-    def stop(self, nodes, metrics_store, skip_telemetry=False):
+    def stop(self, nodes, metrics_store=None):
         self.logger.info("Shutting down [%d] nodes running in Docker on this host.", len(nodes))
         for node in nodes:
             self.logger.info("Stopping node [%s].", node.node_name)
-            if metrics_store:
+            if metrics_store is not None:
                 telemetry.add_metadata_for_node(metrics_store, node.node_name, node.host_name)
                 node.telemetry.detach_from_node(node, running=True)
+
             process.run_subprocess_with_logging(self._docker_compose(node.binary_path, "down"))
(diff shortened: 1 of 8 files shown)


gbanasiak:
I gave it a quick look and think the skip should also apply to `detach_from_node()` methods. There's no point to collect anything, if we're not storing it afterwards.

esrally/mechanic/launcher.py
```diff
@@ -84,16 +84,16 @@ def _wait_for_healthy_running_container(self, container_id, timeout):
         logging.error(msg)
         raise exceptions.LaunchError(msg)
 
-    def stop(self, nodes, metrics_store):
+    def stop(self, nodes, metrics_store, skip_telemetry=False):
```
fressi-elastic: Instead of using two params to handle metric_store you could make it optional and skip using it when it is `None`. It actually makes no big sense anyway having to pass a parameter than later is not used. Example:

```
def stop(self, nodes, metrics_store=None):

  # ...
  if metrics_store is not None:
    # ...
  
```

esrally/mechanic/launcher.py
```diff
@@ -223,12 +223,12 @@ def _start_process(binary_path, env):
 
         return wait_for_pidfile(pid_path)
 
-    def stop(self, nodes, metrics_store):
+    def stop(self, nodes, metrics_store, skip_telemetry=False):
```
fressi-elastic: Same as above.

esrally/mechanic/mechanic.py
```diff
@@ -147,7 +146,7 @@ def stop(cfg: types.Config):
         current_race = None
         metrics_store = None
 
-    node_launcher.stop(nodes, metrics_store)
+    node_launcher.stop(nodes, metrics_store, cfg.opts("mechanic", "skip.telemetry"))
```
fressi-elastic: `skip.telemetry` option should be optional to avoid invalidating already existing configuration files.
Also consider skipping the logic that constructs metrics_store in the case `skip.telemetry` is false.

Please also use `[esrally.utils.convert.to_bool](https://github.com/elastic/rally/blob/master/esrally/utils/convert.py#L231)` function to parse the option. Example:

```python



```Python
skip_telemetric = cfg.opts("mechanic", "skip.telemetry", False, mandatory=False)

metrics_store = None
if not skip_telemetry:
  try:
     metrics_store = metrics.metrics_store_class(cfg)(cfg)
   except exceptions.NotFound:
     
   
  node_launcher.stop(nodes, metrics_store)
    _delete_node_file(root_path)

    if current_race:
        if metrics_store is not None:
           metrics_store.flush(refresh=True)
        for node in nodes:
            results = metrics.calculate_system_results(metrics_store, node.node_name)
            current_race.add_results(results)
            metrics.results_store(cfg).store_results(current_race)
        
       if metrics_store is not None:
           metrics_store.close()
  
```
Please note that in the current implementation when `bool(current_race)` is True then it will fail because the None case has to be supported yet. So the verification for the value to be None has to be implemented anyway or fix the [like 48](https://github.com/elastic/rally/blob/master/esrally/mechanic/mechanic.py#L148).

esrally/rally.py
```diff
@@ -1159,6 +1162,7 @@ def dispatch_sub_command(arg_parser, args, cfg: types.Config):
             mechanic.start(cfg)
         elif sub_command == "stop":
             cfg.add(config.Scope.applicationOverride, "mechanic", "preserve.install", convert.to_bool(args.preserve_install))
+            cfg.add(config.Scope.applicationOverride, "mechanic", "skip.telemetry", convert.to_bool(args.skip_telemetry))
```
fressi-elastic: Please verify this is not overriding the value of the configured file when no value is given by command line.
This makes the option value not mandatory, but it creates a strict dependency between this initialization code and the place where the value is finally consumed. This introduce an hidden technical dept that will makes a problem appear as soon the code is used without being initialized. So please look one of my previous comments.

esrally/types.py
```diff
@@ -145,6 +145,7 @@
     "serverless.mode",
     "serverless.operator",
     "skip.rest.api.check",
+    "skip.telemetry",
```
fressi-elastic, line 148: Please add documentation for this value to explain what this is expected to do and which is the default value.

fressi-elastic (changes requested):
I think there are issues introduced here and issues that should be fixed in the existing code with very little effort in the same are being change here. Please have a look to my comments for more details.

NickDris:
@fressi-elastic thanks a lot for the valuable feedback. Tech debt was imminent as my familiarity with configuration parameters is poor for the time being.
Give it a look now and tell me if the issues were resolved or if there's anything else we should cater for.

esrally/mechanic/launcher.py
NickDris: You are right. Changed as suggested.

esrally/mechanic/mechanic.py
NickDris: You're right. Now mechanic sets its default value to false:
  https://github.com/elastic/rally/blob/fd4b96ce0cee6fb00edc0fc00c9a829a252ac467/esrally/mechanic/mechanic.py#L131-L133

and takes account the case where metrics store has not been set (None)
https://github.com/elastic/rally/blob/fd4b96ce0cee6fb00edc0fc00c9a829a252ac467/esrally/mechanic/mechanic.py#L151-L161

esrally/rally.py
NickDris: Oh, you're very right. That link between parameters specified through command line AND configuration file was not that clear to me. 
Thanks for that wonderful insight. 

I think now it's under the correct initialization places. Give it a look.

esrally/types.py
NickDris: Check if the documentation is sufficient now. If something is not clarified let me know.

fressi-elastic (approved):

gbanasiak:
I think we're missing persisting `--skip-telemetry` in [here](https://github.com/elastic/rally/blob/master/esrally/rally.py#L1167-L1170).

PR #1960, opened 2025-07-16
Update storage tools with last changes from `storage` branch.

fressi-elastic:
This change introduces some improvements from this new feature development branch:
- `Head` class is now a `dataclass` used to pass parameters between methods and to verify responses with the requested parameters.
- It uses a generic `match_url` class method to maps URLS to adapter classes.
- `AdapterRegistry` class skips non-existing modules.
- It adds `list` and `put` methods to `Adapter` interface.
- It adds `MirrorList.from_config` factory method.
- It adds `put` method to `Client` class.
- It adds a `ThreadPoolExecutor` subclass with its `from_config` factory method.
- It makes the `HTTPAdapter` easier to extend.

### DIFF
--- a/esrally/storage/_adapter.py
+++ b/esrally/storage/_adapter.py
@@ -21,6 +21,7 @@ import importlib
 import logging
 import threading
 from abc import ABC, abstractmethod
+from collections.abc import Container
 from dataclasses import dataclass
 from typing import Protocol, runtime_checkable
 
@@ -39,6 +40,9 @@ class Writable(Protocol):
 
     def write(self, data: bytes) -> None:
         pass
+
+
+_HEAD_CHECK_IGNORE = frozenset(["url"])
 
 
 @dataclass
@@ -76,11 +80,14 @@ class Head:
             date=date,
         )
 
-    def check(self, other: Head) -> None:
+    def check(self, other: Head, ignore: Container[str] = _HEAD_CHECK_IGNORE) -> None:
         for field in ("url", "content_length", "accept_ranges", "ranges", "document_length", "crc32c", "date"):
+            if ignore is not None and field in ignore:
+                continue
             want = getattr(self, field)
             got = getattr(other, field)
-            if not ({got, want} & {None, NO_RANGE}) and got != want:
+            # If both got abd want are specified, then they have to match.
+            if all([got, want]) and got != want:
                 raise ValueError(f"unexpected '{field}': got {got}, want {want}")
 
 
--- a/esrally/storage/_adapter.py
+++ b/esrally/storage/_adapter.py
@@ -100,7 +100,7 @@ class Adapter(ABC):
     """Base class for storage class client implementation"""
 
     @classmethod
-    def match_url(cls, url: str) -> str:
+    def match_url(cls, url: str) -> bool:
         """It returns a canonical URL in case this adapter accepts the URL, None otherwise."""
         raise NotImplementedError
 
@@ -184,19 +184,16 @@ class AdapterRegistry:
             self._classes.append(cls)
         return cls
 
-    def get(self, url: str) -> tuple[Adapter, str]:
-        with self._lock:
-            for cls in self._classes:
-                try:
-                    actual_url = cls.match_url(url)
-                    break
-                except NotImplementedError:
-                    continue
-            else:
-                raise ValueError(f"No adapter found for url '{url}'")
+    def get(self, url: str) -> Adapter:
+        for cls in self._classes:
+            if cls.match_url(url):
+                break
+        else:
+            raise ValueError(f"No adapter found for URL: {url}")
 
         adapter = self._adapters.get(cls)
-        if adapter is None:
-            adapter = cls.from_config(self._cfg)
-        self._adapters[cls] = adapter
-        return adapter, actual_url
+        if adapter is not None:
+            return adapter
+
+        self._adapters[cls] = adapter = cls.from_config(self._cfg)
+        return adapter
--- a/esrally/storage/_adapter.py
+++ b/esrally/storage/_adapter.py
@@ -21,7 +21,6 @@ import importlib
 import logging
 import threading
 from abc import ABC, abstractmethod
-from collections.abc import Iterator
 from dataclasses import dataclass
 from typing import Protocol, runtime_checkable
 
@@ -39,13 +38,6 @@ class ServiceUnavailableError(Exception):
 class Writable(Protocol):
 
     def write(self, data: bytes) -> None:
-        pass
-
-
-@runtime_checkable
-class Readable(Protocol):
-
-    def read(self, size: int = -1) -> bytes:
         pass
 
 
@@ -120,13 +112,6 @@ class Adapter(ABC):
         """
 
     @abstractmethod
-    def list(self, url: str) -> Iterator[Head]:
-        """It gets list of file headers.
-        :return: the Head of the remote file.
-        :raises ServiceUnavailableError: in case on temporary service failure.
-        """
-
-    @abstractmethod
     def get(self, url: str, stream: Writable, head: Head | None = None) -> Head:
         """It downloads a remote bucket object to a local file path.
 
@@ -137,20 +122,6 @@ class Adapter(ABC):
             - document_length: the number of bytes to transfer.
             - crc32c the CRC32C checksum of the file.
             - date: the date the file has been modified.
-        :raises ServiceUnavailableError: in case on temporary service failure.
-        """
-
-    @abstractmethod
-    def put(self, stream: Readable, url: str, head: Head | None = None) -> Head:
-        """It uploads a local file object to a remote bucket.
-
-        :param stream: it represents the local file stream where to read data from.
-        :param url: it represents the URL of the remote file object.
-        :param head: it allows to specify optional parameters:
-            - range: the portion of the file to transfer (it must be empty or a continuous range).
-            - document_length: the number of bytes to transfer.
-            - crc32c the CRC32C checksum of the file.
-            - date: the date the file was modified.
         :raises ServiceUnavailableError: in case on temporary service failure.
         """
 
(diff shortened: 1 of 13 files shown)


esrally/storage/_adapter.py
```diff
@@ -78,21 +73,24 @@ def create(
...
             document_length=document_length,
             crc32c=crc32c,
+            date=date,
         )
 
+    def check(self, other: Head) -> None:
+        for field in ("url", "content_length", "accept_ranges", "ranges", "document_length", "crc32c", "date"):
+            want = getattr(self, field)
+            got = getattr(other, field)
+            if not ({got, want} & {None, NO_RANGE}) and got != want:
```
favilo: Maybe put a comment here explaining that this is checking if one of the sets is `None` or `NO_RANGE`? This confused me for a bit.

This is clever code, I generally write it as an `any()` call over a list comprehension, but this is more concise, it's just not as clear, to me, on initial read

esrally/storage/_adapter.py
```diff
@@ -114,33 +112,34 @@ def head(self, url: str) -> Head:
...
+        :param head: it allows to specify optional parameters:
+            - range: portion of the file to transfer (it must be empty or a continuous range).
+            - document_length: the number of bytes to transfer.
+            - crc32c the CRC32C checksum of the file.
+            - date: the date the file has been modified.
         :raises ServiceUnavailableError: in case on temporary service failure.
         """
 
 
 ADAPTER_CLASS_NAMES = ",".join(
```
favilo: Hmm, I'm not sure I like this interface. Could we annotate it as `Final[list[str]]` and just use a list? I don't know if it's the best idea to join and split later.

esrally/storage/_client.py
```diff
@@ -90,7 +93,7 @@ def head(self, url: str, ttl: float | None = None) -> Head:
                     # cached value or error is enough recent to be used.
                     return _head_or_raise(value)
 
-        adapter = self._adapters.get(url)
+        adapter, url = self._adapters.get(url)
```
favilo: I know it's an internal method, but should we use `self.adapters.get()` here? Since you made the `@property`? 

Actually, what was the motivation for having this be a `@property`? Is it just to prevent external modification? If so, that's a perfectly valid reason, I was just curious.

esrally/storage/_http.py
```diff
@@ -82,101 +87,121 @@ def head(self, url: str) -> Head:
...
-    if not got:
-        return None
-    return got == "bytes"
+    @classmethod
+    def _make_head(cls, url: str, headers: CaseInsensitiveDict) -> Head:
+        content_length = cls._content_length_from_headers(headers)
+        accept_ranges = cls._accept_ranges_from_headers(headers)
+        ranges, document_length = cls._content_range_from_headers(headers)
+        hashes = cls._hashes_from_headers(headers)
+        crc32c = hashes.get("crc32c")
```
favilo: Maybe add a comment about specifically not including `date` yet, because it's not supported?

esrally/storage/_http.py
```diff
@@ -82,101 +87,121 @@ def head(self, url: str) -> Head:
...
-    try:
-        value, document_length_text = content_range_text[len("bytes ") :].strip().replace(" ", "").split("/", 1)
-        document_length: int | None = None
-        if document_length_text != "*":
-            document_length = int(document_length_text)
-        return rangeset(value), document_length
-    except ValueError:
-        raise ValueError(f"invalid content range in '{content_range_text}'")
+        if "," in content_range_text:
+            raise ValueError("multi range value is not unsupported")
```
favilo: I don't think you want the double negative here.

esrally/storage/_http.py
```diff
@@ -82,101 +87,121 @@ def head(self, url: str) -> Head:
...
 
+        try:
+            value, document_length_text = content_range_text[len("bytes ") :].strip().replace(" ", "").split("/", 1)
```
favilo: `"bytes "` here feels like a magic value, you've used it twice. Mind pulling it into a variable? Maybe a `Final` called `PREFIX` or something like it?

esrally/storage/_http.py
```diff
@@ -82,101 +87,121 @@ def head(self, url: str) -> Head:
...
+            try:
+                for entry in hashes_text.split(","):
+                    name, value = entry.split("=", 1)
+                    hashes[name] = value
+            except ValueError as e:
+                raise ValueError(f"invalid X-Goog-Hash value: {hashes_text}") from e
+
+        for k, value in headers.items():
+            if k.startswith("x-amz-checksum-"):
+                name = k[len("x-amz-checksum-") :]
```
favilo: Same magic value thing here. Let's pull it into a constant or `Final` variable.

Though I really like the `len(AMZ_PREFIX)` pattern, it make it very clear what we are skipping

tests/storage/adapter_test.py
```diff
@@ -74,31 +92,31 @@ def registry(cfg: Config) -> AdapterRegistry:
 @dataclass()
 class RegistryCase:
     url: str
-    want: type[Adapter] | type[Exception]
+    want_type: type[Adapter] | None = None
```
favilo, line 87: I like this. Thanks for splitting out the wants into valid and error. It feels closer to actual union types in languages with actual union types.

favilo (commented):
Looking pretty good. I've left some comments and questions

esrally/storage/_adapter.py
fressi-elastic: I replaced it with:
```if all([got, want]) and got != want:```

I guest this is simpler, faster and easier to understand.

File: esrally/storage/_client.py (hunk 3 above)
Comment (fressi-elastic) [reply esrally/storage/_client.py]: I just removed the `adapters` property as it is never used. I think it is better if adapters are only used as an implementation details of Client class to keep high and low level separated where possible.

File: esrally/storage/_adapter.py [hunk 9]
```diff
@@ -114,70 +124,79 @@ def head(self, url: str) -> Head:
...
-    ]
-)
-
-
-class AdapterClassEntry(NamedTuple):
-    url_prefix: str
-    cls: type[Adapter]
+ADAPTER_CLASS_NAMES = [
+    "esrally.storage._tracks:TracksRepositoryAdapter",
+    "esrally.storage._s3:S3Adapter",
```
gbanasiak, line 143: These will come later?

esrally/storage/_http.py
fressi-elastic: True!

esrally/storage/_adapter.py
```diff
@@ -114,70 +124,79 @@ def head(self, url: str) -> Head:
...
 
-    def register_class(self, cls: type[Adapter], prefixes: Iterable[str]) -> type[Adapter]:
+    def register_class(self, cls: type[Adapter]) -> type[Adapter]:
         with self._lock:
-            for p in prefixes:
-                self._classes.append(AdapterClassEntry(p.strip(), cls))
-            # The list of adapter classes is kept sorted from the longest prefix to the shorter to ensure that matching
-            # a shorter URL prefix will never hide matching a longer one.
-            self._classes.sort(key=lambda e: len(e.url_prefix), reverse=True)
+            self._classes.append(cls)
```
gbanasiak, line 184: What is the rationale for abandoning the longest-match approach?

esrally/storage/_http.py
fressi-elastic: Done.

esrally/storage/_adapter.py
```diff
@@ -114,70 +124,79 @@ def head(self, url: str) -> Head:
...
         return cls
 
-    def get(self, url: str) -> Adapter:
+    def get(self, url: str) -> tuple[Adapter, str]:
```
gbanasiak: Why this change of returned type? Wouldn't it be enough to keep returning adapter alone?

gbanasiak (commented):
This looks good. I left some questions to understand this change better.

esrally/storage/_adapter.py
fressi-elastic: This gives the adapter class the option of changing the URL, but I think it is not very nice to have as it could generate confusion when debugging configuration problems. I am going to revert it.

fressi-elastic: Corrected.

esrally/storage/_adapter.py
fressi-elastic: The main reason is because the now `s3` adapter (not included here) was matching also some https URLS that was directed to amazon, following the virtual host schema, where the bucket name is embedded at the beginning of the hostname. But after thinking it twice I think we should always bind https adapter for https URLs and s3 adapter for s3 urls so that it would be simpler to understand the expected behavior when debugging mirrors configuration. So There is no more strong reasons not to restore the original implementation.
Anyway it is simpler to understand if the order of adapters is explicitly given, so that the user has the option to fix it in case he wants by editing the configuration file.

esrally/storage/_adapter.py
fressi-elastic: Yes it is coming later. Having this here will no make difference at this point as a failure importing it, it will make it just skipping its adapter class. This is desired as after having it, `esrally` should be able to work with other adapters in case s3 library (boto3) is not installed.

esrally/storage/_adapter.py
fressi-elastic: @gbanasiak This make sense to you?

esrally/storage/_adapter.py
fressi-elastic: @ gbanasiak Shall I remove it for now?

gbanasiak: It's fine, thanks for explanations.

esrally/storage/_adapter.py
gbanasiak: Yes, thanks for explanations.

gbanasiak (approved):
LGTM

PR #1962, opened 2025-07-16
feat: Enable uv for venv management.

favilo:


gbanasiak:
Quick comments:
* I would set `PY_VERSION` to `3.12.2` to reflect current end state after running `make prereq`
* We need corresponding change in internal tools before this gets merged

favilo:
> > Do we need to do this? `pip` should still work after this change.
> 
> You are right. I thought we rely on `make install`. It's enough to confirm all works fine with the PR branch, but there's no reason it wouldn't.

Hmm, I suppose I could add `uv` to the Docker images. Building docker images faster is a reason I wanted `uv` in the first place.

Makefile
```diff
@@ -17,64 +17,60 @@
...
-	@if [[ ! -x $$(command -v pyenv) ]]; then \
-		printf $(PYENV_ERROR); \
-		exit 1; \
-	fi;
-	@if [[ ! -f $(VENV_ACTIVATE_FILE) ]]; then \
-		eval "$$(pyenv init -)" && eval "$$(pyenv init --path)" && $(PY_BIN) -mvenv $(VIRTUAL_ENV); \
-		printf "Created python3 venv under $(PWD)/$(VIRTUAL_ENV).\n"; \
-	fi;
+
+PY_VERSION = 3.12.2
```
gbanasiak: I would still use `.ci/variables.json` for this, as docs and IT keep reading this file. Although `prerequisite_versions` section from `.ci/variables.json` could go away I think.

Makefile
```diff
@@ -98,18 +94,18 @@ serve-docs: check-venv
 	@. $(VENV_ACTIVATE_FILE); cd docs && $(MAKE) serve
 
 test: check-venv
-	. $(VENV_ACTIVATE_FILE); nox -s test-3.9
-	. $(VENV_ACTIVATE_FILE); nox -s test-3.12
+	uv run --python=3.9 nox -s test-3.9
+	uv run --python=3.12 nox -s test-3.12
```
gbanasiak: Aren't we potentially installing Python twice here? Once by `uv`, and then by `nox`? I think I prefer the previous version with `VENV_ACTIVATE` more.

Makefile
```diff
@@ -17,64 +17,60 @@
...
-export HATCHLING_VERSION := $(shell jq -r '.prerequisite_versions.HATCHLING' .ci/variables.json)
-export PIP_VERSION := $(shell jq -r '.prerequisite_versions.PIP' .ci/variables.json)
-export WHEEL_VERSION := $(shell jq -r '.prerequisite_versions.WHEEL' .ci/variables.json)
 VIRTUAL_ENV ?= .venv
 VENV_ACTIVATE_FILE := $(VIRTUAL_ENV)/bin/activate
-VENV_ACTIVATE := . $(VENV_ACTIVATE_FILE)
-VEPYTHON := $(VIRTUAL_ENV)/bin/$(PY_BIN)
-PYENV_ERROR := "\033[0;31mIMPORTANT\033[0m: Please install pyenv.\n"
-PYENV_PREREQ_HELP := "\033[0;31mIMPORTANT\033[0m: please type \033[0;31mpyenv init\033[0m, follow the instructions there and restart your terminal before proceeding any further.\n"
-VE_MISSING_HELP := "\033[0;31mIMPORTANT\033[0m: Couldn't find $(PWD)/$(VIRTUAL_ENV); have you executed make venv-create?\033[0m\n"
```
gbanasiak, line 38: `VE_MISSING_HELP` is used in `check-venv`.

gbanasiak (commented):
Thank you for this change. I will do some testing tomorrow, just dropping a few comments.

Makefile
favilo: Good point.

docs/developing.rst
```diff
@@ -6,7 +6,7 @@ Prerequisites
 
 Install the following software packages:
 
-* pyenv installed, Follow the instructions in the output of ``pyenv init`` to setup your shell and then restart it before proceeding. For more details please refer to the pyenv `installation instructions <https://github.com/pyenv/pyenv#installation>`_.
+* uv installed, Follow `the instructions <https://docs.astral.sh/uv/getting-started/installation/>`_ to install.
```
gbanasiak:
```suggestion
* `uv <https://docs.astral.sh/uv/getting-started/installation/>`_ 
```

gbanasiak (approved):
LGTM

PR #1964, opened 2025-07-17
It adds `TransferManager` and `S3Adapter` classes.

fressi-elastic:
* It adds the new `list` and `put` methods to the `Adapter` interface.
* It introduce a new high level `TransferManager` class used to orchestrate storage transfers that:
  * it controls transfer creation
  * it periodically monitors transfers in progress o print statistics and to control the maximum number of connections allowed for each transfer.
* It introduces the new `S3Adapter` class to allow transfers using the boto3 client libraries for AWS storage services.
* It updates configuration documentation.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -122,10 +122,7 @@ storage
 This section defines how client is configured to transfer corpus files. The main
 advantages of this downloader implementation are:
 
-* It supports configuring mirrors servers to download files from there instead of the original source URL. There could
-  be multiple services from where download files. The client will load balance between these services giving priority
-  to those servers with the lower latency. In case of failures reaching the mirrors services it will download files
-  from the original source server.
+* It supports configuring multiple mirror URLs. The client will load balance between these URLs giving priority to those with the lower latency. In case of failures the client will download files from the original URL.
 * It supports multipart downloading of pieces of files from multiple servers at the same time, increasing the impact of
   a failing network infrastructure. Those parts that experience errors downloading will be eventually re-downloaded from
   another server of from the original source.
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -119,8 +119,23 @@ Define a secure connection to an Elastic Cloud cluster::
 storage
 ~~~~~~~
 
-This section defines how client is configured to transfer big files (I.E. track files) from
-remote servers. Available options are:
+This section defines how client is configured to transfer big files (I.E. track files) from remote servers. The main
+advantages of this downloader implementation are:
+
+* It supports configuring mirrors servers to download files from there instead of the original source URL. There could
+  be multiple services from where download files. The client will load balance between these services giving priority
+  to those servers with the lower latency. In case of failures reaching the mirrors services it will download files
+  from the original source server.
+* It supports multipart downloading of pieces of files from multiple servers at the same time, increasing the impact of
+  a failing network infrastructure. Those parts that experience errors downloading will be eventually re-downloaded from
+  another server of from the original source.
+* It supports local caching of files, and file downloading resuming from previous status in case of connection
+  interruption. It can also resume file downloading between consecutive rally executions.
+
+*NOTE*: This transfers manager implementation is still experimental and under active development. It is not used yet by
+default and it will require some additional works before final delivery.
+
+Configuration options are:
 
 * ``storage.adapters`` is a comma-separated list of storage adapter implementations specified using the following
   format:
@@ -187,6 +202,8 @@ remote servers. Available options are:
 HTTP Adapter
 ************
 
+This adapter can be used only to download files from public HTTP or HTTPS servers.
+
 * ``storage.http.chunk_size`` is used to specify the size of the buffer is being used for transferring chunk of files.
 
 * ``storage.http.max_retries`` is used to configure the maximum number of retries for making HTTP adapter requests.
@@ -203,6 +220,40 @@ HTTP Adapter
 
   .. _urllib3.Retry: https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html
 
+
+S3 Adapter
+**********
+
+This adapter can be used only to download files from `S3 cloud storage service`_. It requires 'boto3' client library to be
+installed and it accepts only URLs with the following format ``s3://<bucket-name>/<object-key>``. It is intended to be
+used to configure S3 buckets as mirror for track file downloads. Example of ``rally.ini`` configuration::
+
+    [storage]
+    storage.mirror_files = ~/.rally/storage-mirrors.json
+
+Example of ``~/.rally/storage-mirrors.json`` file::
+
+    {
+        "mirrors": [
+            {
+                "sources": [
+                    "https://rally-tracks.elastic.co/"
+                ],
+                "destinations": [
+                  "s3://rally-tracks-eu-central-1",
+                  "s3://rally-tracks-us-west-1/"
+                ]
+              }
+            ]
+          }
+    }
+
+Configuration options:
+
+* ``storage.aws.profile`` is used to specify the profile name to be used for connecting to the S3 service. By default
+  it will use credentials detected from the environment as specified by 'boto3' client.
+
+  .. _S3 cloud storage service: https://aws.amazon.com/es/s3/
 
 tracks
 ~~~~~~
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -140,6 +140,8 @@ remote servers. Available options are:
 * ``storage.max_connections`` represents the maximum number of client connections to be made against the same server or
   bucket. The default value is 8.
 
+* ``storage.max_workers`` indicates the maximum number of worker threads used for making storage files transfers.
+
 * ``storage.mirror_files`` is used to provide a json file that specify the mapping for mirrors URLs resolution.
   Example::
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -119,21 +119,21 @@ Define a secure connection to an Elastic Cloud cluster::
 storage
 ~~~~~~~
 
-This section defines how client is configured to transfer big files (I.E. track files) from remote servers. The main
+This section defines how client is configured to transfer corpus files. The main
 advantages of this downloader implementation are:
 
-* It supports configuring mirrors servers to download files from there instead of the original source URL. There could
-  be multiple services from where download files. The client will load balance between these services giving priority
-  to those servers with the lower latency. In case of failures reaching the mirrors services it will download files
-  from the original source server.
-* It supports multipart downloading of pieces of files from multiple servers at the same time, increasing the impact of
-  a failing network infrastructure. Those parts that experience errors downloading will be eventually re-downloaded from
-  another server of from the original source.
-* It supports local caching of files, and file downloading resuming from previous status in case of connection
-  interruption. It can also resume file downloading between consecutive rally executions.
+* It supports configuring multiple mirror URLs. The client will load balance between these URLs giving priority to
+  those with the lower latency. In case of failures the client will download files from the original URL.
+* It supports multipart downloading of pieces of files from multiple servers at the same time, mitigating the impact of
+  network failures. Those file parts that experience errors downloading will be eventually re-downloaded from
+  another server or, as the very last resource, from the original source.
+* It supports local caching of files and it can resume downloads when connections are interrupted. The download state
+  is preserved between Rally executions, allowing downloads to continue from where they left off rather than starting
+  over.
 
-*NOTE*: This transfers manager implementation is still experimental and under active development. It is not used yet by
-default and it will require some additional works before final delivery.
+.. warning::
+
+    This transfers manager implementation is experimental and under active development.
 
 Configuration options are:
 
@@ -186,14 +186,15 @@ Configuration options are:
   are not mirrored or they have a different size from the source one. The client will prefer endpoints with the lower
   latency fetching the head of the file.
 
-* ``storage.monitor_interval`` represents the time interval (in seconds) `TransferManager` should wait two consecutive
+* ``storage.monitor_interval`` represents the time interval (in seconds) `TransferManager` should wait for consecutive
   monitor operations (log transfer and connections statistics, adjust the maximum number of connections, etc.).
 
-* ``storage.multipart_size`` represents the size in bytes to be used for making file parts separation to distribute
-  them to worker threads. Each part will be downloaded separately and in parallel using a dedicated connection by a
-  worker threads. In case there will be more parts that the maximum allowed connections and threads (see
-  ``storage.max_workers`` and ``storage.max_connections`` options), then the transfer of the parts exceeding these limits
-  will be performed as soon a worker thread or a connection get released by other part transfer.
+* ``storage.multipart_size`` When the file size measured in bytes is greater than this value the file is split in chunk
+  of this size plus the last one ()that could be smaller). Each part will be downloaded separately and in parallel using
+  a dedicated connection by a worker thread and eventually from a different mirror server to load balance the network
+  traffic between multiple servers. If the resulting number of parts is greater than ``storage.max_workers`` and
+  ``storage.max_connections`` options, then the transfer of those parts exceeding these limits will be performed as
+  soon as a worker thread gets available or a HTTP connection get released by another thread.
 
 * ``storage.random_seed`` a string used to initialize the client random number generator. This could be used to make
   problems easier to reproduce in continuous integration. In most of the cases it should be left empty.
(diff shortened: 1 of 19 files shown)


esrally/storage/_adapter.py
```diff
@@ -105,7 +90,7 @@ def match_url(cls, url: str) -> bool:
         raise NotImplementedError
 
     @classmethod
-    def from_config(cls, cfg: Config) -> Adapter:
+    def from_config(cls: type[A], cfg: Config, **kwargs: dict[str, Any]) -> A:
```
favilo: I think `typing_extensions` has a `Self` type that works exactly like this for any class. So you can just write `def from_config(cls, cfg: Config, **kwargs: Any) -> Self:`

Also, `**kwargs` like this is typed by the value type only. [Reference](https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values)

esrally/storage/_aws.py
```diff
@@ -0,0 +1,162 @@
...
+AWS_PROFILE: str | None = None
+
+A = TypeVar("A", "S3Adapter", "S3Adapter")
+
+
+class S3Adapter(HTTPAdapter):
+    """Adapter class for s3:// scheme protocol"""
+
+    @classmethod
+    def from_config(cls: type[A], cfg: Config, **kwargs: dict[str, Any]) -> A:
```
favilo: Ditto from above.

esrally/storage/_aws.py
```diff
@@ -0,0 +1,162 @@
...
+        assert issubclass(cls, S3Adapter)
```
favilo: Is this necessary? I'm not sure how to call this method without it being a subclass. I would love to have an example of how to accomplish that.

esrally/storage/_aws.py
```diff
@@ -0,0 +1,162 @@
...
+        session: requests.Session | None = None,
+        chunk_size: int = CHUNK_SIZE,
+    ) -> None:
+        super().__init__(session=session, chunk_size=chunk_size)
+        self.aws_profile = aws_profile
+        self._s3_client = s3_client
+
+    @classmethod
+    def match_url(cls, url: str) -> bool:
+        return url.startswith("s3://")
```
favilo, line 67: Since we do support `https` down in `S3Address` below, should we use `S3Address.from_url` here and just return false on caught exception?

esrally/storage/_http.py
```diff
@@ -112,36 +116,33 @@ def get(self, url: str, stream: Writable, head: Head | None = None, headers: Map
...
+
     @classmethod
-    def _ranges_to_headers(cls, ranges: RangeSet, headers: CaseInsensitiveDict) -> None:
+    def _ranges_to_headers(cls, ranges: RangeSet, headers: MutableMapping[str, str]) -> None:
         if not ranges:
             return
         if len(ranges) > 1:
             raise NotImplementedError(f"unsupported multi range requests: ranges are {ranges}")
-        headers["range"] = f"bytes={ranges[0]}"
+        headers[cls._RANGE_HEADER] = f"bytes={ranges[0]}"
```
favilo: Hmm, do we want this to be hard coded to the first range in this set? How is this going to be used in the future?

tests/storage/aws_test.py
```diff
@@ -0,0 +1,229 @@
...
+class HeadCase:
+    response: dict[str, Any]
+    want: Head
+    url: str = SOME_URL
+    want_bucket: str = SOME_BUCKET
+    want_key: str = SOME_KEY
+
+
+@pytest.fixture
+def s3_client():
```
favilo: Could you include a type annotation?

favilo (commented):
Left some comments and questions

esrally/storage/_adapter.py
fressi-elastic: Cool!

esrally/storage/_http.py
fressi-elastic: The range set can only contain one or zero ranges. More than one ranges will never be supported as it is not supported from S3 service servers. The protocol would be too complex to implement and it is not necessary as `Transfer` class will distribute part to downloads to multiple available storage services.

esrally/storage/_aws.py
fressi-elastic: The purpose of this adapter is only the integration with the python `boto3/S3` client. It the user want to use the raw `https` protocol without the S3 client it should use the plain `HTTPAdapter` instead. So it will be simpler to identify bugs location according to the type of the URL used. I would like to move out of HTTPAdapter to a shared base class those methods that are used by S3Adapter, to avoid confusion and unexpected interferences between the two implementations.

docs/configuration.rst
```diff
@@ -119,8 +119,23 @@ Define a secure connection to an Elastic Cloud cluster::
 storage
 ~~~~~~~
 
-This section defines how client is configured to transfer big files (I.E. track files) from
-remote servers. Available options are:
+This section defines how client is configured to transfer big files (I.E. track files) from remote servers. The main
+advantages of this downloader implementation are:
```
gbanasiak:
```suggestion
This section defines how client is configured to transfer corpus files. The main
advantages of this downloader implementation are:
```

docs/configuration.rst
```diff
@@ -119,8 +119,23 @@ Define a secure connection to an Elastic Cloud cluster::
...
+
+* It supports configuring mirrors servers to download files from there instead of the original source URL. There could
+  be multiple services from where download files. The client will load balance between these services giving priority
+  to those servers with the lower latency. In case of failures reaching the mirrors services it will download files
+  from the original source server.
```
gbanasiak:
```suggestion

* It supports configuring multiple mirror URLs. The client will load balance between these URLs giving priority to those with the lower latency. In case of failures the client will download files from the original URL.
```

docs/configuration.rst
```diff
@@ -119,8 +119,23 @@ Define a secure connection to an Elastic Cloud cluster::
...
+* It supports multipart downloading of pieces of files from multiple servers at the same time, increasing the impact of
+  a failing network infrastructure. Those parts that experience errors downloading will be eventually re-downloaded from
+  another server of from the original source.
```
gbanasiak: What do you mean by "increasing the impact of a failing network infrastructure"? If this is a potential drawback, let's move it to "disadvantages" list.

docs/configuration.rst
```diff
@@ -119,8 +119,23 @@ Define a secure connection to an Elastic Cloud cluster::
...
+* It supports local caching of files, and file downloading resuming from previous status in case of connection
+  interruption. It can also resume file downloading between consecutive rally executions.
```
gbanasiak:
```suggestion
* It supports local caching of files and can resume downloads when connections are interrupted. The download state is preserved between Rally executions, allowing downloads to continue from where they left off rather than starting over.
```

docs/configuration.rst
```diff
@@ -119,8 +119,23 @@ Define a secure connection to an Elastic Cloud cluster::
...
+
+*NOTE*: This transfers manager implementation is still experimental and under active development. It is not used yet by
+default and it will require some additional works before final delivery.
+
```
gbanasiak: We use the following documentation call-out in other places to signal experimental / under-development features.

```suggestion

.. warning::

    This transfers manager implementation is experimental and under active development.

```

docs/configuration.rst
```diff
@@ -169,13 +186,24 @@ remote servers. Available options are:
   are not mirrored or they have a different size from the source one. The client will prefer endpoints with the lower
   latency fetching the head of the file.
 
+* ``storage.monitor_interval`` represents the time interval (in seconds) `TransferManager` should wait two consecutive
+  monitor operations (log transfer and connections statistics, adjust the maximum number of connections, etc.).
```
gbanasiak:
```suggestion
* ``storage.monitor_interval`` represents the time interval (in seconds) `TransferManager` should wait for consecutive
  monitor operations (log transfer and connections statistics, adjust the maximum number of connections, etc.).
```

docs/configuration.rst
```diff
@@ -169,13 +186,24 @@ remote servers. Available options are:
...
+
+* ``storage.multipart_size`` represents the size in bytes to be used for making file parts separation to distribute
+  them to worker threads. Each part will be downloaded separately and in parallel using a dedicated connection by a
```
gbanasiak: The "file parts separation" is unclear. Can you rephrase?

docs/configuration.rst
```diff
@@ -193,6 +221,40 @@ HTTP Adapter
   .. _urllib3.Retry: https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html
 
 
+S3 Adapter
+**********
+
+This adapter can be used only to download files from `S3 cloud storage service`_. It requires 'boto3' client library to be
+installed and it accepts only URLs with the following format ``s3://<bucket-name>/<object-key>``. It is intended to be
```
gbanasiak: This sounds like an extra step is needed to use S3 adapter (`boto3` client library install), which is not the case because `boto3` is simply one of Rally dependencies, and gets installed automatically.

esrally/storage/_aws.py
```diff
@@ -0,0 +1,175 @@
...
+        if u.scheme not in ("s3", "https"):
+            raise ValueError(f"invalid URL scheme '{url}'")
+
+        if u.scheme == "s3":
+            bucket = u.netloc
+        elif u.scheme == "https":
+            bucket, right = u.netloc.split(".s3.", 1)
+            if not right.endswith("amazonaws.com"):
+                raise ValueError(f"https URL doesn't ends with 'amazonaws.com': '{url}'")
+            region = right[: -len("amazonaws.com")].rstrip(".")
```
gbanasiak, line 121: Do we need `u.scheme == "https"` block at all?

esrally/storage/_aws.py
```diff
@@ -0,0 +1,175 @@
...
+        return S3Address(bucket=bucket, key=key, region=region)
+
+    def host(self, scheme: str = "s3") -> str:
+        if scheme == "s3":
+            return self.bucket
+        elif scheme == "https":
+            if self.region:
+                return f"{self.bucket}.s3.{self.region}.amazonaws.com"
+            else:
+                return f"{self.bucket}.s3.amazonaws.com"
```
gbanasiak, line 137: Same here, if this adapter is only for S3, do we need `scheme == "https"` block ?

esrally/storage/_aws.py
```diff
@@ -0,0 +1,175 @@
...
+        ranges=ranges,
+        document_length=document_length,
+        crc32c=crc32,
+    )
+
+
+@runtime_checkable
+class S3Client(Protocol):
+
+    def head_object(self, Bucket: str, Key: str) -> Mapping[str, Any]:
```
gbanasiak: Why `NotImplementedError` instead of ellipsis (`...`)?

esrally/storage/_manager.py
```diff
@@ -0,0 +1,200 @@
...
+        with self._lock:
+            transfers = self._transfers
+            # It removes finished transfers and update max connections
+            self._update_transfers()
+        if transfers:
+            LOG.info("Transfers in progress:\n  %s", "\n  ".join(tr.info() for tr in transfers))
+        self._client.monitor()
+
+    def _update_transfers(self):
+        available_workers = max(1, int(self._max_workers * 0.8))
```
gbanasiak: Why 0.8?

esrally/storage/_manager.py
```diff
@@ -0,0 +1,200 @@
...
+        self._transfers = transfers = [tr for tr in self._transfers if not tr.finished]
+        if transfers:
+            max_connections = min(self._max_connections, max(1, available_workers // len(transfers)))
```
gbanasiak: This bit is a bit puzzling. When we initiate a transfer in `_transfer()`, the transfer's maximum number of connections is set to manager's max connections, but here we reduce the number of maximum connections per transfer to the number of available workers divided by the number of transfers. So `storage.max_connections` is ultimately the total maximum number of connections across all transfers / workers not per transfer, correct? Can we reflect that in documentation?

tests/storage/aws_test.py
```diff
@@ -0,0 +1,185 @@
...
+@dataclass()
+class PutCase:
+    response: dict[str, Any]
+    want: Head
+    url: str = SOME_URL
+    ranges: str = ""
+    content_length: int | None = None
+    want_bucket: str = SOME_BUCKET
+    want_key: str = SOME_KEY
+    want_range: str = ""
```
gbanasiak: Not needed.

esrally/storage/_manager.py
```diff
@@ -0,0 +1,200 @@
...
+from esrally.storage._executor import MAX_WORKERS, Executor, ThreadPoolExecutor
+from esrally.storage._transfer import Transfer
+from esrally.utils.threads import ContinuousTimer
+
+LOG = logging.getLogger(__name__)
+
+LOCAL_DIR = "~/.rally/storage"
+MONITOR_INTERVAL = 2.0  # Seconds
+THREAD_NAME_PREFIX = "esrally.storage.transfer-worker"
+MULTIPART_SIZE = 8 * 1024 * 1024
```
gbanasiak: We also have a multipart size default in `_transfer.py`, do we need both?

gbanasiak (commented):
This looks good. I left a few comments and suggested some documentation changes.

docs/configuration.rst
fressi-elastic: The sentence is wrong. `increasing` should be `mitigating`

docs/configuration.rst
fressi-elastic: boto3 is an optional dependency. This statement states that in case it is not installed, the user should better use the HTTP adapter.

esrally/storage/_aws.py
fressi-elastic: Ooohhh

esrally/storage/_manager.py
fressi-elastic: I would like to keep some extra threads always available for other tasks that are not network transfer. Anyway this behavior is not documented and will mostly get the user by surprise so I am going to remove it.

docs/configuration.rst
gbanasiak: Yes, my mistake. Can we link to https://esrally.readthedocs.io/en/stable/install.html?highlight=s3#optional-dependencies to guide users how to install `boto3`?

esrally/storage/_manager.py
fressi-elastic: Pending tasks:
- [x] create new transfers with an up to date value for `max_connections` attribute. 
- [x] fix configuration documentation related to `max_connections` option.

esrally/storage/_manager.py
fressi-elastic: We need only one definition for both locations.

docs/configuration.rst
```diff
@@ -169,13 +189,25 @@ remote servers. Available options are:
   are not mirrored or they have a different size from the source one. The client will prefer endpoints with the lower
   latency fetching the head of the file.
 
+* ``storage.monitor_interval`` represents the time interval (in seconds) `TransferManager` should wait for consecutive
+  monitor operations (log transfer and connections statistics, adjust the maximum number of connections, etc.).
+
+* ``storage.multipart_size`` When the file size measured in bytes is greater than this value the file is split in chunk
+  of this size plus the last one ()that could be smaller). Each part will be downloaded separately and in parallel using
```
gbanasiak, line 196:
```suggestion
  of this size plus the last one (that could be smaller). Each part will be downloaded separately and in parallel using
```

docs/configuration.rst
```diff
@@ -169,13 +189,25 @@ remote servers. Available options are:
...
+  a dedicated connection by a worker thread and eventually from a different mirror server to load balance the network
+  traffic between multiple servers. If the resulting number of parts is greater than ``storage.max_workers`` and
+  ``storage.max_connections`` options, then the transfer of those parts exceeding these limits will be performed as
```
gbanasiak, line 199:
```suggestion
  ``storage.max_connections`` options, then the transfer of parts exceeding these limits will be performed as
```

docs/configuration.rst
```diff
@@ -193,6 +225,54 @@ HTTP Adapter
   .. _urllib3.Retry: https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html
 
 
+S3 Adapter
+**********
+
+This adapter can be used only to download files from `S3 Cloud Storage Service`_. It is intended to be used to configure
+S3 buckets as mirror for track file downloads.
+
+It requires `Boto3 Client`_ to be installed and it accepts only URLs with the following format::
```
gbanasiak, line 234: I think this is too generic. We have specific instructions for `boto3` installation: https://github.com/elastic/rally/blob/5b0ca49fa293014424cfc7200db322a15f2cf7d5/docs/install.rst?plain=1#L101-L104

which makes use of: https://github.com/elastic/rally/blob/5b0ca49fa293014424cfc7200db322a15f2cf7d5/pyproject.toml#L97-L105

We want specific `boto3` version, not just anything which can be found in generic AWS documentation.

gbanasiak (approved):
LGTM. I've left some non-critical comments.

PR #1966, opened 2025-07-21
Esrally/stop params/fix #1957

NickDris:
Load skip.telemetry from cli arguments
- #1957 was not loading skip.telemetry parameter correctly leaving 
skip.telemetry unused even if it was set as a cli flag.
- There is no value loading skip.telemetry parameter from cfg file
since setting this parameter statically to true is not a plausible option, 
so it should be set to true only when given in the cli args.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -67,7 +67,6 @@ This section defines how metrics are stored.
 * ``sample.queue.size`` (default: 2^20): The number of metrics samples that can be stored in Rally's in-memory queue.
 * ``metrics.request.downsample.factor`` (default: 1): Determines how many service time and latency samples should be kept in the metrics store. By default all values will be kept. To keep only e.g. every 100th sample, specify 100. This is useful to avoid overwhelming the metrics store in benchmarks with many clients (tens of thousands).
 * ``output.processingtime`` (default: false): If set to "true", Rally will show the additional metric :ref:`processing time <summary_report_processing_time>` in the command line report.
-* ``skip.telemetry`` (default: None): Determines whether telemetry data collection should be skipped when stopping a benchmark. Can be set through the command line flag ``--skip-telemetry``.
 
 The following settings are applicable only if ``datastore.type`` is set to "elasticsearch":
 
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -67,7 +67,7 @@ This section defines how metrics are stored.
 * ``sample.queue.size`` (default: 2^20): The number of metrics samples that can be stored in Rally's in-memory queue.
 * ``metrics.request.downsample.factor`` (default: 1): Determines how many service time and latency samples should be kept in the metrics store. By default all values will be kept. To keep only e.g. every 100th sample, specify 100. This is useful to avoid overwhelming the metrics store in benchmarks with many clients (tens of thousands).
 * ``output.processingtime`` (default: false): If set to "true", Rally will show the additional metric :ref:`processing time <summary_report_processing_time>` in the command line report.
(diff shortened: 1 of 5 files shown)


docs/configuration.rst
```diff
@@ -67,7 +67,7 @@ This section defines how metrics are stored.
 * ``sample.queue.size`` (default: 2^20): The number of metrics samples that can be stored in Rally's in-memory queue.
 * ``metrics.request.downsample.factor`` (default: 1): Determines how many service time and latency samples should be kept in the metrics store. By default all values will be kept. To keep only e.g. every 100th sample, specify 100. This is useful to avoid overwhelming the metrics store in benchmarks with many clients (tens of thousands).
 * ``output.processingtime`` (default: false): If set to "true", Rally will show the additional metric :ref:`processing time <summary_report_processing_time>` in the command line report.
-* ``skip.telemetry`` (default: None): Determines whether telemetry data collection should be skipped when stopping a benchmark. This can be set either via configuration param ``skip.telemetry`` the command line flag ``--skip-telemetry``.
+* ``skip.telemetry`` (default: None): Determines whether telemetry data collection should be skipped when stopping a benchmark. Can be set through the command line flag ``--skip-telemetry``.
```
fressi-elastic: *It can be set...

fressi-elastic (approved):
Please look at the typo.

docs/configuration.rst
gbanasiak: `skip.telemetry` is added under `mechanic` configuration section, not `reporting`, but documenting it under `rally.ini` configuration is not correct as we are always setting it based on `--skip-telemetry` flag now. Please remove.

docs/recipes.rst
```diff
@@ -328,6 +328,6 @@ When a benchmark is executed with ``--enable-assertions`` and this query returns
 Skipping telemetry data collection when hitting an error.
 --------------------------------------------------------------
 
-In some cases, you may want to skip telemetry data collection when Rally hits an error. This can be useful in cases you want to avoid storing telemetry results when a benchmark got unexpected results. To enable this behavior, you can either use the ``skip.telemetry`` configuration setting in your ``rally.ini`` file or pass the ``--skip-telemetry`` command line flag when stopping Rally. This will prevent Rally from sending telemetry data to the metrics store. 
+In some cases, you may want to skip telemetry data collection when Rally hits an error. This can be useful in cases you want to avoid storing telemetry results when a benchmark got unexpected results . To enable this behavior, you can pass the ``--skip-telemetry`` command line flag when executing `esrally stop`. This will prevent Rally from sending telemetry data to the metrics store.
```
gbanasiak: I would remove this section too, because there's not enough context here (i.e. skipping telemetry only makes sense when Rally installs, starts, and stops ES separately from running a race).

It's enough to document it in:
* command reference https://esrally--1966.org.readthedocs.build/en/1966/command_line_reference.html#skip-telemetry
* cluster management https://esrally--1966.org.readthedocs.build/en/1966/cluster_management.html

gbanasiak (commented):
I've tested this, and it seems to work fine now. I've left some comments regarding documentation.

docs/recipes.rst
gbanasiak: Done in https://github.com/elastic/rally/pull/1966/commits/2a4bd8d14120346d548d4cb8fb1456262bad046d.

docs/configuration.rst
gbanasiak: Done in https://github.com/elastic/rally/pull/1966/commits/2a4bd8d14120346d548d4cb8fb1456262bad046d.

PR #1967, opened 2025-07-21
Removing negative duration constraint, it might be needed in comparing results

AI-IshanBhatt:
It removes -ve duration constraint.
Currently it is not possible to compare races where contender is having smaller duration in some cases.


gbanasiak (approved):
LGTM

PR #1971, opened 2025-07-24
Support API key authentication for reporting

### LINKED ISSUE #1866: Support Rally metric store in serverless project
Rally supports writing metrics to Elasticsearch metric store. This could be enhanced to support writing to a serverless project.

inqueue:
Add support for API key authentication to reporting. Can be utilized for any Elasticsearch on-prem, Cloud, or Cloud Serverless target that supports API key authentication.

Closes https://github.com/elastic/rally/issues/1866 ES-12479


esrally/metrics.py
```diff
@@ -233,24 +233,45 @@ def __init__(self, cfg: types.Config):
...
+        if user:
+            try:
+                password = os.environ["RALLY_REPORTING_DATASTORE_PASSWORD"]
+            except KeyError:
+                try:
+                    password = self._config.opts("reporting", "datastore.password")
+                except exceptions.ConfigError:
+                    raise exceptions.ConfigError(
+                        "No password configured through [reporting] configuration or "
+                        "RALLY_REPORTING_DATASTORE_PASSWORD environment variable."
```
fressi-elastic: I do prefer a coding style without exception handling where is not necessary as here. For example on this case I would write:

```
from __future__ import annotations

# ...

DATASTORE_PASSWORD: str = os.environ.get("RALLY_REPORTING_DATASTORE_PASSWORD", "")

# ...

  if user:
    password: str =  DATASTORE_PASSWORD or self._config.opts("reporting", "datastore.password", default_value="", mandatory=False)
   if not password:
     raise exceptions.ConfigError(
         "No password configured through [reporting] configuration or "
         "RALLY_REPORTING_DATASTORE_PASSWORD environment variable."
     )
```

esrally/metrics.py
```diff
@@ -233,24 +233,45 @@ def __init__(self, cfg: types.Config):
...
+            api_key = os.environ["RALLY_REPORTING_DATASTORE_API_KEY"]
         except KeyError:
             try:
-                password = self._config.opts("reporting", "datastore.password")
+                api_key = self._config.opts("reporting", "datastore.api_key", mandatory=False)
             except exceptions.ConfigError:
-                raise exceptions.ConfigError(
-                    "No password configured through [reporting] configuration or RALLY_REPORTING_DATASTORE_PASSWORD environment variable."
-                ) from None
+                api_key = None
```
fressi-elastic: I found code complexity here is not necessary. Please have a look to my previous comment about the password configuration.

esrally/metrics.py
```diff
@@ -233,24 +233,45 @@ def __init__(self, cfg: types.Config):
...
+
+        # Enforce that one of (user and password) or api_key is provided, but not both
+        if (user and password) and api_key:
```
fressi-elastic: Parenthesis here make any difference in the code. You can remove them. Anyway when user is set then password is set because of previous check. Therefore the following expression is already good:

`if user and api_key`

Even better, because you are already checking `if user` above, please handle api_key configuration before user configuration and then raise this in the same conditional block before handling the password. Example:

```
# in the global section:
DATASTORE_API_KEY: str = os.environ.get("...", "")
DATASTORE_API_USER: str = os.environ.get("...", "")
DATASTORE_API_PASSWORD: str = os.environ.get("...", "")

  # In this function:
  api_key = DATASTORE_API_KEY or self._config.opts("reporting", "datastore.api_key", default_value="", mandatory=False)
  user = DATASTORE_USER or self._config.opts("reporting", "datastore.user", default_value="", mandatory=False)
  if user:
    if api_key:
        raise ConfigError("...")
    password = DATASTORE_PASSWORD or self._config.opts("reporting", "datastore.password", default_value="", mandatory=False)
    if not password:
       raise ConfigError("...")
```

The code would result much more compact and easy to read on my opinion.

esrally/metrics.py
```diff
@@ -261,6 +282,8 @@ def __init__(self, cfg: types.Config):
         if user and password:
```
fressi-elastic: I would only check the user here, as the password should be already checked to be not empty when the user is not.

tests/metrics_test.py
```diff
@@ -211,6 +211,103 @@ def test_config_opts_parsing(self, client_esclientfactory, password_configuratio
...
 
+    @pytest.mark.parametrize("apikey_configuration", ["config", "environment"])
+    @mock.patch("esrally.client.EsClientFactory")
+    def test_config_opts_parsing_apikey(self, client_esclientfactory, apikey_configuration, monkeypatch):
+        cfg = config.Config()
+
+        _datastore_host = ".".join([str(random.randint(1, 254)) for _ in range(4)])
+        _datastore_port = random.randint(1024, 65535)
+        _datastore_secure = random.choice(["True", "true"])
+        _datastore_apikey = "".join([random.choice(string.ascii_letters + string.digits + "_-@#$/") for _ in range(60)])
```
fressi-elastic: We should test both, not only one. with not doing it we potentially have 50% of chances to have a false success in case of a refactory of the code and having then it merged if unlucky. This thing of using random values instead of testing fixed value scenario I saw it in many places in this company and it is something that I think makes test cases less reliable. We shouldn't consider a condition tested if in a range of random values. It makes the test case randomly valid, instead of some reliable source of true.

tests/metrics_test.py
```diff
@@ -211,6 +211,103 @@ def test_config_opts_parsing(self, client_esclientfactory, password_configuratio
...
+            distribution_flavor=None,
+        )
+
+    @mock.patch("esrally.client.EsClientFactory")
+    def test_config_opts_parsing_both_password_apikey(self, client_esclientfactory, monkeypatch):
+        cfg = config.Config()
+
+        _datastore_host = ".".join([str(random.randint(1, 254)) for _ in range(4)])
+        _datastore_port = random.randint(1024, 65535)
+        _datastore_secure = random.choice(["True", "true"])
```
fressi-elastic: I have the same concerns related to the use of random values as in the previous comments.

tests/metrics_test.py
```diff
@@ -211,6 +211,103 @@ def test_config_opts_parsing(self, client_esclientfactory, password_configuratio
...
+        _datastore_user = "".join([random.choice(string.ascii_letters) for _ in range(8)])
```
fressi-elastic: Why not to check for `"foo"` or `"some user"` or `"toodles"`... I generally like simple values with a very easy to spot purpose instead of expressions like this.

fressi-elastic (changes requested):

tests/metrics_test.py
inqueue: > We should test both, not only one. with not doing it we potentially have 50% of chances to have a false success in case of a refactory of the code and having then it merged if unlucky.

Can you elaborate on "test both"? Thanks

esrally/metrics.py
```diff
@@ -228,39 +228,50 @@ class EsClientFactory:
     """
 
     def __init__(self, cfg: types.Config):
+        DATASTORE_API_KEY: str = os.environ.get("RALLY_REPORTING_DATASTORE_API_KEY", "")
```
fressi-elastic: I would put these two definitions in a global section, for example before of this class definition.

esrally/metrics.py
```diff
@@ -228,39 +228,50 @@ class EsClientFactory:
...
+        user = self._config.opts("reporting", "datastore.user", mandatory=False)
         distribution_version = None
         distribution_flavor = None
-        try:
-            password = os.environ["RALLY_REPORTING_DATASTORE_PASSWORD"]
-        except KeyError:
-            try:
-                password = self._config.opts("reporting", "datastore.password")
-            except exceptions.ConfigError:
+        api_key: str = DATASTORE_API_KEY or self._config.opts("reporting", "datastore.api_key", mandatory=False)
```
fressi-elastic: The same as for `user` variable definition.

esrally/metrics.py
```diff
@@ -228,39 +228,50 @@ class EsClientFactory:
...
+        password: str = DATASTORE_PASSWORD or self._config.opts("reporting", "datastore.password", mandatory=False)
```
fressi-elastic: The same as for `user` variable definition.

esrally/metrics.py
```diff
@@ -228,39 +228,50 @@ class EsClientFactory:
...
+        DATASTORE_API_KEY: str = os.environ.get("RALLY_REPORTING_DATASTORE_API_KEY", "")
+        DATASTORE_PASSWORD: str = os.environ.get("RALLY_REPORTING_DATASTORE_PASSWORD", "")
         self._config = cfg
         host = self._config.opts("reporting", "datastore.host")
         port = self._config.opts("reporting", "datastore.port")
         hosts = [{"host": host, "port": port}]
         secure = convert.to_bool(self._config.opts("reporting", "datastore.secure"))
-        user = self._config.opts("reporting", "datastore.user")
+
+        user = self._config.opts("reporting", "datastore.user", mandatory=False)
```
fressi-elastic: If the value is not found in the configuration, then `user` variable would assume the value `None` that is not of type `str`. To ensure this doesn't happen, add parameter `default_value=""` to this call (preferred), or in alternative use the exact type in the annotation like the following:

```
user: str | None = ...
api_key: str | None = ...
password: str | None = ...
```

In this way when running static analysis (I.E. running `make lint`) it would have better hints to help us catching bugs before we merge our code.

fressi-elastic (approved):

fressi-elastic (approved):

PR #1972, opened 2025-07-30
[ES-12520] Add --user-tags to list races

gbanasiak:
This allows to filter races based on arbitrary user tags. It is a generalization of `--benchmark-name` option which matches either `name` or `benchmark-name` tags today.

Additionally, list races command gets new `--format {text,json}` option to make automated parsing easier. The JSON format provides full race document as persisted in the metric store.


AI-IshanBhatt (approved):
LGTM

PR #1975, opened 2025-08-14
Allow configuring the process startup method to be used for creating thespian actors.

fressi-elastic:
The main purpose for this is to be able to prevent using the fork function to be able to use threads for multipart downloads.

### DIFF
--- a/esrally/actor.py
+++ b/esrally/actor.py
@@ -17,18 +17,14 @@
 from __future__ import annotations
 
 import logging
-import os
 import socket
 import traceback
-from typing import Any
 
 import thespian.actors  # type: ignore[import-untyped]
 import thespian.system.messages.status  # type: ignore[import-untyped]
 
 from esrally import exceptions, log
 from esrally.utils import console, net
-
-LOG = logging.getLogger(__name__)
 
 
 class BenchmarkFailure:
@@ -97,7 +93,7 @@ def no_retry(f, actor_name):
             return f(self, msg, sender)
         except BaseException:
             # log here as the full trace might get lost.
-            LOG.exception("Error in %s", actor_name)
+            logging.getLogger(__name__).exception("Error in %s", actor_name)
             # don't forward the exception as is because the main process might not have this class available on the load path
             # and will fail then while deserializing the cause.
             self.send(sender, BenchmarkFailure(traceback.format_exc()))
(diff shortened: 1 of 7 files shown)


esrally/actor.py
```diff
@@ -193,71 +199,115 @@ def is_current_status_expected(self, expected_status):
...
-            else:
-                logger.debug("Creating new actor system with system base [%s] on coordinator node.", system_base)
-                # if we try to join we can only run on the coordinator...
-                return thespian.actors.ActorSystem(system_base, logDefs=log.load_configuration(), capabilities={"coordinator": True})
+            if system_base == "multiprocTCPBase" and actor_system_already_running(
+                ip=coordinator_ip, port=coordinator_port, system_base=system_base
+            ):
+                if coordinator_ip:
+                    coordinator_ip = net.resolve(coordinator_ip)
+                    coordinator_port = coordinator_port or 1900
```
gbanasiak: Can we move the `1900` default to method arguments? We're specifying it twice now I think.

esrally/actor.py
```diff
@@ -193,71 +199,115 @@ def is_current_status_expected(self, expected_status):
...
                 raise exceptions.SystemSetupError("local IP is required")
-            # always resolve the public IP here, even if a DNS name is given. Otherwise Thespian will be unhappy
-            local_ip = net.resolve(local_ip)
-            coordinator_ip = net.resolve(coordinator_ip)
-
-            coordinator = local_ip == coordinator_ip
+            # always resolve the public IP here, even if a DNS name is given, otherwise Thespian will be unhappy
 
-        capabilities = {"coordinator": coordinator}
+        # if we try to join we can only run on the coordinator...
```
gbanasiak: This comment does not make sense after the refactor I think. The code that follows applies both to coordinator and non-coordinator cases.

esrally/rally.py
```diff
@@ -960,9 +962,22 @@ def race(cfg: types.Config, kill_running_processes=False):
     with_actor_system(racecontrol.run, cfg)
 
 
+def actor_system_already_running() -> bool:
+    try:
+        return actor.actor_system_already_running()
+    except ValueError:
+        return False
```
gbanasiak: What's the benefit of catching an exception here instead of not throwing it in `actor.actor_system_already_running()` in the first place? Unit tests only? Looks like we are only calling `actor.actor_system_already_running()` via these wrappers, and we need to define them twice, both in `rally.py` and rallyd.py`.

gbanasiak (approved):
LGTM. I've left some minor comments and questions, but nothing blocking.

esrally/rally.py
fressi-elastic: I removed both wrappers

PR #1977, opened 2025-09-17
Fix ITs after pytest-rally plugin modification

gbanasiak:
Aligns Rally with https://github.com/elastic/pytest-rally/pull/5.

### DIFF
--- a/it/track_repo_compatibility/conftest.py
+++ b/it/track_repo_compatibility/conftest.py
@@ -28,18 +28,6 @@ REMOTE_TRACK_REPO = "https://github.com/elastic/rally-tracks"
 def pytest_addoption(parser):
     group = parser.getgroup("rally")
     group.addoption(
-        "--track-repository",
-        action="store",
-        default=TRACK_REPO_PATH,
-        help=f"Path to a local track repository\n(default: {TRACK_REPO_PATH})",
-    )
-    group.addoption(
-        "--track-revision",
-        action="store",
-        default="master",
-        help="Track repository revision to test\ndefault: `master`",
-    )
-    group.addoption(
         "--track-repository-test-directory",
         action="store",
         dest="track_repo_test_dir",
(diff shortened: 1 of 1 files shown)


it/track_repo_compatibility/conftest.py
```diff
@@ -50,7 +38,12 @@ def pytest_addoption(parser):
 
 
 def pytest_cmdline_main(config):
+    # override the default track repository and revision if not provided by the user
     repo = config.option.track_repository
+    if repo is None:
+        repo = config.option.track_repository = TRACK_REPO_PATH
+    if config.option.track_revision is None:
+        config.option.track_revision = "master"
```
gbanasiak: We can probably skip this one as plugin default will work?

gbanasiak: Changed in https://github.com/elastic/rally/pull/1977/commits/8c91c9d7c868126bbf03a36c7ef5e9125c911bf5.

it/track_repo_compatibility/conftest.py
```diff
@@ -50,7 +38,10 @@ def pytest_addoption(parser):
 
 
 def pytest_cmdline_main(config):
+    # override the default track repository if not provided by the user
     repo = config.option.track_repository
+    if repo is None:
+        repo = config.option.track_repository = TRACK_REPO_PATH
```
NickDris, line 44: I saw that you deleted the track_revision modification
``` 
if config.option.track_revision is None: 
    config.option.track_revision = "master"
```
while in pytest-rally this default value is [None](https://github.com/elastic/pytest-rally/blob/e58fa6bd18f045746bd1a4528564d10b73b28666/pytest_rally/plugin.py#L63-L66), which leaves us with the wrong value?
Maybe it would make sense to change the default track_revision value from pytest-rally since most users of pytest-rally have a 'master' or a 'main' branch?

NickDris (changes requested):
The default value of track_revision of pytest-rally is _None_. So in order to maintain the former functionality we should change be adding 'master' either on pytest-rally or in conftest.py

it/track_repo_compatibility/conftest.py
gbanasiak: The default value in the plugin is the current branch: https://github.com/elastic/pytest-rally/blob/e58fa6bd18f045746bd1a4528564d10b73b28666/pytest_rally/plugin.py#L84-L86. In newly spawned environments where Rally tracks repo gets checked out this will translate to `master`. Which is why I removed the earlier hardcoded `master`.

NickDris: Oof, I missed it, ok LGTM then

NickDris (approved):
Good, let's merge it

PR #1980, opened 2025-09-23
Add ability to download offsets that are stored along with corpus

gareth-ellis:
Currently it can take a long time to generate offsets for large corpora - though they dont really need updating until we go and change the corpus - so this adds the ability to try and download an offset file from the base-url of a track, before trying to calculate it locally.


esrally/utils/io.py
```diff
@@ -520,6 +520,59 @@ def is_valid(self) -> bool:
...
+            # Ensure the directory exists
+            os.makedirs(os.path.dirname(self.offset_table_path), exist_ok=True)
+
+            # Download the offset file using Rally's existing HTTP download functionality
+            if remote_offset_url.startswith(("http://", "https://")):
+                net.download_http(remote_offset_url, self.offset_table_path)
+            elif remote_offset_url.startswith(("s3://", "gs://")):
+                # Extract scheme for bucket downloads
+                scheme = remote_offset_url.split("://")[0]
+                net.download_from_bucket(scheme, remote_offset_url, self.offset_table_path)
```
gbanasiak: This looks like repetition of `download()` from `esrally.utils.net`, can we reuse it?

gbanasiak (commented):
I've tested this without offset file and it seems to work fine. I just left one comment regarding code reuse.

@fressi-elastic FYI, this is another entry point to `net.download*()` _outside_ of `Downloader` but still within `DefaultTrackPreparator.prepare_docs()`.

esrally/utils/io.py
gareth-ellis: Addressed in fe7114930ee10e7cb0248547de9e555fa25d153f

gbanasiak:
Trying to address the test failure with https://github.com/elastic/rally/pull/1983.

gbanasiak (approved):
LGTM

esrally/utils/io.py
```diff
@@ -520,6 +520,41 @@ def is_valid(self) -> bool:
...
+
+        logger.info("Attempting to download offset file from [%s]", remote_offset_url)
+
+        # Ensure the directory exists
+        os.makedirs(os.path.dirname(self.offset_table_path), exist_ok=True)
+        try:
+            net.download(remote_offset_url, self.offset_table_path)
+            logger.info("Successfully downloaded offset file from [%s]", remote_offset_url)
+            return True
+        except BaseException as be:
```
fressi-elastic: You shouldn't capture and silence `BaseException`, you should typically re-raise them. I think in this case you should capture `Exception` instead of `BaseException`, because are not intended for execution interruption.

esrally/utils/io.py
```diff
@@ -520,6 +520,41 @@ def is_valid(self) -> bool:
         """
         return self.exists() and os.path.getmtime(self.offset_table_path) >= os.path.getmtime(self.data_file_path)
 
+    def try_download_from_corpus_location(self, corpus_base_url: Optional[str] = None) -> bool:
```
fressi-elastic: You can now use `str | None` instead of `Optional[str]`.

esrally/utils/io.py
```diff
@@ -596,16 +631,32 @@ def remove(data_file_path: str) -> None:
         os.remove(f"{data_file_path}.offset")
 
 
-def prepare_file_offset_table(data_file_path: str) -> Optional[int]:
+def prepare_file_offset_table(data_file_path: str, corpus_base_url: Optional[str] = None) -> Optional[int]:
```
fressi-elastic: Same as above:
-  `Optional[str]` -> 'str | None'
-  `Optional[int]` -> 'int | None'

fressi-elastic (changes requested):

esrally/utils/io.py
gareth-ellis: I was excepting on BaseException, since net.download excepts on that - if the download fails, I dont really mind, as I can just carry on and calculate the offset as before. 

Is there a reason to except on Exception instead of BaseException, since as far as I can see the only risk is that an exception is raised that doesnt inherit from Exception, but instead directly from BaseException? If Exception is still preferred, why do we except on BaseException in the below method, that I am calling?

https://github.com/elastic/rally/blob/master/esrally/utils/net.py#L272

fressi-elastic: Subclasses of BaseException and not of exception are a restricted number of errors types for which it is not recommended capturing them without reraising them as typically are intended for managing execution interruption of the workflow. It is very uncommon you will capture them without finally reraise (unless you know what you are doing). Subclasses of Exception are perfectly fine as they are typically real errors. Example of things raising BaseException are exit function, stopping iterations, cancelling async tasks, stopping generators, keyboard interruption, etc.

gbanasiak: > If Exception is still preferred, why do we except on BaseException in the below method, that I am calling?

Only to remove a temporary file, and raise `BaseException` again. Which looks like an attempt to cleanup on program termination, but not prevent it.

TIL, in Python [documentation](https://docs.python.org/3/library/exceptions.html#Exception), `Exception` is described as follows:
> All built-in, **non-system-exiting** exceptions are derived from this class. All user-defined exceptions should also be derived from this class.

From this we can infer that `BaseException` which is a base class for `Exception` is inherited from directly in case of system-existing exceptions, and indeed when we look at the source:

```
class BaseException:
[..]

class GeneratorExit(BaseException): ...
class KeyboardInterrupt(BaseException): ...

class SystemExit(BaseException):
    code: sys._ExitCode

class Exception(BaseException): ...

class StopIteration(Exception):
    value: Any

[..]
```

So in practice, if we catch `BaseException` the extras we catch are `GeneratorExit`, `KeyboardInterrupt` and `SystemExit`. I guess the last 2 are the ones the author of `net.download()` cared about :)

gareth-ellis: I've updated

esrally/utils/io.py
```diff
@@ -520,6 +520,41 @@ def is_valid(self) -> bool:
...
+        logger = logging.getLogger(__name__)
+        data_file_name = os.path.basename(self.data_file_path)
+        offset_file_name = f"{data_file_name}.offset"
+
+        # Construct the remote offset file URL
+        if corpus_base_url.endswith("/"):
+            separator = ""
+        else:
+            separator = "/"
+        remote_offset_url = f"{corpus_base_url}{separator}{offset_file_name}"
```
fressi-elastic: Or simply write these two lines. :-)
```
offset_file_name = os.path.basename(f"{self.data_file_path}.offset")
remote_offset_url = f"{corpus_base_url.rstrip('/')}/{offset_file_name}"
```

fressi-elastic (approved):

gareth-ellis:
@elasticmachine update branch

gareth-ellis:
@elasticmachine update branch

tests/utils/io_test.py
```diff
@@ -125,3 +125,43 @@ def test_decompress_manually_external_fails_if_tool_missing(self, mocked_run):
...
+
+    def test_is_valid_with_well_formed_content(self, tmp_path):
+        table = self._make_table(tmp_path, "line1\n", "0;0\n50000;1234\n100000;5678\n")
+        assert table.is_valid() is True
+
+    def test_is_valid_with_single_line(self, tmp_path):
+        table = self._make_table(tmp_path, "line1\n", "0;0\n")
+        assert table.is_valid() is True
+
+    def test_is_valid_rejects_empty_file(self, tmp_path):
```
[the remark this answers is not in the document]
gareth-ellis: Fixed in 000a4205aa4f3b2ce2efdffeb9fab2fe326ac33a

esrally/utils/io.py
```diff
@@ -590,16 +645,32 @@ def remove(data_file_path: str) -> None:
...
+            if file_offset_table.try_download_from_corpus_location(corpus_base_url):
+                console.println("[DOWNLOADED]")
+                # Verify the downloaded file is valid
+                if file_offset_table.is_valid():
+                    return None
+                else:
+                    console.println("[INVALID - will create locally]")
+            else:
+                console.println("[NOT FOUND - will create locally]")
+
```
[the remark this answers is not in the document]
gareth-ellis: @copilot apply changes based on [this feedback](https://github.com/elastic/rally/pull/1980#discussion_r3154374813)

esrally/utils/io.py
```diff
@@ -510,9 +510,64 @@ def exists(self) -> bool:
...
+                for _ in range(3):
+                    line = f.readline()
+                    if not line:
+                        break
+                    parts = line.strip().split(";")
+                    if len(parts) != 2:
+                        return False
+                    int(parts[0])
+                    int(parts[1])
+            return True
```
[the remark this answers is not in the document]
gareth-ellis: Fixed in 000a4205aa4f3b2ce2efdffeb9fab2fe326ac33a

gareth-ellis:
Variants tested:
Invalid offset downloaded:
```
INFO] Race id is [b9478fd7-bd53-4312-8690-9f8863857909]
[INFO] Attempting to download offset file for [/home/esbench/.rally/benchmarks/data/wikipedia/pages.json] from corpus location ... [DOWNLOADED]
[INVALID - will create locally]
[INFO] Preparing file offset table for [/home/esbench/.rally/benchmarks/data/wikipedia/pages.json] ... [OK]
[INFO] Racing on track [wikipedia], challenge [index-and-search] and car ['external'] with version [9.5.0-SNAPSHOT].
```
Missing from remote:
```
[INFO] Race id is [85d39e0e-d1ea-40b5-871c-25b6a183ddc1]
[INFO] Attempting to download offset file for [/home/esbench/.rally/benchmarks/data/wikipedia/pages.json] from corpus location ... [NOT FOUND - will create locally]
[INFO] Preparing file offset table for [/home/esbench/.rally/benchmarks/data/wikipedia/pages.json] ... [OK]
[INFO] Racing on track [wikipedia], challenge [index-and-search] and car ['external'] with version [9.5.0-SNAPSHOT].
```

Present, downloaded succesfully:
```
[INFO] Race id is [3aa0a3b9-7bf9-4899-a72d-241afa1b1d9b]
[INFO] Attempting to download offset file for [/home/esbench/.rally/benchmarks/data/wikipedia/pages.json] from corpus location ... [DOWNLOADED]
[INFO] Racing on track [wikipedia], challenge [index-and-search] and car ['external'] with version [9.5.0-SNAPSHOT].
```

tests/utils/io_test.py
```diff
@@ -125,3 +125,170 @@ def test_decompress_manually_external_fails_if_tool_missing(self, mocked_run):
...
+        assert result is None
+        assert os.path.exists(offset_file_path)
+
+    def test_falls_back_to_local_when_download_fails(self, tmp_path):
+        data_file_path = self._write_data_file(tmp_path)
+
+        with mock.patch("esrally.utils.net.download", side_effect=Exception("404 Not Found")):
+            result = io.prepare_file_offset_table(data_file_path, "http://example.com/corpora")
+
+        assert result is not None
```
gbanasiak: nit, similarly for other `result is not None`.

```suggestion
        assert result == 100
```

gbanasiak (approved):
still LGTM :)

PR #1981, opened 2025-09-30
It drops support for Python3.9, it add support for Python 3.13, and it drops nox dependencies in favor of uv.

fressi-elastic:
In October 2025 Python.org is planning the release of Python 3.14 and dropping support for Python 3.9. This PR is following the trend dropping version 3.9 and adding the support of the now 1 year old Python 3.13.

It also remove testing Python 3.9 support in the `make test` goal. It instead provide goals to run tests with specific python versions (`test-3.10`, `test-3.11`, `test-3.12` and `test-3.13`, one for each supported version).
'make test' is now equivalent to 'make test-3.12`. 

The main idea is developers will test changes with their most convenient Python version, while the whole range of supported python versions should be better tested at continuous integration stage.


.github/workflows/ci.yml
```diff
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
```
gbanasiak: Can we drop `3.9`? In general let's drop the oldest when adding a new one.

uv.lock
```diff
@@ -12,9 +12,9 @@ resolution-markers = [
 name = "aiohappyeyeballs"
 version = "2.6.1"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/26/30/f84a107a9c4331c14b2b586036f40965c128aa4fee4dda5d3d51cb14ad54/aiohappyeyeballs-2.6.1.tar.gz", hash = "sha256:c3f9d0113123803ccadfdf3f0faa505bc78e6a72d1cc4806cbd719826e943558", size = 22760, upload-time = "2025-03-12T01:42:48.764Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/26/30/f84a107a9c4331c14b2b586036f40965c128aa4fee4dda5d3d51cb14ad54/aiohappyeyeballs-2.6.1.tar.gz", hash = "sha256:c3f9d0113123803ccadfdf3f0faa505bc78e6a72d1cc4806cbd719826e943558", size = 22760 }
```
gbanasiak: What is your `uv` version? The fact that we're dropping `upload-time` field suggests you're not up to date.

gareth-ellis (commented):
Can we update readme too? You mention that 3.12 is the preferred version, we should perhaps call that out in the readme too?

https://github.com/elastic/rally/blob/master/README.md?plain=1#L27

And also docs:
https://github.com/elastic/rally/blob/master/docs/install.rst?plain=1#L21

It may be worth grepping for 3.9, see if there's other occurances too

.github/workflows/ci.yml
fressi-elastic: No problem. I will do both at the same time then. I wasn't sure removing old version could have been a blocker to adding a new version. But if it is not the case then I will add dropping of 3.9 to this change. Cheers :-)

uv.lock
fressi-elastic: I have no idea. I uploaded the file only because building docker image from one of the integration tests was failing. I'll update uv and re-upload the file again.

.buildkite/it/pipeline.yml
```diff
@@ -4,7 +4,9 @@ agents:
   machineType: "n1-standard-8"
 
 steps:
-  - label: "Run 3.9 integration tests :test_tube:"
-    command: bash .buildkite/it/run.sh 3.9
+  - label: "Run 3.10 integration tests :test_tube:"
+    command: bash .buildkite/it/run.sh 3.10
   - label: "Run 3.12 integration tests :test_tube:"
     command: bash .buildkite/it/run.sh 3.12
+  - label: "Run 3.13 integration tests :test_tube:"
```
gareth-ellis, line 10: I think we should replace 3.12 with 3.13 here, rather than adding it as an extra. The pipeline is already reasonably expensive to run, and testing intermediate versions does not really offer much more

.ci/variables.json
```diff
@@ -1,10 +1,10 @@
 {
   "python_versions": {
-    "PY39": "3.9.16",
     "PY310": "3.10.10",
     "PY311": "3.11.7",
     "PY312": "3.12.2",
-    "MIN_PY_VER": "3.9.16",
+    "PY313": "3.13.1",
```
gbanasiak: 3.13.7 is the latest I see in `uv python list`

esrally/storage/_aws.py
```diff
@@ -56,7 +54,7 @@ def __init__(
         self,
         aws_profile: str | None = AWS_PROFILE,
         chunk_size: int = CHUNK_SIZE,
-        s3_client: S3Client | None = None,
+        s3_client: Optional["S3Client"] = None,
```
gbanasiak: `Optional` is not needed since 3.10, so why introducing it now?

esrally/storage/_transfer.py
```diff
@@ -349,7 +348,7 @@ def close(self):
                 LOG.warning("error closing file descriptor %r: %s", fd.path, ex)
 
     @contextmanager
-    def _open(self) -> Iterator[FileDescriptor | None]:
+    def _open(self) -> Iterator[Optional["FileDescriptor"]]:
```
gbanasiak: same here, is `Optional` required with 3.10?

uv.lock
gbanasiak: I'm getting different results with the latest uv 0.8.22. I can see `upload-time` parameter in the updated lines. Can you run `uv self update` followed by `make uv-lock`?

gbanasiak (commented):
I think it would make sense to bump `DEFAULT_PY_VER` to 3.13 in this PR moving development environment to 3.13. If not now then when? Please also update Python versions in `mypy` and `black` configuration in `pyproject.toml`.

fressi-elastic:
Ok!

esrally/storage/_aws.py
fressi-elastic: Being S3Client defined later, I have to replace with a string, which doesn't support the | operator anymore after removing the from __future__ statement, so linters or module importation are failing (can't remember which one). In this cases I could leave the __future__ statement or reintroduce Optional/Union. Which one is better for you?

gbanasiak: Thanks for explaining, I'm good with what we have now :)

esrally/storage/_transfer.py
fressi-elastic: The same as above. I think I would better revert these files changes that require more than just removing the `from __future__` statement as at the end at some moment in time when dropping 3.10 or 3.11 we will be able to remove it with a nice nothing else to do.

esrally/__init__.py
```diff
@@ -72,9 +73,25 @@
...
+    minor: int = 0
+    micro: int = 0
+
+    def __str__(self) -> str:
+        return f"{self.major}.{self.minor}.{self.micro}"
+
+
+MIN_PYTHON_VERSION = Version(3, 10)
+
+MAX_PYTHON_VERSION = Version(3, 14)
```
gareth-ellis: Why not define this as 3,13 and then use >  in the comparison? It feels a little confusing (I assume its because you dont know what the max version may be....but that could be got around another way - e.g sys.version_info.minor ).

gbanasiak (commented):
Python upgrade PR turned into Python upgrade + Makefile refactor + nox replacement (although not complete as `noxfile.py` is still there). The title does not reflect the content. I suggest to reduce the scope to the original one in order to achieve the initial goal of enabling 3.13 while removing 3.9, and move nox removal to another PR. I don't feel sentimental about nox, if uv is faster let's do it.

.buildkite/it/run_serverless.sh
```diff
@@ -17,22 +17,26 @@ export DEBIAN_FRONTEND=noninteractive
...
 
-"python${PYTHON_VERSION}" -m venv .venv
-source .venv/bin/activate
-pip install nox
+curl -LsSf https://astral.sh/uv/install.sh | sh
+source "${HOME}/.local/bin/env"
+
+echo "--- Create virtual environment"
+
+make venv
```
gbanasiak, line 39: This bit is unnecessary as `uv` and venv are already created in the Docker image which is used to run this workload.

esrally/__init__.py
fressi-elastic: I dropped the check for maxinum python version as it makes no sense to me. Anyway the information in `pyproject.toml` file should already help catching potential issues.

.buildkite/it/run_serverless.sh
fressi-elastic: The test cases require a virtual env that was before provided by `nox`. UV is a replacement for it. It is required for running `make it_serverless`. The idea is letting things simpler and with lesser places to update every year we have a new python version to drop. This mean we don't need extra complexity and failure points like having to depend on `nox`.

pyproject.toml
```diff
@@ -92,6 +92,8 @@ dependencies = [
     "hatch==1.3.1",
     "hatchling==1.6.0",
     "wheel==0.37.1",
+    "standard-imghdr>=3.13.0",
```
gareth-ellis: Perhaps a silly question - but where are we using this library? I cant find any use of it in the code base?

FedericoRessi: It has been required because imghdr has been removed from Python3.13. But eventually later by changing/removing some other requirement it got no more required too. So I am going to remove it. Cheers.

gareth-ellis (approved):
LGTM

.buildkite/it/run_serverless.sh
gbanasiak: Why https://github.com/elastic/ci-agent-images/blob/aec1570393a1c25fcb3240cefcb318128262964f/container-images/es-perf/python-base/Dockerfile#L11-L16 is not enough?

FedericoRessi: It test cases need rally virtual environment to be executed. They later use the Docker image to run rally. A bit unexpected to mee too. Pytest could perfectly be exectued from the container too, but it is our of scope here.

.buildkite/it/run_serverless.sh
```diff
@@ -45,12 +49,13 @@ export THESPLOG_THRESHOLD="INFO"
 
 trap upload_logs ERR
 
+
 case $TEST_NAME in
     "user")
-        nox -s it_serverless
+        make -s it_serverless
```
gbanasiak, line 55: `-s` in nox has different meaning than in `make`, is this intentional?

Makefile
```diff
@@ -16,105 +16,211 @@
...
-docs-clean:
-	cd docs && $(MAKE) clean
+# --- venv goals ---
+
+# It creates the project virtual environment.
+venv: uv $(VENV_ACTIVATE_FILE)
+
+$(VENV_ACTIVATE_FILE):
+	uv venv --allow-existing --seed
+	uv sync --locked --extra=develop
```
gbanasiak, line 104: Any particular reason for reintroducing `pip` with `--seed`? Isn't the 2nd command enough?

Makefile
```diff
@@ -16,105 +16,211 @@
...
+# It delete the project virtual environment from disk.
+clean-venv:
+	rm -fR '$(VIRTUAL_ENV)'
+
+# It installs the Rally PyTest plugin.
+install_pytest_rally_plugin: venv
+	$(VENV_ACTIVATE); uv pip install 'pytest-rally @ git+https://github.com/elastic/pytest-rally.git'
+
+# Old legacy alias goals
+install: venv
```
gbanasiak, line 115: Previously `install` was unconditional which allowed going back and forth with the venv state when PR updates dependencies. Now once venv is installed, the only way to update dependencies is running `reinstall`. Is this intentional?

.buildkite/it/run_serverless.sh
gbanasiak: Right, `make venv` not only creates the venv but also populates it according to the uv lock (I think I assumed it's only venv creation earlier). But what about `curl -LsSf https://astral.sh/uv/install.sh | sh` ? It downloads and installs `uv` in Docker container where `uv` has already been installed. This seems unnecessary.

.buildkite/it/run_serverless.sh
fressi-elastic: It is not intentional! It's a mistake.

Makefile
fressi-elastic: I guess I copied from somewhere. I must check what happen removing there options. Also the `--allow-existing` should not be necessary anymore.

Makefile
fressi-elastic: For `venv` I wanted to give a behavior similar to what I would expect from a regular `Makefile` when creating something. It may be I should restore the old behaviour for `make install`. Thank for reporting.

.buildkite/it/run_serverless.sh
fressi-elastic: [.buildkite/it/run_serverless.sh](https://github.com/elastic/rally/pull/1981/files/930ac3ac916b2d9d86389239a8650de3ba2799f2#diff-9edfb7d62f81a60518e6b1b276fd1309e9479aa1f2b18e244586e79b8359aaee) is executed outside of the container, therefore UV is installed outside of the container, where it is needed. In the container is installed when the image is being created copying it from another image.

gbanasiak: Ah, of course, too many pipelines! Thanks for explaining.

gbanasiak:
@pquentin Many thanks for comments.

> You didn't replace nox with uv, but with make, which is where the logic to test different Python versions exists.

Indeed.

> That said, if uv makes it easy to provide Python and install Rally now on Linux and macOS, maybe it's time to force a specific version and not deal with the overhead of supporting older versions.

`uv` definitely makes it easy for dev purposes, but how to ship Rally to end-users? Should we stop releasing PyPI packages and provide an installation script which uses `uv` and `git` or Docker under the hood? `curl -fsSL https://elastic.co/esrally | sh` ?

Makefile
gbanasiak: We had alignment between Rally and esbench on that one. Now we don't. I don't mind changing my habits from `make install` to `make reinstall` (although `make install` was communicated earlier to people using the latest snapshot as a proper way to update dependencies, which is now broken), but let's keep the same across all related projects.

PR #1984, opened 2025-10-08
Add ability to continue on network issues

gareth-ellis:
When running long tests we occasionally see network issues, seemingly arising from either the load driver instance or load balancers. This gives the option to be a little more leniant and not abort the whole test immediately. I haven't built in a retry mechanism, but that would probably be a good future improvement (along with back off).

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -34,7 +34,18 @@ from typing import Callable, Optional
 
 import thespian.actors
 
-from esrally import PROGRAM_NAME, actor, client, config, exceptions, metrics, paths, telemetry, track, types
+from esrally import (
+    PROGRAM_NAME,
+    actor,
+    client,
+    config,
+    exceptions,
+    metrics,
+    paths,
+    telemetry,
+    track,
+    types,
+)
 from esrally.client import delete_api_keys
 from esrally.driver import runner, scheduler
 from esrally.track import TrackProcessorRegistry, load_track, load_track_plugins
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -50,15 +50,7 @@ from esrally.client import delete_api_keys
 from esrally.driver import runner, scheduler
 from esrally.track import TrackProcessorRegistry, load_track, load_track_plugins
 from esrally.utils import console, convert, net
-
-
-class OnErrorBehavior(Enum):
-    ABORT = "abort"
-    CONTINUE = "continue"
-    CONTINUE_ON_NETWORK = "continue-on-network"
-
-    def __str__(self):
-        return self.value
+from esrally.utils.error_behavior import OnErrorBehavior
 
 
 ##################################
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -1235,7 +1235,7 @@ class Worker(actor.RallyActor):
         self.client_contexts = None
         self.current_task_index = 0
         self.next_task_index = 0
-        self.on_error = None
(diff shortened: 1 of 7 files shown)


esrally/driver/driver.py
```diff
@@ -2045,8 +2061,8 @@ async def execute_single(runner, es, params, on_error):
     except elasticsearch.TransportError as e:
         # we *specifically* want to distinguish connection refused (a node died?) from connection timeouts
         # pylint: disable=unidiomatic-typecheck
-        if type(e) is elasticsearch.ConnectionError:
-            fatal_error = True
+        is_connection_error = type(e) is elasticsearch.ConnectionError
```
dpifke-elastic, line 2052: Nit: fatal_error doesn't seemed to be used anywhere else except the check at the bottom of this function, maybe just rename it to is_connection_error (which is more clear)?

dpifke-elastic (approved):
LGTM.

esrally/driver/driver.py
```diff
@@ -2026,6 +2035,13 @@ async def execute_single(runner, es, params, on_error):
     import elasticsearch
 
     fatal_error = False
+    # Accept both string and Enum for backward compatibility
+    if isinstance(on_error, str):
+        try:
+            on_error = OnErrorBehavior(on_error)
+        except ValueError:
+            on_error = OnErrorBehavior.CONTINUE
+
```
gbanasiak: I like `OnErrorBehavior` enum as it improves code readability. It would be nice to convert the type consistently and avoid this BWC bit, also adding type annotations for `on_error` and `self.on_error`.

gareth-ellis: Addressed in 06ee5c1949de4bd07601d0b1208f93edd287a025

esrally/track/track.py
```diff
@@ -1090,14 +1090,14 @@ def error_behavior(self, default_error_behavior):
...
 
-        behavior = "continue"
-        if default_error_behavior == "abort":
-            if self.ignore_response_error_level != "non-fatal":
-                behavior = "abort"
-
+        behavior = default_error_behavior if default_error_behavior in ("abort", "continue", "continue-on-network") else "continue"
+        if behavior == "abort":
+            if self.ignore_response_error_level == "non-fatal":
+                behavior = "continue"
```
gbanasiak: Can we set the type returned by `error_behavior()` to `OnErrorBehavior`?

tests/track/track_test.py
```diff
@@ -379,11 +379,21 @@ def test_task_continues_with_global_continue(self):
         effective_on_error = task.error_behavior(default_error_behavior="continue")
         assert effective_on_error == "continue"
 
+    def test_task_continues_with_global_continue_on_network(self):
+        task = self.task()
+        effective_on_error = task.error_behavior(default_error_behavior="continue-on-network")
+        assert effective_on_error == "continue-on-network"
```
gbanasiak: If `error_behavior()` is modified as per earlier comment we could use `OnErrorBehavior` here too.

gareth-ellis: Take a look at adff8d3, i had to move OnErrorBehaviour due to circular import issues

gbanasiak: I guess at some point we may want to collect those small modules like `esrally.utils.serverless` and `esrally.utils.error_behavior` into a single one, e.g. `esrally.types` but I'm good with current state. Thanks.

gbanasiak (approved):
LGTM

PR #1986, opened 2025-10-10
Enable a release build of an Elasticsearch revision from sources (source.build.release) 

NickDris:
Introducing --source-build-release argument along with source.build.release parameter which defaults to false.

This parameter is designed to build a release version of an Elasticsearch revision when user builds from sources.
Being built as a "release" version Elasticsearch will enable all the feature flags that would be imposed in a regular release build.

### DIFF
--- a/esrally/mechanic/supplier.py
+++ b/esrally/mechanic/supplier.py
@@ -40,7 +40,7 @@ def create(cfg: types.Config, sources, distribution, car, plugins=None):
     caching_enabled = cfg.opts("source", "cache", mandatory=False, default_value=True)
     revisions = _extract_revisions(cfg.opts("mechanic", "source.revision", mandatory=sources))
     source_build_method = cfg.opts("mechanic", "source.build.method", mandatory=False, default_value="default")
-    source_build_snapshot = cfg.opts("mechanic", "source.build.snapshot", mandatory=False, default_value=True)
+    source_build_release = cfg.opts("mechanic", "source.build.release", mandatory=False, default_value=False)
     distribution_version = cfg.opts("mechanic", "distribution.version", mandatory=False)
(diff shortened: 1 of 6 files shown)


tests/mechanic/supplier_test.py
```diff
@@ -692,6 +692,22 @@ def test_build(self):
 
         builder.build.assert_called_once_with(["./gradlew clean", "./gradlew assemble"])
 
+    def test_build_no_snapshot(self):
+        car = team.Car(
+            "default",
+            root_path=None,
+            config_paths=[],
+            variables={"clean_command": "./gradlew clean", "system.build_command.no-snapshot": "./gradlew assemble"},
```
gbanasiak: Nit: I would add `-Dbuild.snapshot=false` to the command to make it easier to understand the code. Same for the other test case.

esrally/mechanic/supplier.py
```diff
@@ -407,17 +409,17 @@ def prepare(self):
...
-                ]
+            system_build_command_var = "system.build_command.arch" if self.template_renderer.arch != "x86_64" else "system.build_command"
+            if getattr(self.builder, "release_build", None) and self.builder.release_build is True:
+                system_build_command_var += ".no-snapshot"
             else:
-                commands = [
-                    self.template_renderer.render(self.car.mandatory_var("clean_command")),
-                    self.template_renderer.render(self.car.mandatory_var("system.build_command")),
-                ]
+                console.warn("Building a SNAPSHOT version of Elasticsearch, source.build.release was not set")
```
gbanasiak: Can we switch to `logger.info()` here, and also add a log in the `if` clause - `Building a release version...`?

gbanasiak (approved):
LGTM, just minor comment about logging left.

PR #1987, opened 2025-10-15
 Recover old silenced loggers instantiated before actor start is complete.

fressi-elastic:
It is convenient to use global references to loggers in python modules so they are available to the module functions without having to pass them as variables. This is a very common use case that is not working in rally because for some reason the logging system is squashed once the actors are being initialized, so those loggers get lost in a old logger manager not more used. When used they send messages to stderr instead of the logging forwarding mechanism implemented in thespian.

This patch implements a workaround to be executed during actors initialization to fix old logger references created by the old manager, with new ones created using the new logger manager. To perform it, it first check the issue checking if the module logger is not the one provided by the new manager. On such case it look at all modules referenced by the old manager and tries to replace their per-module references with new loggers provided by the new manager. The tricky part is to found the old references. The method used is looking for the modules by the name of lost loggers, and then look for the first reference to the logger interacting all the module attributes. The method is not optimal because it requires, after every actor iteration, to iterate tens of modules and for each tens of attributes to see if they refer to the logger.

A better solution has to be found, may be understanding what is going wrong during logging setup.

### DIFF
--- a/esrally/log.py
+++ b/esrally/log.py
@@ -14,6 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+import copy
 import importlib
 import json
 import logging
@@ -21,7 +22,6 @@ import logging.config
 import os
 import time
 import typing
-from copy import deepcopy
 
 import ecs_logging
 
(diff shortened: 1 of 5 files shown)


esrally/log.py
```diff
@@ -113,27 +116,37 @@ def log_config_path():
...
+    updated = copy.deepcopy(original)
+    updated.setdefault("disable_existing_loggers", template.get("disable_existing_loggers", False))
+
+    template_loggers = template.get("loggers", {})
+    config_loggers = updated.setdefault("loggers", template_loggers)
+    for name, logger in template_loggers.items():
+        config_loggers.setdefault(name, logger)
+
+    if original != updated:
+        LOG.info("Update logging configuration file with new values from template: '%s' -> '%s'", template_path, config_path)
```
gbanasiak, line 147: This log is issued before logging configuration is loaded, so does not land in `rally.log`. How about `console` instead?

gbanasiak (approved):
Hacky but works. I prefer full log coverage over beauty :) I've noticed one log is missing, see below.

esrally/log.py
fressi-elastic: I think the code before should update it as LOG is defined in the parent scope. Isn't it? I'll verify and fix it accordingly. Thank you for spotting it.

PR #1988, opened 2025-10-16
Recover all loggers left orphan after thespian logging setup on actors startup.

fressi-elastic:
This goes a step further on top of PR #1987. It is able to recover all lost loggers by letting them use the same object __dict__ as their loggers replacements, so that they will use same filtering and records forwarding as the new ones. Any incremental configuration on top of new ones will be reflected on old loggers too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,7 @@ install_pytest_rally_plugin: venv
 
 # Old legacy alias goals
 install: venv
+	uv sync --locked --extra=develop
 
 reinstall: clean-venv
 	$(MAKE) venv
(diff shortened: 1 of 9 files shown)


gbanasiak (commented):
This works well in my local testing, but IT failure suggests the log recovery does not always work correctly. I'll take another look once IT failure is addressed.

fressi-elastic:
I didn't saw any noisy line from the `elastic_transport`. The integration test did succeeded too. I tested it using `fork` method.

Makefile
```diff
@@ -113,6 +113,7 @@ install_pytest_rally_plugin: venv
 
 # Old legacy alias goals
 install: venv
+	uv sync --locked --extra=develop
```
gbanasiak: Makefile changes are unrelated, can we remove this?

pyproject.toml
```diff
@@ -130,6 +130,8 @@ develop = [
     "types-urllib3==1.26.19",
     "types-requests<2.32.0",
     "types-jsonschema==3.2.0",
+    # Python dead library removed in version 3.13 and used to build Rally docs.
+    "standard-imghdr==3.13.0",
```
gbanasiak: This change is unrelated, can we clean it?

uv.lock
```diff
@@ -635,6 +635,7 @@ develop = [
     { name = "pytest-benchmark" },
     { name = "pytest-httpserver" },
     { name = "sphinx" },
+    { name = "standard-imghdr" },
```
gbanasiak: Same for `uv.lock`...

gbanasiak (commented):
Looks great. I'm happy with current state, please just remove unrelated changes, thanks!

gbanasiak (approved):

PR #1989, opened 2025-10-17
Fix Makefile

fressi-elastic:
It fixes `Makefile`:

- It restores the old `make install` behavior.
- It allows passing arguments to `make it`
- It uses `uv run` to run `make` against the docs sub-folder.
- It fixes `make docs` dependency issue with Python v3.13

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,7 @@ install_pytest_rally_plugin: venv
 
 # Old legacy alias goals
 install: venv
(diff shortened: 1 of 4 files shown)


gbanasiak (commented):
Thank you. After switching `docs` to `uv` I finally see `imghdr` issue. I think it makes sense to add it here.

gbanasiak (approved):

PR #1991, opened 2025-10-24
Create CLI for Rally storage manager.

fressi-elastic:
Whit the purpose of getting post-mortem information of issues that storage manager have found accessing mirror servers I created a small command line interface to list downloaded files and get mirrors errors from their status file. This is a simple but powerful debugging tool useful for getting insight on files produced by the transfer manager. The CLI also extends the pre-existing one to invoke files download using the transfer manager.

Example of use:
```
python3 -m esrally.storage get apm/documents-1k.ndjson.bz2
python3 -m esrally.storage ls
[
  {
    "crc32": "83jA8A==",
    "document_length": 63458,
    "done": "0-63457",
    "errors": [],
    "finished": true,
    "mirror_failures": {
      "s3://es-perf-fressi-eu-west-1/apm/documents-1k.ndjson.bz2": "ClientError:An error occurred (400) when calling the HeadObject operation: Bad Request",
      "s3://es-perf-fressi-us-west-1/apm/documents-1k.ndjson.bz2": "ClientError:An error occurred (400) when calling the HeadObject operation: Bad Request"
    },
    "path": "/Users/fressi/.rally/storage/https:/rally-tracks.elastic.co/apm/documents-1k.ndjson.bz2",
    "progress": 100.0,
    "status": "DONE",
    "todo": "",
    "url": "https://rally-tracks.elastic.co/apm/documents-1k.ndjson.bz2"
  }
]
```


esrally/storage/_cli.py
```diff
@@ -0,0 +1,223 @@
...
+            "status": str(tr.status),
+            "progress": tr.progress,
+            "errors": [str(e) for e in tr.errors],
+            "document_length": tr.document_length,
+            "crc32": tr.crc32c,
+            "done": str(tr.done),
+            "mirror_failures": tr.mirror_failures,
+            "todo": str(tr.todo),
+            "finished": tr.finished,
+        }
```
gbanasiak, line 145: Can we add a simple `was_mirrored` (or similar) boolean that would tell us whether the file was downloaded from a mirror, or not, without having to parse `mirror_failures` dict?

esrally/storage/_config.py
```diff
@@ -69,6 +80,13 @@ def local_dir(self) -> str:
     def local_dir(self, value: str) -> None:
         self.add(config.Scope.applicationOverride, "storage", "storage.local_dir", value)
 
+    def transfer_file_path(self, url: str) -> str:
+        path = os.path.join(self.local_dir, url)
+        return os.path.normpath(os.path.expanduser(path))
```
gbanasiak, line 85: This leaves the colon from URL in the path, e.g. `<path>/.rally/storage/https:/rally-tracks.elastic.co/apm/span.json.bz2`. This looked a bit weird to me at first, but seems to work fine. Just making sure that's intentional.

esrally/storage/_manager.py
```diff
@@ -92,49 +103,106 @@ def shutdown(self):
...
+        self,
+        url: str,
+        *,
+        path: str | None = None,
+        document_length: int | None = None,
+        crc32c: str | None = None,
+        start: bool = True,
+        local_dir: str | None = None,
+        todo: RangeSet | None = None,
+    ) -> Transfer:
```
gbanasiak, line 118: When testing with `esrally race --track=geonames --challenge=append-no-conflicts-index-only --distribution-version=8.15.5 --enable-assertions --kill-running-processes --on-error="abort" --test-mode` I'm getting:

```
2025-10-27 18:30:05,696 ActorAddr-(T|:49482)/PID:7754 esrally.driver.driver ERROR Worker failed. Notifying parent...

  File "/Users/grzegorz/src/rally/esrally/track/loader.py", line 554, in download
    manager.get(data_url, target_path, size_in_bytes).wait()
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: TransferManager.get() takes 2 positional arguments but 4 were given
```

I think there needs to be Rally IT (independently from storage ITs) with transfer manager enabled to catch such refactor issues.

gbanasiak (changes requested):
This is great. I found one glitch in testing due to `manager.get()` refactor. I also have one additional request regarding `esrally-storage ls` output.

esrally/storage/_cli.py
fressi-elastic: I a file is mirrored and mirror worked, this shouldn't be included in the list of files to uploaded. I would expect in an ideal world every file to be uploaded only once the first day the issue is found. We could have some inconsistence uploading it multiple times during the same day if the same file is used by multiple run in the same time, but it shouldn't affect. 🙏

gbanasiak: > If a file is mirrored and mirror worked, this shouldn't be included in the list of files to uploaded.

I know. My ask is in the context of ingestion of this data by Filebeat for statistical purposes. This will end-up in a generic Filebeat index, and I want to keep mappings simple by ingesting only selected fields. This is auxiliary ask that has nothing to do with how `esrally-storage --mirror-failures put <dest>` works.

fressi-elastic: Can you resume which flag should I add and what it would implies? If has a copy in any mirror, one of the mirror will be tried first. At this point another worker will pick another mirror or the same mirror again. At the end there is no info recorded that says which servers provided pieces of the file. I would like to record stats on the use of mirrors in another change as the urgency is uploading files. At this point I would configure only one mirror for every job, so that the job will end populating it (it fails then uses the original server and records failures). As soon a pair of mirrors are well populated, we can use multiple mirrors as help for having more connections and more reliability (if numbers are good).

esrally/storage/_config.py
fressi-elastic: This is intentional. I want to be able to reconstruct the URL. The column help me separating the schema from the path (I replace ':' with ':/' and I get the URL back). I pass the URL to normpath and it gives me back the relative path from the URL.

esrally/storage/_manager.py
fressi-elastic: Agree! I will add an integration test to reproduce this issue! 🙏🏻
Anyway this bug should habe been catch by linters!!! 😓

esrally/storage/_cli.py
gbanasiak: I understand the complexity. For now all we need is a trivial flag that sets `true` if there were no mirror errors, so no entries under `mirror_failures` dictionaries. I know this is not perfect due to multi-part download but that's enough.

gbanasiak: > At this point I would configure only one mirror for every job [..]

Agreed. With a single mirror the interpretation of `mirror_failures` is easier.

gbanasiak (approved):
LGTM

PR #1992, opened 2025-10-24
Add run-until runner

gareth-ellis:
This runner allows a user to make an API call and evaluate part of the JSON response. The aim is to include in e.g a parallel block - where you could loop a corpus and ingest until ingested size reaches X bytes, or Y documents

There needs to be some documentation added and tests, but a review on the initial approach would be appreciated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,7 @@ check-all: all
 	lint \
 	format \
 	pre-commit \
+	precommit \
 	install-pre-commit \
 	docs \
 	serve-docs \
@@ -139,7 +140,7 @@ lint: venv
 	uv run -- pre-commit run --all-files
 
 # It run all linters on changed files using pre-commit.
-pre-commit: venv
+precommit pre-commit: venv
 	uv run -- pre-commit run
 
 # It install a pre-commit hook in the project .git dir so modified files are checked before creating every commit.
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -105,6 +105,7 @@ def register_default_runners(config: Optional[types.Config] = None):
     register_runner(track.OperationType.TransformStats, Retry(TransformStats()), async_runner=True)
     register_runner(track.OperationType.CreateIlmPolicy, Retry(CreateIlmPolicy()), async_runner=True)
     register_runner(track.OperationType.DeleteIlmPolicy, Retry(DeleteIlmPolicy()), async_runner=True)
+    register_runner(track.OperationType.RunUntil, Retry(RunUntil()), async_runner=True)
 
 
 def runner_for(operation_type):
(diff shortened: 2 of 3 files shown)


esrally/driver/runner.py
```diff
@@ -3048,3 +3049,74 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+    async def __call__(self, es, params):
+        params, request_params, transport_params, headers = self._transport_request_params(params)
+        es = es.options(**transport_params)
+        es.return_raw_response()
+        path = mandatory(params, "path", self)
+        method = params.get("method", "GET")
+        body = params.get("body", None)
+        json_path = mandatory(params, "json-path", self)
+        criterion = mandatory(params, "criterion", self)
+        check_interval = params.get("check-interval", 60)
```
inqueue: Can we reuse `retry-wait-period`?

esrally/driver/runner.py
```diff
@@ -3048,3 +3049,74 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+                Supported operators: 'exists', 'not_exists', 'gte', 'lte', 'eq', 'neq'.
+            value: The actual value to compare against the criterion.
+
+        Returns:
+            bool: True if the criterion is met, False otherwise.
+        """
+        operator = criterion.get("operator")
+        target_value: int | float = criterion.get("value", -1)
+        return_value = False
+        if operator == "exists":
```
inqueue: A dictionary-based approach or match statements could be used in place of the `elif` chain. I let Cursor generate a snip if you want to use it:

```python
# Dictionary mapping operators to their comparison functions
    operators = {
        "exists": lambda v, _: v is not None,
        "not_exists": lambda v, _: v is None,
        "gte": lambda v, t: int(v) >= int(t),
        "lte": lambda v, t: int(v) <= int(t),
        "eq": lambda v, t: int(v) == int(t),
        "neq": lambda v, t: int(v) != int(t),
    }
    
    if operator not in operators:
        raise ValueError(f"Unsupported operator: {operator}")
    
    return operators[operator](value, target_value)
```

inqueue (changes requested):
I like the approach and will give it test today.

Makefile
```diff
@@ -139,7 +140,7 @@ lint: venv
 	uv run -- pre-commit run --all-files
 
 # It run all linters on changed files using pre-commit.
-pre-commit: venv
+precommit pre-commit: venv
```
fressi-elastic, line 143: You can remove pre-commit if you want. I don't think others a part than me realized this goal exists.

esrally/driver/runner.py
```diff
@@ -3048,3 +3049,72 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+class RunUntil(Runner):
+    """
+    Performs an API call until a target value is seen.
+    """
+
+    def __init__(self, *args, config=None, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.numerical_operators = {
+            "gte": lambda v, t: int(v) >= int(t),
```
fressi-elastic, line 3063: Why are we converting values to integers? Could you please reflect it in the class doc string at least?

esrally/driver/runner.py
```diff
@@ -3048,3 +3049,72 @@ async def __aexit__(self, exc_type, exc_val, exc_tb):
...
+        Args:
+            criterion (dict): Dictionary containing 'operator' and optionally 'value' keys.
+                Supported operators: 'exists', 'not_exists', 'gte', 'lte', 'eq', 'neq'.
+            value: The actual value to compare against the criterion.
+
+        Returns:
+            bool: True if the criterion is met, False otherwise.
+        """
+        operator = criterion.get("operator")
+        target_value: int | float = criterion.get("value")
```
fressi-elastic, line 3116: At this time also an 'str' could be valid as it is being consumed by `int()` constructor function.

fressi-elastic (approved):
I left some suggestion, but overall it looks good to me.

inqueue (approved):
LGTM

PR #1995, opened 2025-11-07
Initalize all loggers in rally daemon

gbanasiak:
Follow-up to https://github.com/elastic/rally/pull/1988.

The log recovery mechanism can only work when loggers are defined. When using rally daemon, actor processes are spawned from the daemon which has not enough imports to cover all the loggers. This fixes this problem.


NickDris (approved):
LGTM

PR #1997, opened 2025-11-10
Address pytest deprecations

gbanasiak:
Addresses deprecation warnings visible when running `pytest` with the exception of `aiohttp` which should be upgraded separately.


esrally/utils/io.py
```diff
@@ -405,16 +395,20 @@ def _do_decompress_manually_with_lib(target_directory: str, filename: str, compr
         compressed_file.close()
 
 
-def _do_decompress(target_directory: str, compressed_file: Union[zipfile.ZipFile, tarfile.TarFile]) -> None:
+def _do_tar_decompress(target_directory: str, compressed_file: tarfile.TarFile) -> None:
+    try:
+        compressed_file.extractall(path=target_directory, filter="tar")
```
gbanasiak, line 400: it is now recommended to pass `filter` kwarg when calling `extractall()` for tar files

.github/workflows/ci.yml
```diff
@@ -106,28 +106,6 @@ jobs:
...
-      - name: Run Storage manager test cases.
-        run: make it ARGS=it/storage
-      - uses: elastic/es-perf-github-status@v2
-        if: ${{ failure() && ( github.event_name == 'schedule' || ( github.event_name == 'push' && github.ref_name == env.DEFAULT_BRANCH ) ) }}
-        with:
-          slack_bot_token: ${{ secrets.SLACK_BOT_TOKEN }}
-          slack_channel: ${{ secrets.SLACK_CHANNEL }}
-          status: FAILED
-
-
```
gbanasiak, line 130: This is a duplicate, same runs under Buildkite as part of `it/` tests.

.github/workflows/ci.yml
```diff
@@ -83,13 +83,28 @@ jobs:
...
+        continue-on-error: true
+        uses: jlumbroso/free-disk-space@54081f138730dfa15788a46383842cd2f914a1be
+        with:
+          android: true
+          dotnet: true
+          haskell: true
+          large-packages: false
+          docker-images: false
+          swap-storage: false
+          tool-cache: false
```
gbanasiak, line 99: With this package upgrade we are running out of disk space on GH-hosted runners. Perhaps pytest cache is consuming a bit more now? There's 18 GB of free disk space to use at the beginning of tests. Comparisons before/after show we're using 15 GB of disk, but that's after ES installation is removed as part of `esrally stop`.

As a workaround, this removes around 18 GB of disk space doubling the amount of free space. With all options enabled this frees 30 GB of disk space but full cleanup takes longer (4 minutes instead of 1).

gbanasiak: I've tried using `/mnt` as Rally home (there's 60 GB of free disk space there), but it does not seem to be configured consistently among GH-hosted runners. Sometimes writes to `/mnt` work fine, and sometimes they result in `permission denied`.

fressi-elastic (approved):

PR #1998, opened 2025-11-12
Fix transfer finished condition when worker is stuck downloading.

fressi-elastic:
When a worker is blocked because of a network issue (for example because it is getting the file from another continent), it could happen that the last worker to start sets the transfer as finished even if there is still another worker blocked downloading another fragment.

This should fix it by checking if other workers are still running before setting the transfer as finished.
It also add a connect timeout and read timeout parameters when performing a GET request when using HTTPAdapter.

It sets the default connect timeout to 15 seconds and the read timeout to 10 seconds. Either values can be configured in `rally.ini`.
It also increases the default multipart size to 128 MB.


esrally/storage/http.py
```diff
@@ -97,8 +97,11 @@ def get(self, url: str, *, check_head: Head | None = None) -> tuple[Head, Iterat
             check_head.check(head)
 
         def iter_content() -> Iterator[bytes]:
-            with res:
-                yield from res.iter_content(chunk_size=self.chunk_size)
+            try:
+                with res:
+                    yield from res.iter_content(chunk_size=self.chunk_size)
+            except requests.exceptions.Timeout as ex:
+                raise TimeoutError(f"Timed out reading content from URL={url}: {ex}") from ex
```
gbanasiak, line 104: another possibility is `requests.exceptions.ConnectionError` which does not inherit from `requests.exceptions.Timeout` but directly from `requests.exceptions.RequestException`

gbanasiak (approved):
LGTM

PR #1999, opened 2025-11-13
Fix parallel corpora files download.

fressi-elastic:
This created multiple tasks for multiple files to be dowloaded so that each file can be downloaded in parallel during track preparation.

### DIFF
--- a/esrally/track/loader.py
+++ b/esrally/track/loader.py
@@ -473,23 +473,28 @@ class DefaultTrackPreparator(TrackProcessor):
         self.decompressor = None
         self.track = None
 
-    @staticmethod
-    def prepare_docs(cfg: types.Config, track, corpus, preparator):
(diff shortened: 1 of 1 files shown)


esrally/track/loader.py
```diff
@@ -473,23 +473,28 @@ def __init__(self):
...
+                    yield prepare_document, {
+                        "cfg": self.cfg,
+                        "track": track,
+                        "corpus": corpus,
+                        "preparator": prep,
+                        "document_set": document_set,
+                    }
+
+
+def prepare_document(cfg: types.Config, track, corpus, preparator, document_set):
```
gbanasiak, line 490: Just curious: any specific reason this method was moved out of `DefaultTrackPreparator` class?

gbanasiak (approved):
LGTM

PR #2000, opened 2025-11-14
Add Google Cloud Storage adapter.

fressi-elastic:
This adds the new `GSAdapter` class to download files from Google Cloud buckets using authentication.

It allows to download bucket blobs via URLs like the following:
- `gs://{bucket_name}/{blob_name}`
- `https://storage.cloud.google.com/{bucket_name}/{blob_name}`

For downloading files publicly available from a Google Cloud bucket, the following type of URLs is supported by the already existing `HTTPAdapter`:
- `https://storage.googleapis.com/{bucket_name}/{blob_name}`


pyproject.toml
```diff
@@ -103,10 +104,25 @@ s3 = [
...
+    # transitive dependencies:
+    #   google-auth
+    #   google-api-core
+    #   google-cloud-core
+    #   google-resumable-media
+    #   requests
+    #   google-crc32c
+    "google-cloud-storage==3.5.0",
+]
+
```
gbanasiak: This replicates `s3` setup, but there's inconsistency. We introduced `google-resumable-media[requests]` and `google-auth` into mandatory dependencies earlier (see https://github.com/elastic/rally/pull/1094), so why not `google-cloud-storage`? I would be either for moving all to optional, or keeping all mandatory. Which one do you find more logical?

The historical rationale for making S3 dependencies optional is here: https://github.com/elastic/rally/issues/1095. It's very generic. I could paraphrase as: "the less dependencies, the smaller the collision probability". But we know global Python environment is not a solution. If we had an application deployment model that guarantees dedicated venv, that's not needed at all.

If we want to keep this optional for now, please update https://esrally.readthedocs.io/en/stable/install.html#optional-dependencies section. The `esrally[gs,s3]` notation might work, please check.

tests/storage/gc_test.py
```diff
@@ -0,0 +1,141 @@
...
+    url: str = SOME_URL
+    blob: storage.Blob = dataclasses.field(default_factory=DummyBlob)
+    want_bucket: str = SOME_BUCKET
+    want_blob: str = SOME_BLOB
+
+
+# @pytest.fixture(autouse=True)
+# def default_credentials(monkeypatch: pytest.MonkeyPatch):
+#     credentials = mock.create_autospec(google.auth.credentials.Credentials, instance=True, universe_domain="googleapis.com")
+#     monkeypatch.setattr(google.auth, "default", mock.create_autospec(google.auth.default, return_value=[credentials, None]))
```
gbanasiak: Remove if not needed.

esrally/storage/gc.py
```diff
@@ -0,0 +1,154 @@
...
+        self.client = client
+        self.chunk_size = chunk_size
+        self.multipart_size = multipart_size
+        self.user_project = user_project
+        self.connect_timeout = connect_timeout
+        self.read_timeout = read_timeout
+
+    def blob(self, url: str, fetch: bool = True) -> storage.Blob:
+        # chunk_size value has to be multiple of 256K
+        chunk_size = self.multipart_size + (self.multipart_size % (1024 * 256))
```
gbanasiak: Shouldn't this be moved to constructor and logged? Casual user might be surprised chunk size from config is not followed here.

esrally/storage/gc.py
```diff
@@ -0,0 +1,154 @@
...
+            case "gs":
+                bucket_name = hostname
+                blob_name = path
+            case "https":
+                if hostname != "storage.cloud.google.com":
+                    raise ValueError(f"unexpected hostname: {url}")
+                if "/" in path:
+                    bucket_name, blob_name = path.split("/", maxsplit=1)
+                else:
+                    bucket_name = path
```
gbanasiak: Nice, I haven't thought about `https://storage.cloud.google.com` access.

gbanasiak (changes requested):
Looks good, I left small comments below.

Local testing with small files were OK, but tests with larger downloads show data corruption during decompression. `pzstd` reports `Unknown frame descriptor` while standard library reports `zstd.ZstdError: zstd decompress error: Data corruption detected`. This is easily reproducible. Legacy download works fine.

tests/storage/gc_test.py
fressi-elastic: Removed.

gbanasiak (commented):
This works fine in testing now, I've checked both Linux and MacOS.

Can you add relevant tests to `test_adapter_registry_get()` in `adapter_test.py`?

Unrelated: when testing directly with `esrally-storage` I need to pass the bucket URL as base with `--base-url`, example:

```
esrally-storage --base-url=gs://<bucket> -v get gs://<bucket>/<something>/<file>
```

Is this avoidable?

gbanasiak (approved):
LGTM

PR #1959, opened 2025-07-16
Capture statistics during data dowload so that we can profile mirrors configuration in CI.

fressi-elastic:
This captures additional statistics in storage client and export some of them in transfers status files for later use in CLI.

PR #2008, opened 2025-12-04
It implements `esrally-storage prune` sub-command.

fressi-elastic:
This following up PR #1959

The main purpose is cleaning up all local files written by storage manager on the local directories. It also allows to list file names using the `ls` subcommand.

Example of use:

`esrally-storage ls --filenames` It lists all downloaded data filenames.
`esrally-storage ls --status-filenames` It lists all status filenames.
`esrally-storage ls --filenames --status-filenames` It lists all filenames.

`esrally-storage prune` It deletes all downloaded and status files.
`esrally-storage ls --filenames` It deletes all downloaded data files.
`esrally-storage ls --status-filenames` It deletes all status files.


esrally/storage/_cli.py
```diff
@@ -47,16 +47,19 @@ def main():
...
-    for p in (parser, ls_parser, get_parser, put_parser):
-        p.add_argument("-v", "--verbose", action="count", required=False, default=0, help="Increase verbosity level.")
-        p.add_argument("-q", "--quiet", action="count", required=False, default=0, help="Decrease verbosity level.")
-        p.add_argument("--local-dir", type=str, default=cfg.local_dir, help="destination directory for downloading files")
-        p.add_argument("--base-url", type=str, default=None, help="base URL for remote storage.")
-        p.add_argument("--mirror-failures", action="store_true", help="It upload only files that have recorded mirror failures.")
+    ls_parser = subparsers.add_parser("ls", help="It lists file(s) downloaded from ES Rally remote storage services.")
+    get_parser = subparsers.add_parser("get", help="It downloads file(s) from ES Rally rem1ote storage services.")
+    put_parser = subparsers.add_parser("put", help="It uploads file(s) to mirror server.")
+    prune_parser = subparsers.add_parser("prune", help="It delete transfer files from local directories.")
```
gbanasiak:
```suggestion
    prune_parser = subparsers.add_parser("prune", help="It deletes transfer files from local directories.")
```

esrally/storage/_transfer.py
```diff
@@ -727,3 +727,17 @@ def _check_finished_document(self) -> None:
                 checksum = _crc32c.Checksum.from_filename(self.path)
                 if checksum != want_checksum:
                     raise ValueError(f"Unexpected checksum: {checksum}, want {want_checksum}")
+
+    def prune(self):
+        self.close()
+        errors: list[Exception] = []
+        for p in [self.path, self.status_file_path]:
```
gbanasiak: This wipes both the status and the _downloaded_ files. In practice that is OK because what remains is the _decompressed_ file, so Rally can run another race without downloading anything. However, in case I'm missing something, it wouldn't hurt to have an option to prune status files only. WDYT?

fressi-elastic: OK no problem. It is easy to add a flag like `--status-files` to select only those files.

fressi-elastic: I added two new flags: `--filenames` for transferred files and `--status-filenames` for status files. You can use both and you can also pass these flags to the `ls` command for listing these files.

gbanasiak (approved):
LGTM

PR #2009, opened 2025-12-04
Fixes best_match versions algorithm in case of no major version

NickDris:
- Fixed a best_match case where X.0 was not selected if X major is absent in the alternatives

### DIFF
--- a/esrally/utils/versions.py
+++ b/esrally/utils/versions.py
@@ -177,9 +177,9 @@ def best_match(available_alternatives, distribution_version):
             if version in available_alternatives:
                 return version
             # match nearest prior minor
-            if version_type == "with_minor" and (latest_minor := latest_bounded_minor(available_alternatives, versions)):
-                if latest_minor:
-                    return f"{versions.major}.{latest_minor}"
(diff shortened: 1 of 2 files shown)


esrally/utils/versions.py
```diff
@@ -177,9 +177,9 @@ def best_match(available_alternatives, distribution_version):
             if version in available_alternatives:
                 return version
             # match nearest prior minor
-            if version_type == "with_minor" and (latest_minor := latest_bounded_minor(available_alternatives, versions)):
-                if latest_minor:
-                    return f"{versions.major}.{latest_minor}"
+            latest_minor = latest_bounded_minor(available_alternatives, versions)
+            if version_type == "with_minor" and latest_minor is not None:
+                return f"{versions.major}.{latest_minor}"
```
gbanasiak: Short version still works:

```
if version_type == "with_minor" and (latest_minor := latest_bounded_minor(available_alternatives, versions)) is not None:
  return f"{versions.major}.{latest_minor}"
```

gbanasiak (approved):
LGTM

PR #2012, opened 2025-12-09
Fix `it/storage/cli_test.py`

fressi-elastic:
Don't rely on `find` command output order for comparing results with `want_files`.


gbanasiak (approved):
LGTM

PR #2015, opened 2025-12-11
Modify storage CLI to produce multiple objects for transference stats and mirror failures when using `--filebeat` option. 

fressi-elastic:
This small refractory of storage CLI and Transfer class cleans up how storage manager writes the status to JSON and filebeats formats. There are changes how the mirror failures and transfer statistics are being written to status file and how they are presented in the `esrally-storage ls` command. It also fixes the `--stats` and `--mirror-failures` flags by hiding respective details when these flags are not present. These details are always written to status files, so they can always be shown in a further command execution.

Issue: https://elasticco.atlassian.net/browse/ES-13743

### DIFF
--- a/esrally/storage/_cli.py
+++ b/esrally/storage/_cli.py
@@ -17,7 +17,6 @@
 from __future__ import annotations
 
 import argparse
-import dataclasses
 import json
 import logging
 import os.path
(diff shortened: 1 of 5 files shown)


esrally/storage/_cli.py
```diff
@@ -181,34 +183,110 @@ def ls(
...
-            "has_mirror_failures": bool(tr.mirror_failures),
-            "todo": str(tr.todo),
-            "finished": tr.finished,
-            "stats": {u: dataclasses.asdict(s) for u, s in tr.stats.items()},
-        }
-        for tr in transfers
-    ]
+class BaseTransferDict(TypedDict):
+    url: str
+    path: str
```
gbanasiak, line 205: Thank you for adding typed dicts.

gbanasiak (approved):
LGTM

PR #2016, opened 2025-12-15
Use wait_for_completion:False instead of relying on timeouts in ForceMerge

gareth-ellis:
Relying on a timeout in the force merge runner can result in other errors - e.g if a load balancer kills a connection, typically we would see an SSL related error instead. This change changes polling method to use wait_for_completion:False instead and then polls via the task API.

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -700,15 +700,27 @@ class ForceMerge(Runner):
     """
 
     async def __call__(self, es, params):
+        # pylint: disable=import-outside-toplevel
+        import elasticsearch
+
         max_num_segments = params.get("max-num-segments")
         mode = params.get("mode")
         merge_params = self._default_kw_params(params)
         if max_num_segments:
             merge_params["max_num_segments"] = max_num_segments
         if mode == "polling":
-            complete = False
-            merge_params["wait_for_completion"] = False
-            await es.indices.forcemerge(**merge_params)
+            es_info = await es.info()
+            es_version = Version.from_string(es_info["version"]["number"])
+            if es_version < Version(8, 1, 0):
+                try:
+                    await es.indices.forcemerge(**merge_params)
+                    complete = True
+                except elasticsearch.ConnectionTimeout:
+                    pass
+            else:
+                complete = False
+                merge_params["wait_for_completion"] = False
+                await es.indices.forcemerge(**merge_params)
             while not complete:
                 await asyncio.sleep(params.get("poll-period"))
                 tasks = await es.tasks.list(params={"actions": "indices:admin/forcemerge"})
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -709,6 +709,7 @@ class ForceMerge(Runner):
         if max_num_segments:
             merge_params["max_num_segments"] = max_num_segments
         if mode == "polling":
+            task_id = None
             complete = False
             es_info = await es.info()
             es_version = Version.from_string(es_info["version"]["number"])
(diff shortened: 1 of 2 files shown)


esrally/driver/runner.py
```diff
@@ -700,21 +700,15 @@ class ForceMerge(Runner):
...
         if max_num_segments:
             merge_params["max_num_segments"] = max_num_segments
         if mode == "polling":
             complete = False
-            try:
-                await es.indices.forcemerge(**merge_params)
-                complete = True
-            except elasticsearch.ConnectionTimeout:
-                pass
+            merge_params["wait_for_completion"] = False
```
fressi-elastic: You could pass the parameter directly in the call. To me it looks simpler to reed.

`````Python
await es.indices.forcemerge(wait_for_completion=False, **merge_params)
```

esrally/driver/runner.py
```diff
@@ -700,21 +700,15 @@ class ForceMerge(Runner):
...
         mode = params.get("mode")
         merge_params = self._default_kw_params(params)
         if max_num_segments:
             merge_params["max_num_segments"] = max_num_segments
         if mode == "polling":
             complete = False
-            try:
-                await es.indices.forcemerge(**merge_params)
-                complete = True
-            except elasticsearch.ConnectionTimeout:
```
fressi-elastic, line 716: Is `except elasticsearch.ConnectionTimeout` anymore raised? Would it affect negatively keeping its handling as before?

fressi-elastic (approved):
Please have a look to my comments. Overall it LGTM

fressi-elastic (changes requested):
Resulting unit test failures looks related. Please fix it.

esrally/driver/runner.py
```diff
@@ -709,12 +709,18 @@ async def __call__(self, es, params):
         if max_num_segments:
             merge_params["max_num_segments"] = max_num_segments
         if mode == "polling":
-            complete = False
-            try:
+            es_info = await es.info()
```
gbanasiak, line 714: Can we use `es.distribution_version` instead? I would assume `es` is an instance of `RallyAsyncElasticsearch` (although lack of types does not make it obvious)?

gbanasiak:
```
% rg 'cfg.add.*"mechanic", "distribution.version"'
esrally/rally.py
1107:                cfg.add(config.Scope.applicationOverride, "mechanic", "distribution.version", args.distribution_version)

esrally/racecontrol.py
211:            self.cfg.add(config.Scope.benchmark, "mechanic", "distribution.version", distribution_version)
```

It seems we only set this when running with `--pipeline=benchmark-only` or when `--distribution-version` is provided, so it won't be set when rally builds and runs a specific ES revision. We can scratch that idea I think.

esrally/driver/runner.py
gareth-ellis: I've reworked this to include the previous behaviour for backwards compatibility - the route for 8.1.0 and later doesnt include catching, since the request should now return near instantly, and a timeout should be treated as an error instead

tests/driver/runner_test.py
```diff
@@ -1563,15 +1563,262 @@ async def test_force_merge_with_params(self, es):
     @pytest.mark.asyncio
     async def test_force_merge_with_polling_no_timeout(self, es):
         es.indices.forcemerge = mock.AsyncMock()
+        es.info = mock.AsyncMock(
```
gbanasiak, line 1639: There's some repetition in `es.info` and `es.tasks.list` mocks, pytest fixtures or shared methods could help reduce it.

gbanasiak (approved):
LGTM.

When run with `wait_for_completion=true`, force merge returns task ID. This task ID could be used to filter the list of force merge tasks in case there's parallel unrelated force merge running in the cluster. WDYT about adding it as part of this PR?

tests/driver/runner_test.py
gareth-ellis: Take a look at 3d036462dd8830bcf7751200e55b60a330233ad5

fressi-elastic (approved):

PR #2019, opened 2025-12-16
It always shows mirror failures when there is any.

fressi-elastic:
In order to store connectivity statistics we would like to create a file to be consumed by `filebeat` with all the informations required for investigating corpora download issue and monitor its reliability and speed. With this purpose the `esrally-storage ls` sub-command now always shows mirror failures, without requiring `--mirror-failures` flag which is intended to ignore those transfers that have any mirror failure (for example in order to populate a mirror bucket using `esrally-storage put` sub-command).


gbanasiak (approved):
LGTM

PR #2004, opened 2025-12-02
Add sampling method parameter to downsampling

gmarouli:
As part of https://github.com/elastic/elasticsearch/issues/128357 we introduce a new downsampling method that can be configured via the request body. In order to performance test this method we want to expose this parameter in rally,

PR #2020, opened 2025-12-19
Parse sampling value in DownsampleParamSource

gmarouli:
In #2004 we introduced a new parameter to the downsampling operation but we did not parse it correctly in `DownsampleParamSource`

### DIFF
--- a/esrally/track/params.py
+++ b/esrally/track/params.py
@@ -866,9 +866,12 @@ class DownsampleParamSource(ParamSource):
         params["index"] = params.get("source-index")
         self._source_index = get_target(track, params)
         self._target_index = params.get("target-index", f"{self._source_index}-{self._fixed_interval}")
+        self._sampling_method = params.get("sampling-method", None)
 
     def params(self):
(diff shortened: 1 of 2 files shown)


tests/track/params_test.py
gareth-ellis, line 2968: Should we be asserting that the sampling-method param is available here too?

esrally/track/params.py
```diff
@@ -866,9 +866,12 @@ def __init__(self, track, params, **kwargs):
         params["index"] = params.get("source-index")
         self._source_index = get_target(track, params)
         self._target_index = params.get("target-index", f"{self._source_index}-{self._fixed_interval}")
+        self._sampling_method = params.get("sampling-method", None)
```
gareth-ellis: The default value for dict.get is `None` already

gareth-ellis (approved):
LGTM, just one comment and one nit

tests/track/params_test.py
gmarouli: Yes, 💯

fressi-elastic (commented):
LGTM

PR #2023, opened 2026-01-13
It fiexes `esrally-storage` put command target file names when multiple files are copied

fressi-elastic:
When multiple files have been selected for upload (eventually because of a mirror failure, ```esrally storage put``` command is misformating target directory to copy them to.

It fixes the issue and add integration test cases to cover more realistic user cases.

It fixes bug ES-13906


gbanasiak (approved):
LGTM

PR #2024, opened 2026-01-13
Upgrade Elasticsearch Python client to 9.x and urllib3 2.x

### LINKED ISSUE #2018: Update urllib to something greater than 2.5
https://github.com/elastic/rally/blame/87d729ba5ef9da753eb4f435f9d20ad29b1f553e/pyproject.toml#L50

URLlib is currently pinned at 1.26 which has known security vulnerabilities. Update to modern version.

fressi-elastic:
## Summary

Upgrades Rally’s Elasticsearch dependencies to the 9.x Python client stack and urllib3 2.x, updates the sync/async Rally client subclasses to match current transport APIs and stricter Elasticsearch 9 request expectations, and tightens documented Elasticsearch version support for Rally 2.14.0+.

## Dependency changes
- elasticsearch[async] ==9.3.0
- elastic-transport ==9.2.1
- urllib3 ==2.6.3

Client factory uses connections_per_node (replacing the old pool maxsize-style parameter) — see `esrally/client/factory.py`

## Behavior and compatibility

Minimum benchmark target is Elasticsearch 8.0.0 `esrally/min-es-version.txt`; support for 6.8 / 7.x as benchmark clusters is dropped with this release line.

Rally extensions around the official client remain focused on compatibility mode (talking to ES 8 with a 9.x client where appropriate), header handling (including bulk application/x-ndjson where ES 9 is strict), and serverless behavior (e.g. not sending compatible-with where it does not apply).

Legacy _ProductChecker / product-header workaround code paths are removed; client code is aligned with upstream patterns where possible.

## Documentation

`docs/versions.rst`: EOL / support policy for 2.14.0+ (current + previous major) and a Rally ↔ Elasticsearch compatibility matrix.

Other doc updates align examples and migration/add-track guidance with 8.x / 9.x (no OSS-distribution wording that no longer applies).

# Testing

Integration tests exercise version.minimum_es_version() (8.0.0) and a 9.x distribution — see `it/__init__.py`.

Unit tests cover the refactored client helpers and perform_request behavior.

# Related work

Track dependency fix for elastic/security (pinned elasticsearch / elastic-transport in tracks) lives in rally-tracks (see linked PR there); Rally does not need a long-term loader workaround for that case.

Closes #2018 (if still accurate when merging).

# Release / rollout notes

Intended to land in a 2.14.x Rally line after a 2.13.x release so users still on old clusters have a final supported Rally version (per maintainer discussion on the PR).

Performance A/B vs the old client was not explicitly benchmarked in isolation; rely on CI, integration tests, and optional nightly/staging soak as needed.

### DIFF
--- a/it/conftest.py
+++ b/it/conftest.py
@@ -30,7 +30,7 @@ def check_prerequisites():
     print("Checking prerequisites...")
     if process.run_subprocess_with_logging("docker ps") != 0:
         raise AssertionError("Docker must be installed and the daemon must be up and running to run integration tests.")
-    if process.run_subprocess_with_logging("docker compose --help") != 0:
+    if process.run_subprocess_with_logging("docker compose version") != 0:
         raise AssertionError("Docker Compose is required to run integration tests.")
 
 


esrally/client/asynchronous.py
```diff
@@ -330,12 +335,10 @@ async def put_lifecycle(self, *args, **kwargs):
...
 class RallyAsyncElasticsearch(AsyncElasticsearch, RequestContextHolder):
-    def __init__(self, *args, **kwargs):
-        distribution_version = kwargs.pop("distribution_version", None)
-        distribution_flavor = kwargs.pop("distribution_flavor", None)
-        super().__init__(*args, **kwargs)
+    def __init__(self, hosts: Any = None, *, distribution_version: str | None = None, distribution_flavor: str | None = None, **kwargs):
+        super().__init__(hosts, **kwargs)
         # skip verification at this point; we've already verified this earlier with the synchronous client.
-        # The async client is used in the hot code path and we use customized overrides (such as that we don't
+        # The async client is used in the hot code path, and we use customized overrides (such as that we don't
```
gareth-ellis: There shouldnt be a comma here -normally you don't use a comma after and.

esrally/track/loader.py
```diff
@@ -231,19 +230,90 @@ def load_track(cfg: types.Config, install_dependencies=False):
...
-                    [sys.executable, "-m", "pip", "install", *dependencies, "--upgrade", "--target", paths.libs()],
-                    stdout=install_log,
-                    stderr=install_log,
-                )
-        except subprocess.CalledProcessError:
-            raise exceptions.SystemSetupError(f"Installation of track dependencies failed. See [{install_log.name}] for more information.")
+DEPRECATED_PACKAGES = (
+    "elasticsearch",
+    "elastic-transport",
+)
```
gbanasiak: I would be for changing the `elastic/security` track instead of adding whole new functionality in Rally. I think this is an overkill.

gbanasiak (dismissed):
A few areas of concern:
* How does this change affect the range of supported Elasticsearch versions? This needs to be reflected in Rally documentation.
* If the intention is to drop support for ES 6.8 and 7.x as part of client upgrade, we should release one more Rally version before merging this update IMO. It's been a while since we released.
* What is the test plan here? Go to nightly staging, and then nightly production, and revert if things go south? I would prefer to test this more extensively before the merge (e.g. by running all nightlies in test mode).
* Why `_ProductChecker` is kept in `synchronous.py`? This client upgrade is an opportunity to simplify, because the following no longer holds:

> Because Rally needs to support versions of ES >= 6.8.0, we resurrect the previous
logic for determining the authenticity of the server, which does not rely exclusively
on this header.

esrally/track/loader.py
fressi-elastic: Track are actively maintained and the issue could be reintroduced again. The intention of this is to add a limiter so that this will not happen again. There can't be a collision between rally requirements and tracks requirements because they are using the same venv. The same problem could be happening with other tracks we are not looking at. The intention of this is deprecating a bad thing that was working in the past and it is very simple to implement. If you want to remove this then we have to check all the other tracks too. I would leave it for a following up step.

gbanasiak: `elastic/security` is the only exception:

```
[26-01-22 12:05] grzegorz:src/rally-tracks% rg '"(elasticsearch|elastic-transport)( )?[=>]='
elastic/security/track.json
357:    "elastic-transport==8.4.1",
358:    "elasticsearch==8.6.1"
```

These dependencies should not be there. They don't make sense, because Rally includes them. They were introduced by mistake in https://github.com/elastic/rally-tracks/pull/675 which wasn't caught or understood in the review. To me a natural solution in such case is to remove the one-time bug in a track instead of adding new bits into Rally code, hence my first comment.

Now thinking how to solve this in a generic way... what we could do is detect if someone declared any track dependency that is already met in running Rally code (why make ES client special?) and throw an error. The idea of track dependencies is that they bring _extras_ i.e. dependencies that do not exist in Rally at all. We cannot support multiple versions of any dependencies because everything runs in the same Python environment. _But_ this is not a priority, and I don't think should be done as part of ES client upgrade.

If we want to keep current proposal, I would change the following message from "deprecate" to "forbid" (and "deprecated" to "forbidden"), and throw an error.

> DEPRECATED_MESSAGE = (
    "Track dependencies are deprecated: {deprecated}. "
    "Please update the track to rely on the 'elasticsearch' libraries versions as required by Rally. "
    "Rally no longer supports installing per-track Elasticsearch client libraries."
)

I would advise to remove the listed libraries instead of updating them. I would also not mention ES libraries explicitly, perhaps we would like to extend this list in the future. The last sentence suggest per-track ES client libraries were OK at some point, but were they? What was done in `elastic/security` only worked because the versions were equal, but an arbitrary version was never supported.

Either way, let's fix `elastic/security` too by removing these dependencies. It's a bad example.

fressi-elastic: You convinced me! I am going to fix rally-tracks and remove this part from rally.

fressi-elastic: I created PR https://github.com/elastic/rally-tracks/pull/1000 to handle this issue. I

It is pending:
- [ ] to get the PR merged.
- [ ] to remove the workaround logic here.

fressi-elastic: Done. Now waiting for merge of PR elastic/rally-tracks#1000

esrally/client/asynchronous.py
```diff
@@ -356,41 +360,35 @@ def options(self, *args, **kwargs):
...
         if not self.is_serverless:
-            if versions.is_version_identifier(self.distribution_version) and (
-                versions.Version.from_string(self.distribution_version) >= versions.Version.from_string("8.0.0")
-            ):
-                _mimetype_header_to_compat("Accept", request_headers)
-                _mimetype_header_to_compat("Content-Type", request_headers)
+            # Converts all parts of a Accept/Content-Type headers
+            # from application/X -> application/vnd.elasticsearch+X
+            # see https://github.com/elastic/elasticsearch/issues/51816
+            mimetype_headers_to_compat(request_headers, self.distribution_version)
```
gbanasiak: Do we need `if not self.is_serverless` condition? I don't see anything like it in 9.2.1 version of ES Python client, see https://github.com/elastic/elasticsearch-py/blob/2022b70b5c2139c3c5d9c47d038f022a1cb1f7ca/elasticsearch/_sync/client/_base.py#L307-L308 which does support serverless I presume.

If this goes away, can we get rid of `_perform_request()` completely?

pquentin: Not sending `compatible-with` on serverless is more correct, that's a missing feature in the Python client. It's harder to fix because we don't *know* if something is serverless when talking to it initially, so the only thing we can do is add an opt-in serverless mode.

fressi-elastic: In the new client there are new features like supporting open telemetry. I decided to keep it and have the client code as much similar to the official version to make its maintenance simpler for the next upgrades. `_perform_request` is the actual new implementation in the base client, it is being wrapped to support open telemetry and I keep it the same. `is_serverless` condition makes things bit more complex but required to avoid sending an unsupported header to serverless. Anyway the header is only send if the server version is known that could not be the case (and it should work anyway I guess). The fact no CI job was broken by testing it with either 8 and 9 family give me confidence this approach could actually work and it is enough simple.

esrally/client/asynchronous.py
```diff
@@ -356,41 +360,35 @@ def options(self, *args, **kwargs):
...
             request_headers = self._headers.copy()
             request_headers.update(headers)
         else:
             request_headers = self._headers
 
-        # Converts all parts of a Accept/Content-Type headers
-        # from application/X -> application/vnd.elasticsearch+X
-        # see https://github.com/elastic/elasticsearch/issues/51816
-        # Not applicable to serverless
+        if body is not None:
```
fressi-elastic: I think this block is only required to avoid updating our tracks to include these headers. I wonder why the headers handling wasn't moved to an higher level to leave this client patching much simpler. @pquentin What do you think?

esrally/client/asynchronous.py
```diff
@@ -356,41 +360,35 @@ def options(self, *args, **kwargs):
...
+            # It ensures content-type and accept headers are set.
+            mimetype = "application/json"
+            if path.endswith("/_bulk"):
```
fressi-elastic: API protocol 9 is more picky and requires `application/x-ndjson` for bulk requests.

esrally/client/asynchronous.py
```diff
@@ -430,6 +429,19 @@ async def perform_request(
 
             raise HTTP_EXCEPTIONS.get(meta.status, ApiError)(message=message, meta=meta, body=resp_body)
 
+        # 'X-Elastic-Product: Elasticsearch' should be on every 2XX response.
```
fressi-elastic: This is intended to keep the code much similar to the official client.

esrally/client/common.py
```diff
@@ -3,26 +3,33 @@
...
-        request_headers[header] = _COMPAT_MIMETYPE_RE.sub(_COMPAT_MIMETYPE_SUB, mimetype)
+def mimetype_headers_to_compat(
+    headers: elastic_transport.HttpHeaders,
+    distribution_version: str | None = None,
+) -> None:
+    if not headers:
+        return
+
+    major_version = 8
+    if versions.is_version_identifier(distribution_version):
```
fressi-elastic: It only uses compatibility version 9 when it does knows server version is 9, else it always uses version 8.

docs/adding_tracks.rst
```diff
@@ -295,11 +295,11 @@ You can also show details about your track with ``esrally info --track-path=~/ra
     5. force-merge
     6. query-match-all (8 clients)
 
-Congratulations, you have created your first track! You can test it with ``esrally race --distribution-version=7.14.1 --track-path=~/rally-tracks/tutorial``.
+Congratulations, you have created your first track! You can test it with ``esrally race --distribution-version=9.2.4 --track-path=~/rally-tracks/tutorial``.
```
gbanasiak: Could we use a variable for this and other explicit versions to avoid updating this in the next cycle? As we make some effort to scan documentation for version references, it is a good opportunity to make it easier in the future. Could we use `|min_es_version|`, or another variable with the most recent ES version?

docs/command_line_reference.rst
```diff
@@ -109,22 +109,22 @@ Because ``--quiet`` is specified, Rally will suppress all non-essential output (
 
 This subcommand can be used to download Elasticsearch distributions. Example::
 
-    esrally download --distribution-version=6.8.0 --quiet
+    esrally download --distribution-version=9.2.4 --quiet
 
-This will download the OSS distribution of Elasticsearch 6.8.0. Because ``--quiet`` is specified, Rally will suppress all non-essential output (banners, progress messages etc.) and only return the location of the binary on the local machine after it has downloaded it::
+This will download the OSS distribution of Elasticsearch 9.2.4. Because ``--quiet`` is specified, Rally will suppress all non-essential output (banners, progress messages etc.) and only return the location of the binary on the local machine after it has downloaded it::
```
gbanasiak: OSS distributions of Elasticsearch were discontinued in 7.x so "OSS distribution of ES 9.2.4" does not make sense, please update.

docs/migrate.rst
```diff
@@ -213,7 +213,7 @@ Previously a subcommand was optional when running a benchmark. With Rally 2.1.0
 
 Invoke Rally with the ``race`` subcommand instead::
 
-    esrally race --distribution-version=7.10.0
+    esrally race --distribution-version=9.2.4
```
gbanasiak: Do not change versions in `migrate.rst`, it is append-only history. For instance, the lines I highlighted modify remarks for version `2.1.0` which did not support version 9.

I would document breaking changes in a new section for 2.14.0 at the top: dropping support for ES 6.8 and 7.x, and changing ES Python client to 9.x which might potentially be breaking custom runner code in some private tracks.

docs/versions.rst
```diff
@@ -10,9 +10,22 @@ However, Rally does not support Elasticsearch clusters using the OSS license.
...
+   :widths: 40 60
+
+   * - Rally release
+     - Minimum supported Elasticsearch
+   * - < 2.14.0
+     - 6.8.0 and above
+   * - >= 2.14.0
+     - |min_es_version| and above
+
+As a rule of thumbs for the future as `Elasticsearch version reaches end-of-life <https://www.elastic.co/support/eol>`_, Rally may drop support in a future release with the upgrade of Elasticsearch Python client on which rally depends on.
```
gbanasiak: "As a rule of thumb" and "may drop" sounds a bit too vague and frivolous. Rally EOL policy needs to be deterministic IMO.

docs/adding_tracks.rst
```diff
@@ -295,11 +295,11 @@ You can also show details about your track with ``esrally info --track-path=~/ra
...
 
 .. note::
 
-    To test the track with Elasticsearch prior to 7.0.0 you need to update ``index.json`` and ``track.json`` as specified in notes above and then execute ``esrally race --distribution-version=6.5.3 --track-path=~/rally-tracks/tutorial``.
+    Rally supports Elasticsearch 8.0.0 and above. The example above uses 9.2.4. For older track format (e.g. mapping types), see the notes earlier in this section.
```
gbanasiak: Keeping ES version 7 references doesn't make sense with this update. I would remove them. That's the purpose of getting rid of old version support, simplification.

For instance this fragment:

```
If you want to benchmark Elasticsearch prior to 7.0.0 you need to add the ``types`` property above so ``track.json`` will look like::

      ...
      "indices": [
        {
          "name": "geonames",
          "body": "index.json",
          "types": [ "docs" ] <---- THIS
        }
      ],
      ...
```

The `types` property refers to something that no longer exists. See https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html for historical background. Types support should be removed from Rally if not as part of this PR, then in a follow-up.

docs/versions.rst
```diff
@@ -10,9 +10,22 @@ However, Rally does not support Elasticsearch clusters using the OSS license.
 End-of-life Policy
 ==================
 
-The latest version of Rally allows to benchmark all currently supported versions of Elasticsearch. Once an `Elasticsearch version reaches end-of-life <https://www.elastic.co/support/eol>`_, Rally will support benchmarking its last minor version until a new major version comes out. For example, after the release of Elasticsearch 8.0.0, Rally dropped support for Elasticsearch < 6.8.0.
+The following table summarizes the minimum supported Elasticsearch version per Rally release:
```
gbanasiak: We should state the new policy here. Previous policy was "all currently supported ES versions" (which leads to N-2). Now we want to change it to N-1.

docs/versions.rst
```diff
@@ -10,9 +10,22 @@ However, Rally does not support Elasticsearch clusters using the OSS license.
...
+.. list-table:: Minimum supported Elasticsearch version
+   :header-rows: 1
+   :widths: 40 60
+
+   * - Rally release
+     - Minimum supported Elasticsearch
+   * - < 2.14.0
+     - 6.8.0 and above
+   * - >= 2.14.0
+     - |min_es_version| and above
```
gbanasiak: Let's provide supported ES range instead of minimum supported ES. I would also provide one historic step more:

* Rally 2.7.1 was using 7.14.0 ES client, and supported ES range was 6.8 - 8.x
* Rally 2.8.0 <= 2.13.0 was using 8.6.1 ES client, and supported ES range was 6.8 - 9.x
* Rally >= 2.14.0 uses 9.2.1 ES client, and supports ES range is 8.0 - 10.x

esrally/client/asynchronous.py
```diff
@@ -361,101 +345,20 @@ async def perform_request(
...
         *,
-        params: Optional[Mapping[str, Any]] = None,
-        headers: Optional[Mapping[str, str]] = None,
-        body: Optional[Any] = None,
+        params: Mapping[str, Any] | None = None,
+        headers: Mapping[str, str] | None = None,
+        body: Any = None,
+        endpoint_id: str | None = None,
+        path_parts: Mapping[str, Any] | None = None,
+        compatibility_mode: int | None = None,
```
gbanasiak, line 353: I would make a comment that `compatibility_mode` is a Rally extension.

esrally/client/asynchronous.py
```diff
@@ -361,101 +345,20 @@ async def perform_request(
...
-            request_timeout=self._request_timeout,
-            max_retries=self._max_retries,
-            retry_on_status=self._retry_on_status,
-            retry_on_timeout=self._retry_on_timeout,
-            client_meta=self._client_meta,
+            compatibility_mode=compatibility_mode
+            or common.compatibility_mode_from_distribution(version=self.distribution_version, flavour=self.distribution_flavor),
+        )
+        return await super().perform_request(
+            method, path, params=params, headers=headers, body=body, endpoint_id=endpoint_id, path_parts=path_parts
```
gbanasiak, line 382: Does this actually work the way we want considering headers are touched _again_ in parent's method?

esrally/client/common.py
```diff
@@ -3,26 +3,55 @@
...
-def _mimetype_header_to_compat(header, request_headers):
-    # Converts all parts of a Accept/Content-Type headers
-    # from application/X -> application/vnd.elasticsearch+X
-    mimetype = request_headers.get(header, None) if request_headers else None
-    if mimetype:
-        request_headers[header] = _COMPAT_MIMETYPE_RE.sub(_COMPAT_MIMETYPE_SUB, mimetype)
+def compatibility_mode_from_distribution(*, version: str | None = None, flavour: str | None = None) -> int | None:
+    if flavour and versions.is_serverless(flavour):
+        # Serverless doesn't need compatibility mode.
+        return None
```
gbanasiak: This should also work:

```suggestion
    if versions.is_serverless(flavour):
        # Serverless doesn't need compatibility mode.
        return None
```

esrally/client/common.py
```diff
@@ -3,26 +3,55 @@
...
+    if version and versions.is_version_identifier(version):
+        return int(versions.Version.from_string(version).major)
+
+    # To improve compatibilit it prefers to ignore compatibility mode in case it can't understand distribution version.
```
gbanasiak: Small typo...

```suggestion
    # To improve compatibility it prefers to ignore compatibility mode in case it can't understand distribution version.
```

BTW, should we standardize the style of comments, PR titles and commit titles? For instance in case of comments we have 3 different styles now:
- with "it" like in the example above,
- without "it", e.g. "Avoids using compatibility headers by default because..." (or "Avoid" for imperative),
- with "we", e.g. https://github.com/elastic/rally/blob/8ce95d7d74b9cfd087e606d4eba56f35d41fad8f/esrally/utils/git.py#L57

My preference would be to keep imperative mode in commit titles and PR titles (e.g. "Upgrade Elasticsearch Python client..."), and avoid using "we" or "it" in code comments. I am sure I've been using "we" myself, so this only works if we all align and check mutually :)

esrally/client/common.py
```diff
@@ -3,26 +3,55 @@
...
+    headers = elastic_transport.HttpHeaders(headers or {})
+
+    if body is not None:
+        # It ensures content-type and accept headers are initialized.
+        mimetype = "application/json"
+        if path is not None and path.endswith("/_bulk"):
+            # Server version 9 is picky about this. This should improve compatibility.
+            mimetype = "application/x-ndjson"
+        for header in ("content-type", "accept"):
+            headers.setdefault(header, mimetype)
```
gbanasiak, line 41: This seems weird. Why is this required? Is there anything similar in Elasticsearch Python client code? If 9.x version of the client can talk to 9.x version of the server by default, why do we need any extra modifications, e.g. for `_bulk`?

esrally/client/synchronous.py
```diff
@@ -145,113 +46,20 @@ def perform_request(
...
-            request_timeout=self._request_timeout,
-            max_retries=self._max_retries,
-            retry_on_status=self._retry_on_status,
-            retry_on_timeout=self._retry_on_timeout,
-            client_meta=self._client_meta,
+            compatibility_mode=compatibility_mode
+            or common.compatibility_mode_from_distribution(version=self.distribution_version, flavour=self.distribution_flavor),
+        )
+        return super().perform_request(
+            method=method, path=path, params=params, headers=headers, body=body, endpoint_id=endpoint_id, path_parts=path_parts
```
gbanasiak, line 82: Same remark as in async version.

it/__init__.py
```diff
@@ -31,10 +31,7 @@
 from esrally.utils import process
 
 CONFIG_NAMES = ["in-memory-it", "es-it"]
-DISTRIBUTIONS = ["8.4.0"]
-# There are no ARM distribution artefacts for 6.8.0, which can't be tested on Apple Silicon
-if platform.machine() != "arm64":
-    DISTRIBUTIONS.insert(0, "6.8.0")
+DISTRIBUTIONS = ["8.4.0", "9.2.4"]
```
gbanasiak: I would use `8.0.0` as the lowest test version.

tests/client/synchronous_test.py
```diff
@@ -0,0 +1,153 @@
...
+        client,
+        case.method,
+        case.path,
+        params=None,
+        headers=mock.ANY,
+        body=case.body,
+        endpoint_id=None,
+        path_parts=None,
+    )
+    passed_headers = mock_super.call_args[1]["headers"]
```
gbanasiak: This verifies how parent's `perform_request()` is called but it can modify the headers again...

docs/adding_tracks.rst
fressi-elastic: You are right. Which version should I use here? I didn't tested this change. I should probably skip this modification instead?

docs/adding_tracks.rst
fressi-elastic: I think this document requires editing from somebody that actually knows what it is about and it is out of scope for this upgrade. It is a tutorial that explain how to create a new track. It is hasn't the same importance and priority than upgrading the client. What do you think?

esrally/client/asynchronous.py
fressi-elastic: The headers will be touched only once if the compatibility mode is set here. Elsewhere they will be modified by the superclass with 9 as a value, which will lead to failures with server version 8. As a rule I would try to use version 8 compatibility mode as much as possible, unless the track specify elsewhere to keep compatibility.

esrally/client/common.py
fressi-elastic: Older versions of server where assuming JSON type has to be assumed if not specified when the body was clearly a Json. Newer versions are getting more strict, so we have to provide the headers which the track is not providing to allow posting jsons or bulk objects to the server. Version 9 is more picky than version 8.

fressi-elastic: Rally already had this bofore my change for the reason I explained here.

tests/client/synchronous_test.py
fressi-elastic: In the new implementation of the client, the passed header is alway copied, before being modified. Please look at common.py.

docs/migrate.rst
fressi-elastic: I dropped changes to migrate.rst

esrally/client/asynchronous.py
fressi-elastic: I now mentioned in the docstrings.

esrally/client/common.py
fressi-elastic: OK

it/__init__.py
fressi-elastic: I am now using `version.minimum_es_version()` to have it automatically updated.

tests/client/synchronous_test.py
fressi-elastic: I am thinking to actually add checks for the final header passed to the transport. Thank for the suggestion.

esrally/client/asynchronous.py
pquentin: > I wonder why the headers handling wasn't moved to an higher level to leave this client patching much simpler. @pquentin What do you think?

To be clear, what Rally is doing here is totally unsupported by the client. It's not meant to be patched at all.

gbanasiak:
@fressi-elastic I've worked on style and content of EOL policy, see https://github.com/elastic/rally/pull/2024/commits/cdc94e90671c966e1f49347d69547b93e3b99ae5.

docs/adding_tracks.rst
gbanasiak: I've raised https://github.com/elastic/rally/issues/2036, and added https://github.com/elastic/rally/pull/2024/commits/ea1311aa385fe8597b1c1009dbc7676354a89d4d.

docs/adding_tracks.rst
gbanasiak: I would use `|ES_CLIENT_VER|`. This needs `parsed-literal` block, see example in https://github.com/elastic/rally/pull/2024/changes/ea1311aa385fe8597b1c1009dbc7676354a89d4d.

esrally/client/asynchronous.py
gbanasiak: As per earlier discussion in https://github.com/elastic/rally/pull/2024#discussion_r2741344203 I thought the desirable behavior while overriding Python client is the following:
* initially do not send `compatible-with` header, this applies to first synchronous `GET /` which is used to determine ES version and flavor (serverless or not),
* for the following requests use `compatible-with=8` or `compatible-with=9` depending on the detected stateful ES version, or do not use `compatible-with` at all in case of serverless.

With the most recent simplification, unless I'm mistaken, we're initially sending `compatible-with=9` and then drop to `compatible-with=8` for stateful ES 8, or keep using `compatible=with=9` elsewhere. I assume this is passing all the tests but is not the cleanest option. I'm just making sure this is intentional. We are trading simplicity for "correctness" here.

tests/client/synchronous_test.py
fressi-elastic: I made test case checking headers passed to transport.

docs/adding_tracks.rst
fressi-elastic: I skipped changes to this file as this goes outside the scope of this change.

esrally/client/common.py
gbanasiak: `BulkIndex` runner calls Python client's `bulk()` which [sets the right headers](https://github.com/elastic/elasticsearch-py/blob/2022b70b5c2139c3c5d9c47d038f022a1cb1f7ca/elasticsearch/_async/client/__init__.py#L747-L759), so the above code seems redundant. Does this address a specific failure? In which test, track, operation?

docs/adding_tracks.rst
gbanasiak: I don't think there are many changes required, so I did this change for you, see https://github.com/elastic/rally/pull/2024/commits/ea1311aa385fe8597b1c1009dbc7676354a89d4d.

esrally/client/asynchronous.py
fressi-elastic: TODO: verify this requirement

fressi-elastic: > As per earlier discussion in https://github.com/elastic/rally/pull/2024#discussion_r2741344203 I thought the desirable behavior while overriding Python client is the following:
> 
> * initially do not send `compatible-with` header, this applies to first synchronous `GET /` which is used to determine ES version and flavor (serverless or not),
> 
> * for the following requests use `compatible-with=8` or `compatible-with=9` depending on the detected stateful ES version, or do not use `compatible-with` at all in case of serverless.
> 
> 
> 
> With the most recent simplification, unless I'm mistaken, we're initially sending `compatible-with=9` and then drop to `compatible-with=8` for stateful ES 8, or keep using `compatible=with=9` elsewhere. I assume this is passing all the tests but is not the cleanest option. I'm just making sure this is intentional. We are trading simplicity for "correctness" here.

TODO: look at this

tests/client/common_test.py
```diff
@@ -14,11 +14,179 @@
...
+@cases(
+    version_8=CompatibilityModeCase(version="8.0.0", want=8),
+    version_9=CompatibilityModeCase(version="9.1.0", want=9),
+    version_7_raises=CompatibilityModeCase(version="7.17.0", want_error=ValueError),
+    version_int_7_raises=CompatibilityModeCase(version=7, want_error=ValueError),
+    version_int_8=CompatibilityModeCase(version=8, want=8),
+    version_int_9=CompatibilityModeCase(version=9, want=9),
+    version_int_10=CompatibilityModeCase(version=10, want=10, want_error=ValueError),
+    no_version=CompatibilityModeCase(version=None, want=8),
+    empty_version_raises=CompatibilityModeCase(version="", want_error=ValueError),
```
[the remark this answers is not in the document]
fressi-elastic: @copilot apply changes based on [this feedback](https://github.com/elastic/rally/pull/2024#discussion_r3012655241)

Remove want

esrally/client/common.py
fressi-elastic: I experience the error but it could be it come from an intermediate state. I am unsure if this is still required, especially for `_bulk` requests.

PR #2029, opened 2026-01-28
Fail benchmark when driver sample queue exceeds configurable limit

fressi-elastic:
## Summary

Avoids scalability stalls when the metrics pipeline cannot keep up by failing the benchmark when the driver's accumulated sample queue exceeds a configurable limit, instead of stalling.

## Changes

- **New config:** `reporting.sample.queue.failure.size` (default: 100000, 0 = disabled). When set to a positive value, the driver fails the benchmark if `raw_samples` exceeds it.
- **DriverActor:** On each wakeup, checks `len(raw_samples)`; if over threshold, closes driver and sends `BenchmarkFailure` with a message suggesting fewer measured events (e.g. iterations instead of duration, or target throughput).
- **Logging:** When the sample queue exceeds the limit, the driver logs an error with the threshold and queue size. When the queue grows (without failing), the driver logs an info message with the new maximum historical sample queue size.
- **Docs:** `docs/configuration.rst` updated; option registered in `esrally.types`.
- **Tests:** Single parameterized test with `@cases.cases` and `monkeypatch` fixture (failure when exceeded, no failure when disabled, no failure when at limit). Log messages are verified via `want_log_error_calls` and `want_log_info_calls` in the case structure (lists of `mock.call`).

## Usage

Set e.g. `sample.queue.failure.size = 5000` in `reporting` to fail early instead of stalling when the queue grows too large.

## Verification

To verify it works I executed the following command:

```
esrally race --kill-running-processes --track=nyc_taxis --pipeline=from-distribution --distribution-version=9.2.4
```

In case `sample.queue.failure.size` is `1000` I got desired failure and a clean test execution message:

```
Running index                                                                  [  6% done]
[ERROR] Cannot race. Sample queue exceeded limit
	Driver accumulated 1020 samples (limit: 1000). Consider reducing measured events (e.g. iterations instead of duration, or target throughput).

Getting further help:
*********************
* Check the log files in /Users/fressi/.rally/logs for errors.
* Read the documentation at https://esrally.readthedocs.io/en/latest/.
* Ask a question on the forum at https://discuss.elastic.co/tags/c/elastic-stack/elasticsearch/rally.
* Raise an issue at https://github.com/elastic/rally/issues and include the log files in /Users/fressi/.rally/logs.

---------------------------------
[INFO] FAILURE (took 108 seconds)
---------------------------------
```

In case `sample.queue.failure.size` is 0 or bigger than the maximum historical size, the benchmark complete sucessefully. I got the following log messages after it:
```
2026-01-28 14:42:34,812 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 144
2026-01-28 14:42:39,820 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 177
2026-01-28 14:42:44,832 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 341
2026-01-28 14:42:49,838 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 510
2026-01-28 14:42:54,844 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 677
2026-01-28 14:42:59,848 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 835
2026-01-28 14:43:04,854 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 1002
```

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -311,8 +311,8 @@ class DriverActor(actor.RallyActor):
                 self.driver.close()
                 self.logger.error(
                     "Sample queue failure size [%d] exceeds queue size limit [%d]. "
-                    "You can change the limit editing the sample.queue.failure.size config option in [reporting] section of rally.ini."
-                    "For the context, you can also see the historical maximum queue size in the log messages of level INFO, e.g. 'New maximum historical sample queue size: %d'.",
+                    "Set sample.queue.failure.size in [reporting] of rally.ini to change it. "
+                    "See INFO log 'New maximum historical sample queue size: %d'.",
                     self.sample_queue_failure_size,
                     queue_size,
                     self.max_historical_sample_queue_size,
--- a/tests/driver/driver_test.py
+++ b/tests/driver/driver_test.py
@@ -379,9 +379,12 @@ class SampleQueueFailureCase:
         want_failure_cause_contains=("1500", "1000"),
         want_log_error_calls=[
             mock.call(
-                "Sample queue failure size [%d] exceeds queue size [%d].",
+                "Sample queue failure size [%d] exceeds queue size limit [%d]. "
+                "Set sample.queue.failure.size in [reporting] of rally.ini to change it. "
+                "See INFO log 'New maximum historical sample queue size: %d'.",
                 1000,
                 1500,
+                0,
             ),
         ],
         want_log_info_calls=[],


esrally/driver/driver.py
```diff
@@ -301,6 +302,25 @@ def receiveMsg_WakeupMessage(self, msg, sender):
         if msg.payload == DriverActor.RESET_RELATIVE_TIME_MARKER:
             self.driver.reset_relative_time()
         elif not self.driver.finished():
+            sample_queue_failure_size = int(
+                self.driver.config.opts("reporting", "sample.queue.failure.size", mandatory=False, default_value=0)
```
gbanasiak: Can you pick a non-0 default that will protect against Rally stall while allowing regular use? We want to protect _unaware_ user from Rally getting stuck. How would a user know `sample.queue.failure.size` matters?

Another thing: we are retrieving this on every wake-up. Why not move it to the constructor?

esrally/driver/driver.py
```diff
@@ -301,6 +302,25 @@ def receiveMsg_WakeupMessage(self, msg, sender):
...
+            queue_size = len(self.driver.raw_samples)
+            if 0 < sample_queue_failure_size < queue_size:
+                self.driver.close()
+                self.logger.error("Sample queue failure size [%d] exceeds queue size [%d].", sample_queue_failure_size, queue_size)
+                self.send(
+                    self.benchmark_actor,
+                    actor.BenchmarkFailure(
+                        "Sample queue exceeded limit",
+                        f"Driver accumulated {queue_size} samples (limit: {sample_queue_failure_size}). "
+                        "Consider reducing measured events (e.g. iterations instead of duration, or target throughput).",
```
gbanasiak, line 325: I would also mention sampling here.

esrally/driver/driver.py
fressi-elastic: You are right. The user should discover a safe size by looking at the log. There are some lines that will star appearing when the size of the queue is growing. These values finally gets steady to a fair enough value. On my laptop this value for the track I used get stalled at 1000+ samples (which was the pick size of the queue along the whole benchmark). I would try something enough big for much slower hardware and considering performance variability I would use for example 100'000. But to fill up this queue the track has to produce at least those number of samples so it is highly depending on the track and reference hardware. Why not to store this value in the track itself basing on the result we get during nightlies? What do you think? 

The following is the log line appears (that we can use to monitor it, bug we could also store as a telemetry/metric value):

```
2026-01-28 14:43:04,854 ActorAddr-(T|:65182)/PID:31247 esrally.driver.driver INFO New maximum historical sample queue size: 1002
```

gbanasiak: > The user should discover a safe size by looking at the log.

The user will don't know what to look for. We need a safe default which fails the benchmark with a useful message.

> I would try something enough big for much slower hardware and considering performance variability I would use for example 100'000. But to fill up this queue the track has to produce at least those number of samples so it is highly depending on the track and reference hardware.

I would run some tests with an unbounded track (no target throughput, no iterations) and see how much time is needed to realistically ingest a given number of samples to a metric store and compare it with the `POST_PROCESS_INTERVAL_SECONDS` (30s). I actually still think that measuring the `post_process_samples()` processing time is the easiest way to determine if Rally will handle the stream of samples, or not. That's because `POST_PROCESS_INTERVAL_SECONDS` is the _only_ certainty here. But we can proxy the same with the number of stored samples. 100k sounds OK, but I would run some tests to get a better understanding.

> Why not to store this value in the track itself basing on the result we get during nightlies?

Tracks are run on different hardware and might be ingesting metrics into different clusters. What we see in nightlies is just one specific execution. I think we just need a generic value, high enough to not cause trouble for regular use, and low enough to not derail Rally before a telling failure takes place.

BTW, we do have a historical precedent with per-track tunables: target throughputs. We don't have a process of updating them periodically and adjust to current hardware, so even the latencies reported in _our_ nightlies are sometimes completely off, not to mention executions in the wild. Let's not repeat this.

esrally/driver/driver.py
fressi-elastic: Please be more specific.

esrally/driver/driver.py
fressi-elastic: I see. Please have a look again now.

fressi-elastic:
@gareth-ellis @inqueue Can you provide some feedback? Which is your opinion regarding this approach?

gareth-ellis (commented):
My opinion is that I think the message is the most useful feature - i think cancelling the benchmark would likely annoy me more than anything.

Ie, currently the issue is if we are generating too many metrics we occasionally can get into a position where rally hangs - now the course of action will be that rally will exit at an arbitrary point - it may be after it has hung (if our default value is too high) or before it hangs (maybe even when the benchmark would not have hung).

The big question is if i want to run with say 10000 req / second, and that then causes rally to hang - what should i do instead? Reducing target throughput doesnt help me, since I want to run 10000 req / second. Could we instead of aborting, increase the down sample factor? Does that fix the issue? 

I know we have some more things planned for future to avoid the issue, however I think aborting the benchmark seems wrong - if we can trigger an automatic downsample, then we should do that - if not we should print a very clear message (perhaps that even gets printed in the post test results).

fressi-elastic:
What about sending a back pressure method to the sending actor when the
queue reaches a 1st limit so it can adjust the subsampling and scale down,
and fail the execution when it reaches another limit or it can downsample
much more?

esrally/log.py
```diff
@@ -170,7 +170,6 @@ def install_default_log_config():
     io.ensure_dir(paths.logs())
 
 
-# pylint: disable=unused-argument
```
fressi-elastic, line 173: Unrelated

.gitignore
```diff
@@ -108,6 +108,8 @@ target/
 
 # Rally specific
 .rally_it/
+# uv cache (when using Makefile / UV_CACHE_DIR)
+.uv-cache/
```
fressi-elastic: Unrelated

Makefile
```diff
@@ -25,6 +25,7 @@ VENV_ACTIVATE := source $(VENV_ACTIVATE_FILE)
 PY_VERSION := $(shell jq -r '.python_versions.DEFAULT_PY_VER' .ci/variables.json)
 export UV_PYTHON := $(PY_VERSION)
 export UV_PROJECT_ENVIRONMENT := $(VIRTUAL_ENV)
+export UV_CACHE_DIR := $(CURDIR)/.uv-cache
```
fressi-elastic: Unrelated

esrally/driver/driver.py
gbanasiak: I mean `metrics.request.downsample.factor` documented in https://esrally.readthedocs.io/en/stable/configuration.html?highlight=sampling#reporting.

fressi-elastic: @gbanasiak Decreasing `metrics.request.downsample.factor` factor should increase the speed of the consumer therefore improving a stalle situation (when the bottleneck is sending messages to the metric store), but what it is weird to me why we don't apply it at the producer so we avoid transferring them between actors? Is it because of the throughput or percentile calculation?

gbanasiak: This code predates me so I know as much as you do. It seems likely throughput calculation was the reason, yes.

gbanasiak:
@fressi-elastic Fly-by comment: 100k queue limit default requires empirical verification with all the nightlies.

PR #2028, opened 2026-01-27
Set the pbzip2 -m flag to the highest supported limit of 2000MB

inqueue:
Set the maximum memory flag (`-m`) to a supported value supported by pbzip2. Fixes a warning when decompressing `bz2` corpus files:

```
2026-01-26 20:00:17,22 ActorAddr-(T|:56141)/PID:34824 esrally.utils.io WARNING Failed to decompress [/Users/jbryan/.rally/benchmarks/data/github_archive0/documents-20211001-1k.json.bz2] with [['pbzip2', '-d', '-k', '-m10000', '-c', '/Users/jbryan/.rally/benchmarks/data/github_archive0/documents-20211001-1k.json.bz2']]. Error [b'pbzip2: *ERROR: Memory usage size Min: 1MB and Max: 2000MB!  Aborting...\n']. Falling back to standard library.
```

The upper limit according to `pbzip2.cpp` [source](https://compression.great-site.net/pbzip2/?i=1) is 2000MB.

```
3569                                         if ((maxMemory < 1000000) || (maxMemory > 2000000000))
3570                                         {
3571                                                 fprintf(stderr,"pbzip2: *ERROR: Memory usage size Min: 1MB and Max: 2000MB!  Aborting...\n");
3572                                                 return 1;
3573                                         } 
```

PR #2032, opened 2026-01-30
Remove -m flag from pbzip2 decompression (fixes #2030)

### LINKED ISSUE #2030: pbzip2 decompression failure due to memory limit flags
### Summary
When running an Elastic Rally track, the decompression of data files (e.g., `documents.json.bz2`) fails when using `pbzip2`. This occurs because Rally use a flag (`-m`) with a RAM value that is too big for some installable version of the tool. On certain systems (notably macOS/Homebrew), pbzip2 has a hard-coded memory limit of 2000MB.

### Environment
- **Tool:** Elastic Rally
- **Dependency:** `pbzip2`
- **Error observed on:** macOS (Homebrew), various Linux distributions.

### Error Message
```text
esrally.utils.io WARNING Failed to decompress [.../documents.json.bz2] with 
[['pbzip2', '-d', '-k', '-m10000', '-c', '.../documents.json.bz2']]. 
Error [b'pbzip2: *ERROR: Memory usage size Min: 1MB and Max: 2000MB!  Aborting...\n']. 
Falling back to standard library.

### Outcome

Despite rally writes to the log file he will fall back to another mechanism, it actually ends up idefenitivily looping on retrying the same and failing forever.

fressi-elastic:
`pbzip2` on some systems (e.g. macOS/Homebrew) has a hard-coded memory limit of 2000MB. Passing -m with a larger value caused decompression to fail and Rally to loop instead of falling back to the standard library. Dropping the `-m` flag lets `pbzip2` use its default memory limit.

Fixes #2030

### DIFF
--- a/esrally/utils/io.py
+++ b/esrally/utils/io.py
@@ -337,7 +337,7 @@ def decompress(zip_name: str, target_directory: str) -> None:
     if extension == ".zip":
         _do_zip_decompress(target_directory, zipfile.ZipFile(zip_name))
     elif extension == ".bz2":
-        decompressor_args = ["pbzip2", "-d", "-k", "-m2000", "-c"]
+        decompressor_args = ["pbzip2", "-d", "-k", "-c"]
         decompressor_lib_bz2 = bz2.open
         _do_decompress_manually(target_directory, zip_name, decompressor_args, decompressor_lib_bz2)
     elif extension == ".zst":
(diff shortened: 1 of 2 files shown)


fressi-elastic:
I didn't realize there where already a fix for this from Jason: https://github.com/elastic/rally/pull/2028

PR #2033, opened 2026-02-02
Add ingest-doc-count option to index an exact number of documents

davidkyle:
I often want to run a benchmark with an exact number of documents (e.g. ingest 1M docs). This can be achieved approximately with the `ingest-percentage` operation but if the corpus size is not a round number you have to figure out what percentage you want and end up with strange values that need explaining ("16.7% equals 10,000 docs").

The intent is much clearer when using `ingest-doc-count`; "ingest exactly X docs".

`ingest-doc-count` controls the number of bulks so `ingest-doc-count` must be a multiple of `bulk-size`, given that we mostly use round numbers this won't be a problem for most users. That ability could be added as an enhancement


carlosdelest (approved):
LGTM! 🙌 Thanks for doing this!

Let's have a review from the perf team as well

esrally/track/params.py
```diff
@@ -634,13 +634,35 @@ def __init__(self, track, params, **kwargs):
...
+                if self.ingest_doc_count <= 0:
+                    raise exceptions.InvalidSyntax("'ingest-doc-count' must be positive but was %d" % self.ingest_doc_count)
+            except ValueError:
+                raise exceptions.InvalidSyntax("'ingest-doc-count' must be numeric")
+            if self.ingest_doc_count % self.bulk_size != 0:
+                raise exceptions.InvalidSyntax(
+                    "'ingest-doc-count' must be a multiple of 'bulk-size' (%d) but was %d" % (self.bulk_size, self.ingest_doc_count)
+                )
+            total_docs = sum(corpus.number_of_documents(source_format="bulk") for corpus in self.corpora)
+            if self.ingest_doc_count > total_docs:
```
mromaios, line 651: 💭 minor: we could perhaps also relax this and just index all the available documents if it exceeds the corpus size

esrally/track/params.py
```diff
@@ -634,13 +634,35 @@ def __init__(self, track, params, **kwargs):
...
+        if "ingest-doc-count" in params:
+            if "ingest-percentage" in params:
+                raise exceptions.InvalidSyntax("'ingest-doc-count' and 'ingest-percentage' are mutually exclusive")
+            try:
+                self.ingest_doc_count = int(params.get("ingest-doc-count"))
+                if self.ingest_doc_count <= 0:
+                    raise exceptions.InvalidSyntax("'ingest-doc-count' must be positive but was %d" % self.ingest_doc_count)
+            except ValueError:
+                raise exceptions.InvalidSyntax("'ingest-doc-count' must be numeric")
+            if self.ingest_doc_count % self.bulk_size != 0:
```
mromaios, line 646: 💭 since we use `ceil` below to set the total number of bulks:
https://github.com/elastic/rally/pull/2033/changes#diff-8a91ed11f7faea5a5fc7be073889e6afc93c6b38a5ad198460e83fbc30fc20bfR830
 do we need `ingest_doc_count` to be a multiple of `bulk_size`?

mromaios (approved):
Nice! LGTM ✅ 
Left a couple of minor comments

esrally/track/params.py
davidkyle: As a general principle we prefer to avoid leniency in the API. If we silently accept a value the corpus cannot satisfy then the user walks away thinking they have ingested all the docs they asked for which may produce confusing results down the line. If the value is too high the users can read the error and pick a smaller number.

mromaios: I see, thanks!

esrally/track/params.py
davidkyle: The same comment about leniency applies here. If `ingest_doc_count` is not a multiple of `bulk_size` then you won't have exactly `ingest_doc_count` docs ingested. We could eliminate this requirement by modifying the final bulk size but I felt that was a larger change and not really necessary

mromaios: Ah, just re-read the code. Ignore me, we only support multiples of `bulk_size`. Thanks for the reply

gareth-ellis (approved):
LGTM

PR #2034, opened 2026-02-02
Disk usage stats timeout

gareth-ellis:
If disk-usage-stats times out,then rally exits and you lose a whole run. This allows us to avoid a timeout by implementing the ability to override timeout, as well as avoids a Rally Error being raised. We then still exit, since if we fail to collect any stats we raise a RallyError - 

Should we change this so a fatal error in a telemetry device is not fatal for the whole run?

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2326,6 +2326,7 @@ class DiskUsageStats(TelemetryDevice):
             May specify:
             ``disk-usage-stats-indices``: Comma separated list of indices who's disk
                 usage to fetch. Default is all indices in the track.
+            ``disk-usage-stats-timeout``: Timeout in seconds for the disk usage API call. Default is 600s.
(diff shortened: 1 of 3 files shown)


esrally/telemetry.py
```diff
@@ -2367,6 +2370,10 @@ def on_benchmark_stop(self):
                 msg = f"Requested disk usage for missing index {index}"
                 self.logger.warning(msg)
                 continue
+            except elastic_transport.ConnectionTimeout:
+                msg = f"Timeout occurred while collecting disk usage for {index}"
+                self.logger.warning(msg)
+                continue
```
gbanasiak: The `elastic_transport.ConnectionTimeout` exception is mapped to `elasticsearch.ConnectionTimeout` so we don't need this extra import.

gbanasiak (approved):
LGTM

fressi-elastic (approved):

PR #2039, opened 2026-02-13
Log warning on BadRequestError

jordan-powers:
This PR updates the Retry runner to log a warning message when a delegate runner raises an `elasticsearch.BadRequestError`.

I was recently experimenting with a change to the elastic/logs track, and I spent several hours debugging why the `validate-package-template-installation` step was failing. It turns out I had a typo in my changes to one of the component index templates, but the bad request was being silently ignored. This PR should hopefully make it much easier to debug similar issues.


gareth-ellis (approved):
LGTM

PR #2041, opened 2026-02-17
Track certain headers if present in error responses

gareth-ellis:
When a race encounters issues, it would be beneficial to report some of the headers we get in the response. I have added two headers I am currently interested in, I am open to other headers if others are of interest too.

This will currently only add to failed requests - if a request succeeds, then we do not capture the headers in that case. If we wanted to track the headers in successful calls, we would need to tie into each runner, which currently seems quite convoluted. At the same time, I cannot see any value in capturing these headers for successful requests.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -2028,6 +2028,16 @@ async def execute_single(runner, es, params, on_error: OnErrorBehavior):
     # pylint: disable=import-outside-toplevel
     import elasticsearch
 
+    def _parse_headers(e):
+        tracked_headers = [{"request_id": "X-Cloud-Request-Id", "handling_node": "X-Found-Handling-Instance"}]
+        headers_meta = {}
+        if hasattr(e, "errors") and e.errors and len(e.errors) > 0 and hasattr(e.errors[0], "headers"):
+            for header in tracked_headers:
+                for key, value in header.items():
+                    if value in e.errors[0].headers:
+                        headers_meta[key] = e.errors[0].headers[value]
+        return headers_meta
+
     is_connection_error = False
 
     try:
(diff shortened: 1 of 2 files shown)


gbanasiak (approved):
LGTM

esrally/driver/driver.py
```diff
@@ -2028,6 +2028,22 @@ async def execute_single(runner, es, params, on_error: OnErrorBehavior):
...
+    def _parse_headers(e):
+        tracked_headers = [
+            {
+                "request_id": "X-Cloud-Request-Id",
+                "handling_node": "X-Found-Handling-Instance",
+                "handling_cluster": "X-Found-Handling-Cluster",
+            }
+        ]
+        headers_meta = {}
+        if hasattr(e, "errors") and e.errors and len(e.errors) > 0 and hasattr(e.errors[0], "headers"):
```
fressi-elastic, line 2040: `if e.errors and len(e.errors) > 0:` is equivalent to `if e.errors:` therefore you can just write the second one.

I actually would suggest a simpler implementation of this function:

```Python
# Output key -> header mapping.
headers_meta = {
     "request_id": "X-Cloud-Request-Id",
     "handling_node": "X-Found-Handling-Instance",
     "handling_cluster": "X-Found-Handling-Cluster",
}
# It iterate errors
for err in getattr(e, 'errors', []):
    headers = getattr(err, 'headers')
    if not headers:
        continue    
 
    for key, header in tracked_headers.items():
         # Get only tracked header values
         value = headers.get(header)
         if value is not None:
             headers_meta[key] = headers[header]

return headers_meta
```

fressi-elastic: @gareth-ellis I reviewed this too late I see :-D

Issue #1796, opened 2023-11-02
Add 'enrich' operator for setting up enrich indices


craigtaverner:
The recent work to benchmark `ES|QL` has benefitted greatly from the addition of the `esql` operator in https://github.com/elastic/rally/pull/1791. However, the benchmarks that make use of the `enrich` command need enrich indices set up, and this is currently achieved using a complex set of multiple `raw` operations:

* Create enrich source index and ingest data
* Delete previous enrich policy
* Create new enrich policy
* Execute the enrich policy
* Use the enrich policy within an ingestion pipeline

The three steps for delete/create/execute of the enrich policy can be combined into a single `enrich` policy operation. This operation should combine the following raw steps:

== Delete

```
    {
      "name": "delete-enrich-nyc_vendors",
      "operation-type": "raw-request",
      "ignore": [404],
      "include-in-reporting": false,
      "method": "DELETE",
      "path": "/_enrich/policy/nyc_vendors"
    }
```

== Create

```
    {
      "name": "create-enrich-nyc_vendors",
      "include-in-reporting": false,
      "operation-type": "raw-request",
      "method": "PUT",
      "path": "/_enrich/policy/nyc_vendors",
      "body": {
        "match": {
          "indices": "nyc_vendors",
          "match_field": "id",
          "enrich_fields": [ "name" ]
        }
      }
    }
```

== Execute

```
    {
      "name": "execute-enrich-nyc_vendors",
      "operation-type": "raw-request",
      "method": "POST",
      "path": "/_enrich/policy/nyc_vendors/_execute"
    }
```

== Combined

The above operations could be combined into one:

```
    {
      "name": "setup-enrich-nyc_vendors",
      "operation-type": "enrich_policy",
      "policy_name": "nyc_vendors",
      "delete": true,                              // default to true
      "enrich_type": "match",             // could be default to 'match'
      "indices": "nyc_vendors",          // could be default to value of 'policy_name'
      "match_field": "id",
      "enrich_fields": [ "name" ]
    }
```

== Multiple policies

An interesting alternative might be to allow the definition of multiple policies with a more flexible syntax:

```
    {
      "name": "setup-enrich-policies",
      "operation-type": "enrich_policy",
      "delete": true,
      "policies": {
        "nyc_vendors": {
            "match": {
              "indices": "nyc_vendors",
              "match_field": "id",
              "enrich_fields": [ "name" ]
          }
        },
        "nyc_payment_types_fares": {
          "match": {
[...]

PR #2045, opened 2026-02-26
Adds an operator to create enrich policies

slDias:
As suggested in Issue #1796, this commit adds a new operator that combines the delete, creation and execution of enrich policies within a single operation.

I've added a new challenge to nyc_taxis track to test this it is equivalent to the esql challenge, it can be found [here](https://github.com/slDias/rally-tracks/blob/master/nyc_taxis/challenges/default.json#L1778) (I can also open a MR replacing the 'esql' one, if desired).

Happy to change anything if needed.


docs/track.rst
```diff
@@ -3189,9 +3189,13 @@ esql-profile
 
 With the operation type ``esql-profile`` you can execute an `ES|QL query <https://www.elastic.co/guide/en/elasticsearch/reference/current/esql.html>`_ with profiling enabled (``profile: true``). This captures detailed timing information across all phases of query execution.
 
-Properties
-""""""""""
 
+Meta-data
```
craigtaverner: This meta-data section seems to have been incorrectly placed inside the `esql` section, instead of the `create-enrich-policy` section below.

docs/track.rst
```diff
@@ -3228,6 +3232,59 @@ The following metrics are included when profile data is present:
...
+
+create-enrich-policy
+~~~~~~~~~~~~~~~~~~~~
+
+The operation ``create-enrich-policy`` will delete (if it exists), create and execute a list of enrich policies.
+
+Properties
+""""""""""
+
+It's properties follows the request body of the `Create an enrich policy API <https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-enrich-put-policy>`_
```
craigtaverner: This link should rather be moved into the `policies` list item below.

docs/track.rst
```diff
@@ -3228,6 +3232,59 @@ The following metrics are included when profile data is present:
...
+**Examples**
+
+The following snippet will create all enrich policies that have been defined in the ``policies`` section.
+It will attempt to delete them beforehand if it exists::
+
+
+    {
+      "name": "create-enrich-policy",
+      "operation": {
+        "operation-type": "create-enrich-policy",
```
craigtaverner: Since the operation does more than just create, I think it should be named `enrich-policy`, and it can delete, create and execute the policy (or any combination of those actions).

docs/track.rst
```diff
@@ -3228,6 +3232,59 @@ The following metrics are included when profile data is present:
...
+              "indices": "nyc_payment_types",
+              "match_field": "type",
+              "enrich_fields": [
+                "name"
+              ]
+            }
+          }
+        }
+      },
+      "delete": true
```
craigtaverner: Shouldn't the flag be inside the operation? Having it outside the operation makes it seem like it is not tied to enrich policy management at all, but part of generic challenges.

esrally/driver/runner.py
```diff
@@ -3385,3 +3386,50 @@ def compare_value(self, criterion: dict, value: str | int | float) -> bool:
 
     def __repr__(self, *args, **kwargs):
         return "run-until"
+
+
+class CreateEnrichPolicy(Runner):
```
craigtaverner: Since this can do much more than just create the policy, I think it should be renamed `EnrichPolicy`, and take optional boolean flags for `delete`, `create` and `execute`, and only `create` requires the policy bodies, while all the rest require at least the policy name.

docs/track.rst
```diff
@@ -3228,6 +3232,59 @@ The following metrics are included when profile data is present:
...
+
+The following snippet will create all enrich policies that have been defined in the ``policies`` section.
+It will attempt to delete them beforehand if it exists::
+
+
+    {
+      "name": "create-enrich-policy",
+      "operation": {
+        "operation-type": "create-enrich-policy",
+        "policies": {
```
craigtaverner, line 3258: I wonder if it would be nicer to have a list instead of an object as the value of `policies`. Then we could, for example, just list the names if all we want to do it re-execute existing policies:
```
{
  "name": "re-execute-enrich-policies",
  "operation": {
    "operation-type": "enrich-policy",
    "policies": [ "nyc_rate_codes", "nyc_payment_types" ],
    "delete": false,
    "create": false,
    "execute": true
  }
}
```

If we have `"create": true`(which can be the default), then the values inside the list of policies must be objects with a single key, the name of the policy, and it's value would be the policy body. This is one level nesting deeper than your example.

esrally/driver/runner.py
```diff
@@ -3385,3 +3386,50 @@ def compare_value(self, criterion: dict, value: str | int | float) -> bool:
...
+        reqs = []
+        for policy_name in policy_data:
+            reqs.append(es.enrich.execute_policy(name=policy_name, wait_for_completion=True))
+        res = await asyncio.gather(*reqs)
+        self.logger.debug("Executed %s enrich policies: %s", len(res), [r.body for r in res])
+
+    async def __call__(self, es, params):
+        enrich_policies = mandatory(params, "policies", self)
+
+        if params.get("delete", True):
```
craigtaverner, line 3500: I think it would be nicer to have three booleans: delete, create, execute. The defaults could be: false, true, true.

esrally/driver/runner.py
```diff
@@ -3385,3 +3386,50 @@ def compare_value(self, criterion: dict, value: str | int | float) -> bool:
...
+
+    async def _execute_enrich_policy(self, es, policy_data):
+        reqs = []
+        for policy_name in policy_data:
+            reqs.append(es.enrich.execute_policy(name=policy_name, wait_for_completion=True))
+        res = await asyncio.gather(*reqs)
+        self.logger.debug("Executed %s enrich policies: %s", len(res), [r.body for r in res])
+
+    async def __call__(self, es, params):
+        enrich_policies = mandatory(params, "policies", self)
```
craigtaverner, line 3498: If we supported the nested arrays approach I described above, we could have a normalize option here to bring the multiple formats into a single version so the functions below do not need to care about that.

craigtaverner (changes requested):
This looks like a nice addition. But there are a few things to fix or change. At the very least fix the docs. The suggestions I made for supporting more flexible combinations of the delete/create/execute actions could be done here or in a followup PR.

.gitignore
```diff
@@ -119,6 +119,7 @@ recipes/ccr/ccr-target-hosts.json
 /.project
 /.pydevproject
 /.vscode
+/.zed
```
craigtaverner: What does this mean? Should this be in this PR?

gareth-ellis:
@slDias let us know if you have any questions regarding @craigtaverner 's review. I'm going to change this to draft, please switch back to "ready for review" when you have addressed the feedback - thanks!

docs/track.rst
slDias: Thanks for reviewing it. The docs were mixed during a merge, I've rebased the changes instead so its fixed now.

docs/track.rst
slDias: done

docs/track.rst
slDias: renamed it, the combinations will be part of another PR

docs/track.rst
slDias: fixed, good catch. it was already part of the operation but I've misplace it in the docs.

esrally/driver/runner.py
slDias: Renamed it for now, I'll work on a separate PR for the new format. 
I gave it some thought, wdyt of removing `policies` and then having lists on the `delete` and `execute`? that gives more control over what is done to each policy.

example:

```json
{
  "operation-type": "enrich-policy",
  "delete": [
    "a-policy-name",
    "another-policy",
    "yet-another-one"
  ],
  "create": {
    "a-policy-name": {
      "match": {
        "indices": "nyc_rate_codes",
        "match_field": "id",
        "enrich_fields": [
          "name"
        ]
      }
    }
  },
  "execute": [
    "a-policy-name",
    "a-completely-different-policy"
  ]
}
```

I will work on this version if I find the time, since it's easy to change to your suggestion if you prefer later

docs/track.rst
slDias: will do that or my suggestion in another PR

esrally/driver/runner.py
slDias: will do that or my suggestion in another PR

esrally/driver/runner.py
slDias: will do that or my suggestion in another PR

.gitignore
slDias: this is a folder to be ignored from the zed IDE, which I use. I've removed it to avoid friction

esrally/driver/runner.py
slDias: Got to implement this, it's as a draft in case you decide it's no good, but it is working and ready to review: https://github.com/elastic/rally/pull/2101

craigtaverner (approved):
LGTM

gareth-ellis (approved):

gareth-ellis:
@elasticmachine update branch

gareth-ellis:
Thanks @slDias !

PR #2047, opened 2026-03-03
Adds an ESQL profile runner to rally

carlosdelest:
Adds EsqlProfile as a built-in Rally runner (`esql-profile` operation type).

This runner is already in use in 4 different tracks in rally-tracks: wikipeda, so_vector, msmarco-v2-vector and elastic/logs. With this change, we can perform changes to the runner in a single place.

The `esql-profile` runner executes an ES|QL query with `profile: true` and extracts detailed timing metrics from the profile response, returning them as measurable Rally metrics:

  - Query phase timings: query.took_ms, planning.took_ms, parsing.took_ms, preanalysis.took_ms, analysis.took_ms
  - Driver metrics (per named driver, aggregated across instances):
    - <driver>.number — count of driver instances
    - <driver>.took_ms / <driver>.cpu_ms — max across instances
    - <driver>.took_total_ms / <driver>.cpu_total_ms — sum across instances
  - Operator metrics (per operator within each driver):
    - <driver>.<operator>.process_ms
    - <driver>.<operator>.processed_slices
  - Plan optimization timings:
    - <plan>.logical_optimization.took_ms
    - <plan>.physical_optimization.took_ms
    - <plan>.reduction.took_ms

  Usage in a track

  {
    "operation-type": "esql-profile",
    "query": "FROM logs-* | STATS count = COUNT(*) BY host.name"
  }

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3171,6 +3171,50 @@ Meta-data
 * ``unit``: The unit in which to interpret ``weight``, in this case ``pages``.
 * ``success``: A boolean indicating whether the query has succeeded.
 
+esql-profile
+~~~~~~~~~~~~~
+
+With the operation type ``esql-profile`` you can execute an `ES|QL query <https://www.elastic.co/guide/en/elasticsearch/reference/current/esql.html>`_ with profiling enabled (``profile: true``). This captures detailed timing information across all phases of query execution.
+
+Properties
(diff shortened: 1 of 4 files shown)


inqueue (approved):
I tested the runner and it LGTM overall. Did we consider docs for it?

PR #2048, opened 2026-03-03
Add took time for CPS search responses

### LINKED ISSUE #2110: Collect Cross-Project Search took times
Cross-project search responses contain timing details for individual clusters participating in the search. Each named cluster in the `_clusters.details` object contains the individual took times for each cluster. The `_origin` is the coordinating cluster servicing the request and coordinating the response.

inqueue:
Add telemetry for took time from CPS search responses.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -1043,7 +1043,7 @@ If ``detailed-results`` is ``true`` the following meta-data are returned in addi
 The following meta-data are only present when returned by Elasticsearch (e.g. cross-cluster search or reduce phases):
 
 * ``num_reduce_phases``: Number of reduce phases executed (when applicable).
-* ``clusters``: Cross-cluster search summary and per-cluster details when the response includes ``_clusters``. The ``clusters`` object may contain ``total``, ``successful``, ``skipped``, ``running``, ``partial``, ``failed``, and ``details``. The ``details`` key maps each cluster name to an object with ``status``, ``indices``, ``took``, ``timed_out``, and ``_shards``.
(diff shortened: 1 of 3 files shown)


luigidellaquila (commented):
Thanks @inqueue, functionally it looks very good.

With my limited experience on this codebase, the only thing I noticed is that `parse` has an early termination clause, while `parse_optional_props` performs a full scan of the response.
I guess it could take a few milliseconds per query at most on very large responses. I'm not sure it's a real concern or if it's worth benchmarking the benchmark 😄 

Anyway, it looks good to me.

luigidellaquila (approved):

pquentin (commented):
Agreed, performance is a concern here, since this applies to all search queries with detailed results.

Why do we have to parse the JSON object twice? The existing parse() function is more efficient as it only builds Python objects when the prefix is right and terminates early.

inqueue:
> Agreed, performance is a concern here, since this applies to all search queries with detailed results.
> 
> Why do we have to parse the JSON object twice? The existing parse() function is more efficient as it only builds Python objects when the prefix is right and terminates early.

This was an oversight. There is no reason to parse the JSON object twice. The problem is different properties are present with and without CPS for the same request. I've committed an attempt to preserve early termination when we do not want to wait for the full response.

pquentin (commented):
Thanks! Great improvement. We're still parsing the response twice when `extract_cluster_details` is called. Is there a way to avoid it? Or is this OK as it only affects CPS?

inqueue:
> Thanks! Great improvement. We're still parsing the response twice when `extract_cluster_details` is called. Is there a way to avoid it? Or is this OK as it only affects CCS?

I am not thrilled with it. There is one more thing I would like to try. It is true that it only affects CPS, and only when requesting detailed results.

PR #2051, opened 2026-03-05
Use data streams for easier metric handling

NickDris:
# Rally migration to data streams 

Continuous and/or recurrent benchmarking can produce huge volume of metrics. Right now rally implements metric storage management using date based indices with a monthly rollover, also providing the capability to configure `number_of_shards` and `number_of_replicas`. This is not optimal in some cases where a central metric store is receiving huge amounts of metrics, both indexing and search wise. I am introducing data streams with this PR in order to provide for better metric storage management. As of this PR ILM managed data streams will be enabled by default and  versioned composable index templates (e.g. `rally-metrics-template-v1`) will provide customizable behavior with an empty editable @**custom** component. A default component enforces `max_primary_shard_size` setting equal to 50GB and a @**custom** component template that can be edited by users to change settings accordingly to their metric volumes. 

Affects only rally-races, rally-results and rally-metrics, not rally-annotations.

## Use case

An obvious example would be to add retention policy (>90 days) or frozen tier.

### What is currently available

Rally indexes data to date-based normal indices with the rally-*-YYYY-MM pattern. 
1. Number of shards/replicas imposed by datastore.number_of_shards, datastore.number_of_replicas
2. datastore.overwrite_existing_templates defaults to false and is only for rally-metrics.
3. Current rollover By applying a date-based pattern a monthly rollover is implied 

### Functional Changes

1. New `datastore.use_data_streams` rally param **_defaults to true_**
2. If enabled rally will index data in `rally-metrics-v1`, `rally-races-v1`, `rally-results-v1` data streams instead of date based indices. _Note:_ We are adopting a version-like scheme, in order to a. Keep track of the _major changes in index templates_ and b. preserve _backwards compatibility for dashboards_ using index patterns.
4. In order to create the data stream we use composable templates with 2 component templates: a default one (mappings + ilm) and an empty custom one.  mappings are the same as before but we dynamically add `@timestamp` field to the mappings which is mandatory for data stream creation. An ilm default policy is added that enforces rollover at 50gb primary shard size for search and storage optimization. The custom component is conveniently configured at will.
6. Asterisk for index-pattern, to discern between the two, e.g. Old-style index template `rally-races` had the `rally-races-*` index pattern and now data stream with index template of the same name has `rally-races-v*` index pattern.
7. Annotations stay untouched.

### Create `IndexHandler` abstraction to add data stream and lifecycle management to metrics.py

Introduce the `IndexHandler` class that that owns template management for all three ES store types (metrics, races, results). Removes duplicated index/template logic that was scattered across `EsMetricsStore`, `EsRaceStore`, and `EsResultsStore`.

### Key changes

- **`EsClient`**: Renamed `template_exists` → `index_template_exists`. Added new client methods for component templates (`get_component_template`, `put_component_template`, `component_template_exists`) and ILM lifecycle (`get_lifecycle`, `put_lifecycle`, `lifecycle_exists`).
- **`IndexTemplateProvider`**: Use the `use_data_streams` property to apply a mandatory `@timestamp` field in the case of data streams. Changed `_read()` to return a `dict` instead of a JSON string; callers now wrap with `json.dumps()`. 
- **`ComponentTemplateProvider`** *(new)*: Subclass of `IndexTemplateProvider` that decomposes each store type template into three component templates: mapping, lifecycle and a custom placeholder that can be overriden after creation (user custom management).
- **`EsStoreType`** *(new)*: Enum encoding per-store-type constants (`index_prefix`, `index_template_name`, `ilm_default`, `data_stream_version`) — eliminates the scattered `INDEX_PREFIX` / `TEMPLATE_VERSION` class attributes.
- **`IndexHandler`** *(new)*: Single entry point for index management. Selects `ComponentTemplateProvider` (data streams) or `IndexTemplateProvider` (date-based) based on config. Provides `index_name()`, `ensure_index_template()`, and internal methods for idempotent upsert of ILM policies, component templates, and index templates with diff-based skip/overwrite logic.
- **`EsMetricsStore` / `EsRaceStore` / `EsResultsStore`**: Replaced inline `index_name()`, `_ensure_index_template()`, and `_get_template()` with delegation to `self._index_handler`. Removed `self._index` mutable state in `EsMetricsStore` — index name is now computed on demand. Dropped `index_template_provider_class` constructor parameter (no longer needed for testing; tests mock `_index_handler` directly).

## Tests

Replace repetitive `@pytest.mark.parametrize` with structured `@dataclass`-backed `@cases.cases` decorators, as adopted in many test units. Each case is a named, self-documenting dataclass instance.

**Key changes:**

- **`TestIndexTemplateProvider`**: Collapsed 8 test methods into 2 (`test_shard_settings` with 5 cases, `test_data_streams` with 4 cases) using `ShardSettingsCase` and `DataStreamsCase` dataclasses.
- **`TestComponentTemplateProvider`** *(new)*: Tests `ComponentTemplateProvider` directly — validates 3-component structure (mapping, lifecycle, custom) per store type, shard setting propagation, and annotation template inheritance.
- **`TestIndexHandler`**: Replaced tuple-parametrize with `IndexNameCase` (6 cases) and `EnsureTemplateCase` (2 cases), unifying previously split test methods.
- **`TestEsMetricsStore`**: Converted `put_value_*` and `put_doc_*` parametrize to `StoreModeCase`; added `create_false_date_based` open case that documents refresh-on-read behavior for date-based indices.
- **`TestEsRaceStore` / `TestEsResultsStore`**: Same `StoreModeCase` pattern for `store_race` and `store_results` variants.
- **Removed** former `DummyIndexTemplateProvider` classes.

## Manual actions required for migration

Rally will store metrics to data streams by default. In order to migrate older indices to data streams, reindexing must be applied but keep in mind the mandatory "@timestamp" field and "op_type": "create"


docs/configuration.rst
```diff
@@ -90,9 +90,10 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
...
 * ``datastore.number_of_replicas`` (default: `Elasticsearch default value <https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#_static_index_settings>`_): The number of replicas each primary shard has. Defaults to . Any updates to this setting after initial index creation will only be applied to new ``rally-*`` indices. An error is raised if set for Elastic Cloud Serverless projects.
-* ``datastore.overwrite_existing_templates`` (default: ``false``): Existing Rally index templates are replaced only when this option is ``true``.
+* ``datastore.overwrite_existing_templates`` (default: ``false``): Existing Rally index teplates are replaced only when this option is ``true``. In the case of ``datastore.use_data_streams`` is ``true``, this option applies to both the default composable index template and the default ILM policy.
```
pquentin:
```suggestion
* ``datastore.overwrite_existing_templates`` (default: ``false``): Existing Rally index templates are replaced only when this option is ``true``. When ``datastore.use_data_streams`` is ``true``, this option applies to both the default composable index template and the default ILM policy.
```

docs/configuration.rst
```diff
@@ -90,9 +90,10 @@ The following settings are applicable only if ``datastore.type`` is set to "elas
...
 * ``datastore.probe.cluster_version`` (default: true): Enables automatic detection of the metric store's version.
+* ``datastore.use_data_streams`` (default: ``true``): If ``true``, Rally writes to ``rally-metrics-v1``, ``rally-races-v1`` and ``rally-results-v1`` data streams and if ``false`` to date based ``rally-metrics-YYYY-MM``, ``rally-races-YYYY-MM`` and ``rally-results-YYYY-MM`` indices. When ``true`` Rally creates a default ILM policy with ``50 GB`` rollover and a composable index template for each stream, which is composed of mappings, a default ILM policy and an empty @custom component which can be used to apply custom settings.
```
pquentin:
```suggestion
* ``datastore.use_data_streams`` (default: ``true``): If ``true``, Rally writes to ``rally-metrics-v1``, ``rally-races-v1`` and ``rally-results-v1`` data streams and if ``false`` to date based ``rally-metrics-YYYY-MM``, ``rally-races-YYYY-MM`` and ``rally-results-YYYY-MM`` indices. When ``true`` Rally creates a default ILM policy with ``50 GB`` rollover and a composable index template for each stream, which is composed of mappings, a default ILM policy and an empty `@custom` component which can be used to apply custom settings.
```

docs/migrate.rst
```diff
@@ -14,6 +14,11 @@ The metrics store keeps existing index templates
 
 Existing Rally index templates are replaced only when option ``datastore.overwrite_existing_templates`` in section ``reporting`` is ``true``.
 
+Migrate rally indices to data streams
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Rally 2.12.0 introduces support for data streams in the metrics store. We recommend reindexing existing indices to data streams to benefit from the advantages of data streams ( stay aware of the mandatory `@timestamp` field and "op_type": "create" when reindexing ) . In order to start writing on data streams, ``datastore.use_data_streams`` must be set to ``true`` in section ``reporting`` ( it is already set to ``true`` by default ). This will cause Rally to create new composable index templates for data streams, along with a @custom index template, which can be configured to set custom index settings.
```
pquentin: > stay aware of the mandatory `@timestamp` field and "op_type": "create" when reindexing

What does that mean? What do users need to do?

> In order to start writing on data streams, ``datastore.use_data_streams`` must be set to ``true`` in section ``reporting`` ( it is already set to ``true`` by default )

If it's true, we don't need to set it. Instead, we need to make it clear that Rally will write to a data stream unless we set it to `false`. Which makes me wonder if this should be off by default.

docs/metrics.rst
pquentin, line 1: I think we need a better explanation of the data stream logic, including the ``@custom`` component. Users won't necessarily be familiar with ILM, data streams, etc. (While `migrate.rst` is important, new users won't read that at all.)

esrally/metrics.py
```diff
@@ -76,17 +93,20 @@ def exists(self, index):
...
         self.guarded(elasticsearch.helpers.bulk, self._client, items, index=index, chunk_size=5000)
 
-    def index(self, index, item, id=None):
+    def index(self, index, item, id=None, use_data_streams=True):
         doc = {"_source": item}
-        if id:
+        if not use_data_streams and id:
             doc["_id"] = id
-        self.bulk_index(index, [doc])
+        self.bulk_index(index, [doc], use_data_streams=use_data_streams)
```
pquentin: I had a few comments about those functions, but they don't appear to be used at all?

pquentin (commented):
Good work! Sorry, this is a very superficial initial review.

docs/configuration.rst
gareth-ellis: Two questions regarding this:

- Do we want to change the default for this so it uses data streams by default?
- We shouldn't create the @\custom component, I don't think. We can refer to a non existing template, and that will be accepted still. That way we don't need to worry about overwriting templates that are specifically custom.

gareth-ellis (commented):
Looks good so far - i have some thoughts regarding what we auto create, and whether we should enable by default

docs/migrate.rst
NickDris: > What does that mean? What do users need to do?
> If it's true, we don't need to set it. Instead, we need to make it clear that Rally will write to a data stream unless we set it to `false`

That's true, thanks for both comments.

> Which makes me wonder if this should be off by default.

For the default value of this flag my reasoning is the following: 
- Both old and new users get ilm managed data streams right out of the box.
- Users that are not familiar with Elasticsearch may not know what a data stream is. They have no incentive to set this to true and get ilm benefits if they don't hit a wall, so by default they get a less optimized metric storage behavior.
- Having date-based indices may lead to mind bogging problems for non-experts, while ilm managed data streams are easier to maintain and provide for alternatives if storage problems arise. With the old behavior its easier for users to delete or "freeze" the old indices, but setting a delete phase is not that hard either, let alone that they get to decide if they add a frozen phase too. Set this once and never care for it again. 

 The only downside I can think of, is in the case of existing users having both old indices and data streams immediately after the upgrade. In the case of users already having the knowledge to be managing their metric, this can be a frustrating situation at the beginning, but is it really a problem for this audience?

Overall the behavior is practically the same with more alternatives, in case of problems, that's why I prefer it.

I am open for discussion of course, and If we reach a conclusion on the topic I'll make sure I correct the documentation to depict what we're doing.

docs/configuration.rst
NickDris: > Do we want to change the default for this so it uses data streams by default?

That question was also posed by @pquentin and it's valid. Let's discuss it on https://github.com/elastic/rally/pull/2051#discussion_r2982384092

NickDris: > We shouldn't create the @\custom component, I don't think. We can refer to a non existing template, and that will be accepted still. That way we don't need to worry about overwriting templates that are specifically custom.

That's a valid point. Do you mean that, it is already in need of replacement if users decide to set this, so what's the point of adding an empty template? 
I am now a bit worried about expressing the purpose of both cases: a _non-existent_ or _empty_ `@custom` component...
I am wondering about the alternatives:
- Same but never overwrite `@custom` components.
- Refer to a non-existing `@custom` template which still serves the main purpose: point out the customizable usage of composable templates.
- Don't set a `@custom` component at all and describe how to do this only in the documentation.

esrally/metrics.py
NickDris: EsMetricsStore:
https://github.com/NickDris/rally/blob/56e9eaf5c96b6ba39d62affb9d0668fa62d72667/esrally/metrics.py#L1181-L1185

EsRaceStore
https://github.com/NickDris/rally/blob/56e9eaf5c96b6ba39d62affb9d0668fa62d72667/esrally/metrics.py#L2035-L2040

https://github.com/NickDris/rally/blob/56e9eaf5c96b6ba39d62affb9d0668fa62d72667/esrally/metrics.py#L2070-L2082

EsResultsStore:
https://github.com/NickDris/rally/blob/56e9eaf5c96b6ba39d62affb9d0668fa62d72667/esrally/metrics.py#L2269-L2273

Something I'm missing?

docs/configuration.rst
NickDris: There's actually a problem with metrics@custom. This is pre-existing and used by many indices in a new elasticsearch cluster, so this has to change for sure.

gareth-ellis: yea - we should use something like `rally-metrics@custom`

esrally/metrics.py
pquentin: Oh, wow, sorry, I thought `self.client` and `self._client` referred to the Elasticsearch Python client.

esrally/metrics.py
```diff
@@ -76,17 +93,20 @@ def exists(self, index):
     def refresh(self, index):
         return self.guarded(self._client.indices.refresh, index=index)
 
-    def bulk_index(self, index, items):
+    def bulk_index(self, index, items, use_data_streams=True):
```
pquentin: What are your thoughts on avoiding the default value here to ensure the only default is set at the configuration level (and in specific tests)?

```suggestion
    def bulk_index(self, *, index, items, use_data_streams):
```

esrally/metrics.py
```diff
@@ -76,17 +93,20 @@ def exists(self, index):
...
         # pylint: disable=import-outside-toplevel
         import elasticsearch.helpers
 
+        if use_data_streams:
+            for item in items:
+                item["_op_type"] = "create"
         self.guarded(elasticsearch.helpers.bulk, self._client, items, index=index, chunk_size=5000)
 
-    def index(self, index, item, id=None):
+    def index(self, index, item, id=None, use_data_streams=True):
```
pquentin: Same questions as above:

```suggestion
    def index(self, *, index, item, id=None, use_data_streams):
```

esrally/metrics.py
```diff
@@ -76,17 +93,20 @@ def exists(self, index):
...
         doc = {"_source": item}
-        if id:
+        if not use_data_streams and id:
             doc["_id"] = id
```
pquentin, line 111: Silently keeping out the id seems wrong. It's used at least for the race store:
* I don't think we can have collisions given we use uuids?
* Some Rally commands rely on this id, for example to delete a race

esrally/metrics.py
```diff
@@ -199,7 +219,7 @@ def guarded(self, target, *args, **kwargs):
                     node = self._client.transport.node_pool.get()
                     msg = (
                         "An error [%s] occurred while running the operation [%s] against your Elasticsearch metrics store on host [%s] "
-                        "at port [%s]." % (e.error, target.__name__, node.host, node.port)
+                        "at port [%s]. args: %s, kwargs: %s" % (e.error, target.__name__, node.host, node.port, args, kwargs)
```
pquentin: Let's use the Elasticsearch/Rally convention of brackets to clearly delineate dynamic data.

```suggestion
                        "at port [%s]. args: [%s], kwargs: [%s]" % (e.error, target.__name__, node.host, node.port, args, kwargs)
```

esrally/metrics.py
```diff
@@ -213,7 +233,7 @@ def guarded(self, target, *args, **kwargs):
                     err = e
                 msg = (
                     "Transport error(s) [%s] occurred while running the operation [%s] against your Elasticsearch metrics store on "
-                    "host [%s] at port [%s]." % (err, target.__name__, node.host, node.port)
+                    "host [%s] at port [%s]. args: %s, kwargs: %s" % (err, target.__name__, node.host, node.port, args, kwargs)
```
pquentin: Let's use the Elasticsearch/Rally convention of brackets to clearly delineate dynamic data.

```suggestion
                    "host [%s] at port [%s]. args: [%s], kwargs: [%s]" % (err, target.__name__, node.host, node.port, args, kwargs)
```

pquentin (commented):
Sorry, more code-level nitpicks. I hope someone else will cover the general approach.

docs/migrate.rst
gbanasiak: I'm OK with defaulting to data streams. I compare this with behavior of known clients such as Filebeat. At certain version they just flipped to data streams by default.

docs/configuration.rst
gbanasiak: > We shouldn't create the @\custom component, I don't think.

That would be nice, but to achieve that we need to use `ignore_missing_component_templates` option which was introduced in 8.7, see previous discussion on the same in https://github.com/elastic/rally/pull/1854. https://github.com/elastic/rally/pull/2024 drops support for ES older than 8.0, but 8.0 is still older than 8.7. We would need to be more strict on metric store ES versions than ES versions that we support for testing. Do we want such duality?

docs/migrate.rst
gbanasiak: Rally 2.12.0 ship has sailed, and we are releasing 2.13.0 as the last version with old ES Python client. We should target 2.14.0 with this change so migration documentation should reflect that.

esrally/metrics.py
gbanasiak: > Some Rally commands rely on this id, for example to delete a race

I thought the same :) but actually races are removed using delete by query based on `race-id` field value ([src](https://github.com/elastic/rally/blob/14ae93754d2a3c2bcc48db0944447ae1832697b2/esrally/metrics.py#L1937-L1939)). Only annotations are removed based on document ID but those are not touched here.

Can we simply drop storing race documents with explicit ID regardless regular index or data stream?

esrally/metrics.py
```diff
@@ -328,7 +370,212 @@ def _read(self, template_name):
                 template["template"]["settings"]["index"]["number_of_shards"] = int(self._number_of_shards)
             if self._number_of_replicas is not None:
                 template["template"]["settings"]["index"]["number_of_replicas"] = int(self._number_of_replicas)
-            return json.dumps(template)
+            if self._use_data_streams and support_data_streams:
+                index_pattern = template["index_patterns"][0]
+                if index_pattern.endswith("-*"):
+                    index_pattern = index_pattern.replace("-*", "-v*")
+                    template["index_patterns"][0] = index_pattern
```
gbanasiak: It seems that `index_patterns` from JSON template are ignored anyway, so why do we update them here? The `v*` bit discussed elsewhere.

esrally/metrics.py
```diff
@@ -328,7 +370,212 @@ def _read(self, template_name):
...
+                if not template["template"]["mappings"]["properties"].get("@timestamp"):
+                    template["template"]["mappings"]["properties"]["@timestamp"] = {"type": "date", "format": "epoch_millis"}
```
gbanasiak: Please add `@timestamp` field to `races-template.json` and `results-template.json`. It cannot be optional as `@timestamp` is now part of race/results documents as per `as_dict()` methods modification.

esrally/metrics.py
```diff
@@ -328,7 +370,212 @@ def _read(self, template_name):
...
+            return json.dumps(json.load(f))
+
+    def _data_stream_template(self, component_templates):
+        return json.dumps(
+            {
+                "index_patterns": [f"{self._es_store_type.index_prefix}v*"],
+                "data_stream": {},
+                "composed_of": component_templates,
+                "priority": self.TEMPLATE_PRIORITY,
+            }
```
gbanasiak, line 497: Regarding `index_patterns`, doesn't `v*` defeat the purpose of versioning? Today there will be `rally-metrics-v1` DS, in the future `rally-metrics-v2` DS. If there's `v*` it will match both.

The non-standard high priority is to take precedence over wide pattern in old templates, correct?

esrally/metrics.py
```diff
@@ -328,7 +370,212 @@ def _read(self, template_name):
...
+        old_template = None
+        if self._client.index_template_exists(f"{self._es_store_type.index_template_name}-ds"):
+            for existing in self._client.get_template(f"{self._es_store_type.index_template_name}-ds").body.get("index_templates", []):
+                old_template = existing.get("index_template", {})
+                break
+
+        if not self._should_apply_update("index template", old_template, new_template):
+            return
+
+        self._client.put_template(f"{self._es_store_type.index_template_name}-ds", self._data_stream_template(component_names))
```
gbanasiak: Shouldn't we version the names of new composable templates? For example, `rally-metrics-ds-v1`? Or, to ask differently for the same, if we bump up version number, do we want to have both old and new composable templates? If we want to version them, then component templates also need to be versioned.

esrally/metrics.py
```diff
@@ -328,7 +370,212 @@ def _read(self, template_name):
...
+                            "index": {
+                                **template.get("settings", {}).get("index", {}),
+                                "lifecycle": {"name": lifecycle_policy_name},
+                            }
+                        }
+                    }
+                }
+            ),
+            f"{name}{self.COMPONENT_TEMPLATE_CUSTOM_SUFFIX}": json.dumps({"template": {}}),
+        }
```
gbanasiak, line 412: We don't have to divide into 3 different component templates. The only thing that is absolutely necessary is handling of `@custom` component template. All the rest could be embedded directly into composable template I think. Less entities to version, should we wish to version.

gbanasiak (commented):
Further thoughts on templates below. This needs testing but I'll wait with that until open discussion points are resolved.

docs/migrate.rst
NickDris: How do we express that this is incorporated at  an upcoming version since it is not released yet? Maybe omit the migration document right now and wait for the next release?

esrally/metrics.py
NickDris: Yeah, that's safer, thanks!

esrally/metrics.py
NickDris: Indeed, I find that this a better practice.

esrally/metrics.py
NickDris: > Silently keeping out the id seems wrong. 

We should keep out the id in the case of `rally-metrics`, for avoiding performance issues on indexing and at the moment, races and annotations only uses doc id, so we are good.

> I don't think we can have collisions given we use uuids?

Keeping uuids is not avoiding the problem of duplication afaiu, because ES can store multiple docs with same fields and values but not same _id.. I think the original design choice was because of that deduplication, does that answer the questoin?

> Can we simply drop storing race documents with explicit ID regardless regular index or data stream?

Using id is serving the purpose of deduplication in case of races and annotations, which seems reasonable. I don't see the reason to remove it? Is it because we can't do it for `rally-metrics`, so selectively using it for  annotations and races. confuses the reader?.

esrally/metrics.py
NickDris: Good catch, thanks a lot!

esrally/metrics.py
NickDris: > Regarding index_patterns, doesn't v* defeat the purpose of versioning? Today there will be rally-metrics-v1 DS, in the future rally-metrics-v2 DS. If there's v* it will match both.

Oh, you're totally right!

> The non-standard high priority is to take precedence over wide pattern in old templates, correct?

Yes, priority takes precedence in ES.

docs/configuration.rst
NickDris: I am more keen of the idea of adding an empty component there, waiting for customization, instead of missing components.

esrally/metrics.py
NickDris: Yeah, I see no reason not to use versioned component templates too, seems more manageable on a future version bump

docs/migrate.rst
pquentin: We just need the Rally release to ship, and then this can go in the `master` branch as "Migrating to Rally 2.14.0 (unreleased)". It won't show in our docs since we show the latest release by default.

esrally/metrics.py
gbanasiak: > Using id is serving the purpose of deduplication in case of races and annotations, which seems reasonable. I don't see the reason to remove it? Is it because we can't do it for rally-metrics, so selectively using it for annotations and races. confuses the reader?.

To be clear, I wanted to drop explicit document IDs with races, not annotations. My reasoning was: any setup with roll-over (doesn't matter legacy or data stream) cannot handle deduplication 100% because the write index changes. The only index without roll-over is for annotations. Now I've looked closer at the code and it turns out we do _update_ race document. It is created once race preparation completes, and then at the end of the race (if no errors/cancellation).
```
esrally/racecontrol.py
254:        self.race_store.store_race(self.race)
280:            self.race_store.store_race(self.race)
```
This worked OK with fixed rollover because the target index depended on the race timestamp, but will _not_ work reliably with data streams because the write index can change at _any_ time.

This means race document update needs to be handled through update by query based on `race-id`. Once we switch to update by query, removal of explicit document ID in races will be possible and logical.

gbanasiak: Looks like race document update is only to include results. Why do we duplicate results data? There's `results` field in race document, and a separate results index. Γιατί; Where I'm getting with this: can we avoid update by query and rely just on results index?

NickDris: Oh I get it now, thanks for elaboration.

Check 5da47fb4f4ce1eaf475760757ac22f42dafc1a1f. Is this what you had in mind?

> Why do we duplicate results data? There's results field in race document, and a separate results index. Γιατί;

I cannot think of any reason tbh. Maybe we had planned to drop results indices and the plan is left mid-flight?  
I am going to leave "curation" of metrics as tech debt for the future, I believe it is not part of this PR.

gareth-ellis: I can imagine that having the results in a single field makes presenting the results a lot quicker - whilst having a seperate index makes things like dashboards a lot simpler. I dont think we should get rid of the results index; you could look into (outside of this PR), the feasibility of presenting the results direct from the results index - but i dont know if it really giives us much, short of dropping one update query

esrally/metrics.py
NickDris: Addressed in 64763fa2a0c3605ed050c17e8928b80c19f0a56d

docs/configuration.rst
NickDris: Addressed in 64763fa2a0c3605ed050c17e8928b80c19f0a56d

docs/migrate.rst
NickDris: Update: 64763fa2a0c3605ed050c17e8928b80c19f0a56d

docs/metrics.rst
NickDris: Check 5f6d3226e21d4494ea9104543d5daed63e540793 , did you have that in mind? Is there anything I could change?

gbanasiak:
## Manual test plan

### Legacy indices
- [x] create test metric cluster and corresponding `rally.ini`
- [x] set `datastore.use_data_streams` setting to `false`
- [x] run a race, confirm data in legacy `rally-*` indices (`esrally list races`, `esrally compare`, manual check)
- [x] delete a race, confirm data removed from legacy `rally-*` indices
- [x] modify legacy `rally-metrics` template
- [x] run another race, confirm `rally-metrics` template _not_ modified
- [x] set `datastore.overwrite_existing_templates` setting to `true`
- [x] run another race, confirm `rally-metrics` template modified
- [x] remove `datastore.overwrite_existing_templates` setting
- [x] add annotation
- [x] remove annotation

### Data streams
- [x] remove `datastore.use_data_streams` setting
- [x] run a race, confirm data in data stream backing indices (`esrally list races`, `esrally compare`, manual check)
- [x] delete a _new_ race, confirm data removed from DS
- [x] delete an _old_ race, confirm data removed from legacy indices
- [x] modify `rally-metrics-v1` template, `rally-metrics-default` ILM policy, and `rally-metrics-v1@custom` component template
- [x] run another race, confirm `rally-metrics-v1` template, `rally-metrics-default` ILM policy, and `rally-metrics-v1@custom` component templates _not_ modified
- [x] set `datastore.overwrite_existing_templates` setting to `true`
- [x] run another race, confirm `rally-metrics-v1` template, and `rally-metrics-default` ILM policy modified, and `rally-metrics-v1@custom` component template _not_ modified
- [x] add annotation
- [x] remove annotation

### Data streams rollback
- [x] remove `datastore.overwrite_existing_templates` setting
- [x] set `datastore.use_data_streams` setting to `false` (again)
- [x] run a race, confirm data in legacy `rally-*` indices (`esrally list races`, `esrally compare`, manual check)

docs/migrate.rst
```diff
@@ -1,6 +1,102 @@
...
+      "source": { "index": "rally-results-*" },
+      "dest": { "index": "rally-results-v1", "op_type": "create" }
+    }
+
+**Step 3 — Verify and clean up**
+
+After reindexing, verify that the document counts match::
+
+    GET _cat/count/rally-metrics-v1
+    GET _cat/count/rally-metrics-*
```
gbanasiak: The second count should _include_ the first, so the counts shouldn't match I think?

docs/metrics.rst
```diff
@@ -6,9 +6,82 @@ Metrics Records
...
+
+Rally creates three versioned data streams:
+
+* ``rally-metrics-v1`` — benchmark metric samples (throughput, latency, etc.)
+* ``rally-races-v1`` — race metadata (track, challenge, car, timestamps)
+* ``rally-results-v1`` — aggregated race results for Kibana reporting
+
+**Index templates and component templates**
+
+Each data stream (e.g. ``rally-metrics-ds``) is backed by an index template that is composed of two `component templates <https://www.elastic.co/docs/manage-data/data-store/templates#component-templates>`_:
```
gbanasiak: Is `rally-metrics-ds` correct? Please provide composable index template name here, component template names are mentioned below.

docs/metrics.rst
```diff
@@ -6,9 +6,82 @@ Metrics Records
...
 At the end of a race, Rally stores all metrics records in its metrics store. Metrics can be kept in memory or written to a dedicated Elasticsearch cluster (not the cluster where Rally ran its benchmarks). This can be configured in the `[reporting] section <https://esrally.readthedocs.io/en/stable/configuration.html#reporting>`_.
 
-Rally stores the metrics in the indices ``rally-metrics-*``. It will create a new index for each month. Here is a typical metrics record::
+By default, Rally stores metrics in the ``rally-metrics-v1`` data stream. When ``datastore.use_data_streams`` is set to false, Rally falls back to monthly indices named ``rally-metrics-YYYY-MM``.
 
 
+.. _metrics_data_streams:
+
+Data Stream Storage
+~~~~~~~~~~~~~~~~~~~
```
gbanasiak: IMO, this section should be promoted to `===` level. At the same level as `Metrics Records` and `Metric Keys`.

docs/metrics.rst
```diff
@@ -6,9 +6,82 @@ Metrics Records
...
+
+**Shard and replica configuration**
+
+You can also configure shard and replica counts globally for all Rally indices via ``rally.ini``::
+
+    [reporting]
+    datastore.number_of_shards = 3
+    datastore.number_of_replicas = 1
+
+These settings are applied to the main component template and affect all Rally data streams.
```
gbanasiak: I think we should ignore `number_of_shards` and `number_of_replicas` settings when using data streams, and rely on ES administrator to provide the right values in templates, e.g. `@custom` component templates. We've learned that having those settings on Rally side just opens up a possibility of desync in a fleet of Rally hosts. Let's fix it as part of this change.

docs/migrate.rst
```diff
@@ -1,6 +1,102 @@
...
+
+    POST _reindex
+    {
+      "source": {
+        "index": "rally-metrics-*"
+      },
+      "dest": {
+        "index": "rally-metrics-v1",
+        "op_type": "create"
+      }
```
gbanasiak, line 65: This and the following reindexing commands specifies a source that _includes_ the destination. Is it safe?

gbanasiak (changes requested):
I've started testing using [this](https://github.com/elastic/rally/pull/2051#issuecomment-4244919649) WIP plan. I'm using the following to trigger a race:

```
esrally race --config=<name> \
  --track=geonames \
  --challenge=append-no-conflicts-index-only \
  --distribution-version=9.2.4 \
  --track-params="ingest_percentage:10"
```

I'm running into the following error early in the plan when using legacy indices:
```
2026-04-14 20:56:31,575 ActorAddr-(T|:61357)/PID:87843 esrally.metrics INFO Opening metrics store for race timestamp=[20260414T205628Z], track=[geonames], challenge=[append-no-conflicts-index-only], car=[['defaults']]
2026-04-14 20:56:31,751 ActorAddr-(T|:61357)/PID:87843 elastic_transport.transport INFO PUT https://grzegorz-test-6eee40.es.europe-west4.gcp.elastic-cloud.com:443/_index_template/rally-metrics [status:200 duration:0.046s]
2026-04-14 20:56:31,751 ActorAddr-(T|:61357)/PID:87843 esrally.mechanic.mechanic ERROR Cannot process message [<esrally.mechanic.mechanic.StartNodes object at 0x10c56d310>]

  File "/Users/grzegorz/src/rally/esrally/mechanic/mechanic.py", line 581, in receiveMsg_StartNodes
    metrics_store.open(ctx=msg.open_metrics_context)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/Users/grzegorz/src/rally/esrally/metrics.py", line 1158, in open
    index_name = self._index_handler.index_name(race_timestamp)

  File "/Users/grzegorz/src/rally/esrally/metrics.py", line 461, in index_name
    return f"{self._es_store_type.index_prefix}{ts.year:04d}-{ts.month:02d}"
                                                ^^^^^^^

AttributeError: 'NoneType' object has no attribute 'year'
```

This problem is not present in recent `master`. Please fix.

Other than that I have more feedback regarding documentation.

docs/metrics.rst
gareth-ellis: could we not default in the rally-metrics template, and allow `@custom` to override? the idea being that anything a user has added to the `@custom` template, they should be aware of that already.

inqueue: Ignoring `number_of_shards` and `number_of_replicas` simplifies things and will be friendlier to serverless where those settings are not permitted anyway. 💯 to relying on the ES administrator to set these in a component template. I expect the ES defaults will be plenty sufficient most of the time. Those of us doing high volume benchmarking with large metrics stores are fully capable of managing these settings in a template.

inqueue: > could we not default in the rally-metrics template, and allow `@custom` to override? 

I vote to avoid the settings. 1P1R would be a reasonable & minimal default for the rally-metrics template, and these settings are already established as the stateful defaults for index settings.

docs/migrate.rst
NickDris: You are right, thanks!
I change the intention to add number `2` in order to scope the second query to date-based pattern

docs/migrate.rst
NickDris: No you are right, thanks!
I changed it, so we scope the source to date-based patterns `rally-metrics-2*`

docs/metrics.rst
NickDris: I was not aware of that possibility of desync and addressed it in 81ddff0e92ee80c0f9ba65839c6785e29666f764. These two settings are ignored when used with data streams.

docs/metrics.rst
```diff
@@ -6,9 +6,76 @@ Metrics Records
...
 At the end of a race, Rally stores all metrics records in its metrics store. Metrics can be kept in memory or written to a dedicated Elasticsearch cluster (not the cluster where Rally ran its benchmarks). This can be configured in the `[reporting] section <https://esrally.readthedocs.io/en/stable/configuration.html#reporting>`_.
 
-Rally stores the metrics in the indices ``rally-metrics-*``. It will create a new index for each month. Here is a typical metrics record::
+By default, Rally stores metrics in the ``rally-metrics-v1`` data stream. When ``datastore.use_data_streams`` is set to false, Rally falls back to monthly indices named ``rally-metrics-YYYY-MM``.
 
 
+.. _metrics_data_streams:
+
+Data Stream Storage
+-------------------
```
gbanasiak: Thanks for lifting to `---` level, but it now cuts previous `Metrics Records` in half, see the list of fields starting from `environment`. This must be kept in `Metrics Records` section.

esrally/metrics.py
```diff
@@ -328,7 +372,218 @@ def _read(self, template_name):
...
+            old_template = None
+            for existing in self._client.get_template(self._es_store_type.date_based_template_name).body.get("index_templates", []):
+                old_template = existing.get("index_template", {}).get("template", {})
+                break
+            new_template = json.loads(_index_template)["template"]
+
+            if not self._should_apply_update("index template", old_template, new_template):
+                return
+
+        self._client.put_template(self._es_store_type.date_based_template_name, _index_template)
```
gbanasiak: In the previous logic, `put_template()` was only executed if `create` was true. BTW, `race_timestamp` is not used it seems, I wonder why pylint does not throw a warning here.

gbanasiak (changes requested):
Thank you for improvements. In the second round of tests I've discovered that `rally-metrics` template gets overwritten with the default value of `datastore.overwrite_existing_templates`. Please fix.

docs/metrics.rst
gbanasiak: Just to be clear, keep fields descriptions in `Metrics Records` and `Data Stream Storage` as a separate section at the same level.

NickDris:
I went over the test plan in https://github.com/elastic/rally/pull/2051/#issuecomment-4244919649 and everything checks out. Tell me if anything else is needed here.

esrally/metrics.py
```diff
@@ -328,7 +372,221 @@ def _read(self, template_name):
...
+
+        new_template = json.loads(self._data_stream_template(component_names))
+        old_template = None
+        if self._client.index_template_exists(self._es_store_type.data_stream_template_name):
+            for existing in self._client.get_template(self._es_store_type.data_stream_template_name).body.get("index_templates", []):
+                old_template = existing.get("index_template", {})
+                break
+
+        if not self._should_apply_update("index template", old_template, new_template):
+            return
```
gbanasiak, line 564: nit: can we add actual template name here, similarly to what is done for component templates?

esrally/metrics.py
```diff
@@ -328,7 +372,221 @@ def _read(self, template_name):
...
+            self.logger.info("Create %s:\n%s", resource_label, pretty.dump(new_resource, pretty.Flag.FLAT_DICT))
+            return True
+
+        diff = pretty.diff(old_resource, new_resource, pretty.Flag.FLAT_DICT)
+        if diff == "":
+            self.logger.info("Keep existing %s (it is identical)", resource_label)
+            return False
+        if not self.overwrite_templates:
+            self.logger.info("Keep existing %s (datastore.overwrite_existing_templates = false):\n%s", resource_label, diff)
+            return False
```
gbanasiak, line 515: I would keep debug log level here instead of info level. It's very useful now when reviewing the PR, but overall will be quite noisy. We can always increase log level when troubleshooting.

esrally/metrics.py
```diff
@@ -293,6 +316,33 @@ def create(self):
...
+    metrics = ("metrics", "v1")
+    results = ("results", "v1")
+    races = ("races", "v1")
+
+    def __new__(cls, metric_name, version):
+        obj = object.__new__(cls)
+        obj.metric_name = metric_name
+        obj.index_prefix = f"rally-{obj.metric_name}-"
+        obj.index_template_resource = f"{obj.metric_name}-template"
+        obj.ilm_default_name = f"{obj.index_prefix}ilm-default"
```
gbanasiak, line 338: super nit: I think "ilm" is repetitive, we know type of object by ES endpoint
```suggestion
        obj.ilm_default_name = f"{obj.index_prefix}default"
```

esrally/metrics.py
```diff
@@ -293,6 +316,33 @@ def create(self):
...
+        obj.ilm_default_resource = "ilm-default"
+        obj.data_stream_template_name = f"rally-{obj.metric_name}-template-{version}"
```
gbanasiak, line 340: super nit: same as above
```suggestion
        obj.data_stream_template_name = f"rally-{obj.metric_name}-{version}"
```

gbanasiak (changes requested):
Thank you.

In the third round of tests I discovered `datastore.overwrite_existing_templates = true` setting leads to custom component template overwrite. Custom templates, once created should never be overwritten.

```
2026-04-17 13:02:13,842 ActorAddr-(T|:54560)/PID:68160 metricsesrally.metrics WARNING Overwrite existing component template [rally-metrics-v1@custom] (datastore.overwrite_existing_templates = true):
+ {}
- {
-   "settings.index.number_of_shards": "7"
- }
```

I think this is the last item. I ran through all the remaining test steps and they look good.

esrally/metrics.py
gbanasiak: feel free to ignore if you think otherwise

NickDris:
Added the following manual test (additionally to the unit tests)

### Custom component templates left untouched

- [x] Run a race with `datastore.use_data_streams = true` (or unset, since it's default)
- [x] Modify `rally-metrics-custom`
- [x] Run a race and ensure `rally-metrics-custom` left untouched
- [x] Set `datastore.overwrite_existing_templates = true`
- [x] Run a race and ensure that `rally-metrics-custom` is still untouched after the modification.

gbanasiak (approved):
LGTM. Many thanks for your patience and multiple iterations.

NickDris:
Thank you @gbanasiak and many thanks for your clear guidance on the testing direction and documentation update.

NickDris:
@elasticmachine update branch

PR #2053, opened 2026-03-06
Fix for ingest-doc-count to Correctly Handle Partition When Calculating Total Bulks 

john-wagster:
When we added ingest_doc_count we didn't take into account the start_index and end_index which only matters when we have a multi-process scenario that needs to take a subset of the docs into account.  In this case previously it was leading to incorrectly calculating total bulks and only ingesting a subset of data.  

This PR should fix that case by taking into account start_index and end_index that same way we do with ingest_percentage.

I was able to manually update rally on a benchmark test and verify that this fixed the issue.

AI helped with this one.


davidkyle (approved):
LGTM

PR #2056, opened 2026-03-10
Add render-track command

pquentin:
The `render-track` command renders a track's Jinja2 templates and outputs the resulting JSON. This is useful for debugging track templates or inspecting the final track definition after template variables have been substituted.

### DIFF
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -255,6 +255,17 @@ def create_arg_parser():
         default="",
     )
     render_track_parser.add_argument(
+        "--build-flavor",
+        help="Define the build flavor to render the track for. This is used to render tracks differently based on the build flavor of the Elasticsearch nodes that will be benchmarked. List possible build flavors with `{PROGRAM_NAME} list tracks`.",
+        choices=["default", "serverless"],
+    )
+    render_track_parser.add_argument(
+        "--serverless-operator",
+        help="Whether to render the track for a serverless operator.",
+        default=False,
+        action="store_true",
+    )
+    render_track_parser.add_argument(
         "--output-path",
         help="Path to the output file. If not specified, the rendered track is written to stdout.",
         default=None,
@@ -1283,7 +1294,9 @@ def dispatch_sub_command(arg_parser, args, cfg: types.Config):
             if args.track:
                 cfg.add(config.Scope.applicationOverride, "track", "track.name", args.track)
             cfg.add(config.Scope.applicationOverride, "track", "params", opts.to_dict(args.track_params))
-            track.render_track(cfg, output_path=args.output_path)
+            track.render_track(
+                cfg, build_flavor=args.build_flavor, serverless_operator=args.serverless_operator, output_path=args.output_path
+            )
         else:
             raise exceptions.SystemSetupError(f"Unknown subcommand [{sub_command}]")
         return ExitStatus.SUCCESSFUL
--- a/esrally/rally.py
+++ b/esrally/rally.py
@@ -901,7 +901,7 @@ def create_arg_parser():
         p.add_argument(
             "--quiet",
             help="Suppress as much as output as possible (default: false).",
-            default=p is not render_track_parser,
+            default=p is render_track_parser,  # disable output for render-track
             action="store_true",
         )
         p.add_argument(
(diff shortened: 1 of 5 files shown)


inqueue (approved):
Great, this is useful! The first thing I did to test was:

```shell
esrally render-track --track-path="$TRACKPATH" | jq '.|keys'
jq: parse error: Invalid numeric literal at line 2, column 9
Exception ignored on flushing sys.stdout:
BrokenPipeError: [Errno 32] Broken pipe
```

Following with `--quiet` stripped the banner and footer and then it was good. What do you think about adding a `sys.stdout.isatty()` check to avoid having to use `--quiet` when piping?

esrally/track/loader.py
```diff
@@ -175,6 +175,23 @@ def list_tracks(cfg: types.Config):
     console.println(tabulate.tabulate(tabular_data=data, headers=headers))
 
 
+def render_track(cfg: types.Config, output_path=None):
+    repo = track_repo(cfg)
+    track_name = repo.track_name
+    track_spec_file = repo.track_file(track_name)
+    track_params = cfg.opts("track", "params", default_value={})
+
+    rendered = render_template_from_file(track_spec_file, track_params)
```
gbanasiak: We're missing `build_flavor` and `serverless_operator` arguments in `render_template_from_file()` so we can only see a stateful version of a track. Would be nice to extend `render-track` arguments to be able to render serverless version of a track with and without operator privileges.

esrally/rally.py
```diff
@@ -883,7 +901,7 @@ def add_track_source(subparser):
         p.add_argument(
             "--quiet",
             help="Suppress as much as output as possible (default: false).",
-            default=False,
+            default=p is not render_track_parser,
```
gbanasiak: I think `p is render_track_parser` was intended here?

The following worked for me:
```
        p.add_argument(
            "--quiet",
            help=f"Suppress as much as output as possible (default: {str(p is render_track_parser).lower()}).",
            default=p is render_track_parser,
            action="store_true",
        )
```

gbanasiak (commented):
Thanks for the change. That's useful. I've left some comments.

> I think it should be the default for commands that output JSON, so I made the change in [#2056 (9879a06d2)](https://github.com/elastic/rally/pull/2056/changes/9879a06d2d539b55ad68915851599ad27f4e2bba) (and adjusted the docs in a later commit). I did not do it initially because the code seemed ugly, but I now prefer it this way. What do you think?

Changing `--quiet` default for `render-track` makes sense to me.

> BTW, the integration tests are broken in the master branch too.

Last couple of periodic ones were fine.

pquentin:
Thanks for the great command, PTAL.

esrally/rally.py
```diff
@@ -883,7 +912,7 @@ def add_track_source(subparser):
         p.add_argument(
             "--quiet",
             help="Suppress as much as output as possible (default: false).",
-            default=False,
+            default=p is render_track_parser,  # disable output for render-track
             action="store_true",
         )
```
gbanasiak, line 917: nit: I would still modify `help` text to determine the default value depending on the command. This seems to work. Also, we could fix the grammar: `Suppress as much as output as possible` -> `Suppress as much output as possible`.

gbanasiak (approved):
LGTM. Thanks for the next iteration.

esrally/rally.py
pquentin: Sorry for not applying your change, I missed the `help` part. The problem with `store_true` is that there was no way to disable quiet mode, so it wasn't really a default but rather an enforced change. Thankfully, `BooleanOptionalAction`, available with Python 3.9+, solves this problem.

esrally/rally.py
```diff
@@ -882,9 +911,9 @@ def add_track_source(subparser):
...
         p.add_argument(
             "--quiet",
-            help="Suppress as much as output as possible (default: false).",
-            default=False,
-            action="store_true",
+            help=f"Suppress as much as output as possible (default: {str(p is render_track_parser).lower()}).",
```
gbanasiak:
```suggestion
            help=f"Suppress as much output as possible (default: {str(p is render_track_parser).lower()}).",
```

esrally/rally.py
gbanasiak: Thanks for explaining!

PR #2058, opened 2026-03-12
ESQL Profile operation - Retrieve timing information for operators that use receive_nanos and emit_nanos

carlosdelest:
Follow up to https://github.com/elastic/rally/pull/2047

Adds timing information for operators like TopNOpeartor, that does not use `process_nanos`


inqueue (approved):
LGTM

PR #2062, opened 2026-03-16
Apply persist-credentials: false to GitHub Actions

pquentin:
By default, credentials are persisted, which can be a security issue.

This was highlighted by the excellent [zizmor](https://github.com/zizmorcore/zizmor) tool. There are other issues, but this seemed like a good first step.


gbanasiak (approved):
LGTM

PR #2069, opened 2026-03-23
Update release preparation to scripts/release

fressi-elastic:
Supersedes https://github.com/elastic/rally/pull/2067

## Summary

- **Release prep scripts** live under `scripts/release/`: `prepare.sh` (NOTICE, AUTHORS, changelog via `changelog.py`, version bump in `esrally/_version.py`, git commit), with `create-notice.sh` as the shared NOTICE generator.
- **Makefile**: `make release RELEASE_VERSION=X.Y.Z` runs `prepare.sh`; `make release-checks` / `make test-release` run `clean`, `lint`, `test-all`, `docs`, then `prepare.sh --dry` with the same version.
- **Removed** root `prepare-release.sh` and `release-checks.sh` in favor of the paths above.
- **`scripts/offline-install.sh`** sources `create-notice.sh` from `scripts/release/`.
- **Changelog automation**: `changelog.py` reads the GitHub token from a file; set **`RALLY_CHANGELOG_TOKEN_FILE`** to that path (default `~/.github/rally_release_changelog.token`). `github3.py` remains a **develop** dependency for this script.
- **Docs**: `CONTRIBUTING.md` (maintainer release pointer), `docs/developing.rst` (e.g. Java 21 for integration tests).
- **Tooling**: `pyproject.toml` / `uv.lock` updates as needed (including a documented **`pip`** pin for track dependency installs); `.pre-commit-config.yaml` runs **shellcheck** on `scripts/release/*.sh` and **pylint** includes `scripts/release/changelog.py`.
- **Small fix**: `esrally/utils/io.py` — tar extraction uses `filter="tar"` on Python 3.12+ with a fallback for 3.10/3.11.

## Out of scope

Per-release generated artifacts for an actual release (e.g. committed `CHANGELOG.md`, `AUTHORS`, `NOTICE.txt`, version bump) can land in a **follow-up release PR** after this merges.

## Maintainer runbook

- [ ] https://github.com/elastic/elasticsearch-team/pull/2568


docs/developing.rst
```diff
@@ -77,6 +78,44 @@ To get a rough understanding of Rally, it makes sense to get to know its key com
...
+
+#. In GitHub: **Settings → Developer settings → Personal access tokens → Fine-grained tokens → Generate new token**.
+#. Set a name and expiration (recommend a short lifetime such as 7 days for release-only tokens).
+#. **Resource owner:** your GitHub user (the usual case).
+#. **Repository access:** **Only select repositories** → choose **elastic/rally**. Your account must have a role on that repository that can manage milestones.
+#. **Repository permissions:** under **Issues**, set **Read and write**. GitHub maps milestone create/update and milestone-scoped issue queries to the Issues permission on fine-grained tokens.
+#. Generate the token and copy it when shown; it is displayed only once.
+
+If the **elastic** organization enforces SAML SSO, open the token on GitHub after creation and use **Configure SSO** / **Authorize** for **elastic**. Without that step, the API can return 403 even when permissions are correct.
```
fressi-elastic: We should remove this part.

pyproject.toml
```diff
@@ -92,6 +92,8 @@ dependencies = [
     "hatch==1.3.1",
     "hatchling==1.6.0",
     "wheel==0.46.2",
+    # Release Docker image pins the same pip (scripts/release/Dockerfile).
+    "pip==26.0.1",
```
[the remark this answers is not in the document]
fressi-elastic: This is intentional as pip is required to install track scripts dependencies after downloading them.

docs/developing.rst
```diff
@@ -8,6 +8,7 @@ Install the following software packages:
 
 * `uv <https://docs.astral.sh/uv/getting-started/installation/>`_ 
 * JDK version required to build Elasticsearch. Please refer to the `build setup requirements <https://github.com/elastic/elasticsearch/blob/main/CONTRIBUTING.md#contributing-to-the-elasticsearch-codebase>`_.
+  For running Rally's integration tests (e.g. ``make it`` or ``make it_tracks_compat``), ensure your environment uses **Java 17 or 21** (recent Rally versions use Java 21 in CI). Set ``JAVA_HOME`` or ``JAVA21_HOME`` accordingly.
```
gareth-ellis: Do we not need specifically Java 21? We build Elasticsearch with java21, I think 17 was the previous build jdk

docs/docker.rst
```diff
@@ -154,3 +154,17 @@ You can then build and test the image with::
 
     docker build --tag=custom-rally .
     docker run -ti custom-rally list tracks
+
+Release preparation image (maintainers)
```
gareth-ellis: So this preparation image, this is just for runing prepare.sh? Is there a need for this vs the above instructions for running prepare.sh from ones workstation?

scripts/release/prepare-docker.sh
```diff
@@ -0,0 +1,134 @@
...
+DOCKER_ARGS+=(
+	-e "HOME=${CONTAINER_HOME}"
+	-e PREPARE_RELEASE_NO_VERIFY=1
+	-e GIT_AUTHOR_NAME
+	-e GIT_AUTHOR_EMAIL
+	-e GIT_COMMITTER_NAME
+	-e GIT_COMMITTER_EMAIL
+	-v "${REPO_ROOT}:/workspace"
+	-v rally-prepare-release-venv:/workspace/.venv
+	-v "${TOKEN_FILE}:${CONTAINER_HOME}/.github/rally_release_changelog.token:ro"
```
gareth-ellis: This looks like it requires the file based method for the token, rather than env var that was allowed above

CONTRIBUTING.md
```diff
@@ -60,7 +71,7 @@ Once your changes and tests are ready to submit for review:
 
     Ensure that all tests pass by running `make check-all`. This runs sequentially lint checks, unit tests and integration tests. These can be executed in isolation using `make lint`, `make test` and `make it` respectively, in case you need to iterate over a subset of tests.
 
-    Note: Integration tests are much slower than unit tests and require `docker-compose`.
+    Note: Integration tests are much slower than unit tests and require `docker-compose`. They also require **Java 17 or 21**; set `JAVA_HOME` (or `JAVA21_HOME`) to match.
```
gareth-ellis: as above, I am fairly sure we _require_ java21 for recent ES commits (last couple of years)

gareth-ellis (commented):
Added some initial comments

docs/docker.rst
fressi-elastic: The `prepare.sh` requires Linux and have some requirements that need to be preinstalled on the workstation. Tho help having a reproducible environment I provided an alternative way of executing `prepare.sh` in a docker container. This should simplify the execution and reproducing issues on a reference environment (the container). This part of the document is being removed because it could create confusion. The image is only intended for our internal use and it is actually derived by the official rally image (just because it has to run `rally --version` for verification).

CONTRIBUTING.md
fressi-elastic: I must figure out and correct this.

docs/developing.rst
fressi-elastic: I must figure out and correct this.

fressi-elastic: Java 17 come out from Elasticsearch 8.x requirements. I removed it from here.

CONTRIBUTING.md
fressi-elastic: Java 17 come out from Elasticsearch 8.x requirements. I am going to remove it from here.

scripts/release/prepare-docker.sh
fressi-elastic: To avoid confusion I renamed the environment variable `RALLY_CHANGELOG_TOKEN` to `RALLY_CHANGELOG_TOKEN_FILE`

CONTRIBUTING.md
```diff
@@ -79,6 +92,10 @@ Then sit back and wait. There will probably be discussion about the pull request
 
 Note: Contributors belonging to the "Elastic" organization on Github can merge PRs themselves after getting a "LGTM" (Looks good to me); this workflow is similar to the established one in the Elasticsearch project.
 
+## Release process (maintainers)
+
+Follow the internal **Rally Release Process** runbook: https://codex.elastic.dev/r/elasticsearch-team/teams/performance/runbooks/rally-release-process
```
gareth-ellis: We should not be referring to internal material in the public repo - this adds a barrier to external contributors.

docs/developing.rst
```diff
@@ -76,8 +77,3 @@ To get a rough understanding of Rally, it makes sense to get to know its key com
 * `Reporter`: A reporter tells us how the race went (currently only after the fact).
 
 There is a dedicated :doc:`tutorial on how to add new tracks to Rally</adding_tracks>`.
-
-How to contribute code
-----------------------
-
-See the `contributors guide <https://github.com/elastic/rally/blob/master/CONTRIBUTING.md>`_. We strive to be PEP-8 compliant but don't follow it to the letter.
```
gareth-ellis, line 83: why is this removed?

fressi-elastic: I was a mistake.

CONTRIBUTING.md
fressi-elastic: Agree. To be removed.

scripts/release/prepare.sh
```diff
@@ -0,0 +1,135 @@
...
+	echo "dry-run: skipping git add/commit, pip install, and esrally --version check." >&2
+	echo ""
+	echo "===================="
+	echo "Dry run finished for ${RELEASE_VERSION}"
+	echo "===================="
+	exit 0
+fi
+
+echo "Updating release version number"
+printf '__version__ = "%s"\n' "$RELEASE_VERSION" > esrally/_version.py
```
gbanasiak, line 99: Will this stop working after https://github.com/elastic/rally/pull/2080? Just making sure we're aligned here, cc @pquentin .

scripts/release/prepare.sh
```diff
@@ -0,0 +1,135 @@
...
+		echo "Unstage unrelated changes (e.g. git restore --staged <file>) or start from a clean tree." >&2
+		exit 1
+		;;
+	esac
+done < <(git diff --cached --name-only)
+
+# Non-empty PREPARE_RELEASE_NO_VERIFY adds --no-verify (e.g. scripts/release/prepare-docker.sh).
+git commit -m "Bump version to $RELEASE_VERSION" ${PREPARE_RELEASE_NO_VERIFY:+--no-verify}
+
+pip install --editable .
```
gbanasiak, line 119: Is `pip install` required considering we're running this under uv via Makefile?

Makefile
```diff
@@ -216,12 +217,18 @@ benchmark: venv
 
 # --- Release goals ---
 
-release-checks: venv
-	$(VENV_ACTIVATE); ./release-checks.sh $(release_version) $(next_version)
+# usage: e.g. make release-checks RELEASE_VERSION=X.Y.Z
+release-checks test-release: venv
+	@if [ -z "$(RELEASE_VERSION)" ]; then echo "error: set RELEASE_VERSION (e.g. make release-checks RELEASE_VERSION=X.Y.Z)" >&2; exit 1; fi
+	$(MAKE) clean lint test-all docs
```
gbanasiak, line 223: I found `make release-checks` a bit lenghty. Can we use `test` instead of `test-all` to reduce the time it takes to run the check?

```suggestion
	$(MAKE) clean lint test docs
```

scripts/release/changelog.py
```diff
@@ -81,26 +109,39 @@ def prs(gh, milestone, with_labels, without_labels=None):
...
 
-    milestone = find_milestone(rally_repo, title=milestone_name)
-    if not milestone:
-        print("No open milestone named [%s] found." % milestone_name, file=sys.stderr)
-        exit(2)
+    milestone = ensure_open_milestone(rally_repo, milestone_name, dry_run=dry_run)
     if milestone.open_issues > 0:
         print("There are [%d] open issues on milestone [%s]. Aborting..." % (milestone.open_issues, milestone_name), file=sys.stderr)
-        exit(2)
+        sys.exit(2)
```
gbanasiak, line 141: WDYT about adding a check if milestone has any PRs / issues attached?

gbanasiak (commented):
Thank you, this works fine to me. Please find minor comments below.

scripts/release/prepare.sh
pquentin: Yes, with #2080 this won't be needed and may even fail.

scripts/release/prepare.sh
pquentin: So the uv docs state this:

> Note that if you use uv run in a project, i.e., a directory with a pyproject.toml, it will install the current project before running the script. If your script does not depend on the project, use the --no-project flag to skip this:
> 
> ```
> # Note: the `--no-project` flag must be provided _before_ the script name.
> uv run --no-project example.py
> ```

That said, the install won't be editable, so we'll need to reinstall to get the new version. It does not need to be editable, though. And it would be cleaner to use `uv pip`.

PR #2076, opened 2026-03-27
Publish to TestPyPI on every commit

pquentin:
And also upgrade the GitHub Actions versions.

This should allow us to see issues sooner, not at every release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,8 @@ name: Release to PyPI
 
 on:
   push:
+    branches:
+      - "main"
     tags:
       - "*"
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -2,6 +2,8 @@ name: Release to PyPI
 
 on:
   push:
+    branches:
+      - "main"
```
fressi-elastic: May be the name here is `master`. Isn't it?

fressi-elastic (changes requested):
Please verify branch name is the correct one: `main` or `master`?

.github/workflows/release.yml
fressi-elastic: Corrected.

fressi-elastic (approved):
LGTM

PR #2082, opened 2026-03-30
Stabilize integration tests and fail fast on persistent REST protocol errors

fressi-elastic:
- **REST:** Cap consecutive protocol-like `ConnectionError`s in `wait_for_rest_layer` so bad HTTP/HTTPS config fails faster; tests updated.
- **IT benchmark port (19200):** Session/module/function fixtures clear Rally-owned ES on the fixed port before/after tests; docstrings explain why the port stays fixed and teardown exists.
- **IT metrics store:** Probe with retries; reuse cluster on 10200 when name matches; fail fast on wrong cluster / non-ES listener; `wait_for_cluster_health` aligns client `request_timeout` with ES wait and treats `timed_out` as failure.
- **IT git:** Session autouse sets `GIT_CONFIG_GLOBAL` to `os.devnull` via `pytest.MonkeyPatch` (avoids `ScopeMismatch` with function-scoped `monkeypatch`).
- **Python / tar:** Use `TarFile.extractall(filter=...)` only on 3.12+; document legacy-path risk on older Pythons.


it/resources/gitconfig-empty
```diff
@@ -0,0 +1,2 @@
+# Intentionally empty: integration tests set GIT_CONFIG_GLOBAL to this file
```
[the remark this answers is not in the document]
fressi-elastic: It is used by conftest.py

esrally/utils/io.py
```diff
@@ -397,7 +398,10 @@ def _do_decompress_manually_with_lib(target_directory: str, filename: str, compr
 
 def _do_tar_decompress(target_directory: str, compressed_file: tarfile.TarFile) -> None:
     try:
-        compressed_file.extractall(path=target_directory, filter="tar")
+        if sys.version_info >= (3, 12):
+            compressed_file.extractall(path=target_directory, filter="tar")
+        else:
+            compressed_file.extractall(path=target_directory)
```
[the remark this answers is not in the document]
fressi-elastic: Acknowledged. On Python <3.12, `TarFile.extractall` has no PEP 706 `filter` parameter, so we cannot apply the same member-path safety profile as on 3.12+. We documented this limitation in `_do_tar_decompress` (docstring plus an inline comment on the legacy branch) so the path-traversal risk is explicit for maintainers. Extraction behavior is unchanged here; a secure-extract helper for older interpreters would be separate work.

esrally/client/factory.py
```diff
@@ -313,32 +348,49 @@ def wait_for_rest_layer(es, max_attempts=40):
...
             raise exceptions.SystemSetupError("Could not connect to cluster via HTTPS. Are you sure this is an HTTPS endpoint?", e)
         except ConnectionError as e:
-            if "ProtocolError" in str(e):
+            is_protocol = _connection_error_is_protocol_like(e)
+            if is_protocol:
+                consecutive_protocol_connection_errors += 1
+                if consecutive_protocol_connection_errors > _MAX_CONSECUTIVE_PROTOCOL_CONNECTION_ERRORS:
+                    raise exceptions.SystemSetupError(
+                        "Received a protocol error. Are you sure you're using the correct scheme (HTTP or HTTPS)?", e
+                    )
```
pquentin, line 365: The two previous checks should cover incorrect HTTP/HTTPS configurations. Are they not useful?

it/conftest.py
```diff
@@ -97,7 +132,29 @@ def stop(self):
...
+
+    Restores the previous ``GIT_CONFIG_GLOBAL`` value (or unsets it) after the session.
+    """
+    previous = os.environ.get("GIT_CONFIG_GLOBAL")
+    os.environ["GIT_CONFIG_GLOBAL"] = _GIT_CONFIG_EMPTY
+    yield
+    if previous is None:
+        os.environ.pop("GIT_CONFIG_GLOBAL", None)
+    else:
+        os.environ["GIT_CONFIG_GLOBAL"] = previous
```
pquentin: Why don't we use `monkeypatch.setenv()` instead? And why do we need a real file? We can use `/dev/null` instead.

it/__init__.py
```diff
@@ -38,6 +41,20 @@
...
+_LOG = logging.getLogger(__name__)
+_CLUSTER_PROBE_ATTEMPTS = 5
+_CLUSTER_PROBE_DELAY_SEC = 0.5
+_CLUSTER_PROBE_REQUEST_TIMEOUT_SEC = 5.0
+
+# Host HTTP port integration tests use for Rally-provisioned benchmark Elasticsearch (see it/* tests, docker pipeline).
+BENCHMARK_IT_HTTP_PORT = 19200
+# Default ``--cluster-name`` for ``esrally race`` / install; keep aligned with ``esrally/rally.py``.
+RALLY_DEFAULT_BENCHMARK_CLUSTER_NAME = "rally-benchmark"
+# Only these cluster names may be torn down on :data:`BENCHMARK_IT_HTTP_PORT` (metrics store uses 10200).
```
pquentin: Why `:data:`?

it/__init__.py
```diff
@@ -168,6 +359,41 @@ def start(self, race_id):
         client.wait_for_rest_layer(es)
         assert es.info()["cluster_name"] == self.cfg
 
+    def wait_for_cluster_health(self, *, wait_for_status: str = "yellow", timeout: float = 120) -> None:
```
pquentin, line 379: We already have this in the Rally runtime, why duplicate it here?

it/__init__.py
```diff
@@ -134,12 +151,186 @@ def wait_until_port_is_free(port_number=39200, timeout=120):
...
+            _terminate_listeners_on_port(http_port)
+    return
+
+
+def ensure_benchmark_http_port_free(http_port: int = BENCHMARK_IT_HTTP_PORT, wait_timeout: int = 120) -> int:
+    """Stop any Rally IT benchmark cluster on ``http_port`` and wait until the port is free."""
+    assert http_port is not None
+    stop_rally_provisioned_es_on_port(http_port)
+    wait_until_port_is_free(port_number=http_port, timeout=wait_timeout)
+    return http_port
```
pquentin, line 319: This pulls a lot of code.

* Is having Elasticsearch already running a cause of issues in CI, actually?
* Why can't we use a free port?

pquentin (commented):
Thanks for the improvements.

it/conftest.py
fressi-elastic: True

it/__init__.py
fressi-elastic: The port is hardcoded in many places and test case has not control on this port. Test cases are failing some time because rally or docker can't open this port as the tear down procedure is not always working fine. For example Rally is closed after a signal, it didn't kill its subprocesses like ES JVM. This is a cleanup mechanism in place to workaround rally issue we know exist but it is out of scope now to fix it.
There is also the problem similar with docker some time taking more tile releasing this port after tear down.

esrally/client/factory.py
fressi-elastic: Connection errors can be raised by different layers of the protocol and not all of them must be threated in the same way because some are transient but some are not. So timeout has to be smarter on this case: consecutive errors of the same time could mean there is a persistent problem and it doesn't worth supposing it is transient. I got many problems of this kind while running tests and I finally got this compromise.

it/__init__.py
fressi-elastic: I create this because I saw random failures running integration test cases which where hard to investigate. Having the test case cleaned it up before makes failures in setting up the metrics store (witch could be slow) much easier to debug and with a timeout policy independent from rally's one. We are not going to relay on rally to know if the fixture did his job or not. Better doing it earlier. This has been failing very often. The reason is probably because tearing up and down the server is not always a reliable operation and has delays (which rally doesn't thread well, and hide the cause of the problem).

fressi-elastic: wait_for_rest_layer in Rally waits until enough nodes answer and applies Rally’s own retry/backoff for connection/protocol errors; it uses cluster.health(wait_for_nodes=...), not a status wait. The integration harness also reuses a metrics-store cluster that may already be running across the session, so after start() we still need an explicit wait_for_status (e.g. yellow) before we treat the metrics store as ready.

I kept this in it/ with a single, explicit pairing of Elasticsearch’s server-side timeout and the HTTP client’s request_timeout so fixture failures surface as clear health/timeout assertions instead of depending on Rally’s broader wait semantics

it/__init__.py
fressi-elastic: The benchmark IT port (19200) is fixed in practice: install/race invocations, existing tests, and CI/docker wiring assume it. Moving to an ephemeral port would mean threading a chosen port through every code path and artifact that references it, which is a larger refactor than this PR.

The extra teardown exists because incomplete teardown does bite us in automation: when the runner gets a signal or a test aborts, Rally does not always reap child ES JVMs, and Docker can hold the published port briefly after stop. That shows up as flaky “address already in use” or confusing bind failures on the same port the next test expects. ensure_benchmark_http_port_free is a targeted workaround for that class of flake without redesigning Rally’s process lifecycle here.

fressi-elastic:
Example of issue this PR should be solving: https://buildkite.com/elastic/rally-it/builds/3054#019d4f98-bc74-439f-9c4c-8777e109b385/L2153

pquentin (approved):
I'm not a big fan because this adds duct tape instead of fixing the real issues. For example, I'd like to understand why we sometimes can't stop Elasticsearch and get it to release port 9200. That said, the situation with the CI is too painful to ignore.

Thanks! LGTM.

PR #2097, opened 2026-04-09
Implement track parameter filtering for metrics reporting

gareth-ellis:
- Introduced a new function `track_params_for_reporting` to omit track parameters starting with `secret_` from metrics documents and race records while retaining them in configuration.
- Updated the `MetricsStore` and `Race` classes to utilize this new filtering function.
- Added unit tests to ensure that secret-prefixed parameters are correctly omitted during reporting.

### DIFF
--- a/docs/command_line_reference.rst
+++ b/docs/command_line_reference.rst
@@ -280,7 +280,7 @@ Example JSON file::
       "clients": 16
    }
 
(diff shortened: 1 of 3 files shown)


pquentin (commented):
The idea makes sense! But I think the key should be kept, and you would only hide the value. This is how logging works in the Elasticsearch Python client, for example.

https://github.com/elastic/elastic-transport-python/blob/411b4dfc74866bc84c63a450f4778015af8c03b8/tests/test_models.py#L79-L90

gareth-ellis:
good idea - i'll make that change

pquentin (approved):
Thanks! LGTM.

PR #2099, opened 2026-04-10
Use esrally/race check= consistently in integration tests

fressi-elastic:
This updates integration tests to rely on `esrally` / `race` with `check=True` by default instead of `assert ... == 0`, and uses `check=False` with `CompletedProcess.returncode` where the test needs a non-zero exit or an explicit success check.

Also fixes the `pytest.fail` message in `it.esrally` (f-string), handles `TestCluster.stop` via `returncode`, imports `CompletedProcess`, and expands docstrings for `esrally` and `race` without Sphinx roles in Python strings.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/it/__init__.py
+++ b/it/__init__.py
@@ -21,6 +21,7 @@ import json
 import logging
 import os
 import random
+import shlex
 import socket
 import subprocess
 import time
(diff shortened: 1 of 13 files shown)


fressi-elastic (commented):
**Resolving Copilot review threads (explanation)**

1. **`esrally_command_line_for`** — The helper was removed; the shell command is built only inside \`it.esrally\`. All former call sites use \`it.esrally\` / \`it.race\` or an explicit full command where subprocess needs a raw string (e.g. interrupt test). There are no remaining references to \`esrally_command_line_for\` on this branch.

2. **`shlex.join(err.args)`** — The failure path now uses \`err.cmd\` in the \`pytest.fail\` message instead of \`shlex.join(err.args)\`.

3. **`json.loads` / \`"".join(result.stdout)\`** — \`TestCluster.install\` now uses \`json.loads(result.stdout)\`.

4. **`tracker_test` \`check=False\`** — Simplified to \`it.race(cfg, cmd)\` with default \`check=True\` for consistent failure reporting.

5. **`esrallyd_test`** — Same: plain \`it.race(...)\` with default \`check=True\`, no redundant manual \`returncode\` check.

PR #2101, opened 2026-04-12
Differentiate processes in the enrich-policy operation

slDias:
As suggested in [MR 2045](https://github.com/elastic/rally/pull/2045), this commit adds a new operator that combines the delete, creation and execution of enrich policies within a single operation.

### DIFF
--- a/docs/track.rst
+++ b/docs/track.rst
@@ -3228,6 +3228,62 @@ The following metrics are included when profile data is present:
 * ``<driver>.<operator>.processed_slices``: Cumulative processed slices for an operator across all driver instances.
 * ``<plan>.<optimization>.took_ms``: Time spent on a plan optimization step (``logical_optimization``, ``physical_optimization``, or ``reduction``), in milliseconds.
 
+
+create-enrich-policy
+~~~~~~~~~~~~~~~~~~~~
+
+The operation ``create-enrich-policy`` will delete (if it exists), create and execute a list of enrich policies.
+It's properties follows the request body of the `Create an enrich policy API <https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-enrich-put-policy>`_
+
+* ``policies`` (mandatory): An object containing the name of the policy to be created as its keys and the definition of the policy as its value following the request body of "Create an enrich policy".
+* ``delete`` (optional): A boolean specifying if the operation should try to delete an existing policy with the same name before creating a new one. Defaults to true.
+
+**Examples**
+
+The following snippet will create all enrich policies that have been defined in the ``policies`` section.
+It will attempt to delete them beforehand if it exists::
+
+
+    {
+      "name": "create-enrich-policy",
+      "operation": {
+        "operation-type": "create-enrich-policy",
+        "policies": {
+          "nyc_rate_codes": {
+            "match": {
+              "indices": "nyc_rate_codes",
+              "match_field": "id",
+              "enrich_fields": [
+                "name"
+              ]
+            }
+          },
+          "nyc_payment_types": {
+            "match": {
+              "indices": "nyc_payment_types",
+              "match_field": "type",
+              "enrich_fields": [
+                "name"
+              ]
+            }
+          }
+        }
+      },
+      "delete": true
+    }
+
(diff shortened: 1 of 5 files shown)


gbanasiak:
Buildkite test this please

docs/track.rst
```diff
@@ -3266,25 +3266,34 @@ The following metrics are included when profile data is present:
 enrich-policy
 ~~~~~~~~~~~~~
 
-The operation ``enrich-policy`` will delete (if they exist), create and execute a list of enrich policies.
+The operation ``enrich-policy`` can delete, create or execute enrich policies.
+
+All operations can occurr on different policies.
+If the same policy is specified within multiple operations, it will follow the execution order: ``delete -> create -> execute``.
+
+A refresh of all indices is performed after deletion and creation of policies and before execution.
```
gbanasiak, line 3274: What is the justification for the refresh? Is a refresh required in all enrich policy scenarios? Rally has a separate refresh operation, which could be used explicitly, if needed.

gbanasiak (commented):
Thanks for the improvement, and sorry for the delay (tagging `@elastic/es-perf` was the right thing to do). The change looks fine to me, I just left one question regarding the refresh.

I would also like to ask @craigtaverner to take a look. My understanding is the PR implements [this](https://github.com/elastic/rally/pull/2045#discussion_r3069400978) idea.

docs/track.rst
slDias: Thank you for your review,

I believe that I did so based on the [nyc_taxis track](https://github.com/elastic/rally-tracks/blob/e2958c2fbb25f8e40c41e52ac233f50e3dc82ea9/nyc_taxis/challenges/default.json#L893) and assumed it was required. It also seemed to make sense as a way to ensure that policies are ready to execute.
But I don't know for sure, as this was my first time working on data enrichment.

Happy to remove if it's not required.

gbanasiak: I think we can skip the refresh indeed.

craigtaverner: I assume since this new operation performs the complete sequence delete->create->execute, and the refresh should occur before the execute, that was the reason for this? However, if that refresh is only for original source indices, then I understand that it does not need to be here, and could simply be moved to before this operation (and therefor not be part of this operation at all).

docs/track.rst
```diff
@@ -3266,25 +3266,34 @@ The following metrics are included when profile data is present:
...
+
+Lastly, it will execute the ``nyc_vendors`` policy but will **not** delete it or create it beforehand.::
 
     {
       "name": "enrich-policy",
       "operation": {
         "operation-type": "enrich-policy",
-        "delete": true,
-        "policies": {
+        "delete": ["nyc_rate_codes", "nyc_trip_types"],
```
craigtaverner, line 3295: While I like the added support for a list of names, I do think that removing the "policies" key does present a backwards compatibility issue. I think my original suggestion in the previous PR would provide backwards compatibility, and not require all policies to be repeated in multiple lists.

craigtaverner (changes requested):
I preferred my original suggestion, keeping the "policies" key, but allowing it to be either an object with policy name keys (BWC with existing operation), or a list of either strings or objects. For delete and execute we only need strings. For create we need the object format.

So, for example, just executing policies:

```
{
  "name": "re-execute-enrich-policies",
  "operation": {
    "operation-type": "enrich-policy",
    "policies": [
      "nyc_rate_codes",
      "nyc_payment_types"
    ],
    "delete": false,
    "create": false,
    "execute": true
  }
}
```

Or, to re-create and re-execute:

```
{
  "name": "re-create-enrich-policies",
  "operation": {
    "operation-type": "enrich-policy",
    "policies": {
      "nyc_rate_codes": {...},
      "nyc_payment_types: {...}"
    },
    "delete": true,
    "create": true,
    "execute": true
  }
}
```

I feel like the use case you new design opens is when we want to delete, create or execute a different set of policies. That is an unlikely use case, and is already fully satisfied by simply using multiple operations. It also opens the risk of people typing the wrong names into the different lists by mistake.

And finally, the main reason for rejecting this approach is that it is not backwards compatible with the existing format, meaning that if someone upgrades rally, they also need to upgrade all their challenges in their tracks to the new format.

PR #2108, opened 2026-04-23
Ensure the correct op_type is used when indexing telemetry

inqueue:
From Claude:

Fixes a thread-safety race in `EsMetricsStore.flush()` that occurs when `bulk_index` first loops over `items` setting `_op_type = "create"` on each item, then calls `helpers.bulk` on the same list. Background sampler threads (`RecoveryStatsRecorder`, `BlobStoreStatsRecorder`) can append new docs to that same list between those two steps. Those late-arriving docs never get `_op_type = "create"` set and `expand_action` defaults them to `"index"`.

The fix is to snapshot and reset `self._docs` before calling `bulk_index`, so background threads' new docs land in a fresh list that `bulk_index` never sees.

### DIFF
--- a/benchmarks/metrics/metrics_store_test.py
+++ b/benchmarks/metrics/metrics_store_test.py
@@ -21,18 +21,20 @@ concurrent access between flush() and background sampler threads (_add()).
 
 Three scenarios are measured:
 
-  uncontended  – single thread calling _add() in a tight loop.  Represents
+  uncontended   single thread calling _add() in a tight loop.  Represents
                  worker processes (InMemoryMetricsStore) and the coordinator
                  between flushes.
 
-  contended    – N sampler threads continuously calling _add() while the
+  contended     N sampler threads continuously calling _add() while the
                  main thread calls _add() too.  Represents the coordinator
                  with active telemetry devices.
 
-  no_lock      – same single-threaded loop but with the lock replaced by a
+  no_lock       same single-threaded loop but with the lock replaced by a
                  no-op context manager.  Provides the baseline cost without
                  locking so the overhead can be quantified.
 """
+
+# pylint: disable=protected-access
 
 import datetime
 import threading
@@ -40,7 +42,6 @@ import threading
 import pytest
 
 from esrally import config, metrics, time
-
 
 # ---------------------------------------------------------------------------
 # Shared fixtures
--- a/benchmarks/metrics/metrics_store_test.py
+++ b/benchmarks/metrics/metrics_store_test.py
@@ -144,7 +144,7 @@ def test_add_no_lock_with_sampler_threads(benchmark, in_memory_store):
     def sampler():
         while not stop.wait(timeout=0.1):
             for _ in range(SAMPLER_DOCS_PER_CALL):
-                in_memory_store._add(DOC)
+                in_memory_store._add_unlocked(DOC)
 
     threads = [threading.Thread(target=sampler, daemon=True) for _ in range(SAMPLER_THREADS)]
     for t in threads:
--- a/benchmarks/metrics/metrics_store_test.py
+++ b/benchmarks/metrics/metrics_store_test.py
@@ -21,17 +21,16 @@ concurrent access between flush() and background sampler threads (_add()).
 
 Three scenarios are measured:
 
-  uncontended   single thread calling _add() in a tight loop.  Represents
-                 worker processes (InMemoryMetricsStore) and the coordinator
-                 between flushes.
+  uncontended   single thread calling _add() in a tight loop. Represents
+                worker processes (InMemoryMetricsStore) and the coordinator
+                between flushes.
 
-  contended     N sampler threads continuously calling _add() while the
-                 main thread calls _add() too.  Represents the coordinator
-                 with active telemetry devices.
+  contended     N sampler threads periodically calling _add() while the
+                main thread calls _add() too. Represents the coordinator
+                with active telemetry devices.
 
-  no_lock       same single-threaded loop but with the lock replaced by a
-                 no-op context manager.  Provides the baseline cost without
-                 locking so the overhead can be quantified.
+  no_lock       uses _add_unlocked() without lock. Provides the baseline
+                cost without locking so the overhead can be quantified.
 """
 
 # pylint: disable=protected-access
@@ -50,16 +49,7 @@ from esrally import config, metrics, time
 DOC = {"name": "indexing_throughput", "value": 5000, "unit": "docs/s"}
 SAMPLER_THREADS = 4  # used by the single contended test; sweep uses its own range
 DOCS_PER_CALL = 1000
-
-
-class _NoOpLock:
-    """Drop-in replacement for threading.Lock that acquires/releases nothing."""
-
-    def __enter__(self):
-        return self
-
-    def __exit__(self, *args):
-        pass
+SAMPLER_DOCS_PER_CALL = 10
 
 
 @pytest.fixture
@@ -95,18 +85,17 @@ def test_add_uncontended(benchmark, in_memory_store):
 
 
 # ---------------------------------------------------------------------------
-# Uncontended baseline: single thread, no-op lock
+# No lock: single thread
 # ---------------------------------------------------------------------------
 
 
 @pytest.mark.benchmark(group="metrics_store_add", warmup="on", warmup_iterations=1000, disable_gc=True)
-def test_add_no_lock_baseline(benchmark, in_memory_store):
-    """Baseline: same loop with the lock replaced by a no-op to isolate lock cost."""
-    in_memory_store._docs_lock = _NoOpLock()
+def test_add_no_lock(benchmark, in_memory_store):
+    """Baseline: same loop without lock."""
 
     def run():
         for _ in range(DOCS_PER_CALL):
-            in_memory_store._add(DOC)
+            in_memory_store._add_unlocked(DOC)
 
     benchmark(run)
 
@@ -116,14 +105,15 @@ def test_add_no_lock_baseline(benchmark, in_memory_store):
 # ---------------------------------------------------------------------------
 
 
-@pytest.mark.benchmark(group="metrics_store_add", warmup="on", warmup_iterations=100, disable_gc=True)
+@pytest.mark.benchmark(group="metrics_store_add", warmup="on", warmup_iterations=100, disable_gc=True, max_time=10)
 def test_add_contended(benchmark, in_memory_store):
     """Cost of _add() while SAMPLER_THREADS background threads also call _add()."""
     stop = threading.Event()
 
     def sampler():
-        while not stop.is_set():
-            in_memory_store._add(DOC)
+        while not stop.wait(timeout=0.1):
+            for _ in range(SAMPLER_DOCS_PER_CALL):
+                in_memory_store._add(DOC)
 
     threads = [threading.Thread(target=sampler, daemon=True) for _ in range(SAMPLER_THREADS)]
     for t in threads:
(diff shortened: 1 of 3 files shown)


esrally/metrics.py
```diff
@@ -1204,25 +1204,26 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
-        if self._docs:
+        docs_to_flush = self._docs
+        self._docs = []
+        if docs_to_flush:
             sw = time.StopWatch()
             sw.start()
             self._client.bulk_index(
                 index=self._index_handler.index_name(self._race_timestamp),
-                items=self._docs,
+                items=docs_to_flush,
```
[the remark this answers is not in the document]
inqueue: @copilot apply changes based on [this feedback](https://github.com/elastic/rally/pull/2108#discussion_r3132830251)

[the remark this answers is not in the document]
inqueue: test added

esrally/metrics.py
```diff
@@ -1204,25 +1204,32 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
             self._client.refresh(index=index_name)
 
     def flush(self, refresh=True):
-        if self._docs:
+        docs_to_flush = self._docs
+        self._docs = []
+        if docs_to_flush:
```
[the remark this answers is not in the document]
inqueue: @copilot it would seem there a very small window for the thread to grab a stale reference. It looks like the issue existed prior to https://github.com/elastic/rally/commit/a0d7e015c51efbd25164cc699d9d033b627d0bf8, and the tradeoff to put a lock on `_add()` is probably not worth it.

esrally/metrics.py
```diff
@@ -1204,25 +1204,32 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
+                self._client.bulk_index(
+                    index=self._index_handler.index_name(self._race_timestamp),
+                    items=docs_to_flush,
+                    use_data_streams=self._index_handler.use_data_streams,
+                )
+            except Exception:
+                # Re-queue at the front so the next flush retries these docs in order,
+                # ahead of any docs added to self._docs during the failed attempt.
+                self._docs = docs_to_flush + self._docs
+                raise
```
gbanasiak: This exception handling adds another repetition layer on top of `guarded()` which is used in `self._client.bulk_index()`. I would assume that if `guarded()` raises then the problem is not transient, and re-queueing of metrics documents is not helpful anyway.

gbanasiak: Actually it's more nuanced, as there are cases where `guarded()` does not repeat, but perhaps those should be handled in `guarded()`. My main point is readability and cognitive load.

esrally/metrics.py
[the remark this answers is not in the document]
gbanasiak: The issue existed before but now the lack of correct `_op_type` is fatal whereas earlier we could only have had missing metrics documents or duplicates. I'm leaning towards adding a lock now that we fully (?) understand the problem. The lock should cover `_add()`, `flush()` and `to_externalizable()` in `InMemoryMetricsStore`.

tests/metrics_test.py
```diff
@@ -1754,6 +1754,63 @@ def _get_error_rate(self, buckets):
         )
         return actual_error_rate
 
+    def test_flush_snapshots_docs_before_bulk_index(self):
+        # Docs appended to _docs between the for-loop in bulk_index (which sets _op_type)
+        # and helpers.bulk's iteration must not slip through without _op_type="create".
+        # The fix captures self._docs and resets it before calling bulk_index so that
+        # concurrent _add calls land in the new buffer and never reach this flush.
```
gbanasiak: The mention of "the fix" is clear now, but will not be later. I would reword this to explain this in more generic terms, not mentioning "the fix".

esrally/metrics.py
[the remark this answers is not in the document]
inqueue: > I'm leaning towards adding a lock now that we fully (?) understand the problem. The lock should cover `_add()`, `flush()` and `to_externalizable()` in `InMemoryMetricsStore`.

I expect it will come with a potential performance hit. I will benchmark it and report back.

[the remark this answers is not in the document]
pquentin: I don't expect high contention, as flushing happens every 30 seconds. That said, the bug already existed before this PR, so we could treat it as a follow-up to revert independently if needed.

(We would have to apply the lock before merging this PR, since it's now required as mentioned by @gbanasiak above.)

esrally/metrics.py
inqueue: Addressed by https://github.com/elastic/rally/pull/2108/changes/fdf2d74c8f0dc41ed2ea0dd92fe799f2e13f511d.

tests/metrics_test.py
inqueue: Addressed by https://github.com/elastic/rally/pull/2108/changes/daad44c2485a0c7a967d1f719579cb72e36b7dea.

esrally/metrics.py
```diff
@@ -1204,31 +1207,34 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
                 index=self._index_handler.index_name(self._race_timestamp),
-                items=self._docs,
+                items=docs_to_flush,
                 use_data_streams=self._index_handler.use_data_streams,
             )
             sw.stop()
             self.logger.info(
                 "Successfully added %d metrics documents for race timestamp=[%s], track=[%s], challenge=[%s], car=[%s] in [%f] seconds.",
-                len(self._docs),
+                len(docs_to_flush),
```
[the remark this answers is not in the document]
inqueue: 1. `guarded()` is already the retry boundary.

1. Re-queuing doesn't help after persistent failure.

1. A reader of `flush()` would need to understand exactly which exception paths inside `guarded()` retry and which raise immediately in order to reason about when the re-queue block actually fires. That logic already lives in `guarded()` — if non-retryable paths ever need re-queue semantics, that's where they belong, not scattered into every call site.

tests/metrics_test.py
```diff
@@ -566,6 +566,55 @@ def raise_bulk_index_error():
...
+        call_count = 0
+
+        def raise_then_succeed():
+            nonlocal call_count
+            call_count += 1
+            if call_count == 1:
+                raise elasticsearch.helpers.BulkIndexError("1 document(s) failed to index", bulk_index_errors)
+
+        client = metrics.EsClient(self.ClientMock([{"host": "127.0.0.1", "port": "9243"}]))
+        client.guarded(raise_then_succeed)
```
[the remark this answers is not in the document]
inqueue: Addressed https://github.com/elastic/rally/pull/2108/changes/4f823e3119584ef8e3b3bf120074b9484c85b30d

esrally/metrics.py
```diff
@@ -1179,6 +1181,7 @@ def __init__(
         self._client = client_factory_class(cfg).create()
         self._index_handler = IndexHandler(self._config, self._client, EsStoreType.metrics)
         self._docs = None
+        self._docs_lock = threading.Lock()
```
gbanasiak: nit: this could be moved to parent class because both children classes use it

gbanasiak (approved):
LGTM. I haven't done any stress testing to assess the slowdown (if any).

esrally/metrics.py
inqueue: Nice catch. It is addressed in https://github.com/elastic/rally/pull/2108/commits/10d94747f9585ed33a40f65a7f33cdbc0b4463d7.

inqueue:
### Benchmarks

```shell
--------------------------- benchmark 'metrics_store_add': 3 tests ---------------------------
Name (time in us)                 Mean                StdDev            OPS (Kops/s)          
----------------------------------------------------------------------------------------------
test_add_contended            337.7060 (4.80)     3,061.1162 (464.92)         2.9612 (0.21)   
test_add_uncontended           70.3748 (1.0)          6.5842 (1.0)           14.2096 (1.0)    
test_add_no_lock_baseline      88.8773 (1.26)        20.8954 (3.17)          11.2515 (0.79)   
----------------------------------------------------------------------------------------------

-------------------------- benchmark 'metrics_store_add_sweep': 4 tests --------------------------
Name (time in us)                     Mean                StdDev            OPS (Kops/s)          
--------------------------------------------------------------------------------------------------
test_add_contended_2_samplers     210.7780 (1.31)     1,575.1583 (2.23)           4.7443 (0.77)   
test_add_contended_4_samplers     359.3272 (2.23)     3,332.4387 (4.72)           2.7830 (0.45)   
test_add_contended_8_samplers     603.1385 (3.74)     6,704.0001 (9.49)           1.6580 (0.27)   
test_add_contended_1_sampler      161.2521 (1.0)        706.5420 (1.0)            6.2015 (1.0)    
--------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

Telemetry devices spawn one `SamplerThread`. While the thread count is not configurable, the sample rate is. The sample rate controls how frequently each thread calls `_add()`. With this change, the user should be mindful of which telemetry devices are used and their sample rates.

benchmarks/metrics/metrics_store_test.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+class _NoOpLock:
+    """Drop-in replacement for threading.Lock that acquires/releases nothing."""
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        pass
```
gbanasiak: Great idea with pytest benchmarks.

`_NoOpLock` has too much overhead I think. It's pure Python vs. C (threading.Lock). This results in `test_add_no_lock_baseline()` slower than `test_add_uncontended()` which is not the expected result. It's better to compare code with lock with code without any locks:

```
    def _add(self, doc):
        with self._docs_lock:
            self.docs.append(doc)

    # for testing purposes only
    def _add_unlocked(self, doc):
        self.docs.append(doc)
```

With this, and some renaming of the first 3 benchmarks, and adding sampler threads benchmark _without_ lock I'm getting the following results:

```
------------------------------- benchmark 'metrics_store_add': 4 tests ------------------------------
Name (time in us)                        Mean                StdDev            OPS (Kops/s)          
-----------------------------------------------------------------------------------------------------
test_add_single_thread_no_lock        21.2965 (1.0)          3.1434 (1.0)           46.9561 (1.0)    
test_add_single_thread_lock           79.4961 (3.73)         7.5931 (2.42)          12.5792 (0.27)   
test_add_sampler_threads_no_lock     109.1919 (5.13)     1,686.1866 (536.42)         9.1582 (0.20)   
test_add_sampler_threads_lock        414.9385 (19.48)    3,379.8023 (>1000.0)        2.4100 (0.05)   
-----------------------------------------------------------------------------------------------------
```

Roughly, by adding a lock, we slow down document addition by the factor of 4 in those tests. That's quite consistent regardless of the presence of sampler threads.

I'm wondering about big increase from 20 microseconds without a lock and without sample threads to 400 microseconds under lock and with background sampler threads. Are the tests representative of what is actually happening within driver process though? I think in reality the proportions between telemetry sampler threads and driver sample processing thread are way different. Samplers add documents once a second the most, and in each round I would expect a handful of documents. Let's assume we add 10 documents per 100 ms in every telemetry sampler, whereas main sample processing thread continues adding 1000 documents.

```
SAMPLER_DOCS_PER_CALL = 10
[..]
@pytest.mark.benchmark(group="metrics_store_add", warmup="on", warmup_iterations=100, disable_gc=True, max_time=10)
def test_add_sampler_threads_lock(benchmark, in_memory_store):
    """Cost of _add() while SAMPLER_THREADS background threads also call _add()."""
    stop = threading.Event()

    def sampler():
        while not stop.wait(timeout=0.1):
            for _ in range(SAMPLER_DOCS_PER_CALL):
                in_memory_store._add(DOC)
[..]
```

Then I'm getting:

```
---------------------------- benchmark 'metrics_store_add': 4 tests ----------------------------
Name (time in us)                       Mean            StdDev            OPS (Kops/s)          
------------------------------------------------------------------------------------------------
test_add_sampler_threads_no_lock     21.9800 (1.0)      8.5470 (1.80)          45.4959 (1.0)    
test_add_single_thread_no_lock       22.3034 (1.01)     7.2304 (1.53)          44.8362 (0.99)   
test_add_sampler_threads_lock        79.5050 (3.62)     4.7388 (1.0)           12.5778 (0.28)   
test_add_single_thread_lock          79.9506 (3.64)     6.6431 (1.40)          12.5077 (0.27)   
------------------------------------------------------------------------------------------------
```

My read of it is sampler thread events are so rare (in relative terms) that what really matters is just the presence of lock in main sample processing thread. This x4 slowdown will affect tasks with (too) many metrics, lowering the point of failure we discussed in https://github.com/elastic/rally/pull/2029.

gbanasiak: Changes included in https://github.com/elastic/rally/pull/2108/commits/0d736a9d1485f046c4f7ad10300ce6ebd3a5b3f1 and https://github.com/elastic/rally/pull/2108/commits/017a5bfd65eb0177dbe3b10e6f9af87eef64e637.

PR #2111, opened 2026-04-29
Ignore ILM settings for Serverless metrics stores

inqueue:
This should be the final PR needed to make Rally compatible with a Serverless metrics store for reporting.

- Check if the target metrics store is serverless
- Do not include ILM in the configuration if it is serverless


NickDris (approved):
LGTM

PR #2112, opened 2026-05-04
docker: add timeout to wget call

pquentin:
They are currently failing frequently due to timeouts. One example: https://buildkite.com/elastic/rally-dev-docker-build/builds/1129#_

### DIFF
--- a/.buildkite/dev-docker/pipeline.yml
+++ b/.buildkite/dev-docker/pipeline.yml
@@ -40,14 +40,14 @@ steps:
     retry:
       automatic:
         - limit: 1
-    timeout_in_minutes: 20
+    timeout_in_minutes: 40
   - label: ":docker: Build Docker Artifacts for Rally arm64"
     command: bash .buildkite/dev-docker/run.sh build arm64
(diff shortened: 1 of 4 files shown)


gareth-ellis (commented):
It seems from the output we seem to be hanging around https://github.com/elastic/rally/blob/master/docker/Dockerfiles/release/Dockerfile#L15-L24 - i'm not convinced extending the job timeout will do more than make us timeout 20 minutes later.

Could we try and add `-T 20` to the wget command? I have a feeling that may be where we are hanging - perhaps we are being blocked from downloading from launchpad? If so, then we could consider hosting this outselves somewhere

pquentin:
Apparently, Launchpad was improved last week, and Ubuntu is generally suffering from a DDoS attack. So hosting it ourselves could help, at least temporarily. Added the timeout flag to see if it helps, though.

gareth-ellis (approved):
LGTM

pquentin:
Reverted the Buildkite level timeout changes, let's see if timeouts in wget improve our retry strategy.

PR #2114, opened 2026-05-06
Fix null client timeout handling

gbanasiak:
With `--client-options='{"default": {"timeout": null}}'` we're getting the following exception:

```
esrally.exceptions.RallyError: Traceback (most recent call last):
  File "/Users/grzegorz/src/rally/esrally/actor.py", line 96, in guard
    return f(self, msg, sender)
  File "/Users/grzegorz/src/rally/esrally/driver/driver.py", line 282, in receiveMsg_PrepareBenchmark
    self.driver.prepare_benchmark(msg.track)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/Users/grzegorz/src/rally/esrally/driver/driver.py", line 728, in prepare_benchmark
    es_clients = self.create_es_clients()
  File "/Users/grzegorz/src/rally/esrally/driver/driver.py", line 636, in create_es_clients
    ).create()
      ~~~~~~^^
  File "/Users/grzegorz/src/rally/esrally/client/factory.py", line 195, in create
    return RallySyncElasticsearch(
        distribution_version=self.distribution_version,
    ...<3 lines>...
        **self.client_options,
    )
  File "/Users/grzegorz/src/rally/esrally/client/synchronous.py", line 30, in __init__
    super().__init__(hosts, **kwargs)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
TypeError: Elasticsearch.__init__() got an unexpected keyword argument 'timeout'
```

### DIFF
--- a/tests/client/factory_test.py
+++ b/tests/client/factory_test.py
@@ -345,6 +345,36 @@ class TestEsClientFactory:
 
         assert client_options == original_client_options
 
+    def test_timeout_is_translated_to_request_timeout(self):
+        hosts = [{"host": "localhost", "port": 9200}]
+        client_options = {"timeout": 60}
+        # make a copy so we can verify later that the factory did not modify it
+        original_client_options = dict(client_options)
+
+        f = client.EsClientFactory(hosts, client_options)
+
+        assert f.client_options["request_timeout"] == 60
+        assert "timeout" not in f.client_options
+        assert client_options == original_client_options
+
+    def test_no_request_timeout_set_when_timeout_absent(self):
+        hosts = [{"host": "localhost", "port": 9200}]
+        client_options = {}
+
+        f = client.EsClientFactory(hosts, client_options)
+
+        assert "request_timeout" not in f.client_options
+        assert "timeout" not in f.client_options
+
+    def test_timeout_none_is_translated_to_request_timeout(self):
+        hosts = [{"host": "localhost", "port": 9200}]
+        client_options = {"timeout": None}
+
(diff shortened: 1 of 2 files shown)


tests/client/factory_test.py
```diff
@@ -345,6 +345,36 @@ def test_create_http_connection_with_path(self):
 
         assert client_options == original_client_options
 
+    def test_timeout_is_translated_to_request_timeout(self):
+        hosts = [{"host": "localhost", "port": 9200}]
+        client_options = {"timeout": 60}
+        # make a copy so we can verify later that the factory did not modify it
+        original_client_options = dict(client_options)
```
pquentin: nit: Using `dict.copy()` is clearer: https://docs.python.org/3/library/stdtypes.html#dict.copy (using the copy module could have worked too).

```suggestion
        # used to verify later that the factory did not modify it
        original_client_options = client_options.copy()
```

pquentin (approved):
Thanks! LGTM. Tested successfully.

I'm wondering, should we start advertising `request_timeout` instead of `timeout`?

gbanasiak:
> I'm wondering, should we start advertising `request_timeout` instead of `timeout`?

It makes sense to me. My understanding is `timeout` appeared in client options only because it used to be the name of timeout kwarg in `Elasticsearch` constructor. As this was changed to `request_timeout`, Rally docs should ultimately follow.

PR #2119, opened 2026-05-13
feat(telemetry): add `disk-usage-stats-flush` param

salvatore-campagna:
## Summary

Add a `disk-usage-stats-flush` telemetry parameter (default `true`, no behavior change) that controls the `flush` argument the `disk-usage-stats` device passes to `_disk_usage`. Setting it to `false` lets the telemetry work on serverless, which today rejects the default call.

## Why

`_disk_usage` defaults to `flush=true`. On serverless the search-tier engine refuses that with `illegal_argument_exception: "Search engine does not support acquiring last index commit with flush_first"`: the data being asked to flush lives on the index tier, not the search tier the call hits. The search tier reads committed segments from object storage and has no in-memory write buffer to flush, so `flush=false` is the architecturally correct call shape there, not a degraded mode.

For Rally benchmarks `flush=false` is also safe in practice. `disk-usage-stats` runs after the steady-state phase. The canonical schedule fires `disk-usage-stats` last, after `index`, `refresh`, `force-merge`, `wait-until-merges-finish`, and a final `refresh`. By telemetry time no writes are in flight, so `flush=false` returns the same bytes `flush=true` would. The two diverge only for live-indexing workloads that fire `disk-usage-stats` while writes are still settling, where bytes would be slightly under-reported.

Today the only workaround on serverless is to drop `disk-usage-stats` entirely, which loses per-field byte attribution. This param adds a knob so callers can opt into `flush=false` instead.

## Change

`DiskUsageStats` reads a new `disk-usage-stats-flush` telemetry param (default `True`) and passes it through to `client.indices.disk_usage(..., flush=...)`. A new unit test asserts `flush=False` is plumbed through when set; existing `TestDiskUsageStats` assertions are updated to include the now-explicit `flush=True` kwarg.

Usage in a benchmark entry's `stats.telemetry.params`:

```yaml
stats.telemetry: ["disk-usage-stats", "shard-stats"]
stats.telemetry.params:
  disk-usage-stats-flush: false
```

Unset means current behavior (`flush=true`). The follow-up consumer is the serverless `tsdb-metricsgen-270m-es95-vs-es819` entry in `elastic/elasticsearch-benchmarks`, filed separately because Rally has its own release cadence. Part of elastic/elasticsearch#148193.

### DIFF
--- a/docs/telemetry.rst
+++ b/docs/telemetry.rst
@@ -302,6 +302,7 @@ Supported telemetry parameters:
 
 * ``disk-usage-stats-indices`` (default all indices in the track): Comma separated string or a list of indices who's disk usage to fetch.
 * ``disk-usage-stats-timeout`` (default 600s): A time value specifying the timeout for the disk usage API call.
+* ``disk-usage-stats-flush`` (default true): Whether to pass ``flush=true`` to the disk usage API. When false, the response may not include uncommitted data.
 
 Example::
 
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2359,12 +2359,8 @@ class DiskUsageStats(TelemetryDevice):
                 usage to fetch. Default is all indices in the track.
             ``disk-usage-stats-timeout``: Timeout in seconds for the disk usage API call. Default is 600s.
             ``disk-usage-stats-flush``: Whether to pass ``flush=true`` to the disk
-                usage API. Default is ``True`` (matches the API default). Set to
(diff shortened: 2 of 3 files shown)


romseygeek (approved):
LGTM

gareth-ellis (approved):
LGTM

gareth-ellis:
@salvatore-campagna can you run `make lint` and then push updated files?

salvatore-campagna:
made another commit to update the docs.

PR #2087, opened 2026-04-01
Add support for multi-clusters.

gareth-ellis:
This adds support for a single rally load driver to be able to run benchmarks against multiple Elasticsearch clusters at the same time. This differs than the current behaviour of providing a list of hosts, or a list of clusters where requests are either spread out across the hosts, or a track has to cater specifically for the multiple clusters.

PR #2126, opened 2026-05-20
Add per window throughput

gareth-ellis:
Whilst working on #2087 I was reminded that our throughput calculation is questionable - where it is an average of all requests since the start - this means that the longer into a test we get, the less impact a change in real throughput has on the reported throughput.

This adds "per window" throughput as an option. I propose at a later date, we switch the default of this so that the per window is the default - but this can be discussed in the PR.

### DIFF
--- a/esrally/driver/driver.py
+++ b/esrally/driver/driver.py
@@ -1647,8 +1647,6 @@ class ThroughputCalculator:
             self.has_samples_in_sample_type = False
             # start relative to the beginning of our (calculation) time slice.
             self.start_time = start_time
-            # used for windowed throughput: stored inside finish_bucket before snapshots are updated
-            self._prev_total_count = 0
             self._prev_interval = 0
             self._windowed_rate = None
 
@@ -1679,7 +1677,6 @@ class ThroughputCalculator:
             delta_count = new_total - self.total_count
             delta_interval = self.interval - self._prev_interval
             self._windowed_rate = delta_count / delta_interval if delta_interval > 0 else None
-            self._prev_total_count = self.total_count
             self._prev_interval = self.interval
             self.unprocessed = []
             self.total_count = new_total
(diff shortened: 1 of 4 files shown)


esrally/driver/driver.py
```diff
@@ -1645,11 +1649,18 @@ def __init__(self, bucket_interval, sample_type, start_time):
...
 
         @property
         def throughput(self):
             return self.total_count / self.interval
 
+        @property
+        def windowed_throughput(self):
+            """Throughput based only on ops and time elapsed since the previous bucket boundary."""
+            return self._windowed_rate if self._windowed_rate is not None else self.throughput
+
```
gbanasiak, line 1663: Can we avoid fallback to running average (`self.throughput`)? I think it would be better to return `None` and handle it at call sites, e.g. by skipping generation of throughput sample.

gbanasiak (commented):
As expected, this is much less stable than current running average. Here are 4 runs of indexing in `geonames` on my local computer - 2 running average (blue/green) vs. 2 window throughput (red/pink).

```
esrally race --track=geonames --challenge=append-no-conflicts-index-only --distribution-version=9.2.7 --config=...
```

Screenshot 2026-06-16 at 18 36 00

Around 10s into the indexing task we can see throughput spike on both windowed runs. I'm guessing this is an artifact of processing early samples which arrived late due to initial ES queueing (?). Can we consider such read-outs as "real"? Say we initiate a bulk request and it takes 10s to complete. If bucket interval (now 1s) is much lower than maximum bulk latency we might always see those types of spikes, right? I'm wondering if post-processing is a better strategy.

We probably want to expose bucket interval to experiment.

At the end of the run we can see how throughput decreases, most likely due to some of the clients finishing earlier.

gbanasiak:
Here is a comparison of post-processed throughput calculated from service time samples (`windowed-start-time`) with proposed throughput (`windowed-proposed`) from one of the races shown earlier. I've also added service time samples. The time buckets are 1s.

Screenshot 2026-06-17 at 10 09 40

It is not clear to me why there's a throughput dip at the beginning of the race with no corresponding dip in the throughput calculated from raw service time samples. If there's a slowdown and clients are waiting for bulk responses new samples should not be created.

We may need to trace throughput calculation closer to understand this.

PR #2127, opened 2026-05-22
Add OTLP support to rally

gareth-ellis:
This commit adds support for indexing to the _otlp/v1/metrics endpoint of elasticsearch.

There's quite a bit more than just a new runner here, as follows:
- To index to _otlp, you need an otel dataset. Currently I have created these via [metricsgenreceiver](https://github.com/elastic/metricsgenreceiver).
- Once you have a json otel dataset, you have two choices:
   - Use the json doc as is, prior to a race starting that required an otlp dataset, then a .pb file will be created - this is the protobuf version of your json doc. 
   - Alternatively you can store this file along with your corpus, and the .pb file will be collected instead. This will save time, especially if running the benchmark multiple times.
 

To do:
- Docs

### DIFF
--- a/esrally/driver/runner.py
+++ b/esrally/driver/runner.py
@@ -46,6 +46,7 @@ __RUNNERS = {}
 
 def register_default_runners(config: Optional[types.Config] = None):
     register_runner(track.OperationType.Bulk, BulkIndex(), async_runner=True)
+    register_runner(track.OperationType.OtlpIngest, OtlpIngest(), async_runner=True)
     register_runner(track.OperationType.ForceMerge, ForceMerge(), async_runner=True)
     register_runner(track.OperationType.IndexStats, Retry(IndicesStats()), async_runner=True)
     register_runner(track.OperationType.NodeStats, NodeStats(), async_runner=True)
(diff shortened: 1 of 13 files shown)


gareth-ellis:
This is split into 6 PRs now:
The recommended review/merge order applies:

[#2134](https://github.com/elastic/rally/pull/2134) and [#2135](https://github.com/elastic/rally/pull/2135) — fully independent, can merge in any order, in parallel.
[#2136](https://github.com/elastic/rally/pull/2136) and [#2137](https://github.com/elastic/rally/pull/2137) — both need [#2135](https://github.com/elastic/rally/pull/2135), can review in parallel once [#2135](https://github.com/elastic/rally/pull/2135) is in.
[#2138](https://github.com/elastic/rally/pull/2138) — needs [#2135](https://github.com/elastic/rally/pull/2135) + [#2136](https://github.com/elastic/rally/pull/2136) + [#2137](https://github.com/elastic/rally/pull/2137).
[#2139](https://github.com/elastic/rally/pull/2139) — needs [#2138](https://github.com/elastic/rally/pull/2138).

esrally/driver/runner.py
```diff
@@ -806,6 +807,179 @@ def __repr__(self, *args, **kwargs):
...
+    async def __call__(self, es, params):
+        """
+        :param es: The Elasticsearch client.
+        :param params: A hash with all parameters.
+
+        Mandatory parameters:
+        * ``body``: Raw binary protobuf payload (bytes).
+
+        Optional parameters:
+        * ``endpoint``: OTLP endpoint path. Defaults to ``/_otlp/v1/metrics``.
```
inqueue, line 857: Could a `signal-type` attribute be considered with possible values of `metrics`, `logs`, or `traces` instead of `endpoint`?

inqueue (changes requested):

PR #2120, opened 2026-05-13
Fully remove document type support

pquentin:
Mapping types were removed in Elasticsearch 7.0 and are rejected by Elasticsearch 8.0, which is Rally's minimum supported version. The doc-type plumbing was therefore dead on every supported server. Remove Index.types, Documents.target_type, the per-bulk type param, and the doc_type URL component from Query._raw_search.  Tracks that still specify types/target-type/search-operation type now fail to load, throwing a TrackSyntaxError that points the user to the field to remove.

All tracks in rally-tracks were migrated years ago. One example: https://github.com/elastic/rally-tracks/commit/5a785504e6f5cd2b70b5d28c2a6b16985035d90d

This pull request cost $23.25 in Claude. (Probably should _not_ have used Opus here.)

TODO:

 - [x] Test manually
 - [x] Fix remaining CI failures

PR #2128, opened 2026-05-25
Fix flaky test_union_document_corpora_is_symmetric test

pquentin:
Example failure: https://github.com/elastic/rally/actions/runs/26035271579/job/76531523012?pr=2120

I can't reproduce locally, and this was only exposed by #2120, but the fix here makes sense without the rest of #2120. Using a set and expecting a given order can only result in flakiness.


gbanasiak (approved):
LGTM

PR #2134, opened 2026-05-29
1/6 Don't crash on non-UTF-8 ApiError bodies

gareth-ellis:
## Summary

When an Elasticsearch ApiError's body is non-UTF-8 (e.g. binary protobuf returned by OTLP endpoints under coordinator-bytes backpressure), `e.body.decode("utf-8")` in `execute_single` raises `UnicodeDecodeError` and crashes the worker.

Replaces six `.decode("utf-8")` call sites in the driver's ApiError handler with `.decode("utf-8", errors="replace")` so undecodable bytes downgrade to U+FFFD instead of crashing. No semantic change for valid UTF-8 (the common case).

Part 1 of 6 in the OTLP ingest series, but a standalone hardening fix that's useful on its own. Has no dependencies on the rest of the series.

## Series
- **#2134 (this PR)** — Don't crash on non-UTF-8 ApiError bodies
- #2136 — Add OtlpParamSource for OTLP corpora (depends on #2135)
- #2137 — Add OtlpIngest runner (depends on #2135)
- #2138 — Parallelize OTLP corpus generation (depends on #2135, #2136, #2137)

## Test plan
- [x] All existing driver tests pass
- [x] pre-commit clean

### DIFF
--- a/tests/driver/driver_test.py
+++ b/tests/driver/driver_test.py
@@ -1988,6 +1988,25 @@ class TestAsyncExecutor:
         assert exc.value.args[0] == ("Request returned an error. Error type: api, Description: Huge error, HTTP Status: 499")
 
     @pytest.mark.asyncio
+    async def test_execute_single_with_http_400_with_non_utf8_raw_response_body(self):
+        es = None
+        params = None
+        body = io.BytesIO(b"\xff")
+        str_literal = str(body)
+        error_meta = elastic_transport.ApiResponseMeta(
+            status=499,
+            http_version="1.1",
+            headers=elastic_transport.HttpHeaders(),
+            duration=0.0,
+            node=elastic_transport.NodeConfig(scheme="http", host="localhost", port=9200),
+        )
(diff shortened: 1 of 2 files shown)


pquentin (approved):
It's wild that OTLP ingest does this. A bug IMHO. Still approving to unblock you.

tests/driver/driver_test.py
```diff
@@ -1987,6 +1987,25 @@ async def test_execute_single_with_http_400_with_raw_response_body(self):
...
         assert exc.value.args[0] == ("Request returned an error. Error type: api, Description: Huge error, HTTP Status: 499")
 
+    @pytest.mark.asyncio
+    async def test_execute_single_with_http_400_with_non_utf8_raw_response_body(self):
+        es = None
+        params = None
+        body = io.BytesIO(b"\xff")
+        str_literal = str(body)
+        error_meta = elastic_transport.ApiResponseMeta(
+            status=499,
```
gareth-ellis: @copilot why 499 and not 400 like the test is called? A server is unlikely to return a 499, since that is typically used as a client timeout

PR #2135, opened 2026-05-29
2/6 Add OTLP binary protobuf core IO and track preparation

gareth-ellis:
## Summary

Adds the foundation for OTLP (OpenTelemetry Protocol) corpora in Rally:

- `OtlpProtobufFile` in `esrally/utils/io.py`: reads/writes length-prefixed `ExportMetricsServiceRequest` protobufs, with a sidecar `.offset` file so worker partitions can seek without rescanning the corpus.
- `prepare_otlp_document_set` in `esrally/track/loader.py`: downloads a pre-built `.pb` from the corpus base URL when available, otherwise converts a local JSON corpus on disk.
- New OTLP document-set fields in `esrally/track/track.py`, `opentelemetry-proto` dependency in `pyproject.toml`.

Part 2 of 6 in the OTLP ingest series. Standalone — no dependencies on other PRs in the series.

## Series
- #2134 — Don't crash on non-UTF-8 ApiError bodies
- **#2135 (this PR)** — Add OTLP binary protobuf core IO and track preparation
- #2136 — Add OtlpParamSource for OTLP corpora (depends on this PR)
- #2137 — Add OtlpIngest runner (depends on this PR)
- #2138 — Parallelize OTLP corpus generation (depends on this PR, #2136, #2137)
- #2139 — Support gzipped OTLP ingest requests (depends on #2138)

## Test plan
- [x] 146 unit tests covering OtlpProtobufFile read/write/offset behaviour and the loader's prepare-track flow
- [x] pre-commit clean (pyupgrade, black, isort, mypy, pylint)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2136, opened 2026-05-29
3/6 Add OtlpParamSource for OTLP corpora

gareth-ellis:
## Summary

Adds `OtlpParamSource` in `esrally/track/params.py` — the parameter source that drives OTLP ingest operations.

- Streams length-prefixed protobuf records out of an `OtlpProtobufFile` rather than materializing into memory (a 1 MB-per-record corpus with 6k records per partition would otherwise consume 6+ GB per worker).
- Partitions records across workers using the offset sidecar so workers seek directly to their slice.
- Surfaces `percent_completed` so Rally's progress bar tracks real progress.
- Supports a `looped` mode that cycles the partition indefinitely for time-bound benchmarks.
- Honors a `corpora` operation param so a track with multiple OTLP corpora can target a specific one.

**Depends on #2135 — merge after #2135.** Part 3 of 6.

## Series
- **#2136 (this PR)** — Add OtlpParamSource for OTLP corpora
- #2138 — Parallelize OTLP corpus generation (depends on #2135, this PR, #2137)

## Test plan
- [x] 12 new unit tests cover partitioning, looping, percent-completed, corpus filtering, StopIteration semantics
- [x] pre-commit clean

### DIFF
--- a/create-notice.sh
+++ b/create-notice.sh
@@ -70,6 +70,9 @@ function main {
     add_license "wheel" "https://raw.githubusercontent.com/pypa/wheel/main/LICENSE.txt"
     add_license "pip" "https://raw.githubusercontent.com/pypa/pip/main/LICENSE.txt"
     add_license "boto3" "https://raw.githubusercontent.com/boto/boto3/develop/LICENSE"
+    # OTLP corpus preparation dependencies (otlp extra)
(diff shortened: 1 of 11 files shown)


ebadyano (approved):
lgtm

PR #2137, opened 2026-05-29
4/6 Add OtlpIngest runner with backpressure-aware retries

gareth-ellis:
## Summary

Adds the `OtlpIngest` runner in `esrally/driver/runner.py` — POSTs the serialized protobuf bytes to the OTLP metrics endpoint.

Key design choices:
- **Transport-level retries disabled** (`es.options(max_retries=0)`). The default elasticsearch-py retry loop fires 4 attempts back-to-back with no sleep, which hammers an already-overloaded ES under 429/503.
- **Explicit retry loop with exponential backoff + full jitter**. Retries 429/502/503/504 and connection errors; configurable via `retries-on-error` (default 5) and `retry-wait-period` (default 0.5s, capped at 30s).
- **Non-retryable ApiErrors return a failure dict** instead of propagating, so the driver records the error without crashing the worker.
- **Custom `application/x-protobuf` serializer** + `AsyncElasticsearch.perform_request` bypass to avoid Rally's REST-compat `Accept` headers that OTLP endpoints reject.

**Depends on #2135 — merge after #2135.** Part 4 of 6.

## Series
- **#2137 (this PR)** — Add OtlpIngest runner
- #2138 — Parallelize OTLP corpus generation (depends on #2135, #2136, this PR)

## Test plan
- [x] 14 unit tests: 429 → retry → success, exhausted retries → failure dict, 400 → no retry, transport max_retries=0 is set, custom retry-wait-period honored, exponential backoff schedule verified
- [x] pre-commit clean

### DIFF
--- a/create-notice.sh
+++ b/create-notice.sh
@@ -70,6 +70,9 @@ function main {
     add_license "wheel" "https://raw.githubusercontent.com/pypa/wheel/main/LICENSE.txt"
     add_license "pip" "https://raw.githubusercontent.com/pypa/pip/main/LICENSE.txt"
     add_license "boto3" "https://raw.githubusercontent.com/boto/boto3/develop/LICENSE"
+    # OTLP corpus preparation dependencies (otlp extra)
+    add_license "opentelemetry-proto" "https://raw.githubusercontent.com/open-telemetry/opentelemetry-python/main/LICENSE"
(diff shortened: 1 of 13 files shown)


ebadyano (approved):
lgtm

PR #2138, opened 2026-05-29
5/6 Parallelize OTLP corpus generation; download compressed .pb when available

gareth-ellis:
## Summary

Two perf improvements for OTLP corpus preparation. Neither changes hot-path benchmark behaviour.

**1. Parallel .pb generation.** `OtlpProtobufFile.create()` now uses a `ProcessPoolExecutor` pipeline — JSON lines stream in, worker processes parse and serialize in parallel, the main process writes batched results in source order. Worker count tunable via `RALLY_OTLP_CONVERSION_WORKERS`. Converting an 82 GB JSON corpus drops from hours to minutes.

**2. Compressed .pb download.** When the JSON corpus is published as a compressed archive (e.g. `.zst`), prepare-track now tries the matching compressed `.pb` from the corpus base URL (e.g. `.pb.zst`) and decompresses locally. Typically 2–4× less network bytes than the raw `.pb`.

**Depends on #2135, #2136, #2137 — merge after all three.** Part 5 of 6.

## Series
- **#2138 (this PR)** — Parallelize OTLP corpus generation
- #2139 — Support gzipped OTLP ingest requests (depends on this PR)

## Test plan
- [x] New tests: parallel/sequential output is byte-identical, RALLY_OTLP_CONVERSION_WORKERS is honored, compressed-then-fallback download path
- [x] All 314 OTLP tests green
- [x] pre-commit clean

### DIFF
--- a/create-notice.sh
+++ b/create-notice.sh
@@ -70,6 +70,9 @@ function main {
     add_license "wheel" "https://raw.githubusercontent.com/pypa/wheel/main/LICENSE.txt"
     add_license "pip" "https://raw.githubusercontent.com/pypa/pip/main/LICENSE.txt"
     add_license "boto3" "https://raw.githubusercontent.com/boto/boto3/develop/LICENSE"
+    # OTLP corpus preparation dependencies (otlp extra)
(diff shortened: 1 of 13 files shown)


ebadyano (approved):
lgtm

PR #2139, opened 2026-05-29
6/6 Support gzipped OTLP ingest requests

gareth-ellis:
## Summary

Adds an optional `gzip` operation parameter that turns on `Content-Encoding: gzip` for OTLP ingest.

Records are pre-compressed at prepare-track time into a sibling `.pbgz` corpus file (each record independently gzipped). The benchmark hot path ships those bytes verbatim — no in-process compression, no decompress/recompress. Operations sharing a corpus can opt in independently; when a corpus has both gzip and non-gzip consumers, Rally generates both `.pb` and `.pbgz`.

Measured impact: ~12% throughput improvement against an ES cluster where coordinator-bytes backpressure is the bottleneck (gzip reduces decompressed body size accounted against the bucket).

**Depends on #2138 — merge after #2138 (which in turn depends on #2135, #2136, #2137).** Part 6 of 6, completes the OTLP ingest series.

## Series
- #2138 — Parallelize OTLP corpus generation
- **#2139 (this PR)** — Support gzipped OTLP ingest requests

## Test plan
- [x] New tests: gzip flag toggles .pbgz path, body bytes are gzip-magic-prefixed, Content-Encoding header set correctly
- [x] All 321 OTLP tests green
- [x] pre-commit clean

### DIFF
--- a/create-notice.sh
+++ b/create-notice.sh
@@ -70,6 +70,9 @@ function main {
     add_license "wheel" "https://raw.githubusercontent.com/pypa/wheel/main/LICENSE.txt"
     add_license "pip" "https://raw.githubusercontent.com/pypa/pip/main/LICENSE.txt"
     add_license "boto3" "https://raw.githubusercontent.com/boto/boto3/develop/LICENSE"
+    # OTLP corpus preparation dependencies (otlp extra)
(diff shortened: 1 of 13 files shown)


ebadyano (approved):
lgtm

PR #2144, opened 2026-06-15
Add track parameter validation hook

### LINKED ISSUE #2143: Add extension point to support per-track parameter validation
Rally tracks can register custom param sources, runners, and schedulers via `register(registry)` in `track.py`. There is currently no equivalent hook for validating track parameters before a benchmark run starts.

When a track uses array parameters that must satisfy a structural constraint - for example, all scheduling arrays must have either 1 element or exactly N elements where N is the number of configured phases - there is no clean way to enforce this.

The proposal is to add a `register_validator(challenge_name, fn)` function to `esrally/track/params.py`, following the same pattern as the existing `register_param_source_for_name`. The validator function receives the resolved params dict and raises `TrackConfigError` with a readable message if the params are invalid. Rally calls all registered validators for the active challenge after the track spec is loaded and params are merged, but before the schedule is built. `TrackConfigError` already exists in `esrally/exceptions.py` for this purpose, so no new exception type is needed.

```python
# In track.py
def _validate_autoscale_params(params):
    phases = params.get("as_phases", 5)
    for name, default in [("as_warmup_time_periods", [600]), ("as_time_periods", [1800])]:
        arr = params.get(name, default)
        if len(arr) != 1 and len(arr) != phases:
            raise TrackConfigError(f"{name} must have 1 or {phases} elements, got {len(arr)}")

def register(registry):
    for challenge in ("search-autoscale", "ingest-autoscale", "ingest-search-autoscale"):
        registry.register_validator(challenge, _validate_autoscale_params)
```

This produces a clean error like `TrackConfigError: as_search_clients must have 1 or 4 elements, got 2` before any cluster operations run.

Without a validation hook, track authors either skip validation (silent wrong results) or implement workarounds that give users misleading error messages. A first-class `register_validator` hook is a small addition that makes the plugin API complete and consistent with the rest of the extension points Rally already provides.

reugn:
Tracks can already register custom param sources, runners, and schedulers, but there was no way to validate track parameters before a benchmark runs. This adds that missing hook.

- Adds `register_validator(challenge_name, fn)` so a track's `register(registry)` can register one or more validators for a challenge.
- A validator receives the resolved track parameters and raises `TrackConfigError` when they are invalid.
- Rally runs the validators for the selected challenge in `BenchmarkCoordinator.setup()`, after the track is loaded but before it provisions nodes or starts the benchmark, so bad parameters fail fast with a clear message.

This change is additive and non-breaking. Tracks that do not register a validator behave exactly as before, since the hook is a no-op for them. No existing function signature changes in a way that affects current callers.

Closes #2143


pquentin (approved):
Thanks! LGTM.

It took me a while to test this, there's a Rally bug where `--track-path` fails with a mysterious error when there's a  `.venv` in the track folder. (I have one because I tested the pip install instructions for a different PR.)

I was a bit worried at the place where the validation was called, and I also noticed the exception is displayed as an internal unexpected error and a traceback, while it should be displayed like other Rally user errors. I explored a few ways to fix that, and ended up with [`e086ed2`](https://github.com/elastic/rally/pull/2144/changes/e086ed26677bdd162f4791c26efa992dc976d7e2). (It failed in CI, so I pushed a fix for that too.)

PR #2150, opened 2026-06-22
Fix : Rally should not overwrite pre existing templates by default (#…

### LINKED ISSUE #1900: Rally should not overwrite pre existing templates by default
Originally a decision was made that rally should always update existing index templates - the reasoning was that the fields in the index could well be updated regularly. We are now in a position where we have a stable layout for the index templates, thus we should now change to only writing templates if they don't already exist, or if a user specifically says they want them overwritten.

The reasoning behind that is it will allow users to set custom lifecycle policies for moving data to frozen and more.

Suggested implementation:
- Check if the index template exists.
- If it does, and the user has not set a new setting - `overwrite_existing_templates` to true in the [reporting] section of rally.ini, then do not do anything
- If the template does not exist we should create the templates

Where we update index template: https://github.com/elastic/rally/blob/master/esrally/metrics.py#L902

Thomas-Belly:
### Description

This PR fixes an issue where Rally would unconditionally overwrite existing index templates (specifically the `rally-annotations` template), entirely ignoring the `datastore.overwrite_existing_templates` setting defined by the user in `rally.ini`.

### What changed?

I updated the logic in `metrics.py` (`EsRaceStore`) to evaluate `self._index_handler.overwrite_templates` before applying the template update. 

Now, if the template already exists and the user has not explicitly opted to overwrite it (which defaults to `False`), Rally will preserve the existing template and simply log a debug message detailing the diff, honoring the intended default behavior.

### Related Issue

Fixes #1900

### How was this tested?

- [x] Verified the logic locally.
- [x] Ran `make format` to ensure code style compliance (Black/isort).
- [x] Ran `make lint test` successfully (all unit tests passed without errors).

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -2152,9 +2152,10 @@ class EsRaceStore(RaceStore):
             )
         else:
             if not self.client.exists(index="rally-annotations"):
-                # create or overwrite template on index creation
-                self.client.put_template("rally-annotations", self._index_handler.annotations_template())
+                if not self.client.index_template_exists(name="rally-annotations") or self._index_handler.overwrite_templates:
+                    self.client.put_template("rally-annotations", self._index_handler.annotations_template())
                 self.client.create_index(index="rally-annotations")
+
             self.client.index(
                 index="rally-annotations",
                 id=annotation_id,


b-deam (changes requested):
Thanks for picking this up, much appreciated!

Sorry the surrounding code makes this trickier than it should be. The template handling has grown over time and `rally-annotations` is a bit of an outlier. Just a couple of things to tighten up and I think we'll be good to merge.

Firstly, the description says Rally will "_log a debug message detailing the diff_," but the current check silently skips the `put_template` when preserving, so no diff is computed or logged. The good news is that this behaviour you want already exists.

`IndexHandler._should_apply_update` handles all the cases (create when missing, skip when identical, keep-and-log-the-diff by default, and overwrite-with-a-warning when `overwrite_existing_templates=true`). `_ensure_date_based_template` is the closest example of feeding a single index template through it, so that's the spot to model on. I'd avoid routing this through `ensure_index_template`, since that one is tied to `EsStoreType` and annotations aren't a store type.

Take a look at what `_ensure_date_based_template` actually does step by step. It checks whether the template exists, fetches the existing one if it does (and treats it as absent if not), then passes the old and new versions to `_should_apply_update`, which decides whether to call `put_template`.

That middle sequence (exists, fetch, compare, conditionally put) is exactly what annotations need too. So rather than copying it, the idea is to lift those shared steps into one small helper on `IndexHandler` that takes the template name and body as inputs. Then `_ensure_date_based_template` calls the new helper with the store's template name, and annotations call it with `rally-annotations` and `annotations_template()`.  

That way `add_annotation` then just calls that path instead of doing the exists/put logic inline. 

Lastly, no good deed goes unpunished, so this needs tests 😄. 

A good example is `test_open` from #1912, which covers each branch (missing, exists but identical, differs and kept, differs and overwritten). Mirroring those four cases for annotations would be ideal.

Happy to help if you get stuck.

b-deam:
Hey @Thomas-Belly - thanks for your contribution, but given the ticket has been inactive for ~3 weeks I am going to close it for now, please feel free to re-open it when you're ready to work on this again :-)

PR #2152, opened 2026-06-24
Pin uv version in it CI

pquentin:
Also switches to UV_UNMANAGED_INSTALL which is documented to be [useful on CI](https://docs.astral.sh/uv/reference/environment/#uv_unmanaged_install) as it disables self-updates and does not affect the $PATH.

As requested in https://github.com/elastic/rally-tracks/pull/1177#discussion_r3459010186.


gbanasiak (approved):
LGTM

Issue #1724, opened 2023-06-02
Improve remote metrics store flushing


b-deam:
Similar to https://github.com/elastic/rally/issues/1723, benchmarks with many concurrent/parallel tasks that generate a lot of samples can end up producing too many samples for our single-threaded metrics store flush method to handle, causing excessive memory usage on the load driver and delaying metrics delivery to the remote store.

The `Driver` already wakes up every `POST_PROCESS_INTERVAL_SECONDS` (30) to flush the collected samples to the metrics store:
https://github.com/elastic/rally/blob/2470328e4167e28b6e3f47ffcf0e8700845522e9/esrally/driver/driver.py#L295-L305
https://github.com/elastic/rally/blob/2470328e4167e28b6e3f47ffcf0e8700845522e9/esrally/driver/driver.py#L950-L955
https://github.com/elastic/rally/blob/2470328e4167e28b6e3f47ffcf0e8700845522e9/esrally/driver/driver.py#L1062-L1068

Once `self._client.bulk_index()` completes we clear the in-memory buffer:
https://github.com/elastic/rally/blob/2470328e4167e28b6e3f47ffcf0e8700845522e9/esrally/metrics.py#L924-L942

We use the `elasticsearch.helper.bulk()` method to iterate over all the documents in the in-memory buffer, sending them in chunks of 5000 docs at a time:
https://github.com/elastic/rally/blob/2470328e4167e28b6e3f47ffcf0e8700845522e9/esrally/metrics.py#L89-L93

This approach works fine for most benchmarks, but for challenges like `logging-indexing-querying` of the `elastic/logs` track we generate a _lot_ of documents, meaning that this in-memory buffer is often full of tens to hundreds of thousands of documents that are indexed by a single client controlled by the `Driver` actor.  This problem is exacerbated in environments where the load driver is a long way away from the metrics store (i.e. cross regional), or if the metrics store itself is overloaded, because our single client throughput is bound by the latency of each request.

[There’s a `parallel_bulk` helper](https://elasticsearch-py.readthedocs.io/en/v8.6.1/helpers.html) that uses `multiprocessing.pool import ThreadPool`, [but the Thespian docs specifically call out that](https://thespianpy.com/doc/using.pdf):
> Actors are not intended to create other processes or threads; attempts to do so may create undefined behavior

We could consider a few things here, but all will require extensive testing:
[...]

PR #2154, opened 2026-07-02
Fix metrics store flush blocking the benchmark actor event loop

reugn:
Rally's `DriverActor` runs in Thespian's single-threaded event loop, so any `time.sleep()` inside a message handler blocks all incoming messages. Every 30 seconds, `receiveMsg_WakeupMessage` calls `EsMetricsStore.flush()`, which calls `guarded()` — a retry wrapper that sleeps between attempts. With the original settings (10 retries, 120s socket timeout), a single stuck connection could block the event loop for ~39 minutes. Thespian's 5-minute message delivery timeout would then fire, returning `PoisonPacket` for every queued `UpdateSamples` and `JoinPointReached` message, silently killing the benchmark. This is what happened twice in my benchmarks: a dead TCP socket (blackhole state) between the EU load driver and the US metrics store caused exactly this. The metrics store itself was green throughout.

The fix bounds worst-case blocking to ~250 seconds by reducing retries in `guarded()` from 10 to 3 and applying a `request_timeout=60` on the flush path. `EsClient` builds a single `options(request_timeout=60)` client once and reuses it across the flush/close path. The same 60s timeout therefore also covers `EsRaceStore.store_race()` and `EsResultsStore.store_results()`, which run in the same actor context at benchmark completion, as well as annotation writes (previously the client default). `EsMetricsStore.flush()` is rewritten to catch transient errors and re-queue docs for the next 30-second cycle rather than crashing, with a consecutive-failure counter that fails the benchmark after 10 unrecoverable cycles. A quiet cycle (no docs to flush) resets the counter. `SystemSetupError` (auth/config failures) bypasses re-queuing and raises immediately. The closing path logs lost docs as an error rather than silently dropping or blocking. Nine new tests cover all the error paths introduced.

### DIFF
--- a/esrally/metrics.py
+++ b/esrally/metrics.py
@@ -81,8 +81,9 @@ class EsClient:
     def delete_by_query(self, index, body):
         return self.guarded(self._client.delete_by_query, index=index, body=body)
 
-    def update_by_query(self, index, body):
-        return self.guarded(self._client.update_by_query, index=index, body=body)
+    def update_by_query(self, index, body, request_timeout=None):
+        client = self._client.options(request_timeout=request_timeout) if request_timeout is not None else self._client
+        return self.guarded(client.update_by_query, index=index, body=body)
 
     def delete(self, index, id):
         # ignore 404 status code (NotFoundError) when index does not exist
@@ -98,34 +99,38 @@ class EsClient:
     def exists(self, index):
         return self.guarded(self._client.indices.exists, index=index)
 
-    def refresh(self, index):
-        return self.guarded(self._client.indices.refresh, index=index)
+    def refresh(self, index, request_timeout=None):
+        client = self._client.options(request_timeout=request_timeout) if request_timeout is not None else self._client
+        return self.guarded(client.indices.refresh, index=index)
 
-    def bulk_index(self, *, index, items, use_data_streams):
+    def bulk_index(self, *, index, items, use_data_streams, request_timeout=None):
         # pylint: disable=import-outside-toplevel
         import elasticsearch.helpers
 
         if use_data_streams:
             for item in items:
                 item["_op_type"] = "create"
-        self.guarded(elasticsearch.helpers.bulk, self._client, items, index=index, chunk_size=5000)
+        client = self._client.options(request_timeout=request_timeout) if request_timeout is not None else self._client
+        self.guarded(elasticsearch.helpers.bulk, client, items, index=index, chunk_size=5000)
 
-    def index(self, *, index, item, id=None, use_data_streams):
+    def index(self, *, index, item, id=None, use_data_streams, request_timeout=None):
         doc = {"_source": item}
         if not use_data_streams and id:
             doc["_id"] = id
-        self.bulk_index(index=index, items=[doc], use_data_streams=use_data_streams)
+        self.bulk_index(index=index, items=[doc], use_data_streams=use_data_streams, request_timeout=request_timeout)
 
     def search(self, index, body):
         return self.guarded(self._client.search, index=index, body=body)
 
-    def guarded(self, target, *args, **kwargs):
+    def guarded(self, target, *args, _max_retries=3, **kwargs):
         # pylint: disable=import-outside-toplevel
         import elasticsearch
         import elasticsearch.helpers
         from elastic_transport import ApiError, TransportError
 
-        max_execution_count = 10
+        # 3 retries × 60s request_timeout + sleep keeps worst-case blocking under
+        # Thespian's 5-minute actor event-loop delivery timeout.
+        max_execution_count = _max_retries
         execution_count = 0
 
         while execution_count <= max_execution_count:
@@ -770,7 +775,7 @@ class MetricsStore:  # pylint: disable=too-many-public-methods
         """
         self._stop_watch.start()
 
-    def flush(self, refresh=True):
+    def flush(self, refresh=True, closing=False):
         """
         Explicitly flushes buffered metrics to the metric store. It is not required to flush before closing the metrics store.
         """
@@ -783,7 +788,7 @@ class MetricsStore:  # pylint: disable=too-many-public-methods
         """
         self.logger.info("Closing metrics store.")
         self.opened = False
-        self.flush()
+        self.flush(closing=True)
         self._clear_meta_info()
 
     def add_meta_info(self, scope, scope_key, key, value):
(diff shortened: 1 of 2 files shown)


esrally/metrics.py
```diff
@@ -1248,33 +1254,79 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
+                    self._car,
+                    sw.total_time(),
+                )
+            except exceptions.SystemSetupError:
+                raise
+            except exceptions.RallyError as e:
+                self._flush_consecutive_failures += 1
+                if closing:
+                    self.logger.error(
+                        "Failed to flush final %d metrics docs on close — these docs are lost: %s",
```
b-deam:
```suggestion
                        "Failed to flush final %d metrics docs on close, these documents have been dropped: %s",
```

_Nit_: The em-dash here is non-ASCII, which could maybe cause issues in some environments with a non-UTF-8 console. Almost certainly a non-issue in practice, but I think it's safest to stick with convention and keep log messages ASCII.
```python
python3 -c "print('—'.isascii())"
False
```

esrally/metrics.py
```diff
@@ -1248,33 +1254,79 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
+                if self._flush_consecutive_failures >= self._MAX_FLUSH_FAILURES:
+                    self.logger.error(
+                        "Metrics store unreachable after %d consecutive flush failures, failing benchmark.",
+                        self._MAX_FLUSH_FAILURES,
+                    )
+                    raise
+                with self._docs_lock:
+                    self._docs = docs_to_flush + self._docs
+                self.logger.warning(
+                    "Failed to flush %d metrics docs (attempt %d/%d), re-queuing for next cycle: %s",
```
b-deam, line 1310: Non-blocking, but this change could compound the [lots-of-samples problem](https://github.com/elastic/rally/issues/1724) if the metrics store is also slow. The re-queue is bounded by consecutive cycles (`_MAX_FLUSH_FAILURES`), so on high sample volume tracks, slow flushes could mean we hold up to ~10 cycles of docs in memory and re-attempt ever-larger batches, with each cycle taking longer to flush (more 5000-doc chunks to get through).

I think your change is a good tradeoff vs. crashing on a transient blip, but I'm just flagging that we probably want a follow-up in https://github.com/elastic/rally/issues/1724 to cap the re-queue by doc count/bytes to avoid this etc.

b-deam (commented):
Thanks a lot for the PR! I've left a few minor comments. 

Regarding `client.options()`, it's unfortunate but the ES Python instantiates a new client object (_and_ namespaced clients) every time it is called.  [On a hot path this shows up as measurable client side overhead](https://github.com/elastic/rally/issues/1673#issuecomment-1526910802), but I suspect that it's probably not actually a problem in this instance since metrics flushing occurs infrequently, in the background, and on a single thread.

esrally/metrics.py
reugn: Agreed. I've written up a suggestion in #1724 that folds this re-queue cap into a broader direction for the single-writer bottleneck.

reugn:
Thanks for the review, @b-deam. On the `client.options()` overhead, would you prefer I cache a single `options(request_timeout=60)` client on the `IndexHandler` and reuse it across the flush path, instead of creating one per call? It'd also let me drop most of the repeated per-call `request_timeout` plumbing.

b-deam:
> Thanks for the review, @b-deam. On the `client.options()` overhead, would you prefer I cache a single `options(request_timeout=60)` client on the `IndexHandler` and reuse it across the flush path, instead of creating one per call? It'd also let me drop most of the repeated per-call `request_timeout` plumbing.

This sounds reasonable to me.

esrally/metrics.py
```diff
@@ -1250,31 +1260,75 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
+                        len(docs_to_flush),
+                        e,
+                    )
+                    return
+                if self._flush_consecutive_failures >= self._MAX_FLUSH_FAILURES:
+                    self.logger.error(
+                        "Metrics store unreachable after %d consecutive flush failures, failing benchmark.",
+                        self._MAX_FLUSH_FAILURES,
+                    )
+                    raise
```
pquentin: Any reason why the error is not part of the exception? Having to read the (long!) log to correlate logs with exceptions is time-consuming for users.

esrally/metrics.py
```diff
@@ -1250,31 +1260,75 @@ def open(self, race_id=None, race_timestamp=None, track_name=None, challenge_nam
...
+                raise
+            except exceptions.RallyError as e:
+                self._flush_consecutive_failures += 1
+                if closing:
+                    self.logger.error(
+                        "Failed to flush final %d metrics docs on close, these documents have been dropped: %s",
+                        len(docs_to_flush),
+                        e,
+                    )
+                    return
```
pquentin: Why are we happy to swallow the exception and drop the docs on close? I think we need to try other things first, such as concurrent indexing or a larger downsampling factor, before just giving up and silently hiding flush exceptions?

pquentin (commented):
Thank you, this is a step in the right direction!

esrally/metrics.py
```diff
@@ -41,8 +41,15 @@ class EsClient:
...
     """
 
+    # Per-request timeout for the flush/close path. Bounds worst-case blocking in the actor
+    # event loop (see guarded()) well under Thespian's 5-minute message delivery timeout.
+    FLUSH_REQUEST_TIMEOUT = 60
+
     def __init__(self, client, cluster_version=None):
         self._client = client
+        # Reused across the flush/close path so we don't rebuild a client (and its namespaced
+        # clients) on every call. The underlying transport/connection pool is shared.
```
b-deam:
```suggestion
        # Reused across the flush/close path so we don't rebuild a client on every call. 
```

Nit: Probably don't need this level of detail in a comment.

esrally/metrics.py
reugn: Agreed. Chained the error as the cause and removed the log print.

esrally/metrics.py
reugn: Fixed to raise a RallyError instead of logging and dropping. This aligns with the pre-PR behavior.

esrally/metrics.py
reugn: Trimmed the comment.

b-deam (approved):
Thanks for iterating @reugn! This LGTM. But let's wait to see if @pquentin has any further feedback before merging.

PR #2155, opened 2026-07-02
Add TCP keepalive to metrics store client connections

reugn:
Adds TCP keepalive to the Elasticsearch client used by the metrics store (`EsMetricsStore`, `EsRaceStore`, `EsResultsStore`).

The root cause of the benchmark crash fixed in #2154 was a NAT gateway silently dropping its connection state while the socket remained `ESTABLISHED` on both ends. The client only discovered the connection was dead after waiting out the full `request_timeout` on every retry attempt. TCP keepalive detects this at the OS level without waiting for a request to time out. After 60 seconds of idle the OS sends a probe. If the server does not respond after 6 probes (one every 10 seconds) the socket is closed and the application receives an immediate connection error — worst case ~120 seconds instead of ~250 seconds. During active flushing (every 30 seconds) keepalive probes never fire since real traffic resets the idle timer.

Subclasses `elastic_transport.Urllib3HttpNode` to inject keepalive socket options into the urllib3 connection pool's `conn_kw` after pool initialization. The subclass is passed as `node_class` to the Elasticsearch client, which is the supported API for customising the transport layer. urllib3's default `TCP_NODELAY` option is preserved by seeding from `HTTPConnection.default_socket_options` before appending keepalive options. Linux sets `TCP_KEEPIDLE=60`, `TCP_KEEPINTVL=10`, `TCP_KEEPCNT=6`. macOS sets `TCP_KEEPALIVE=60` (the macOS equivalent of `TCP_KEEPIDLE`; interval and count are system-wide only). Both are guarded with `hasattr` to handle non-standard builds.

#2154 bounds the worst-case event-loop block to ~250 seconds via `request_timeout=60` and `max_retries=3`. This PR is the root-cause fix: keepalive surfaces dead connections before `request_timeout` fires, making the retry budget rarely needed in the NAT-drop scenario. The two mechanisms are complementary — keepalive catches dead network paths, `request_timeout` catches alive-but-unresponsive servers.

### DIFF
--- a/tests/metrics_test.py
+++ b/tests/metrics_test.py
@@ -22,7 +22,10 @@ import json
 import logging
 import os
 import random
+import socket
+import sys
 import tempfile
+import urllib3.connection
 import uuid
 from dataclasses import dataclass
 from unittest import mock
@@ -638,22 +641,16 @@ class TestKeepaliveUrllib3HttpNode:
         return metrics.KeepaliveUrllib3HttpNode(config=None)
 
     def test_enables_so_keepalive(self, monkeypatch):
-        import socket
-
         node = self._make_node(monkeypatch)
         assert (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1) in node.pool.conn_kw["socket_options"]
 
     def test_preserves_existing_socket_options(self, monkeypatch):
-        import socket
-
         sentinel = (socket.IPPROTO_TCP, 200, 42)
         node = self._make_node(monkeypatch, conn_kw={"socket_options": [sentinel]})
         assert sentinel in node.pool.conn_kw["socket_options"]
         assert (socket.SOL_SOCKET, socket.SO_KEEPALIVE, 1) in node.pool.conn_kw["socket_options"]
 
     def test_does_not_mutate_original_socket_options_list(self, monkeypatch):
-        import socket
-
         # Pass the original list object directly (no copy) so mutation is detectable.
         original = []
         node = self._make_node(monkeypatch, conn_kw={"socket_options": original})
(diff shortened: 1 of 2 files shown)


reugn:
@elasticmachine run tests

pquentin:
For the record, this was on the roadmap for the Python client at some point: https://github.com/elastic/elasticsearch-py/issues/3306, which may be the best place to fix this. But it will be more complex there, so a targeted fix makes sense in Rally.

reugn:
> For the record, this was on the roadmap for the Python client 

Given benchmarks are actively affected, I'd merge this targeted fix now and revisit when/if the client-side enhancement lands.

tests/metrics_test.py
```diff
@@ -621,6 +626,56 @@ def raise_bulk_index_error():
             client.guarded(raise_bulk_index_error)
 
 
+class TestKeepaliveUrllib3HttpNode:
+    """Tests for the TCP keepalive node subclass."""
+
+    def _make_node(self, monkeypatch, conn_kw=None):
+        """Return a KeepaliveUrllib3HttpNode with a mocked pool, bypassing real network setup."""
+        pool = mock.MagicMock()
+        pool.conn_kw = conn_kw if conn_kw is not None else {}
```
b-deam, line 636: _non-blocking_: Although extremely unlikely, by mocking this it means we may miss any internal/upstream changes in urllib3 to `conn_kw` that could cause this to fail silently.

I don't know how much extra work it would be to create an actual socket in a small test and check the options are applied correctly?

b-deam (approved):
Changes LGTM.

tests/metrics_test.py
reugn: Thanks for flagging this. Added a quick integration test that creates a real connection and checks `SO_KEEPALIVE` on the actual socket.

PR #2163, opened 2026-07-27
Ignore own parent process when detecting other Rally processes

### LINKED ISSUE #2026: `find_all_other_rally_processes` can detect its own parent in certain cases
If the process that launches rally happens to be a python process with `esrally` or `rally` in the command line,
then rally detects its own parent as a conflict and refuses to start.

**Example**

When using a VS Code debugpy launch task, the parent process command line looks like this: `/path/to/python .../debugpy/launcher ... -- /path/to/esrally ...`

**Proposed Solution**

I propose modifying `for_all_other_processes` to ignore parent processes in addition to the current process.

I have a change and updated unit tests ready to propose. Would the team be open to a PR for this?

lowbyteguy:
Closes #2026

If the process that launches Rally is itself a Python process whose command line mentions `esrally` (a VS Code `debugpy` launcher, or a wrapper script), `is_rally_process()` matches it, so Rally reports its own parent as a conflicting instance and refuses to start.

`for_all_other_processes()` already skips the current PID; it now skips the parent PID too. Genuine concurrent Rally processes are still detected, and a regression test in `tests/utils/process_test.py` covers the launcher case.

Ran `tests/utils/process_test.py` (7 passed); the new test fails without the `process.py` change.

### DIFF
--- a/esrally/utils/process.py
+++ b/esrally/utils/process.py
@@ -228,9 +228,12 @@ def kill_all(predicate: Callable[[psutil.Process], bool]) -> None:
 def for_all_other_processes(predicate: Callable[[psutil.Process], bool], action: Callable[[psutil.Process], None]) -> None:
     # no harakiri please
     my_pid = os.getpid()
+    # the process that launched us is not a competing Rally process either, even if its command line
+    # happens to mention Rally (e.g. a debugger or wrapper script that invokes esrally)
+    my_ppid = os.getppid()
     for p in psutil.process_iter():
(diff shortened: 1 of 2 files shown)


pquentin (commented):
I haven't reviewed it explicitly yet (only added an upvote), but I think the Copilot comment is correct: it's not that easy. There's also the question of the interaction with rallyd. I'd be interested to see an analysis of this, maybe done by an LLM initially.

PR #2165, opened 2026-07-27
Add validate-track subcommand to check track parameters without running a benchmark

### LINKED ISSUE #2145: Add a subcommand to check track parameters without running a benchmark
The track parameter validation hook (introduced in #2144) runs validators as part of `rally race`, inside the actor system, just before the benchmark starts. This means a bad parameter is only reported after Rally bootstraps the actor system, and the error is shown as a traceback rather than a clean message.

This issue tracks adding a `validate-track` subcommand that loads a track, runs the registered validators for the resolved challenge, and then exits, without provisioning a cluster or running the benchmark.

The proposal is to run the validators in the main process (outside the actor system), reusing the existing `params.invoke_validators` seam. Load the track, resolve the challenge the same way `race` does, and run the validators. On success, exit zero and optionally print a short confirmation (quiet by default); on failure, exit non-zero with the `TrackConfigError` message. This keeps the check fast, avoids any cluster contact, and gives a clean one-line error suitable for automation (e.g. fail before provisioning load drivers).

reugn:
The track parameter validation hook (introduced in #2144) runs validators as part of `rally race`, inside the actor system, just before the benchmark starts. A bad parameter is only reported after Rally bootstraps the actor system, and the error is shown as a traceback rather than a clean message.

This PR adds a `validate-track` subcommand that loads a track, resolves the challenge the same way `race` does, runs registered validators via `params.invoke_validators`, and exits - without provisioning a cluster or running the benchmark. Validation runs in the main process for a fast, automation-friendly check (exit zero on success, quiet by default; non-zero with the `TrackConfigError` message on failure).

Closes #2145

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -312,7 +312,7 @@ To check track parameters without running a benchmark, use the ``validate-track`
 
     esrally validate-track --track-path=/path/to/my-track --challenge=autoscaling --track-params='{"scheduling": [1]}' --build-flavor=serverless --no-quiet
 
-This loads the track (including Jinja rendering, schema checks, plugins, and any track ``dependencies``), resolves the challenge the same way ``race`` does (explicit ``--challenge`` or the default), runs validators registered for that challenge, and exits. Use ``--build-flavor`` / ``--serverless-operator`` when the track's templates depend on serverless conditionals (``validate-track`` cannot probe a cluster the way ``race`` does). On success the exit code is zero; confirmation text is silent by default unless you pass ``--no-quiet``. If no validators are registered for the resolved challenge, exit code 0 still means the track loaded, but no custom parameter checks ran. On failure it exits non-zero with the ``TrackConfigError`` message. See the :ref:`command line reference <clr_validate_track>` for scope limits (no corpora download; use ``--offline`` to skip track-repo updates).
+See the :ref:`command line reference <clr_validate_track>` for details, including serverless flags and scope limits.
 
 .. _adding_tracks_custom_runners:
 
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -308,6 +308,12 @@ Note the following:
 * Raise ``TrackConfigError`` to abort the benchmark. Rally runs the validators after loading the track and selecting the challenge, but before provisioning nodes or starting the benchmark, so a misconfigured parameter is reported without waiting for the engine to start.
 * You may register more than one validator for the same challenge; Rally invokes them in registration order.
 
+To check track parameters without running a benchmark, use the ``validate-track`` subcommand::
+
+    esrally validate-track --track-path=/path/to/my-track --challenge=autoscaling --track-params='{"scheduling": [1]}' --build-flavor=serverless --no-quiet
+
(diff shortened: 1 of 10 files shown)


pquentin (commented):
I'm wondering if this is really necessary.

* The validation added in #2144 happens as part of `BenchmarkCoordinator.setup()`, which is the [very first step of benchmark preparation](https://github.com/elastic/rally/blob/master/docs/architecture/actor_system.md#preparebenchmark). In my local tests, `esrally race` errors out after 6/7s, while `esrally validate-track` errors out after 2/3s.
* The other advantage you mentioned is the error message, but I improved this before merging in #2144, so now it's properly reported as `[ERROR]` in the standard output, even if there's more than just this line in the output.
* Regarding the actor system, benchmark preparation should be updated to not use it at all, even if this can be a bigger lift.
* And finally, I'm hesitant to ecxpand a functionality that we do not even use yet.

reugn:
The main use case for `validate-track` is esbench (and similar) preflight before provisioning: we want to fail on bad track params / Jinja / unused fields / register_validator checks before starting load drivers and burning time and cloud capacity. In that flow, race-time validation (“race fails in ~6s”) isn’t available yet - provision happens first, then Rally runs on the load drivers. I hit this several times with wikipedia search-autoscale (bad as_* params).

So this isn’t meant to replace race-time validation. It’s a small, quiet, exit-code-friendly CLI for orchestration.

inqueue:
> The main use case for `validate-track` is esbench (and similar) preflight before provisioning: we want to fail on bad track params / Jinja / unused fields / register_validator checks before starting load drivers and burning time and cloud capacity. In that flow, race-time validation (“race fails in ~6s”) isn’t available yet - provision happens first, then Rally runs on the load drivers. I hit this several times with wikipedia search-autoscale (bad as_* params).
> 
> So this isn’t meant to replace race-time validation. It’s a small, quiet, exit-code-friendly CLI for orchestration.

I can second the usefulness of this case as it would shorten the discovery time of a parameter related issue in the benchmark-- a problem I have encountered many times. However, the user or the provisioning tool would have to call Rally to perform the validation prior to launching load drivers. I am for this change since it enhances usability.

reugn:
> However, the user or the provisioning tool would have to call Rally to perform the validation prior to launching load drivers.

Right, that’s the intended follow-up. After this merges, esbench will call `esrally validate-track` in `start` before provisioning load drivers, so bad track params fail fast.

docs/advanced.rst
```diff
@@ -308,6 +308,12 @@ Note the following:
...
 
+To check track parameters without running a benchmark, use the ``validate-track`` subcommand::
+
+    esrally validate-track --track-path=/path/to/my-track --challenge=autoscaling --track-params='{"scheduling": [1]}' --build-flavor=serverless --no-quiet
+
+This loads the track (including Jinja rendering, schema checks, plugins, and any track ``dependencies``), resolves the challenge the same way ``race`` does (explicit ``--challenge`` or the default), runs validators registered for that challenge, and exits. Use ``--build-flavor`` / ``--serverless-operator`` when the track's templates depend on serverless conditionals (``validate-track`` cannot probe a cluster the way ``race`` does). On success the exit code is zero; confirmation text is silent by default unless you pass ``--no-quiet``. If no validators are registered for the resolved challenge, exit code 0 still means the track loaded, but no custom parameter checks ran. On failure it exits non-zero with the ``TrackConfigError`` message. See the :ref:`command line reference <clr_validate_track>` for scope limits (no corpora download; use ``--offline`` to skip track-repo updates).
```
pquentin: Should we omit more details here, since everything is already covered in `_clr_validate_track` and we link to it?

pquentin (approved):
Thanks! LGTM.

Approving due to popular demand, but I'd like to see actual usage of this feature in tracks. Also, a schema would catch more issues.

docs/advanced.rst
reugn: Trimmed the validate-track docs to keep the example and point to the CLI reference.

PR #2170, opened 2026-08-05
Upgrade Buildkite agent images from Ubuntu 22.04 to 24.04

pquentin:
The motivation is to fix the current Dev Docker Build alerts, which are caused by deadlocks/hangs in old Docker Compose versions. While this bug won't be fixed until https://github.com/elastic/ci-agent-images/pull/2913 is merged, Ubuntu 22.04 is near end-of-life anyway.

This upgrades agents to Ubuntu 24.04 for now (26.04 can follow later).

This also fixes the Rally IT tests to actually test Python 3.10 and 3.13, instead of testing Python 3.13 twice.

### DIFF
--- a/.buildkite/it/pipeline.yml
+++ b/.buildkite/it/pipeline.yml
@@ -2,6 +2,7 @@ agents:
   provider: "gcp"
   image: family/core-ubuntu-2604
   machineType: "n1-standard-8"
+  diskSizeGb: 50
 
 steps:
   - label: "Run 3.10 integration tests :test_tube:"
--- a/.buildkite/it/pipeline.yml
+++ b/.buildkite/it/pipeline.yml
@@ -1,6 +1,6 @@
 agents:
   provider: "gcp"
-  image: family/core-ubuntu-2204
+  image: family/core-ubuntu-2604
   machineType: "n1-standard-8"
 
 steps:
--- a/.buildkite/it/pipeline.yml
+++ b/.buildkite/it/pipeline.yml
@@ -1,8 +1,7 @@
 agents:
   provider: "gcp"
-  image: family/core-ubuntu-2604
+  image: family/core-ubuntu-2404
   machineType: "n1-standard-8"
-  diskSizeGb: 50
 
 steps:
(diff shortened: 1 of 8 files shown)


b-deam:
IT tests have consistently failed `test_get[gs-big_file]` and hung on `test_create_track[in-memory-it]`:
```
it/storage/manager_test.py::test_get[gs-big_file] FAILED
it/tracker_test.py::test_create_track[in-memory-it]
```

.buildkite/it/pipeline.yml
```diff
@@ -1,7 +1,8 @@
 agents:
   provider: "gcp"
-  image: family/core-ubuntu-2204
+  image: family/core-ubuntu-2604
   machineType: "n1-standard-8"
+  diskSizeGb: 50
```
inqueue: How about pruning downloads once test_get() finishes and passes?

Can it be placed in https://github.com/pquentin/rally/blob/32e88135ad65f650b4bc2bb2e0ee1b6c370c0f02/it/storage/manager_test.py#L68-L75?

```python
def test_get(case: GetCase, base_url: str, transfer_manager: storage.TransferManager):
    tr = transfer_manager.get(f"{base_url}/{case.path}")
    try:
        tr.wait(timeout=300.0)
        assert tr.done
        assert os.path.isfile(tr.path)
        assert os.path.getsize(tr.path) == case.want_size
        assert tr.crc32c is not None
        assert crc32c.Checksum.from_filename(tr.path) == crc32c.Checksum.from_base64(tr.crc32c)
    finally:
        tr.prune()
```

inqueue (dismissed):

.buildkite/it/pipeline.yml
pquentin: Done — `test_get` now prunes the transfer in a `finally` block, so the ~7GB corpora are deleted after each case. I also dropped `diskSizeGb: 50` and switched the agents to Ubuntu 24.04 for now instead of 26.04.

That should also unblock the `test_create_track[in-memory-it]` hang, which looked like a follow-on from the disk filling up.

b-deam (approved):

PR #2171, opened 2026-08-05
Fix intermittent NoCredentialsError in S3 multipart transfer manager

### LINKED ISSUE #2168: Transfer Manager fails with NoCredentialsError in AWS
### Symptom

When using multi-part transfer manager with regional authenticated S3 mirrors:

```
# rally.ini
[..]
[track]
track.downloader.multipart_enabled = true
[..]

# storage-mirrors.json
{
  "mirrors": [
    {
      "sources": ["https://rally-tracks.elastic.co/"],
      "destinations": [ "s3://<mirror>/" ]
    }
  ]
}
```

AWS authentication can fail with `botocore.exceptions.NoCredentialsError: Unable to locate credentials` exception.

Image

```

  File "/home/esbench/rally/esrally/storage/_transfer.py", line 487, in _run
    with self.client.get(self.url, check_head=check_head) as got:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/rally/esrally/storage/_client.py", line 262, in get
    got = self._get(head.url, check_head=check_head)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/rally/esrally/storage/_client.py", line 292, in _get
    got: GetResponse = adapter.get(url, check_head=check_head)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/rally/esrally/storage/aws.py", line 75, in get
    response = self._s3.get_object(Bucket=address.bucket, Key=address.key, **headers)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/client.py", line 565, in _api_call
    return self._make_api_call(operation_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/client.py", line 999, in _make_api_call
    http, parsed_response = self._make_request(
                            ^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/client.py", line 1023, in _make_request
    return self._endpoint.make_request(operation_model, request_dict)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/endpoint.py", line 119, in make_request
    return self._send_request(request_dict, operation_model)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/endpoint.py", line 196, in _send_request
    request = self.create_request(request_dict, operation_model)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/endpoint.py", line 132, in create_request
    self._event_emitter.emit(

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/hooks.py", line 412, in emit
    return self._emitter.emit(aliased_event_name, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/hooks.py", line 256, in emit
    return self._emit(event_name, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/hooks.py", line 239, in _emit
    response = handler(**kwargs)
               ^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/signers.py", line 105, in handler
    return self.sign(operation_name, request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/signers.py", line 197, in sign
    auth.add_auth(request)

  File "/home/esbench/.local/lib/python3.11/site-packages/botocore/auth.py", line 418, in add_auth
    raise NoCredentialsError()

botocore.exceptions.NoCredentialsError: Unable to locate credentials
```

The authentication failure is not due to infrastructure problem. Other Rally runs in the same AWS region in EC2 instances provisioned in the same way succeed. This suggests the problem is with how transfer manager obtains and maintains authentication for S3 access.

Another problem is once the `botocore.exceptions.NoCredentialsError` is seen, Rally does not recover correctly by falling back to unauthenticated `https://rally-tracks.elastic.co`.

### Workaround

Multi-part transfer is optional and not enabled by default. A known workaround is to disable multi-part transfer.

Abmun:
Closes #2168

## Problem

When downloading tracks from a regional, authenticated S3 mirror with
`track.downloader.multipart_enabled = true`, Rally could intermittently
fail with `botocore.exceptions.NoCredentialsError: Unable to locate
credentials`, even though the AWS credentials on the host (e.g. an EC2
instance role) were otherwise valid — other Rally runs on the same
instance succeeded.

Two things contributed to this:

1. `AdapterRegistry` caches a single `S3Adapter` instance per adapter
   class, so every worker thread in the multipart transfer manager's
   thread pool shares one `S3Adapter`, and therefore one lazily created
   boto3 S3 client. Several threads racing to resolve AWS credentials for
   the very first time concurrently (e.g. fetching them from the EC2
   instance metadata service) could intermittently raise
   `NoCredentialsError`.
2. `S3Adapter` didn't catch any botocore errors, so `NoCredentialsError`
   propagated as a raw exception instead of the `ServiceUnavailableError`
   that `Client` already knows how to handle by failing over to another
   mirror or the unauthenticated source URL. This is why a single
   transient S3 auth hiccup aborted the whole transfer instead of Rally
   falling back, as described in the issue.

## Fix

- Make lazy S3 client creation in `S3Adapter` thread-safe (double-checked
  locking) and warm up credentials once, synchronously, before any
  concurrent request is dispatched to the shared client.
- Translate `NoCredentialsError`, `EndpointConnectionError`, and
  retryable `ClientError` responses (throttling, 5xx) into
  `ServiceUnavailableError` in `head()`/`get()`, matching the pattern
  already used by the HTTP adapter, so `Client` correctly fails over to
  another mirror/source instead of crashing. Non-transient errors (e.g.
  `AccessDenied`) still propagate unchanged so real
  configuration/permission problems aren't masked.

## Test plan

- Added unit tests in `tests/storage/aws_test.py` covering:
  - transient errors (`NoCredentialsError`, `EndpointConnectionError`,
    throttling/5xx `ClientError`) being translated to
    `ServiceUnavailableError` in both `head()` and `get()`
  - non-transient `ClientError`s (e.g. `AccessDenied`) still propagating
    unchanged
  - concurrent first-access to the S3 client from multiple threads
    creating only a single boto3 session/client
  - credentials being warmed up exactly once, and client creation
    surviving a warm-up failure
- `make lint test` (black, isort, mypy, full `tests/storage` suite —
  212 tests) passes locally.

### DIFF
--- a/esrally/storage/aws.py
+++ b/esrally/storage/aws.py
@@ -18,16 +18,19 @@ from __future__ import annotations
 
 import logging
 import os
+import threading
 import urllib.parse
 from collections.abc import Mapping
 from typing import Any, NamedTuple, Protocol, runtime_checkable
 
 import boto3
+import botocore.exceptions
 from botocore.response import StreamingBody
 from typing_extensions import Self
 
 from esrally import types
 from esrally.storage import Adapter, GetResponse, Head, StorageConfig
+from esrally.storage._adapter import ServiceUnavailableError
 from esrally.storage.http import (
     head_to_headers,
     parse_accept_ranges,
@@ -36,6 +39,25 @@ from esrally.storage.http import (
 )
 
 LOG = logging.getLogger(__name__)
+
+# botocore error codes that represent a transient/retryable failure of the S3 service (as opposed to e.g. an
+# invalid request or a missing object). They are translated to `ServiceUnavailableError` so that `Client` can fail
+# over to another mirror (or the unauthenticated source URL) instead of aborting the whole transfer.
+_RETRYABLE_CLIENT_ERROR_CODES = frozenset(
+    {
+        "InternalError",
+        "RequestTimeout",
+        "RequestTimeoutException",
+        "PriorRequestNotComplete",
+        "ServiceUnavailable",
+        "SlowDown",
+        "Throttling",
+        "ThrottlingException",
+        "ThrottledException",
+        "RequestLimitExceeded",
+        "TooManyRequestsException",
+    }
+)
 
 
 class S3Adapter(Adapter):
@@ -61,10 +83,19 @@ class S3Adapter(Adapter):
         self.chunk_size = chunk_size
         self.aws_profile = aws_profile
         self._s3_client = s3_client
+        # It protects lazy S3 client creation (and the initial credentials resolution) from being executed
+        # concurrently by more than one thread at once. The multipart transfer manager downloads a single file
+        # from several threads at the same time, all of them sharing this very same `S3Adapter` instance.
+        self._lock = threading.Lock()
 
     def head(self, url: str) -> Head:
         address = S3Address.from_url(url)
-        res = self._s3.head_object(Bucket=address.bucket, Key=address.key)
+        try:
+            res = self._s3.head_object(Bucket=address.bucket, Key=address.key)
+        except botocore.exceptions.ClientError as ex:
+            raise _translate_client_error(ex) from ex
+        except (botocore.exceptions.NoCredentialsError, botocore.exceptions.EndpointConnectionError) as ex:
+            raise ServiceUnavailableError(str(ex)) from ex
         return head_from_response(url, res)
 
     def get(self, url: str, *, check_head: Head | None = None) -> GetResponse:
@@ -72,7 +103,17 @@ class S3Adapter(Adapter):
         head_to_headers(check_head, headers)
 
         address = S3Address.from_url(url)
-        response = self._s3.get_object(Bucket=address.bucket, Key=address.key, **headers)
+        try:
+            response = self._s3.get_object(Bucket=address.bucket, Key=address.key, **headers)
+        except botocore.exceptions.ClientError as ex:
+            raise _translate_client_error(ex) from ex
+        except (botocore.exceptions.NoCredentialsError, botocore.exceptions.EndpointConnectionError) as ex:
+            # Getting credentials can intermittently fail when many threads are requesting them at the same time
+            # (for example while downloading many parts of the same file concurrently via the multipart transfer
+            # manager). It is treated as a transient failure so that `Client` can fail over to another mirror (or
+            # to the unauthenticated source URL) instead of aborting the whole transfer.
+            raise ServiceUnavailableError(str(ex)) from ex
+
         body: StreamingBody | None = response.get("Body")
         if body is None:
             raise RuntimeError("S3 client returned no body.")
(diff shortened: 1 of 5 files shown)


gbanasiak:
Buildkite test this

esrally/storage/aws.py
```diff
@@ -37,6 +40,25 @@
...
+        "PriorRequestNotComplete",
+        "ServiceUnavailable",
+        "SlowDown",
+        "Throttling",
+        "ThrottlingException",
+        "ThrottledException",
+        "RequestLimitExceeded",
+        "TooManyRequestsException",
+    }
+)
```
gbanasiak: How was this list obtained exactly? Looking at https://docs.aws.amazon.com/AmazonS3/latest/developerguide/ErrorResponses.html#RESTErrorResponses, `ServiceUnavailable`, and `SlowDown` are documented and seem to make sense here. I am not sure about the rest. Do they exist? Also see my later comment on `ServiceUnavailableError`.

stevereads-elastic (approved):
This is all pretty straightforward: be more precise about the exceptions we throw, and translate exceptions in such a way that we respond to them appropriately (e.g., don't retry with a 403). LGTM.

esrally/storage/aws.py
```diff
@@ -61,18 +83,37 @@ def __init__(
...
+        try:
+            response = self._s3.get_object(Bucket=address.bucket, Key=address.key, **headers)
+        except botocore.exceptions.ClientError as ex:
+            raise _translate_client_error(ex) from ex
+        except (botocore.exceptions.NoCredentialsError, botocore.exceptions.EndpointConnectionError) as ex:
+            # Getting credentials can intermittently fail when many threads are requesting them at the same time
+            # (for example while downloading many parts of the same file concurrently via the multipart transfer
+            # manager). It is treated as a transient failure so that `Client` can fail over to another mirror (or
+            # to the unauthenticated source URL) instead of aborting the whole transfer.
+            raise ServiceUnavailableError(str(ex)) from ex
```
gbanasiak: Today `ServiceUnavailableError` signals a transient error due to a possibly overwhelmed service which is why it is met with reduction of the number of connections. In other words it is a "slow down but continue" signal. The `NoCredentialsError` and `EndpointConnectionError` exceptions might indicate a _permanent_ client-side issue. Is `ServiceUnavailableError` a good match for these exceptions?

gbanasiak (changes requested):
Many thanks for the `NoCredentialsError` fix. I have concerns about how much we want to pack into `ServiceUnavailableError`, see below.

esrally/storage/aws.py
Abmun: @gbanasiak  I checked it against AWS's actual [S3 error code docs](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses), only InternalError, RequestTimeout, ServiceUnavailable, and SlowDown are real S3 codes. Things like PriorRequestNotComplete (that's a Route53 code) and ThrottledException/TooManyRequestsException (Lambda/API Gateway style) don't apply to S3's REST API at all. Trimmed the list to just the four verified codes, the existing status_code >= 500 fallback still catches anything else server-side.

esrally/storage/aws.py
Abmun: @gbanasiak That exception means "the server is overwhelmed" and Client responds by permanently throttling concurrent connections to that server. A NoCredentialsError or EndpointConnectionError says nothing about server load, it's a local problem. So I added a new ClientUnavailableError in _adapter.py: Client still fails over to another mirror on it (same as before), but no longer reduces the connection count for that server. S3Adapter now raises this instead for those two exceptions.

esrally/storage/aws.py
gbanasiak: Thank you.

esrally/storage/aws.py
gbanasiak: I like this direction. Can we also detect a server exhaustion case when all the servers resulted in `ClientUnavailableError` in `Client:get()` and raise `ClientUnavailableError` instead of `ServiceUnavailableError`? This should fast fail the transfer.

gbanasiak (commented):
Thank you. Additional request raised in https://github.com/elastic/rally/pull/2171#discussion_r3806667903.

gbanasiak:
Buildkite test this

esrally/storage/aws.py
Abmun: @gbanasiak 

Good call, done in cb0099cc.

`Client.get()` now tracks, across the mirrors it actually attempts a GET
against, how many failed with `ClientUnavailableError` vs.
`ServiceUnavailableError` (or were skipped due to the per-server
connection limit). If every attempted mirror failed with
`ClientUnavailableError`, `Client.get()` raises `ClientUnavailableError`
instead of `ServiceUnavailableError` once mirrors are exhausted.

Turns out no changes were needed in `_transfer.py` for the fail-fast
behavior: `TransferTask._run()` already treats any exception other than
`ServiceUnavailableError`/`TimeoutError`/`StreamClosedError` as fatal
(interrupts the transfer and records the error), so a
`ClientUnavailableError` propagating out of `Client.get()` gets that
"fail fast" treatment for free.

If mirrors fail for a mix of reasons (some client-unavailable, some
genuinely overloaded, or some just skipped for being at the connection
limit with unknown status), it keeps the existing
`ServiceUnavailableError` back-off-and-retry behavior, since it isn't
safe to assume every mirror is unusable in that case.

Added two tests in `client_test.py` covering both scenarios.

gbanasiak:
Buildkite test this

esrally/storage/aws.py
```diff
@@ -37,6 +40,20 @@
...
+# S3 REST API, see https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#RESTErrorResponses
+_RETRYABLE_CLIENT_ERROR_CODES = frozenset(
+    {
+        "InternalError",
+        "RequestTimeout",
+        "ServiceUnavailable",
+        "SlowDown",
+    }
+)
+
```
gbanasiak, line 56: In HTTP adapter timeouts are surfaced through `TimeoutError`, see https://github.com/elastic/rally/pull/2173. To align HTTP and S3 adapters `RequestTimeout` could be treated as `TimeoutError`, not `ServiceUnavailableError`.

gbanasiak (approved):
LGTM

esrally/storage/aws.py
gbanasiak: To be clear, this is not a blocker. We can address this later if we see a good reason to align.

PR #2182, opened 2026-09-07
feat(telemetry): include field_type in disk-usage-stats meta

salvatore-campagna:
## TL;DR

`DiskUsageStats` now stores `field_type` in the metrics meta for every field the `_disk_usage` API reports a type for. Downstream queries against the metrics store can filter or aggregate `disk-usage-*` metrics by field type directly, without reconstructing types from index mapping files.

## Summary

The Elasticsearch `_disk_usage` API returns a `type` key alongside storage byte counts for every mapped field: `keyword`, `float`, `date`, `long`, `dense_vector`, and so on. `DiskUsageStats.handle_telemetry_usage` was reading that response dict and discarding the `type` key; the meta dict stored with each metric contained only `index` and `field`.

Without the type in the store, any post-race analysis that breaks storage down by field type must reconstruct types from track mapping files. For tracks with templated mappings (for example `elastic/logs`, which spreads hundreds of ECS fields across Jinja2-conditional files), that reconstruction is complex and fragile.

The fix is a single conditional in `handle_telemetry_usage`: if the API response includes `type`, it is forwarded as `field_type` in the meta dict. Internal fields (`_id`, `_source`, `_seq_no`, and so on) carry no type in the API response and are unaffected, so existing tooling that reads those metrics is unchanged.

## Testing

```
pytest tests/telemetry_test.py::TestDiskUsageStats -v
```

### DIFF
--- a/esrally/telemetry.py
+++ b/esrally/telemetry.py
@@ -2430,6 +2430,8 @@ class DiskUsageStats(TelemetryDevice):
 
             for field, field_info in idx_fields["fields"].items():
                 meta = {"index": index, "field": field}
+                if field_type := field_info.get("type"):
+                    meta["field_type"] = field_type
                 self.metrics_store.put_value_cluster_level("disk_usage_total", field_info["total_in_bytes"], meta_data=meta, unit="byte")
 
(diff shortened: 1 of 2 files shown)


gbanasiak (approved):
LGTM

PR #2183, opened 2026-09-09
Don't overwrite existing rally-annotations index template

### LINKED ISSUE #1900: Rally should not overwrite pre existing templates by default

parh0m2007:
## What does this PR do?

Closes #1900.

Rally should not overwrite pre-existing index templates by default. The bulk of this behavior was already implemented in #2111 for the metrics/races/results stores. However, one code path was left behind: when storing an annotation (`esrally add annotation`), Rally still **unconditionally overwrote** the `rally-annotations` index template whenever the annotations index did not exist yet — even if the template had been customized by the user (e.g. to apply a custom lifecycle policy for moving data to frozen tiers) and even with `datastore.overwrite_existing_templates = false`.

## What does this PR change?

* Adds `IndexHandler.ensure_annotations_template()` in `esrally/metrics.py`, which reuses the existing `_should_apply_update()` logic that all other Rally index templates already go through: the template is written only if it doesn't exist yet, or if it differs **and** the user opted in via `datastore.overwrite_existing_templates = true` in the `[reporting]` section of `rally.ini`. Identical templates are left untouched, differing templates are kept by default and a warning is logged when they are overwritten.
* `EsRaceStore.add_annotation()` now calls this method instead of issuing a raw `put_template` call.
* Documents that `datastore.overwrite_existing_templates` also applies to the `rally-annotations` index template (`docs/configuration.rst`).
* Adds tests covering all four cases: fresh create, identical existing template, differing template without opt-in (kept), and differing template with opt-in (overwritten).

## Why is it important?

This closes the remaining gap in #1900: users who customize the `rally-annotations` template (e.g. to set custom lifecycle policies) no longer have their changes silently reverted whenever Rally adds an annotation.

## Checklist

- [x] I have tested this change locally (unit tests pass: `tests/metrics_test.py`)
- [x] Documentation updated (`docs/configuration.rst`)
- [x] Formatting/lint (`black`, `isort`) passes
- [ ] Any changes to `docs/` have been verified with `make docs`

Co-authored-by: Claude <noreply@anthropic.com>


parh0m2007:
Hi maintainers! I picked this up after noticing the last comment asked if anyone was working on it. While the core of #1900 landed in #2111, I found the annotations code path was still unconditionally overwriting its index template — this PR closes that remaining gap so `rally-annotations` is treated consistently with all other Rally templates under `datastore.overwrite_existing_templates`. Happy to make any adjustments if needed!

pquentin:
The CLA check failed because the commit author is currently `parh0m2007@users.noreply.github.com` instead of your email address. You need to amend the commit and force push to fix this.

pquentin (changes requested):
Thanks! I'm not sure that this is needed. rally-annotations is typically tiny, so there's no need to add custom lifecycles. Going to close this one. Sorry for the delay.

PR #2184, opened 2026-09-09
Introduce named constants for config keys

### LINKED ISSUE #1646: DRY/Refactor hard coded config file keys

parh0m2007:
## Summary

Closes #1646

The `Config` object's keys were referenced as hard-coded strings all over the codebase: they get populated during argument parsing in `rally.py` (e.g. `cfg.add(config.Scope.applicationOverride, "system", "list.max_results", ...)`) and are read back, again as strings, in `metrics.py`, `paths.py`, `driver/driver.py`, the `mechanic` modules, and others. A typo in any of those strings silently resolves to a missing config option instead of raising an error — this is exactly the class of bug seen in #1645.

## Changes

- New module `esrally/config_keys.py` defining every `"system"` section config key as a module-level constant (e.g. `RACE_ID = "race.id"`, `LIST_MAX_RESULTS = "list.max_results"`), plus `SYSTEM_SECTION`. Constants are annotated with `typing.Final`, so mypy preserves their literal types and still accepts them wherever `types.Key` is expected — the static type contract stays fully intact.
- All ~100 call sites across `esrally/` now reference the constants instead of raw strings; a typo becomes a real error (`NameError`/`mypy` failure) instead of a silently missing option.
- New `tests/config_keys_test.py` guards that every constant is a valid member of the `types.Key` Literal and that the constants are unique, keeping the two in sync.

Behavior is unchanged: the constants carry the exact same string values, verified by the full unit test suite.

## Testing

- `make test` equivalent: full unit suite passes (1947 tests) — the only pre-existing failure is `tests/storage/mirror_test.py::test_from_config[mirror_files]`, which fails identically on unmodified `master` in this environment and is unrelated (it depends on a local `mirrors.json` fixture path).
- `make lint` equivalent: `pre-commit run --all-files` passes all hooks (black, isort, mypy, pylint, pyupgrade).

### DIFF
--- a/esrally/config_keys.py
+++ b/esrally/config_keys.py
@@ -1,0 +1,75 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
+# license agreements. See the NOTICE file distributed with
+# this work for additional information regarding copyright
+# ownership. Elasticsearch B.V. licenses this file to you under
+# the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""Named constants for configuration keys.
+
+Instead of scattering hard-coded strings for config keys all over the
+codebase (which is error-prone, as a typo does not raise an error), we
+define the keys as module-level constants so a typo becomes a real
+error (`AttributeError` / ``NameError``) instead of a silently missing
+config option.
+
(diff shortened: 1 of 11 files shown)


pquentin:
buildkite test this please

pquentin:
Sorry for the reviewing delay. This was already done in https://github.com/elastic/rally/pull/1798, which relies on mypy to check the keys. It also differentiates between sections and keys, which your change does not.

PR #2185, opened 2026-09-16
Install track dependencies with Rally's constraints

### LINKED ISSUE #2180: Preserve Rally dependency constraints when installing track dependencies
Rally installs track dependencies with `pip install --upgrade --target ~/.rally/libs`. This can upgrade transitive dependencies independently of Rally’s pinned environment.

For example, the `elastic/security` track installs [geneve](https://github.com/elastic/geneve), which depends on `anyio` without any constraints. Two days ago, [anyio 4.15.0](https://github.com/agronholm/anyio/releases/tag/4.15.0) was released, and that version imports `sentinel` from `typing_extensions`, while Python resolves Rally’s pinned `typing_extensions==4.12.2`, which only provides `Sentinel`. ([This is a recent breaking change in typing extensions.](https://github.com/python/typing_extensions/releases/tag/4.16.0)) The resulting mix of packages from ~/.rally/libs and Rally’s virtualenv fails at runtime:

```
File ".../.rally/libs/anyio/abc/_streams.py", line 8, in <module>
from .._core._typedattr import TypedAttributeProvider

File ".../.rally/libs/anyio/_core/_typedattr.py", line 10, in <module>
from typing_extensions import sentinel

ImportError: cannot import name 'sentinel' from 'typing_extensions' (.../rally/.venv/lib/python3.13/site-packages/typing_extensions.py)
```

(Notice how anyio comes from `.rally/libs`, while `typing_extensions` comes from `.rally/.venv`.)

Track dependency installation should use Rally’s resolved environment as a [pip constraints file](https://pip.pypa.io/en/stable/user_guide/#constraints-files). This would allow track-specific packages to be installed or upgraded while forcing shared dependencies such as `elasticsearch`, `anyio`, and `typing_extensions` to versions compatible with Rally.

As a short-term mitigation, bumping Rally’s pinned dependencies, particularly `typing_extensions`, would restore compatibility with the newer anyio. However, without applying Rally’s constraints during track dependency installation, similar incompatibilities can recur whenever a track introduces or upgrades a shared transitive dependency.

mmustafasenoglu:
Fixes #2180

Track dependencies install with pip --upgrade --target, which can pull shared transitive deps beyond what Rally runs with (e.g. anyio 4.15 vs pinned typing_extensions). This passes Rally's resolved versions as a constraints file so only track-specific packages float.

Changes:
- esrally/track/loader.py: build constraints from installed esrally deps, pass --constraint to pip, clean up the temp file
- tests/track/loader_test.py: cover the --constraint wiring and the constraints helper


pquentin:
Hello, and thank you for your contribution! Sorry for not submitting https://github.com/elastic/rally/pull/2187 earlier; I prepared it a few days ago and did not get to open it until now. There are two key differences compared to this PR:

 1. it records all pinned dependencies, including transitive ones, and
 2. it creates the constraints file at build time, not runtime.

Sorry about that! Closing.

PR #2186, opened 2026-09-16
Add Cold Run Support To Search and ESQL operators

lukewhiting:
This adds a new param to the ESQL and Search operators that, when true and running against serverless, clears the blob cache to simulate a query against a "cold" node. It does the cache clear in a separate timing context to prevent polluting the results with this admin work.

Previously this was implemented on some tracks either by subclassing the relevant runners to add this functionality (a tech debt nightmare as there's multiple copies) or some tracks used a composite operation to support this which was not technically accurate as it would include the cache clear operation's timings in the final reports, skewing results.

The decision was made to fail noisy if the credentials for the run do not have the required operator permissions.

This PR was agent assisted and human reviewed


esrally/driver/runner.py
```diff
@@ -239,6 +239,37 @@ def _transport_request_params(params):
...
+            await es.perform_request(method="POST", path="/_internal/blob_caches/clear")
+        except (elasticsearch.ApiError, elasticsearch.TransportError) as e:
+            exc = e
+        finally:
+            es.restore_context(token)
+        if exc is not None:
+            # Stamp the outer context so driver.py timing arithmetic has valid values.
+            es.on_request_start()
+            es.on_request_end()
+            raise exc
```
gbanasiak, line 271: nit: I would add a comment a measured service time will be close to 0 in the error case

gbanasiak (approved):
LGTM

